PR #423, opened 2022-03-02
Labeled facet

MartinKarp:
Adds support to choose bcs in the parameter file. A rather large revamp of how we treat facets and bcs.


jacwah:
Maybe we should add the new parameter to `doc/pages/case-file.md`

njansson (approved):

PR #437, opened 2022-03-12
Improvements to the source term

MartinKarp:
Adds support for the user to define a "vector valued" source term. This greatly helps when working with accelerators.

Also adds sort


njansson (approved):
We should probably implement some proper sorting instead of Num. Rec. ...

PR #447, opened 2022-03-17
GPU direct MPI in gather scatter

jacwah:
Uses GPU direct feature in MPI to avoid copying back and forth between device and host memory in gather scatter kernel.

Currently always enabled on device backend. Could consider adding a configure flag for this, though MPI support for GPU direct seems fairly good as far as I can tell.

The pack kernel definitely can be improved by launching more threads for each PE.

On nj works with `openmpi/4.0.5-gcc-10.2.0-qakjkep` module.

On dt works with `craype-accel-amd-gfx908` module plus `MPICH_GPU_SUPPORT_ENABLED=1` environment variable.

OpenCL not implemented.

### DIFF
--- a/src/gs/gather_scatter.f90
+++ b/src/gs/gather_scatter.f90
@@ -61,8 +61,6 @@ module gather_scatter
      integer, allocatable :: shared_dof_gs(:)         !< Shared dof to gs map.
      integer, allocatable :: shared_gs_dof(:)         !< Shared gs to dof map.
      integer, allocatable :: shared_blk_len(:)        !< Shared non-facet blocks
-     type(stack_i4_t), allocatable :: send_dof(:)     !< Send dof to shared-gs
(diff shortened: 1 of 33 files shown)


src/gs/gather_scatter.f90
```diff
@@ -95,7 +96,11 @@ subroutine gs_init(gs, dofmap, bcknd)
     
     gs%dofmap => dofmap
     
-    allocate(gs_mpi_t::gs%comm)
+    if ((NEKO_BCKND_HIP .eq. 1) .or. (NEKO_BCKND_CUDA .eq. 1)) then
```
jacwah, line 99: `#ifdef HAVE_DEVICE_MPI`?

njansson: I would rather have a parameter in `neko_config`

njansson (approved):

PR #509, opened 2022-05-18
Fix/bcs

### LINKED ISSUE #508: pnpn formulation diverges for cases with symmetry conditions
After the introduction of #494 `fluid_pnpn` diverges for cases with symmetry conditions e.g. `hemi`, both under `examples` and in the test suite (https://github.com/ExtremeFLOW/test-suite)

Any ideas @MartinKarp?

MartinKarp:
Closes #508 and makes the code a bit nicer

### DIFF
--- a/src/bc/non_normal.f90
+++ b/src/bc/non_normal.f90
@@ -36,6 +36,7 @@ module non_normal
   use neko_config
   use num_types
   use dirichlet
+  use tuple
   use device
   use coefs
   use math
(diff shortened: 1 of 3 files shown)


njansson:
But this will now break `plan4` host and device right? (since it assumes one bc per velocity component)

njansson:
And I guess we can also remove a lot of redundant symmetry related device code?

MartinKarp:
It should work for plan4 as well, I only changed things under the hood for symmetry and non_normal

PR #579, opened 2022-08-17
Feature/ax kernel optimisation on amd gpu

jonvin73:
Not sure the output is done as you would like, and open to suggestions for the structure.

Using environment variables to turn of the autotuning seemed best as we so not have to worry about passing stuff to here and everything is self contained.

Right now for each lx will run the versions 5 times and take the one with the shortest time to run all the time after that.


njansson (approved):
@njansson will fix the logging once #580 is merged

PR #594, opened 2022-09-01
Runtime selection of Sync. / Async. gs pack/unpack w. device MPI

njansson:
**Rework device aware MPI gs alg.**

**Send:**
Either, issue one asynchronous gather-scatter pack kernel per, to be sent, message, instead of packing all buffers at once. or issue a global device sync, sync on each packing stream, followed by issuing the non-blocking send (old way).

**Recv:**
Either, issue one asynchronous gather-scatter unpack kernels per received MPI message in gs, or wait for all messages to arrive. 

The strategy is selected at runtime, and can be different for each `gs_t`, and for each device in a `gs_t`

### DIFF
--- a/src/device/cuda_intf.F90
+++ b/src/device/cuda_intf.F90
@@ -124,6 +124,16 @@ module cuda_intf
   end interface
 
   interface
+     integer (c_int) function cudaStreamCreateWithFlags(stream, flags) &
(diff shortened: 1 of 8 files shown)


njansson:
**Don't merge** until we can verify on LUMI-G and JUWELS

PR #615, opened 2022-09-20
Feature/glsc3 optimisation

jonvin73:
Replace the memcpy and cpu reduction with a GPU reduction then much smaller memcpy in glsc3_many

Comes out significantly ahead on CUDA generally. Looks to be better on HIP as well, Did not touch the OpenCL


njansson:
Relates to https://github.com/ExtremeFLOW/EuroHack22/issues/2

njansson (approved):

PR #618, opened 2022-09-21
Retire plan4 solver, change all examples to pnpn

njansson:


timofeymukha:
@njansson Can we maybe erase plan4 completely? I am not sure what this was, but I guess there is 0 chance someone is using it now, or?

njansson:
> @njansson Can we maybe erase plan4 completely? I am not sure what this was, but I guess there is 0 chance someone is using it now, or?

Indeed, we should probably remove remove both `plan4` and `device_plan4`, or what do you think @MartinKarp

MartinKarp:
Yeah, they can be removed

PR #634, opened 2022-09-28
"2D", but really it is a one element mesh

MartinKarp:


njansson:
Related to https://github.com/orgs/ExtremeFLOW/projects/8

Issue #627, opened 2022-09-27
coef type


pschlatt1:
coef as a class name may be misleading as it mainly contains geometry factors. Therefore, one could rename coef to geom, except h1,h2 which could stay in class coef.

MartinKarp:
I might suggest that we add h1,h2 to the class axhelm

njansson:
I rather still have them in `coef_t`, otherwise it would make ax problem specific. Furthermore, we might want/need more coefficients in the future, which will not only affect Ax operations

PR #654, opened 2022-10-01
Pass down parameters in user_check

### LINKED ISSUE #640: passing down params to user_check
use case: access the Reynolds number in user_check

Potentially make the argument list the same for user_check and user_init_modules (in the spirit of the Nek5000 call to user_check with istep=0)

njansson:
Pass down parameters to `user_check` to allow access to e.g. the Reynolds number, fixes #640

Note: This breaks all existing examples using `user_check`. All provided examples have been updated.

### DIFF
--- a/bench/tgv32/tgv.f90
+++ b/bench/tgv32/tgv.f90
@@ -62,10 +62,11 @@ contains
     uvw(3)   = zero
   end function tgv_ic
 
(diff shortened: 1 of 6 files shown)


njansson:
@pschlatt1,@MartinKarp do we need to pass down anything else than `params` for now? (yes once #627 is addressed we need to pass down additional things)

PR #657, opened 2022-10-03
Consitent symmetric vertex notation

adampep:
The assumption is to keep backward compatibility and do not rewrite file format and converters used at pre-proessing step. Only neko internals are changed and the points are swapped at the file reading step. Periodic bc, dof numberring and grid generation are corrected as well.
Stuff left to change/test:
- mesh file writers (e.g. nmsh_file_write) have to swap points
- curved faces/edges are not tested yet
- vtk file writers
- krylov/pc_hsmg.f90 should be checked as it operates on elements

### DIFF
--- a/src/io/nmsh_file.f90
+++ b/src/io/nmsh_file.f90
@@ -434,6 +434,8 @@ contains
     integer :: nmsh_quad_size, nmsh_hex_size, nmsh_zone_size, nmsh_curve_size
     integer :: nzones, ncurves 
     class(element_t), pointer :: ep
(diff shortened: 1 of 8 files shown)


adampep:
Cylindrical faces and nmsh writer should be fixed, but I cannot test the code (turbpipe is too big for my laptop and it seems there is no test with midpoints). I've swapped vertices in vtk writer as well, but not 100% sure it is sufficient.

njansson (approved):

MartinKarp:
I will test this further today

PR #661, opened 2022-10-05
Fix invalid offset in HIP async. GS pack

njansson:
This fixes the divergence observed on LUMI-G, when an asynchronous gather-scatter strategy was picked.


njansson:
> Don't merge until final verification tests passes on LUMI-G

Passed verification!

PR #677, opened 2022-10-16
Cleanup and reduce (host) memory usage in device GMRES

njansson:
* Use proper `c_sizeof` instead of magic constants
* Don't allocate unnecessary big (unused) host arrays, instead of `allocate(x(n))`, use `allocate(x(1))` and use the scalar array as target in `device_associate`


njansson:
> * Don't allocate unnecessary big (unused) host arrays, instead of `allocate(x(n))`, use `allocate(x(1))` and use the scalar array as target in `device_associate`

This had to be reverted, since we can't assume that `Ax%compute` does not use any host allocated arrays

PR #694, opened 2022-10-28
Add support for non-uniform distributions of elements

MartinKarp:
Adds an environment variable `rank_weight` that can be set in order to weight the distribution of elements across ranks.


timofeymukha:
Did we forget to review this?

timfelle:
@njansson, @timofeymukha What is the plan here?

njansson:
> @njansson, @timofeymukha What is the plan here?

I would say we can close it for now (and reopen when needed). 

This was mainly for a project specific use case

timofeymukha:
Yeah, if nobody bothered with it since Oct 2022, I think we can close it for now.

PR #720, opened 2022-11-11
3D Statistics 

MartinKarp:
Here is a draft for the 3D statistics in Neko. 
It computes all 44 averaged quantities that are present in the Nek5000 toolbox.
It outputs the mean flow field i.e. `<u>, <v>, <w>, <p> `in the file mean_field.nek5000 and all other (40) stats in the stats.nek5000. The order should probably be documented so that we can postprocess the fields more easily.
The time interval over which one computes the averages is decided by the parameter stats_sample_time. We should maybe also add one for how often we sample (Now its every time step, but we can probably reduce it to every 10 or so). 

I have not tested it completely yet so please dont merge. If you have the time, please try it out.


src/simulation.f90
```diff
@@ -126,8 +127,8 @@ subroutine neko_solve(C)
     end do
 
     call profiler_stop
-
-    if (t .lt. C%params%T_end) then
+    call C%sample_means%sample(t,ifforce=.true.)
+    if (C%params%output_chkp .or. t .lt. C%params%T_end) then
```
njansson, line 131: The intent is to let the sampler take cares of the logic so we can keep the time-stepping loop uncluttered

src/case.f90
```diff
@@ -299,27 +304,31 @@ subroutine case_init(C, case_file)
...
-       end if
+       C%f_mf = mean_flow_output_t(C%fluid%mean, C%params%stats_begin, &
+                                   path=C%params%output_dir)
+       call C%sample_means%add(C%f_mf)
+    end if
+    if (C%params%stats_fluid) then
+       call C%q%add(C%fluid%stats)
+       C%f_stats_output = fluid_stats_output_t(C%fluid%stats, C%params%stats_begin, &
+                                              path=C%params%output_dir)
+       call C%sample_means%add(C%f_stats_output)
```
njansson, line 331: See the comment in `simulation.f90` regarding several samplers

src/fluid/fluid_stats.f90
```diff
@@ -0,0 +1,650 @@
+! Copyright (c) 2021, The Neko Authors
```
njansson, line 1:
```suggestion
! Copyright (c) 2022, The Neko Authors
```

src/io/fluid_stats_output.f90
```diff
@@ -0,0 +1,99 @@
+! Copyright (c) 2021-2022, The Neko Authors
```
njansson, line 1:
```suggestion
! Copyright (c) 2022, The Neko Authors
```
Or did we add this before?

src/io/mean_flow_output.f90
```diff
@@ -80,7 +80,12 @@ subroutine mean_flow_output_sample(this, t)
     real(kind=rp), intent(in) :: t
 
     if (t .ge. this%T_begin) then
+        call device_memcpy(this%mf%p%mf%x, this%mf%p%mf%x_d, this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
+        call device_memcpy(this%mf%u%mf%x, this%mf%u%mf%x_d,  this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
+       call device_memcpy(this%mf%v%mf%x, this%mf%v%mf%x_d,  this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
+       call device_memcpy(this%mf%w%mf%x, this%mf%w%mf%x_d,  this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
```
njansson, line 86:
```suggestion
       call device_memcpy(this%mf%p%mf%x, this%mf%p%mf%x_d, this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
       call device_memcpy(this%mf%u%mf%x, this%mf%u%mf%x_d,  this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
       call device_memcpy(this%mf%v%mf%x, this%mf%v%mf%x_d,  this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
       call device_memcpy(this%mf%w%mf%x, this%mf%w%mf%x_d,  this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
```

src/case.f90
```diff
@@ -65,7 +67,9 @@ module case
      real(kind=rp), dimension(10) :: tlag
      real(kind=rp), dimension(10) :: dtlag
      type(sampler_t) :: s
+     type(sampler_t) :: sample_means
```
njansson, line 70: See comment in `simulation.f90` about multiple samplers.

src/field/field.f90
```diff
@@ -54,8 +54,20 @@ module field
      logical :: internal_dofmap = .false. !< Does the field have an own dofmap
      character(len=80) :: name            !< Name of the field
      type(c_ptr) :: x_d = C_NULL_PTR
+  contains
+    procedure, pass(f) :: init => field_init_external_dof
+    procedure, pass(f) :: free => field_free
```
njansson, line 59: Why do we change to type bound procedures here, as far as I can see they are not used throughout the code-base

njansson:
Would be great if also @adampep or @pschlatt1 also have look if this covers the necessary statistics. I've only reviewed the code structure etc.

timofeymukha:
Hello, here are some comments.

* I don't like the tight integration between the statistics and `fluid_method`. It seems like only `init` is called there, and I would perhaps just move that outside, maybe directly to `simulation`. Later, I am hoping we can fit this into #722 , which is one of the reasons I started that discussion now.
* There is a fundamental issue with having a preset number of fields. This is a good start for us now, so that our group can do the standard simulations we are used to and get the needed statistics, but.
  - For 99% of people out there, anything beyond 2nd order statistics is rarely needed. I would say even stuff like <p'v'> is a rare beast.
  - On the other hand, these 44 fields will not cover everything. For example, I compute <t'v'> right now for ABLs, where t is temperature. This is the reason why everyone has their own little version of the statistics code now in our group for Nek :-). So there needs to eventually be a mechanism to easily add a field to statistics. 
Something like this in psuedo-json in the par file
```
statisitcs
{
    fields = ["t", "tv"] # will grab from registry by name, "tv" should be computed in usercheck
    orders = [2, 1] # will compute mean and mean of square for t and only mean for tu.
}
```
* I think there is definitely room for a "prepackaged" set of values like here, because putting everything in manually is tedious. A question here is whether to just compute all the raw statistics or e.g. directly the k budget. We currently do it one way for 2d stats and the other for 3d in Nek :-). Perhaps @pschlatt1 and @adampep could comment on this.

src/simulation.f90
MartinKarp: Yes, but I also want to make sure to sample when the simulation ends, otherwise it is possible to lose a lot of data.

src/simulation.f90
```diff
@@ -117,6 +117,7 @@ subroutine neko_solve(C)
        call neko_log%section('Postprocessing')       
        call C%q%eval(t, C%params%dt)
        call C%s%sample(t)
+       call C%sample_means%sample(t)
```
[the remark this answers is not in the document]
MartinKarp: I think this will be rather difficult as the conditions to sample can be very different between different fields, I think it makes sense to kind of group things in each sampler, with similar conditions for nowe, but maybe thats me.

src/field/field.f90
MartinKarp: I think we should ahve type bound procedures for every class for init and so on. I just feel that would make things more consistent.

MartinKarp:
Yeah, we can move the init. The framework structure seems like a good idea, but would each simulationobject or so also handle all things related to it such as io and so on?

Yeah, we should have a simple way to add statistics. Right now this requires quite some work for it to work. Here I mostly wanted to just output everything so that the dqata from the simulations on LUMI are dumped to file.

timofeymukha:
> Yeah, we can move the init. The framework structure seems like a good idea, but would each simulationobject or so also handle all things related to it such as io and so on?
> 
> Yeah, we should have a simple way to add statistics. Right now this requires quite some work for it to work. Here I mostly wanted to just output everything so that the dqata from the simulations on LUMI are dumped to file.

Difficult to say about IO, since there is so may factors, like formats and file-sizes. But generally I would say each component should be quite independent and use the registry as the communication point.

src/field/field.f90
njansson: Yes, that's a good point, but we should make sure that init is overloaded for both external and internal dof such that we don't break e.g. `examples/poisson`

timofeymukha:
@MartinKarp could you please add a short module documentation to src/io/fld_file_data.f90?

MartinKarp:
Yeah, will do. I am aiming to have this branch "finished" by the end of the week.

MartinKarp:
I have now added two postprocessing tools. 

`average_fields` which takes a series of fluid snapshots and computes the temporal average of the fields stored in the files.

The other is `postprocess_fluid_stats` which takes three files as input: the mesh, a fluid file with the temporal average of the flowfield and a file with the temporal averages outputted by the fluid_stats module. 

Currently, we only post-process and output the reynolds stresses and the gradients of the mean flow field, but as we collect many other moments during the simulation the postprocessing can be easily extended to other statistical quantities that are used in the original KTH stats toolbox.

MartinKarp:
Hi all!
I have updated the PR a bit, particularly regarding how we sample/write/output fields of interest. There are a two things that need to be fixed before merging though.

- [ ] Testing, both on GPU and CPU and on some different systems.
- [ ] Where should field_list_t be? I have left it in field_t as I quite like that the functionality needed to create an array of pointers to fields is included when loading the field module

There are probably many other issues, but please have a look and say what you think.

src/field/mean_field.f90
```diff
@@ -37,34 +37,42 @@ module mean_field
...
+       name = field_name
+    else 
+       write(name, '(A,A)') 'mean_',trim(f%name)
+    end if
 
-    name = 'mean_'//trim(f%name)
-
-    call field_init(this%mf, f%dof, name)
+    call neko_field_registry%add_field(f%dof, name)
+    this%mf => neko_field_registry%get_field(name)
```
njansson, line 75: Is there a reason for using the registry here? Do we need access to this field outside of a case?

src/field/mean_field.f90
```diff
@@ -75,11 +83,26 @@ subroutine mean_field_free(this)
...
+  !> Resets a mean field
+  subroutine mean_field_reset(this)
+    class(mean_field_t), intent(inout) :: this
+    
+    this%time = 0.0
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_rzero(this%mf%x_d,size(this%mf%x))
+    else
+       call rzero(this%mf%x,size(this%mf%x))
+    end if
```
njansson, line 102:
```suggestion
    this%mf = 0.0_rp  
```

src/fluid/fluid_method.f90
```diff
@@ -55,6 +55,7 @@ module fluid_method
   use non_normal
   use krylov_fctry
   use precon_fctry
+  use fluid_stats
```
njansson, line 58: Related to @timofeymukha  earlier comment, should we really have statistics as part of the fluid scheme?

src/field/field.f90
```diff
@@ -54,8 +54,21 @@ module field
      logical :: internal_dofmap = .false. !< Does the field have an own dofmap
      character(len=80) :: name            !< Name of the field
      type(c_ptr) :: x_d = C_NULL_PTR
+  contains
+    generic :: init => field_init_external_dof, field_init_internal_dof
+    procedure, pass(f) :: free => field_free
+    procedure :: field_init_external_dof, field_init_internal_dof
```
njansson, line 60: This depends on what the decision will be on #794, thus not sure we should introduce these changes now in this pr.

MartinKarp: It does not remove the previous functionality, so in my mind I think it is. Otherwise I can change it back and change the inits in fluid_stats.

src/fluid/fluid_method.f90
MartinKarp: I think for these pre-set values that correspond to the KTH toolbox it is ok as they are tightly coupled to the fluid method. When we add userstats I think it makes sense to have them in simulation/case instead.

src/field/mean_field.f90
MartinKarp: Yes, I think I use the field_registry quite extensively to obtain the mean field values and such. It might also be convenient when we start adding user-defined stats to make sure that the mean fields can be accessed in the registry.

MartinKarp: After thinking about it a bit I am more divided whether it is correct to add to the registry.

MartinKarp: I am fixing this

src/common/sampler.f90
```diff
@@ -113,82 +110,188 @@ subroutine sampler_free(this)
...
+       tmpi(1:this%size) = this%nsample_list
+       call move_alloc(tmpi, this%nsample_list)
        this%size = this%size * 2
     end if
 
     this%n = this%n + 1
+    n = this%n
     this%output_list(this%n)%outp => out
     
+    call neko_log%message('---Adding write output---')
```
njansson, line 166: neko_log%section is the way to go

src/case.f90
```diff
@@ -281,38 +284,45 @@ subroutine case_init(C, case_file)
...
     ! Setup sampler
     !
-    call C%s%init(C%params%nsamples, C%params%T_end)
+    call C%s%init(C%params%T_end)
     C%f_out = fluid_output_t(C%fluid, path=C%params%output_dir)
-    call C%s%add(C%f_out)
-
+    if (trim(C%params%fluid_write_control) .eq. 'org') then
+       call C%s%add(C%f_out, real(C%params%nsamples,rp),'nsamples')
+       call neko_warning('nsamples parameter is legacy, please use X_write_par and X_write_control instead')
```
njansson, line 291: TODO, we should also add this to the deprecated doxygen page

src/common/sampler.f90
MartinKarp: I will test, but as the outputs are added at different "levels" it made things look kind of weird before.

njansson: 🍻 Don't forget to call `neko_log%end_section()` otherwise the indention will look very strange.

If it still doesn't look "nice", we can add an optional "no indent" argument to section (such that we get the same `------` banner size everywhere.

src/case.f90
MartinKarp: Now or later?

src/common/sampler.f90
```diff
@@ -113,82 +110,189 @@ subroutine sampler_free(this)
...
+       tmpi(1:this%size) = this%nsample_list
+       call move_alloc(tmpi, this%nsample_list)
        this%size = this%size * 2
     end if
 
     this%n = this%n + 1
+    n = this%n
     this%output_list(this%n)%outp => out
     
+    call neko_log%section('---Adding write output---')
```
njansson, line 166:
```suggestion
    call neko_log%section('Adding write output')
```

THe `---` stuff is added automagically by the sampler

src/case.f90
njansson: Would be nice now, such that we don't forget it

src/common/parameters.f90
```diff
@@ -80,6 +80,14 @@ module parameters
      real(kind=rp) :: user(16)           !< User defined parameters
      integer :: prs_max_iter           !< pressure max iterations
      integer :: vel_max_iter           !< velocity max iterations
+     logical :: stats_fluid        !< Fluid statistics
+     real(kind=rp) :: fluid_write_par   !< Parameter for writes of fld fields
```
njansson, line 84:
```suggestion
     real(kind=rp) :: fluid_write_par   !< Time interval for writes of fld fields
```
Isn't this a more correct description? Also for a release, we should not forget to update the doxygen page

src/common/parameters.f90
```diff
@@ -80,6 +80,14 @@ module parameters
...
+     character(len=20) :: fluid_write_control   !< How to control writes, simulationtime, nsamples, tsteps
+     real(kind=rp) :: stats_write_par   !< Time interval over which we compute averages before output
+     character(len=20) :: stats_write_control   !< Control output simulationtime, nsamples, tsteps
+     real(kind=rp) :: chkp_write_par   !< Time interval between checkpoints
+     character(len=20) :: chkp_write_control   !< Control output chkp, simulationtime, nsamples, tsteps
+     logical :: write_at_end !< Whether sampler should force output at end of simulation
```
njansson, line 90: Same as above, we should not forget to add this to the oxygen page

src/simulation.f90
```diff
@@ -126,8 +126,8 @@ subroutine neko_solve(C)
     end do
 
     call profiler_stop
-
-    if (t .lt. C%params%T_end) then
+    call C%s%sample(t,tstep,ifforce=C%params%write_at_end)
+    if (.not. (C%params%write_at_end) .and. t .lt. C%params%T_end) then
```
njansson, line 130: Is it correct that we should not support stopping a simulation after a given time period?

njansson (changes requested):

src/simulation.f90
MartinKarp: How do you mean? If the simulation stops after 25 t units it will finish then, and by default write a checkpoint and everything in the sampler. And same if we set at joblimit.

njansson: I mean if we set a job limit, or a signal is received to halt. But I guess that breaks sampling of stats, thus should not be supported?

MartinKarp: That should work the same. If we send a signal we will dump everything to file, assmuing write_at_end is true. So no data will be lost

njansson: Unless `write_at_end = .true.` then the job limit check will not be checked as far as I understands it?

MartinKarp: The the sampler will output it since ifforce will be true

njansson: Sure, but it will not catch signal or joblimit things since that's is not in the sampler. Maybe it's ok to accept data loss if one is using job limit

MartinKarp: I think so? Since ifforce is true, and we leave the time step loop it will output whatever is in the sampler, reegardless if it a signal, joblimit or the run simply finishes

njansson: Ok then data loss it is since signal or job limit is not handled by the sampler. fair enough

src/io/mean_flow_output.f90
```diff
@@ -80,7 +81,12 @@ subroutine mean_flow_output_sample(this, t)
     real(kind=rp), intent(in) :: t
 
     if (t .ge. this%T_begin) then
+       call device_memcpy(this%mf%p%mf%x, this%mf%p%mf%x_d, this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
+       call device_memcpy(this%mf%u%mf%x, this%mf%u%mf%x_d,  this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
+       call device_memcpy(this%mf%v%mf%x, this%mf%v%mf%x_d,  this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
+       call device_memcpy(this%mf%w%mf%x, this%mf%w%mf%x_d,  this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
```
njansson, line 87:
```suggestion
       call device_memcpy(this%mf%p%mf%x, this%mf%p%mf%x_d, this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
       call device_memcpy(this%mf%u%mf%x, this%mf%u%mf%x_d, this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
       call device_memcpy(this%mf%v%mf%x, this%mf%v%mf%x_d, this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
       call device_memcpy(this%mf%w%mf%x, this%mf%w%mf%x_d, this%mf%p%mf%dof%size(), DEVICE_TO_HOST)
```

src/simulation.f90
```diff
@@ -69,7 +69,7 @@ subroutine neko_solve(C)
     !> Call stats, samplers and user-init before time loop
     call neko_log%section('Postprocessing')       
     call C%q%eval(t, C%params%dt)
-    call C%s%sample(t)
+    call C%s%sample(t,tstep)
```
njansson, line 72:
```suggestion
    call C%s%sample(t, tstep)
```

njansson (changes requested):

src/simulation.f90
```diff
@@ -116,7 +116,7 @@ subroutine neko_solve(C)
 
        call neko_log%section('Postprocessing')       
        call C%q%eval(t, C%params%dt)
-       call C%s%sample(t)
+       call C%s%sample(t,tstep)
```
njansson, line 119:
```suggestion
       call C%s%sample(t, tstep)
```

njansson (approved):

njansson (approved):

timofeymukha (approved):

PR #724, opened 2022-11-18
Feature/template fdm

jonvin73:
Completed AMD tests, gives about 30% on CUDA for TGV, and 50% on AMD testing on 50 timesteps.

### DIFF
--- a/src/math/bcknd/device/cuda/fdm_kernel.h
+++ b/src/math/bcknd/device/cuda/fdm_kernel.h
@@ -32,105 +32,104 @@
  POSSIBILITY OF SUCH DAMAGE.
 */
 
-template< typename T >
+template< typename T, int NL >
 __global__ void fdm_do_fast_kernel(T  * __restrict__ e,
                                    T * __restrict__ r,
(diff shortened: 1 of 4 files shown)


src/math/bcknd/device/cuda/fdm_kernel.h
```diff
@@ -32,105 +32,104 @@
  POSSIBILITY OF SUCH DAMAGE.
 */
 
-template< typename T >
+template< typename T, int NL >
```
njansson, line 35:
```suggestion
template< typename T, const int NL >
```

src/math/bcknd/device/hip/fdm_kernel.h
```diff
@@ -32,105 +32,104 @@
  POSSIBILITY OF SUCH DAMAGE.
 */
 
-template< typename T >
+template< typename T, int NL >
```
njansson, line 35:
```suggestion
template< typename T, const int NL >
```

njansson (changes requested):

njansson (approved):

PR #725, opened 2022-11-22
Scalar checkpoint

MartinKarp:
Adds a scalar as an optional field in a checkpoint. Maybe we should consider moving chkp up from fluid into case. In some sense we save a case rather than only the fluid fields. 

There is still a risk that one sets incorrect ics for scalars as we do not ensure that there are no jumps between elements, but before addressing that we maybe should consider how we would like to handle user defined ics and bcs for multiple fields.

### DIFF
--- a/src/common/checkpoint.f90
+++ b/src/common/checkpoint.f90
@@ -54,12 +54,15 @@ module checkpoint
      type(field_series_t), pointer :: vlag => null()
      type(field_series_t), pointer :: wlag => null()
 
+     type(field_t), pointer :: s => null()
+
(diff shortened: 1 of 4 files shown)


src/common/checkpoint.f90
```diff
@@ -197,6 +208,16 @@ subroutine chkp_add_lag(this, ulag, vlag, wlag)
     
   end subroutine chkp_add_lag
 
+  !> Add lagged velocity terms
```
njansson, line 211:
```suggestion
  !> Add scalars
```

njansson (changes requested):

njansson (approved):

PR #726, opened 2022-11-23
Larger mesh support

MartinKarp:
This fixes issues with generating nmsh files for meshes with more than 8 or 9 Million elements.

### DIFF
--- a/src/io/nmsh_file.f90
+++ b/src/io/nmsh_file.f90
@@ -46,7 +46,7 @@ module nmsh_file
   private
   !> Specifices the maximum number of elements any rank is allowed to write (for nmsh).
   !! Needed in order to generate large meshes where an individual write might exceed 2GB.
-  integer :: max_write_nel = 8000000 
+  integer, parameter :: max_write_nel = 8000000 
   !> Interface for Neko nmsh files
   type, public, extends(generic_file_t) :: nmsh_file_t
    contains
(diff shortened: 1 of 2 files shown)


src/io/nmsh_file.f90
```diff
@@ -44,7 +44,9 @@ module nmsh_file
   implicit none
   
   private
-
+  !> Specifices the maximum number of elements any rank is allowed to write (for nmsh).
+  !! Needed in order to generate large meshes where an individual write might exceed 2GB.
+  integer :: max_write_nel = 8000000
```
njansson, line 49:
```suggestion
  integer, parameter :: max_write_nel = 8000000 
```

src/io/re2_file.f90
```diff
@@ -417,7 +417,7 @@ subroutine re2_file_read_curve(msh, ncurve, dist, fh, mpi_offset, v2_format)
           end do
        end do
     end do
-  
+    write(*,*) 'reading curved data'
```
njansson, line 420: Can't we do this with neko_log

njansson (changes requested):

src/io/re2_file.f90
MartinKarp: We use write everywhere else in re2

MartinKarp: Maybe they should all be changed

njansson: Ok, should probably change, but in a different pr of course

njansson (approved):

PR #742, opened 2022-12-08
Be nice to the memory during init/resize

njansson:
Reduce the amount of calls to allocate during init/resize operations.

The new structure should also improve cache performance.

### DIFF
--- a/src/adt/htable.f90
+++ b/src/adt/htable.f90
@@ -328,9 +328,9 @@ contains
        end if
        !> Check if entry at this index is empty or if key matches
        if ((.not. this%valid(index)) .or. &
-            htable_eq_key(this, index, key)) then
-          call htable_set_key(this, index, key)
-          call htable_set_data(this, index, data)
+            htable_eq_key(this%key(index), key)) then
+          call htable_set_key(this%key(index), key)
+          call htable_set_data(this%data(index), data)
           if (.not. this%valid(index)) then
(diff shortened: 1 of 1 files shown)


src/adt/htable.f90
```diff
@@ -43,19 +43,15 @@ module htable
...
   !> Base type for a hash table
   type, public, abstract :: htable_t
      integer, private :: size
      integer, private :: entries
-     type(h_tuple_t), private, allocatable :: t(:)
+     class(*), private, allocatable :: key(:)
+     class(*), private, allocatable :: data(:)
+     logical, private, allocatable :: valid(:) 
+     logical, private, allocatable  :: skip(:)
+     !     type(h_tuple_t), private, allocatable :: t(:)
```
njansson, line 54:
```suggestion
```

src/adt/htable.f90
```diff
@@ -283,8 +284,8 @@ end subroutine htable_free
   subroutine htable_clear(this)
     class(htable_t), intent(inout) :: this
 
-    if (allocated(this%t)) then
-       this%t(:)%valid = .false.
+    if (allocated(this%data)) then
+       this%valid = .false.
```
njansson, line 288:
```suggestion
    if (allocated(this%data)) then
       this%valid = .false.
       this%skip = .false.
```

PR #764, opened 2023-01-26
Feature/read fld files

MartinKarp:
Initial draft for reading fld files. I take the .nek5000 file into consideration, but one can also read a series of fld files by specifying the metadata manually.

### DIFF
--- a/src/io/fld_file.f90
+++ b/src/io/fld_file.f90
@@ -47,6 +47,7 @@ module fld_file
   use mpi_types
   use mpi_f08    
   implicit none
+  private
   
 
 
@@ -707,6 +708,10 @@ contains
           x%x(i) = tmp_sp(i) 
        end do     
     end if
+
+    if (allocated(tmp_dp)) deallocate(tmp_dp)
+    if (allocated(tmp_sp)) deallocate(tmp_sp)
+
   end subroutine fld_file_read_field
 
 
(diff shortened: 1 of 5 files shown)


src/io/fld_file.f90
```diff
@@ -46,7 +47,7 @@ module fld_file
   use mpi_types
   use mpi_f08    
   implicit none
-  private
```
njansson, line 49: Why do we need this to be public, can't it be avoided?

src/io/fld_file.f90
```diff
@@ -450,11 +451,329 @@ end subroutine fld_file_write
...
+
+    if (this%dp_precision) then
+       allocate(tmp_dp(n))
+       call MPI_File_read_at_all(fh, byte_offset, tmp_dp, n, &
+            MPI_DOUBLE_PRECISION, status, ierr)
+    else
+       allocate(tmp_sp(n))
+       call MPI_File_read_at_all(fh, byte_offset, tmp_sp, n, &
+            MPI_REAL, status, ierr)
+    end if
```
njansson, line 699: These allocated arrays. even if they are freed at some point, I would rather se a `if allocated` then `deallocate` at the end of the routine

src/io/fld_file.f90
```diff
@@ -450,11 +451,329 @@ end subroutine fld_file_write
...
+
+    if (this%dp_precision) then
+       allocate(tmp_dp(fld_data%gdim*n))
+       call MPI_File_read_at_all(fh, byte_offset, tmp_dp, nd, &
+            MPI_DOUBLE_PRECISION, status, ierr)
+    else
+       allocate(tmp_sp(fld_data%gdim*n))
+       call MPI_File_read_at_all(fh, byte_offset, tmp_sp, nd, &
+            MPI_REAL, status, ierr)
+    end if
```
njansson, line 736: Same comment regarding allocate here as above

njansson (changes requested):

njansson (approved):

PR #786, opened 2023-02-16
Add header file with configuration information

njansson:


timfelle:
Isn't this already implemented?
https://github.com/ExtremeFLOW/neko/blob/develop/src/config/neko_config.f90.in

njansson:
> @njansson Isn't this already implemented?
> 
> https://github.com/ExtremeFLOW/neko/blob/develop/src/config/neko_config.f90.in

For Fortran yes, this was for the C/C++ side, but it was not playing well so we solved it in a different way

PR #799, opened 2023-03-14
Max iter

MartinKarp:


timofeymukha:
We also have a Doxygen page that lists the case parameters, so let's add it there as well.

MartinKarp:
Doesnt really matter with the basic iter IMO, if you don't converge in 800, something is off in any case.
I will add it to a case, and good idea with the Doxygen. I have however never added anything there before and it seems to be rather outdated. I would suggest changing that page to instead point to the doxygen page for the parameter module, which always will be updated.

timofeymukha:
> Doesnt really matter with the basic iter IMO, if you don't converge in 800, something is off in any case. I will add it to a case, and good idea with the Doxygen. I have however never added anything there before and it seems to be rather outdated. I would suggest changing that page to instead point to the doxygen page for the parameter module, which always will be updated.

Yeah, I agree, 800 just looked suspiciously specific :-). But it doesn't matter. I agree regarding Doxygen, pointing to the parameter module is probably the best solution, at least while things are in flux.

So, just to an example, and we merge!

MartinKarp:
added it to the lid example

timofeymukha (approved):

PR #804, opened 2023-03-16
Safe user module finalization

JudithYi:
Some user modules also need safe finalization. For instance, it is necessary to close the engine while using ADIOS2 or free user allocated arrays.


adperezm:
Should we close this pull request? This was implemented in another one and has been in the code for a while now

njansson:
Added in #879

PR #805, opened 2023-03-16
appends name of the case to neko bin

MaansAndersson:
Maybe cut is not the best - could change to awk if preferred.


MartinKarp:
very good, bash magic

timofeymukha:
I've missed out on the discussion on this. Can I ask what's the rationale here?

njansson:
> I've missed out on the discussion on this. Can I ask what's the rationale here?

Something that came up during the hackathon, appending the case name to neko to distinguish it from the standalone model (driven by a case file alone).

Now after some time has passed, I'm not sure about this either

timofeymukha:
I guess I can see how one would sometimes type `neko` instead of `./neko`!

njansson:
> maybe stupid point, but how often would one really use a "vanilla" neko version for a real case?

A valid question, I think it would be great if most things could be tweaked from the case file (or a python driver), no need to compile etc, and one can guarantee that a release is ready for science.

timofeymukha:
> > maybe stupid point, but how often would one really use a "vanilla" neko version for a real case?
> 
> A valid question, I think it would be great if most things could be tweaked from the case file (or a python driver), no need to compile etc, and one can guarantee that a release is ready for science.

That is really my hope as well. I think compilation should in the long run be an exception rather then rule.

P.S. Please `userneko`, not `usrneko` :-P

MartinKarp:
Userneko sounds nice, but as the intent is not to mix up different executables I also think the original idea with neko_CASE makes quite a lot of sense

pschlatt1:
I would clearly vote against userneko. neko_CASE is certainly an option, but I just foresee people running scripts which call neko and don't realise that they don't use their own executable. I would simply go with neko...

timofeymukha:
> I would clearly vote against userneko. neko_CASE is certainly an option, but I just foresee people running scripts which call neko and don't realise that they don't use their own executable. I would simply go with neko...

Wait, but wasn't the name change intended to address just that? But I guess if you run the central neko on a case that should use a local one, you just immediately crash, right? So it's not like you accidentally spend half your allocation or something.

Should we maybe just wait for now and see how big of an issue this is in practice?

njansson:
> > I would clearly vote against userneko. neko_CASE is certainly an option, but I just foresee people running scripts which call neko and don't realise that they don't use their own executable. I would simply go with neko...
> 
> 
> 
> Wait, but wasn't the name change intended to address just that? But I guess if you run the central neko on a case that should use a local one, you just immediately crash, right? So it's not like you accidentally spend half your allocation or something.
> 
> 
> 
> Should we maybe just wait for now and see how big of an issue this is in practice?

Mixing executables is one thing, the other one is QA, at least we must have a flag or something to distinguish our (verified) neko from user build ones, as discussed at the hackathon.

MaansAndersson:
I would say that QA would be helped more by the checksum and a dirty/official message in the log files (see GROMACS) future PR. This was mostly a way of making naming more intuitive for the user.

timfelle:
I realized the user interface is not a dependency for the `logger`module. So there was no reason to export it.
We now print a user interface section to the log file with a list of the modified parts.

<details>
<summary>Example from TGV</summary>

```
   -----Setup user interface-----  
   - Fluid initial condition
   - Mesh setup
   - User check
   - Initialize modules
   - Finalize modules
   End of User defined interface
```

</details>

njansson (commented):
I would suggest that we don't print anything `user` if we run `turboneko` only for `usrneko`, not even the section header

timfelle:
> I would suggest that we don't print anything `user` if we run `turboneko` only for `usrneko`, not even the section header

I was trying to do that. Problem was i needed to build a dynamic array of the modified pieces of code, meaning i needed add quite a lot of extra code everywhere in the function. But i might have a solution.

timfelle:
> I would suggest that we don't print anything `user` if we run `turboneko` only for `usrneko`, not even the section header

Done

njansson (approved):
Happy days!

timofeymukha (approved):

PR #809, opened 2023-03-23
Better temporaries handling in fluid

timofeymukha:
Towards removing temporaries in fluid_pnpn.  WIP, but things seem to work. I don't like that one does a linear search through the scratch registry 8 times per time-step. But this should not be overthought, since the direction is to change the interfaces of the stuff called inside `step` in a way that I can just pass the registry to them.

### DIFF
--- a/src/case.f90
+++ b/src/case.f90
@@ -58,6 +58,7 @@ module case
   use jobctrl
   use user_intf  
   use scalar_pnpn ! todo directly load the pnpn? can we have other
+  use scratch_registry, only : scratch_registry_t, neko_scratch_registry
   implicit none
 
   type :: case_t
@@ -183,6 +184,12 @@ contains
     !
     call fluid_scheme_factory(C%fluid, trim(fluid_scheme))
     call C%fluid%init(C%msh, lx, C%params)
+
+    
+    !
+    ! Setup scratch registry
+    !
+    neko_scratch_registry = scratch_registry_t(C%fluid%dm_Xh, 10, 10)
 
     !
     ! Setup scalar scheme
--- a/src/common/bcknd/cpu/rhs_maker_cpu.f90
+++ b/src/common/bcknd/cpu/rhs_maker_cpu.f90
@@ -1,5 +1,6 @@
 module rhs_maker_cpu
   use rhs_maker
+  use scratch_registry
   implicit none
   private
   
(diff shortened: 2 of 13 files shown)


timofeymukha:
@njansson Now using global scratch. This ties the global scratch to a particular case with a particular dofmap. But perhaps we can live with that for now, probably a we somehow have fields on different dofmaps in the simulation. If this occurs we can probably extend the registry to have multiple arrays, corresponidng to scratch fields on different dofs, or something like that.

timofeymukha:
CI's acting up...

njansson:
> CI's acting up...

Should be fixed now.

njansson:
> @njansson Now using global scratch. This ties the global scratch to a particular case with a particular dofmap. But perhaps we can live with that for now, probably a we somehow have fields on different dofmaps in the simulation. If this occurs we can probably extend the registry to have multiple arrays, corresponidng to scratch fields on different dofs, or something like that.

Ok! I think we can live with that for now. One concern I've however, is if the `pointer` attribute will prevent vectorisation, we need to check the compiler listings. At least we should add the `contiguous` attribute to be on the safe side?

timofeymukha:
Sorry @njansson, I can't quite figure out where I should add the `contiguous`. It should be a pointer to an array, right?

timofeymukha:
@njansson A kind reminder of the above request, sorry.

njansson:
> @njansson A kind reminder of the above request, sorry.

Sorry my bad, we don't need `contiguous`, since the pointer is on a derived type. I checked listing for both CCE and NEC and everything vectorises as expected.

src/case.f90
```diff
@@ -179,6 +180,11 @@ subroutine case_init(C, case_file)
     !
     call fluid_scheme_factory(C%fluid, trim(fluid_scheme))
     call C%fluid%init(C%msh, lx, C%params)
+    
+    !
+    ! Setup scratch registry
+    !
+    neko_scratch_registry = scratch_registry_t(C%fluid%dm_Xh, 20, 10)
```
njansson, line 187: Do we need as many as 20 scratch fields?

src/fluid/fluid_method.f90
```diff
@@ -93,6 +94,7 @@ module fluid_method
      type(chkp_t) :: chkp                      !< Checkpoint
      type(mean_flow_t) :: mean                 !< Mean flow field
      type(mean_sqr_flow_t) :: mean_sqr         !< Mean squared flow field
+     type(scratch_registry_t) :: scratch      !< Manager for temporary fields
```
njansson, line 97:
```suggestion
     type(scratch_registry_t) :: scratch       !< Manager for temporary fields
```

timofeymukha:
I guess the current implementation does not reduce the memory size used, because once a temp field is used for the first time, it never gets deallocated. Since the whole point of the scratch was to reduce the necessity of allocating new fields. So basically, for each temporary we exchange the necessity to allocate for having to fetch a field from the registry (which requires a search for an unused field inside it)

In light of what you said @MartinKarp , I very much agree with "it might just be better for each object to own its temporaries". So, I would perhaps add registries to the `rha_maker` and the residual classes instead of using the global one. I originally envisioned the global one to be used in stand-alone subroutines, like the 2 work arrays in `curl`, for example.

njansson (approved):

PR #813, opened 2023-04-01
Scratch registry improvements

timofeymukha:
This cherry-picks commits pertaining to the scratch registry from #809, so that #809 does not become a monster PR.

- Adds possibility to relinquish several fields at once by passing an array of indices.
- Adds a test for the above.
- Fixes a bug in the destructor, now checks if stuff has been allocated before attempting to free it.


timofeymukha:
Intel acting up again :-(.

njansson (approved):

PR #819, opened 2023-04-04
Harmonize file, module and class name for fluid_scheme and scalar_scheme

timofeymukha:
- I'm not sure if this is the best possible name, but at least everything is consistent.
- So if we decide to refactor again, it will be easier.
- There was a small clash with interface procedure names, so I had to add `_interface` to those.


timofeymukha:
Are there objections to this?

njansson:
> Are there objections to this?

Not sure about adding `_interface`, otherwise `method` -> `scheme` I'm all positive for

timofeymukha:
> > Are there objections to this?
> 
> Not sure about adding `_interface`, otherwise `method` -> `scheme` I'm all positive for

Yeah, something simply had to be done to avoid name collisions. And since it purely internal use, I just took the most obvious approach. But it could be good with a canonical way of marking interface procedures to make naming streamlined. Would you prefer some abbreviation, like `_intrf`?

njansson:
> > > Are there objections to this?
> > 
> > 
> > Not sure about adding `_interface`, otherwise `method` -> `scheme` I'm all positive for
> 
> Yeah, something simply had to be done to avoid name collisions. 

Hmm, but what was the collision? I tried to figured it out from the diff but I can't see it

>And since it purely internal use, I just took the most obvious approach. But it could be good with a canonical way of marking interface procedures to make naming streamlined. Would you prefer some abbreviation, like `_intrf`?

If we have to, I would prefer a shorter one, such that we can keep col 80.

MartinKarp:
I think it looks nice. I like that we move to scheme and use it constistently

njansson (approved):

PR #820, opened 2023-04-04
Cleanup of bdf_time_scheme_t; adding its docs and tests

### LINKED ISSUE #114: Abstract BDF type
In order to be able to use different BDF schemes a new BDF type is necessary with subroutines such as 'bdf%apply'

timofeymukha:
- The computation of the coeffs for the BDF scheme no longer uses a messy linear solver. Instead, they are computed using explicit formulas, taken from a [technical note by Nishikawa](https://www.researchgate.net/publication/351082535_Derivation_of_BDF2BDF3_for_Variable_Step_Size). 
- The above fixes the maximum order to 3, which anyway is the maximum used. 
- However, we can easily extend to higher order later if we want, because the subroutine `fd_weights_full` that we have can be used to compute the coefficients for arbitrary order. I am surprised Nek5000 did not use that, actually! It requires a bit of array reordering, etc, so for now I went for the explicit formulas.
- Quite detailed documentation for the BDF scheme is provided. There are two somewhat surprising conventions one should be aware of.
- Added tests for constant and variable dt cases.

I would say this closes #114 for now.

### DIFF
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -140,7 +140,7 @@ jobs:
         if: matrix.backend == 'cpu' && failure()
         uses: actions/upload-artifact@v2
         with:
-          name: Test report / ${{ matrix.os }} / ${{ matrix.compiler }}
+          name: Test report_${{ matrix.os }}_${{ matrix.compiler }}
           path: tests/test-suite.log
           retention-days: 2
       - name: Dist (CPU backend)
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -136,6 +136,13 @@ jobs:
         if: matrix.backend == 'cpu'
         run: |
           make -j$(nproc) check
+      - name: Archive test report
+        if: matrix.backend == 'cpu' && failure()
+        uses: actions/upload-artifact@v2
+        with:
+          name: Test report / ${{ matrix.os }} / ${{ matrix.compiler }}
+          path: tests/test-suite.log
+          retention-days: 2
       - name: Dist (CPU backend)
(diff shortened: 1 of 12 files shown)


timofeymukha:
Well, that's unfortunate. @njansson can I somehow see exactly what fails in the test?

njansson:
> Well, that's unfortunate. @njansson can I somehow see exactly what fails in the test?

Unfortunately not, via GitHub you can only see which test failed, but we don't keep any of the more detailed logs (which we should fix...)

I would rerunning the tests locally on check what breaks.

njansson:
> By the way, the old messy implementation is moved to the test file, so that we can use it to test the new stuff.

Given the discussions we had earlier in #114, can't we just fix the dimensions to 3 or similar?

timofeymukha:
> > By the way, the old messy implementation is moved to the test file, so that we can use it to test the new stuff.
> 
> Given the discussions we had earlier in #114, can't we just fix the dimensions to 3 or similar?

Yes, makes sense, particularly since `rhs_maker`s kernels are hard-coded to take 3 values.

timofeymukha:
Let's see if our new action works :-)

timofeymukha:
It's still the old yaml file :-(. @njansson do we actually have to merge the yaml to master?

njansson:
> It's still the old yaml file :-(. @njansson do we actually have to merge the yaml to master?

It uses the develop yaml of the to be merged branch, thus you need to cherry pick the changes into this branch

timofeymukha:
image

njansson (approved):

PR #825, opened 2023-04-07
Simulation component framework

timofeymukha:
* This is a blueprint of the simulation component structure, which was discussed in #722 .
* The blueprint relies on a JSON-based case file, although in principle one could try to fit this in the format now, but with quite some pain.
* I was scared that we cannot have an allocatable array of polymorphic objects, but [we can](https://community.intel.com/t5/Intel-Fortran-Compiler/Array-of-polymorphic-objects/td-p/809798), and it is used here.
* What we need to think about is what kind of methods we need. Right now, besides for `init` and `free`, I only have one `compute` routine that runs at the end of every timestep. I think we need to rely on the experience of @adampep with the KTH_FRAMEWORK. What do you think, Adam? One can for example have one routine that runs in the beginning of the time-step and one at the end.
* I have not thought about simulation components with IO, because I have not yet looked into the structure of IO classes. I don't envision much issues, but rather decisions like "output to a separate file" vs "output to the same file as u,v,p, etc."


MartinKarp:
This seems really nice. About I/O, I envision more issues such as how often will it sample, what frequency, and how will it handle if we restart, I think most things will want their own counter compared to how often we output fields. This is something I think we should discuss how we want to do with sampling quite soon as this is something that should be rather easy to add and specify from the user side.

timofeymukha:
I've updated this today to an actual implementation, input is welcome.
* One simcomp for computing vorticity, besides for the base class.
* Used in the `tgv` example.

One question is what to pass to the abstract-interfaced constructor besides for the JSON. I started with just JSON, but then realized I need the  `coef_t` to compute the curl. So, I need `fluid_scheme` to pull it out of there => added fluid to the constructor and as a pointer attribute. But I can easily imagine that some simcomp might, e.g. need the Pr number. So suddenly either we pass `scalar_scheme`, or we pass the *whole* JSON. I don't know, given the generic nature of simcomps, one can make a case for just passing it the `case`... but somehow that always feels like a suboptimal solution :-). Maybe, the whole JSON, and the basic SEM classes (mesh, space, coef, dofmap).

njansson:
> I've updated this today to an actual implementation, input is welcome.
> 
> * One simcomp for computing vorticity, besides for the base class.
> * Used in the `tgv` example.
> 
> One question is what to pass to the abstract-interfaced constructor besides for the JSON. I started with just JSON, but then realized I need the `coef_t` to compute the curl. So, I need `fluid_scheme` to pull it out of there => added fluid to the constructor and as a pointer attribute. But I can easily imagine that some simcomp might, e.g. need the Pr number. So suddenly either we pass `scalar_scheme`, or we pass the _whole_ JSON. I don't know, given the generic nature of simcomps, one can make a case for just passing it the `case`... but somehow that always feels like a suboptimal solution :-). Maybe, the whole JSON, and the basic SEM classes (mesh, space, coef, dofmap).

I guess the only way to cover all possible options would be to pass `case`, agree not optimal, but it would probably make life easier moving forward.

njansson (approved):

Issue #794, opened 2023-03-10
Common format for all derived types


njansson:
### Discussed in https://github.com/ExtremeFLOW/neko/discussions/728

<div type='discussions-op-text'>

<sup>Originally posted by **njansson** November 28, 2022</sup>
(Based on the discussion in PR #720 )

Neko uses various ways of initialising objects, as shown below. Before the code grows too big, it would probably be a good idea to agree on a common format and make changes throughout fo 
r the next-next release (v0.6)

The various formats are shown below
```Fortran
type(apa_t) :: blargh

! F90 style
call apa_init(blargh, args)

! type bound procedures  
call blargh%init(args)

! type interface 
blargh = apa_t(args)
```
Overhead costs of type bound procedures aside, my biggest issue with enforcing all derived types to have `init()` methods is the somewhat reduced readability of the code's intent. Both the f90 and interface indicate what kind of type `myapa` is when calling the initialisation method. Any other thoughts?  
</div>

njansson:
As concluded in #728, we go with the type interface format `apa_t`

timofeymukha:
The coclusion is actually that we use `call blargh%init(args)`, and we are close to having it everywhere, but still some types remain.

PR #837, opened 2023-04-24
Module scope

### LINKED ISSUE #795: Module scope
Recent release of Cray Fortran (v15) exposes our sloppy module usage, throwing several warnings (`ftn-878`) about already associated modules in the same scope

njansson:
Restrict module scope to avoid indirect association in a scoping unit.

This removes most of the `ftn-878` warnings when compiling with `cce > 14` (#795)


timofeymukha:
A much-needed update in terms of best practices, I guess. So, I discern two guidelines here.
1. Always have default scope as `private` in modules, forcing to explicitly declare what is `public.`
2. Directly declare what you need from a module with `only`. There are still imports that don't use this, though. For example, for `utils`, sometimes it is imported completely, and sometimes only concrete stuff. Is there some logic behind this?

So you know why the other compilers are complaining?

timofeymukha:
> So far, I've only seen Cray throwing warnings about this, and I'm not sure what the standard says. But I still think it's good to fix these issues.

What I meant is why the CI is unhappy.

timofeymukha (approved):

njansson:
> > So far, I've only seen Cray throwing warnings about this, and I'm not sure what the standard says. But I still think it's good to fix these issues.
> 
> What I meant is why the CI is unhappy.

Missing interfaces in certain backends, after `private` has been added.

njansson:
Depends on #973 being merged

njansson:
Depends on #974 being merged

njansson:
Depends on #975 being merged

njansson:
In the spirit of #794, almost types follows the same format (type-bound) such that redundant module use can be kept at a minimum

PR #844, opened 2023-05-07
Output CPU type

njansson:
If supported by the platform, retrieve and output CPU type at job start

### DIFF
--- a/src/common/system.f90
+++ b/src/common/system.f90
@@ -36,6 +36,10 @@ module system
   implicit none
 
   interface
+  !> Interface to a C function to retrieve the CPU name (type).
+  !! @param name Stores the retrieved name. Should be a `character` of length 
+  !! `len` and `c_char` kind.
+  !! @param len The maximum anticipated length of the retrieved name.
      subroutine system_cpuid(name, len) &
           bind(c, name='system_cpuid')
        use, intrinsic :: iso_c_binding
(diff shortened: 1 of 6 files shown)


src/common/system.f90
```diff
@@ -0,0 +1,63 @@
...
+          bind(c, name='system_cpuid')
+       use, intrinsic :: iso_c_binding
+       type(c_ptr), value :: name
+       integer(c_int), value :: len
+     end subroutine system_cpuid
+   end interface
+
+ contains
+
+   !> Retrieve the systems' CPU name (type)
```
timofeymukha, line 49:
```suggestion
   !> Retrieve the systems' CPU name (type)
   !! @param name Stores the retrieved name.
```

src/common/system.f90
```diff
@@ -0,0 +1,63 @@
...
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+!
+!> Interface to system information routines
+module system
+  use, intrinsic :: iso_c_binding
+  implicit none
+
+  interface
```
timofeymukha, line 38:
```suggestion
  interface
  !> Interface to a C function to retrieve the CPU name (type).
  !! @param name Stores the retrieved name. Should be a `character` of length 
  !! `len` and `c_char` kind.
  !! @param len The maximum anticipated length of the retrieved name.
```

timofeymukha (commented):
Some suggested docstring improvements

src/common/system.f90
timofeymukha: By the way, is it systems' and not system's because there are somehow many systems?

PR #859, opened 2023-05-15
turbChannel -> turb_channel and remove nsamples from the .case

timofeymukha:
* Renames `turbChannel` to `turb_channel` following the convention for the other examples. Renames .f90 and .case files to be `turb_channel` also.
* Removes `nsamples` from the .case file since it is not used due to `fluid_write_control` being set.


MartinKarp:
Looking good

MartinKarp (approved):

timofeymukha:
CI needs a kick.

njansson (approved):

PR #864, opened 2023-05-22
Postpend output_dir with a forward-slash

### LINKED ISSUE #863: output_dir does not work properly
* Setting output_dir does not actually transfer the output to a directory of that name.
* Instead, it prepends all output files with the name. So there is like a forward-slash missing. 
* Adding / gives an error though, e.g. 
```
Fortran runtime error: Cannot open file 'outdir/bdry0.nek5000': No such file or directory
```

timofeymukha:
Fixes #863, although it is still up to the user to create the output directory on disk.


timofeymukha:
Oh wait, one should of course avoid the modification when the output_dir is empty.

njansson:
> Oh wait, one should of course avoid the modification when the output_dir is empty.

and when the parameter already contains a trailing `/`

njansson (approved):

PR #867, opened 2023-05-25
NEKO v0.6.0

njansson:


njansson:
Maybe we should wait for #867

njansson:
> Ah, I forgot about that one. OK, I can work on it today.

Thanks, now merged!

Anything else we should wait for before tagging this as done?

timofeymukha:
Not development-wise, I think.

PR #871, opened 2023-05-27
Added some implicit none

timofeymukha:
Took a few files and added implicit none to all procedures. Is that what we should aim for, or do we rely on propagation from module level? Somehow, I would like to have it everywhere.

### DIFF
--- a/src/adt/htable.f90
+++ b/src/adt/htable.f90
@@ -229,6 +229,7 @@ contains
 
   !> Initialize a hash table of type @a data
   subroutine htable_init(this, size, key, data)
+    implicit none
     class(htable_t), intent(inout) :: this
     integer, value :: size              !< Initial size of the table
     class(*), target, intent(in) :: key            !< Type of key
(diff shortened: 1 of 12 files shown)


njansson:
All procedures within a module are in the same scope, thus `implicit none` in the module will cover everything.
Most of the files in this pr has `implicit none` in the module except `neko.f90`, where I guess there is a discussion if one should provide an implicit `implicit none` for each user or do it per subroutine.

timofeymukha:
OK, so should we generally rely on module-level `implicit none`?

njansson:
> OK, so should we generally rely on module-level `implicit none`?

Yes I would say so. The question is how to deal with interfaces, these are not covered by the `implicit none` under the modules scope.

MartinKarp:
I also think it is enough with implicit none in the module.

timofeymukha:
So perhaps to interfaces we need to add it explicitly. Although it is not that often one would create something there by accident.

PR #879, opened 2023-06-02
Feature/user finalize

vbaconnet:
Added functionality to finalize modules started by the user.

An example would be that a user starts ADIOS2 in the user file and needs to close it to release the MPI ranks in asynchronous execution.

Examples TGV, LID have been modified to illustrate the use.

The naming convention is open to discussion.

### DIFF
--- a/src/common/user_intf.f90
+++ b/src/common/user_intf.f90
@@ -100,7 +100,7 @@ module user_intf
 
   !> Abstract interface for finalizating user variables
   abstract interface
-     subroutine userfinalize(t, u, v, w, p, coef, param)
+     subroutine user_finalize_modules(t, u, v, w, p, coef, param)
        import field_t
        import param_t
        import coef_t
@@ -112,7 +112,7 @@ module user_intf
        type(field_t), intent(inout) :: p
        type(coef_t), intent(inout) :: coef
        type(param_t), intent(inout) :: param
-     end subroutine userfinalize
+     end subroutine user_finalize_modules
   end interface
 
   type :: user_t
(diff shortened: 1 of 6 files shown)


njansson:
Isn't this similar to #804?

src/simulation.f90
```diff
@@ -134,7 +134,10 @@ subroutine neko_solve(C)
     if (.not. (C%params%write_at_end) .and. t .lt. C%params%T_end) then
        call simulation_joblimit_chkp(C, t)
     end if
-    
+
+    call C%usr%user_final_modules(t, C%fluid%u, C%fluid%v, C%fluid%w, &
+         C%fluid%p, C%fluid%c_Xh, C%params)
+
```
njansson, line 140: Wouldn't this allow for a user to deallocate the case's fluid fields?, which is a bit strange since we later call the case's `free` routine.

src/common/user_intf.f90
```diff
@@ -98,11 +98,29 @@ subroutine usercheck(t, tstep, u, v, w, p, coef, param)
...
+       import rp
+       real(kind=rp) :: t
+       type(field_t), intent(inout) :: u
+       type(field_t), intent(inout) :: v
+       type(field_t), intent(inout) :: w
+       type(field_t), intent(inout) :: p
+       type(coef_t), intent(inout) :: coef
+       type(param_t), intent(inout) :: param
+     end subroutine user_finalize_modules
+  end interface
```
njansson, line 116: Not sure it's a good idea to pass `u,v,w,p,coef` here since that would allow for strange deallocations. For init modules they are needed to have all the necessary data.

njansson (commented):
The main question is if we should expose, `u,v,w,p,coef` in finalise, allowing for potential badness

timofeymukha:
I agree with Niclas that there is probably no need to pass flow variable to this function.

vbaconnet:
Removed `u,v,w,p` and flipped the names so that `user_finalize_modules` is the exposed one.

njansson (approved):

MartinKarp (approved):

Issue #831, opened 2023-04-17
JSON case file roadmap


timofeymukha:
Let's use this as a metaissue to track the progress towards a JSON parfile.

- [x] Add [the JSON library](https://github.com/jacobwilliams/json-fortran) to the build.
- [x] Resolve issues with `json-fortran` compilation with all necessary compilers.
- [x] Develop the JSON scheme for the case file. 
- [x] Add JSON-based `case_t` class. At least partially done [here](https://github.com/ExtremeFLOW/neko/blob/develop/contrib/pyneko/src/json_case.f90).
- [x] Fix all examples.
- [x] #891

timofeymukha:
@njansson If I remember correctly, introducing the JSON library somehow forced a change to using Neko as a .so instead of a static library? Or am I confusing things with Python... In any case, I fear that you will have to be assigned to the first task, unless @MartinKarp feels up to it. I still use the build system as a monkey, more or less.

njansson:
> @njansson If I remember correctly, introducing the JSON library somehow forced a change to using Neko as a .so instead of a static library? Or am I confusing things with Python... In any case, I fear that you will have to be assigned to the first task, unless @MartinKarp feels up to it. I still use the build system as a monkey, more or less.

No it will not force `libneko` to be a `.so` file instead, that was due to Python.

Adding the JSON library to `configure.ac` is as simple as

```
PKG_CHECK_MODULES([JSON_Fortran],[json-fortran >= 8.3.0])
LIBS="$JSON_Fortran_LIBS $LIBS"
FCFLAGS="$JSON_Fortran_CFLAGS $FCFLAGS"
```

njansson:
> But the above would rely on one already having installed `json-fortran` in a way that `pkg-config` finds it, right?
> 
> I guess the question is how we want to approach this, since `json-fortran` will be a hard dependency. Should building it be incorporated directly in Neko's build? Or should we maybe have a `thirdparty` directory with important deps as git submodules and associated build scripts?

I would strongly argue against having it as part of Neko's build, or in a `thirdparty` directory. Either of these is a recipe for disaster I would say.

timofeymukha:
[...]

PR #886, opened 2023-06-08
Change case file to json format

### LINKED ISSUE #858: JSON case file layout
Here is an attempt to produce a layout, which incorporates everything we have in the parameter file, except for old statistics stuff. It is difficult to define sometime what is general and what should be in e.g. `fluid`. I tried to make all the names as obvious as possible, without resorting to abbreviations or trying to save space. Please share your thoughts!

```json
{
  "general": {
    "mesh_file": "hemi.nmsh",
    "output_boundary": false,
    "output_partitions": false,
    "output_checkpoints": true,
    "output_at_end": true,
    "output_directory": "",
    "checkpoint_control": "simulationtime",
    "checkpoint_parameter": 10000, !we need a better name for these...
    "load_balance": false,
    "job_timelimit": "00:00:00",
    "restart_file": "",
    "end_time": 0.0,
    "timestep": 0.0
  },
  "numerics": {
    "time_order": 3,
    "polynomial_order": 7,
    "dealias": true,
    "dealiased_space_size": 0
  },
  "fluid": {
    "scheme": "pnpn",
    "rho": 1.0,
    "mu": 1.0,
    "Re": 1.0,
    "source_term": "noforce",
    "blasius": {
      "approximation": "sin",
      "freestream_velocity": [0.0, 0.0, 0.0],
      "delta": 1.0
    },
    "initial_condition": {
      "type": "blasius"
    },
    "inflow_condition": {
      "type": "default"
    },
    "outflow_condition": {
      "type": "dong",
      "velocity_scale": 1.0,
      "delta": 0.01
    },
    "boundary_types": ["w", "v"],
    "flow_rate_force": {
      "direction": 0,
      "rate": 0.0,
      "use_averaged_flow": true
    },
    "velocity_solver": {
      "type": "cg",
      "preconditioner": "jacobi",
      "projection_space_size": 0,
      "absolute_tolerance": 1e-9,
      "max_iterations": 800
    },
    "pressure_solver": {
      "type": "gmres",
      "preconditioner": "hsmg",
      "projection_space_size": 20,
      "absolute_tolerance": 1e-9,
      "max_iterations": 800
    },
    "output_control": "org",
    "output_value": 10000
  },
  "scalar": {
    "enabled": false,
    "source_term": "noforce",
    "Pr": 1.0,
    "boundary_types": ["t", "f"]
  },
  "statistics": {
    "enabled": false,
    "output_value": 10000,
    "start_time": 0.0,
    "sampling_interval": 10
  }
}
```

timofeymukha:
This PR will eventually close #858 and #831. My general intention here is to get something that we can use and expand on. I expect the JSON layout to be somewhat in flux during the 0.7 dev cycle, and there are also things that can be done better internally.

Current status and to-dos
- [x] Compiles locally without using `case%params`.
- [x] Reasonable output for most the examples (those I could run locally). 
- [ ] Clean-up unnecessary things pertaining to the old format.
- [ ] Modify CI and fix issues that pop up.
- [ ] Make it compile with Cray 14 and 15 on Dardel / Lumi.
- [ ] Run the bigger examples.


njansson:
Very nice! 

Some suggestions
* We should be nice to lustre and only parse the json file on one rank, then broadcast the serialised string, and later create the json object from the string on the other ranks.
* To remove a lot of the boilerplate load test etc, maybe we can hide this inside a module at a later stage, e.g. inside `param` once we go full json

timofeymukha:
@njansson @MartinKarp, I am working on the documentation for the case file. Just committed som WIP. There are a few things I need help on, which I marked with `todo` in the modified `.md`. Could you please take a look and provide some insight? Thanks!

src/fluid/fluid_pnpn.f90
```diff
@@ -232,21 +242,32 @@ subroutine fluid_pnpn_init(this, msh, lx, param)
     call bc_list_add(this%bclst_dw, this%bc_vel_res)
 
     !Intialize projection space thingy
-    if (param%proj_prs_dim .gt. 0) then
-       call this%proj_prs%init(this%dm_Xh%size(), param%proj_prs_dim)
+    if (this%ksp_pr_projection_dim .gt. 0) then
+       call this%proj_prs%init(this%dm_Xh%size(), this%ksp_pr_projection_dim)
```
njansson, line 246: Not sure `ksp_` is a good naming here, since it is not directly tied to the Krylov solvers, as the name then would suggrest

src/fluid/fluid_pnpn.f90
```diff
@@ -232,21 +242,32 @@ subroutine fluid_pnpn_init(this, msh, lx, param)
...
     end if
     
-    if (param%proj_vel_dim .gt. 0) then
-       call this%proj_u%init(this%dm_Xh%size(), param%proj_vel_dim)
-       call this%proj_v%init(this%dm_Xh%size(), param%proj_vel_dim)
-       call this%proj_w%init(this%dm_Xh%size(), param%proj_vel_dim)
+    if (this%ksp_vel_projection_dim .gt. 0) then
+       call this%proj_u%init(this%dm_Xh%size(), this%ksp_vel_projection_dim)
+       call this%proj_v%init(this%dm_Xh%size(), this%ksp_vel_projection_dim)
+       call this%proj_w%init(this%dm_Xh%size(), this%ksp_vel_projection_dim)
```
njansson, line 252: Not sure `ksp_` is a good naming here, since it is not directly tied to the Krylov solvers, as the name then would suggrest

src/fluid/fluid_scheme.f90
```diff
@@ -82,6 +83,10 @@ module fluid_scheme
      class(ksp_t), allocatable  :: ksp_prs     !< Krylov solver for pressure
      class(pc_t), allocatable :: pc_vel        !< Velocity Preconditioner
      class(pc_t), allocatable :: pc_prs        !< Velocity Preconditioner
+     integer :: ksp_vel_maxiter                !< Max iterations in ksp_vel
+     integer :: ksp_pr_maxiter                 !< Max iterattions in ksp_pr
+     integer :: ksp_vel_projection_dim         !< Size of the projection space for ksp_vel
+     integer :: ksp_pr_projection_dim          !< Size of the projection space for ksp_pr
```
njansson, line 89: Not sure `ksp_` is a good naming here, since it is not directly tied to the Krylov solvers, as the name then would suggrest

src/fluid/fluid_pnpn.f90
```diff
@@ -334,20 +365,19 @@ subroutine fluid_pnpn_step(this, t, tstep, ext_bdf)
...
          ulag => this%ulag, vlag => this%vlag, wlag => this%wlag, &
-         params => this%params, msh => this%msh, prs_res => this%prs_res, &
+         msh => this%msh, prs_res => this%prs_res, &
          vel_res => this%vel_res, sumab => this%sumab, &
          makeabf => this%makeabf, makebdf => this%makebdf, &
-         prs_max_iter => this%params%prs_max_iter, &
-         vel_max_iter => this%params%vel_max_iter)
-         
+         rho => this%rho, Re => this%Re, mu => this%mu)
+
```
njansson, line 372: If we associated some of the params now stored in various derived types, I suggest we do the same for the project space as well to keep the step loop rather clean from `this%`

src/fluid/fluid_scheme.f90
```diff
@@ -396,27 +506,46 @@ subroutine fluid_scheme_init_all(this, msh, lx, params, &
...
     if (kspp_init) then
+       call json_get(params, 'case.fluid.pressure_solver.type', string_val1)
+       call json_get(params, 'case.fluid.pressure_solver.preconditioner', &
+                     string_val2)
+       call json_get(params, 'case.fluid.pressure_solver.absolute_tolerance', &
+                     real_val)
+
        call fluid_scheme_solver_factory(this%ksp_prs, this%dm_Xh%size(), &
-            params%ksp_prs, params%abstol_prs)
+            "gmres", real_val)
```
njansson, line 546: Shouldn't this be `string_val1` ?

src/fluid/fluid_scheme.f90
```diff
@@ -426,6 +555,7 @@ end subroutine fluid_scheme_init_all
 
   !> Deallocate a fluid formulation
   subroutine fluid_scheme_free(this)
+    implicit none
```
njansson, line 558:
```suggestion
```

Module wide `implicit none`

src/fluid/fluid_scheme.f90
```diff
@@ -482,7 +612,11 @@ end subroutine fluid_scheme_free
   !> Validate that all fields, solvers etc necessary for
   !! performing time-stepping are defined
   subroutine fluid_scheme_validate(this)
+    implicit none
```
njansson, line 615:
```suggestion
```

Module wide implicit none

src/fluid/fluid_scheme.f90
```diff
@@ -618,13 +753,8 @@ subroutine fluid_scheme_set_source(this, source_term_type, usr_f, usr_f_vec)
...
     else if (trim(source_term_type) .eq. 'user_vector' .and. present(usr_f_vec)) then
        call source_set_type(this%f_Xh, usr_f_vec)
-    else if (trim(source_term_type) .eq. '') then
-       if (pe_rank .eq. 0) then
-          call neko_warning('No source term defined, using default (noforce)')
-       end if
-       call source_set_type(this%f_Xh, source_eval_noforce)
     else
-       call neko_error('Invalid source term')
+       call neko_error('Invalid fluid source term '//source_term_type)
```
njansson, line 757: Does this mean that users now has to choose a source term?

src/fluid/fluid_pnpn.f90
```diff
@@ -165,35 +175,37 @@ subroutine fluid_pnpn_init(this, msh, lx, param)
       
     end associate
     
+    ! Read boundary type labels
+    call read_boundary_labels(params, bc_labels)
+
```
njansson, line 180: Can't we store bc_labels somewhere such that we don't need to re read it three times?

src/fluid/fluid_scheme.f90
```diff
@@ -653,5 +783,48 @@ function fluid_compute_cfl(this, dt) result(c)
...
+          call core%get(bc_ptr, bc_label)
+          
+          ! Assign non-empty label
+          if (len(bc_label) .gt. 0) then
+            bc_labels(i) = bc_label
+          end if
+       end do
+    end if
+
+  end subroutine read_boundary_labels
```
njansson, line 828: As mentioned earlier, can't we store the bc_labels somewhere such that we don't need to call this functions at least three times, and keep duplicates

src/fluid/fluid_scheme.f90
```diff
@@ -562,9 +697,9 @@ end subroutine fluid_scheme_bc_apply_prs
   subroutine fluid_scheme_solver_factory(ksp, n, solver, abstol)
     class(ksp_t), allocatable, target, intent(inout) :: ksp
     integer, intent(in), value :: n
-    character(len=20), intent(inout) :: solver
-    real(kind=rp) :: abstol
-
+    character(len=*), intent(in) :: solver
+    real(kind=rp), intent(in) :: abstol
+
```
njansson, line 702: Are we sure `solver` always is a properly defined string? Otherwise, assumed len can cause issues

src/fluid/fluid_scheme.f90
```diff
@@ -577,7 +712,7 @@ subroutine fluid_scheme_precon_factory(pc, ksp, coef, dof, gs, bclst, pctype)
     type(dofmap_t), target, intent(inout) :: dof
     type(gs_t), target, intent(inout) :: gs
     type(bc_list_t), target, intent(inout) :: bclst
-    character(len=20) :: pctype
+    character(len=*) :: pctype
```
njansson, line 715: Are we sure `pctype` always is a properly defined string? Otherwise, assumed len can cause issues

src/fluid/fluid_scheme.f90
```diff
@@ -527,18 +661,19 @@ subroutine fluid_scheme_validate(this)
...
+               this%mean%v, this%mean%w, this%mean%p)
+       end if
     end if
 
-    if (this%params%stats_mean_sqr_flow) then
-       call this%mean_sqr%init(this%u, this%v, this%w, this%p)
-    end if
+!    if (this%params%stats_mean_sqr_flow) then
+!       call this%mean_sqr%init(this%u, this%v, this%w, this%p)
+!    end if
```
njansson, line 676: Don't we support this anymore, or is it due to other reasons?

src/krylov/krylov_fctry.f90
```diff
@@ -52,15 +52,15 @@ module krylov_fctry
...
+    character(len=*), intent(in) :: solver
     real(kind=rp), optional :: abstol
     class(pc_t), optional, intent(inout), target :: M
  
     if (allocated(ksp)) then
        call krylov_solver_destroy(ksp)
        deallocate(ksp)
     end if
-
+
```
njansson, line 63:
```suggestion

```

src/scalar/scalar_pnpn.f90
```diff
@@ -156,16 +161,23 @@ subroutine scalar_pnpn_init(this, msh, coef, gs, param)
 
     ! @todo not param stuff again, using velocity stuff
     ! Intialize projection space thingy
-    if (param%proj_vel_dim .gt. 0) then
-       call this%proj_s%init(this%dm_Xh%size(), param%proj_vel_dim)
+    if (this%ksp_projection_dim .gt. 0) then
+       call this%proj_s%init(this%dm_Xh%size(), this%ksp_projection_dim)
```
njansson, line 165: Still not convinced that projections should be related to `ksp` as in Krylov Solver Preconditioner

src/common/json_utils.f90
```diff
@@ -0,0 +1,258 @@
...
+  private
+
+  public :: json_get, json_get_or_default
+  
+  !> Retrieves a parameter by name or throws an error
+  interface json_get
+    module procedure json_get_real, json_get_integer, json_get_logical, &
+                     json_get_string, json_get_real_array,&
+                     json_get_integer_array, json_get_logical_array
+  end interface json_get
```
njansson, line 48: Something to think about. I would argue that it's more convenient to have functions returning the values instead of subroutines. That would for example remove the redundant `logical_val` or `string1_val` we have now in various init routines

src/case.f90
```diff
@@ -84,94 +88,76 @@ module case
...
-    namelist /NEKO_CASE/ mesh_file, fluid_scheme, lx,  &
-         source_term, initial_condition, scalar, scalar_source_term
-    
     integer :: ierr
     type(file_t) :: msh_file, bdry_file, part_file
     type(mesh_fld_t) :: msh_part
     integer, parameter :: nbytes = NEKO_FNAME_LEN + (4 * 80) + 4 + 4
-    character buffer(nbytes)
+    character :: buffer(nbytes)
     integer :: pack_index
```
njansson, line 109: This can be removed now, since we don't use `MPI_PACK`

src/case.f90
```diff
@@ -194,57 +185,75 @@ subroutine case_init(C, case_file)
     !
     ! Setup scalar scheme
     !
-    ! @todo no scalar factroy for now, probably not needed
+    ! @todo no scalar factory for now, probably not needed
+    if (C%params%valid_path('case.scalar')) then
+       call json_get_or_default(C%params, 'case.scalar.enabled', scalar,&
+                                .false.)
+    end if
+
```
njansson, line 193: Why can't we enable scalars based on the existence of such a block in the parameter file? e.g.

if value_pat (case.scalar) then => Allocated and setup scalars?

src/case.f90
```diff
@@ -269,107 +278,146 @@ subroutine case_init(C, case_file)
...
+!       call C%q%add(C%fluid%mean_sqr%pp)
+
+!       if (C%params%output_mean_sqr_flow) then
+!          C%f_msqrf = mean_sqr_flow_output_t(C%fluid%mean_sqr, &
+!                                             C%params%stats_begin, &
+!                                             path=output_directory)
+!          call C%s%add(C%f_msqrf, C%params%stats_write_par, &
+!               C%params%stats_write_control)
+!       end if
+!    end if
```
njansson, line 412: Don't we support this anymore? Or are the other reasons behind it being disabled?

src/scalar/scalar_pnpn.f90
```diff
@@ -257,8 +273,8 @@ subroutine scalar_pnpn_step(this, t, tstep, ext_bdf)
 
       ! compute scalar residual
       call profiler_start_region('Scalar residual')
-      call res%compute(Ax, s,  s_res, f_Xh, c_Xh, msh, Xh, params%Pr, &
-          params%Re, params%rho, ext_bdf%diffusion_coeffs(1), params%dt, &
+      call res%compute(Ax, s,  s_res, f_Xh, c_Xh, msh, Xh, this%Pr, &
+          this%Re, this%rho, ext_bdf%diffusion_coeffs(1), dt, &
```
njansson, line 277: Can't we associated `Re, Pr, rho` and friends to make this a bit more prettier?

src/scalar/scalar_scheme.f90
```diff
@@ -215,9 +262,32 @@ subroutine scalar_scheme_init(this, msh, c_Xh, gs_Xh, params, scheme)
...
+          call core%get(bc_ptr, bc_label)
+          
+          ! Assign non-empty label
+          if (len(bc_label) > 0) then
+            bc_labels(i) = bc_label
+          end if
+       end do
+    end if
+    call scalar_scheme_add_bcs(this, msh%labeled_zones, bc_labels) 
+
```
njansson, line 290: We need to sort out this bc_labels at some point, having multiple copies scattered across several modules isn't nice. 

How about (in the future) we move it to `mesh`?

src/scalar/scalar_scheme.f90
```diff
@@ -320,7 +390,7 @@ end subroutine scalar_scheme_bc_apply
   subroutine scalar_scheme_solver_factory(ksp, n, solver, abstol)
     class(ksp_t), allocatable, target, intent(inout) :: ksp
     integer, intent(in), value :: n
-    character(len=20), intent(inout) :: solver
+    character(len=*), intent(in) :: solver
```
njansson, line 393: Same comment as for fluid, are we sure that `solver` never can be a "heap" allocated string?

src/scalar/scalar_scheme.f90
```diff
@@ -335,7 +405,7 @@ subroutine scalar_scheme_precon_factory(pc, ksp, coef, dof, gs, bclst, pctype)
     type(dofmap_t), target, intent(inout) :: dof
     type(gs_t), target, intent(inout) :: gs
     type(bc_list_t), target, intent(inout) :: bclst
-    character(len=20) :: pctype
+    character(len=*) :: pctype
```
njansson, line 408: Same comment as for fluid, are we sure that `pctype` never can be a "heap" allocated string?

njansson (changes requested):
First set of comments

src/fluid/fluid_scheme.f90
timofeymukha: No, but I moved the logic for selecting the default value into case.f90, as for the rest of the stuff. I think it is better for type constructor to actually crash if we give it a random string. Could potentially lead to hard to find bugs otherwise.

src/fluid/fluid_scheme.f90
timofeymukha: Open question :-)! I ignored it in the json layout, thinking that the new statistics has superseded this. I think I asked on zulip at some point, but we did not arrive to a conclusion

src/fluid/fluid_scheme.f90
timofeymukha: What I remember is that I was somehow forced to do this change to make things work... But I would say yes, the solver and pc type should be properly defined.

src/scalar/scalar_pnpn.f90
timofeymukha: I am glad to remove the ksp everywhere for the projections. This was just lack of knowledge on my side,

src/common/json_utils.f90
timofeymukha: Yeah, I started with functions, but then bumped into the fact that you can't have a generic interface that distinguishes purely on the return type of a function. So, switched to a subroutine. If we sack the beauty of the generic interface, then functions are an option.

src/case.f90
timofeymukha: We can, having the "enabled" parameter was a deliberate choice on my part. But there is a mistake here, the default value should be true. So, the enabled parameter is optional for the user. 

I carried over the "enabled" from experience with OF. It is nice to have just one switch to turn in case you need to kill some part of the simulation. For example, your case crashes, and you want to check if it will crash without the scalar. So you flip enabled to false, instead of having to cut out the whole scalar from the file, and then eventually restoring it later. Same with other stuff, like statistics.

src/scalar/scalar_pnpn.f90
timofeymukha: yes!

src/scalar/scalar_scheme.f90
timofeymukha: I agree. Not sure about "mesh" because I guess we want to in fact decouple the bc stuff from the mesh in the future. But maybe just the "utils" module? Not sure

src/fluid/fluid_scheme.f90
njansson: Strange, I recall that I had to give it a size, since sometimes the pc type was passed by value, directly as a dummy argument

src/scalar/scalar_scheme.f90
njansson: > I agree. Not sure about "mesh" because I guess we want to in fact decouple the bc stuff from the mesh in the future. But maybe just the "utils" module? Not sure

Or maybe we simply store it somewhere in `bc.f90`

src/case.f90
njansson: Ok then I understand, and I agree that it's rather nice to have the flexibility to disable parts of a simulation, without having to rip out part of a case description.

njansson: But then I put the default as `.true.` and we are happy

src/case.f90
```diff
@@ -194,57 +185,75 @@ subroutine case_init(C, case_file)
     !
     ! Setup scalar scheme
     !
-    ! @todo no scalar factroy for now, probably not needed
+    ! @todo no scalar factory for now, probably not needed
+    if (C%params%valid_path('case.scalar')) then
+       call json_get_or_default(C%params, 'case.scalar.enabled', scalar,&
+                                .false.)
```
njansson, line 191:
```suggestion
                                .true.)
```

src/common/json_utils.f90
njansson: so true my bad

src/case.f90
```diff
@@ -182,7 +157,12 @@ subroutine case_init(C, case_file)
     !
     ! Setup fluid scheme
     !
-    call fluid_scheme_factory(C%fluid, trim(fluid_scheme))
+    call json_get(C%params, 'case.fluid.scheme', string_val)
+    call fluid_scheme_factory(C%fluid, trim(string_val))
+
+    call C%params%get('case.numerics.polynomial_order', lx)
+    call json_get(C%params, 'case.numerics.polynomial_order', lx)
```
njansson, line 164: @timofeymukha Not sure if I understand why we need to read `lx` twice? Or is it some old leftover before the json interface was added?

doc/pages/case-file.md
```diff
@@ -1,106 +1,218 @@
...
+The value of the keyword is an array of strings, with the following possible
+values:
+* `w`, a no-slip wall.
+* `v`, a Dirichlet boundary.
+* `sym`, a symmetry boundary.
+* `on`, Dirichlet for the boundary-parallel velocity and homogeneous Neumann for
+   the wall-normal. todo: WHERE IS THE VELOCITY AND AMBIENT PRESSURE SET?
+* `o`, outlet boundary. 
+* `o+dong`, outlet boundary using the Dong condition. 
+* `on+dong`, an `on` boundary using the Dong condition. todo: NOT SURE WHAT THIS IS EXACTLY.
```
MartinKarp, line 156: It sets the pressure boundary condition to make sure that the boundary normal velocity is always going out.

MartinKarp (commented):
Some comments on the doc

doc/pages/case-file.md
```diff
@@ -1,106 +1,218 @@
...
+directly inside the mesh file.
+In particular, this happens when Nek5000 `.re2` files are converted to `.nmsh`.
+Periodic boundary conditions are *always* defined inside the mesh file.
+
+The value of the keyword is an array of strings, with the following possible
+values:
+* `w`, a no-slip wall.
+* `v`, a Dirichlet boundary.
+* `sym`, a symmetry boundary.
+* `on`, Dirichlet for the boundary-parallel velocity and homogeneous Neumann for
```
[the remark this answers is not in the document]
MartinKarp: Kididng, it is the initialcondtion that sets the ambient velocity

src/case.f90
timofeymukha: yeah,  looks like 163 should be deleted.

src/scalar/scalar_scheme.f90
njansson: I'll mark conversation related to `bc_labels` as resolved for now, and similar with currently disabled features (stats etc)

njansson (approved):

PR #892, opened 2023-06-14
Edit documentation in fast3d

vbaconnet:
Just a quick documentation fix for `setup_intp`

### DIFF
--- a/src/math/fast3d.f90
+++ b/src/math/fast3d.f90
@@ -233,9 +233,11 @@ contains
   !!
   !! @param jh Matrix of the interpolation weights.
   !! @param jht Same as `jh` but transposed.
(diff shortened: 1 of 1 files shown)


njansson (approved):

timofeymukha:
Now comes the most important question: should there or should there not be a period in the end of
>@param z_from Quadrature points

and similar. As you can see, I am a proponent of having a period :D!

timofeymukha:
@njansson Should the CI have not been smart in this case and skip everything but doxygen?

njansson:
> @njansson Should the CI have not been smart in this case and skip everything but doxygen?

No, that fix had to be reverted since it caused issues with the branch protection rules (read we need to rethink this...)

PR #893, opened 2023-06-15
Fix memory leak in device pipelined CG

njansson:
Leak and redundant device allocation of `alpha` and `beta` in each solve step

### DIFF
--- a/src/krylov/bcknd/device/pipecg_device.F90
+++ b/src/krylov/bcknd/device/pipecg_device.F90
@@ -296,7 +296,7 @@ contains
        call device_free(this%alpha_d)
     end if
     if (c_associated(this%beta_d)) then
-       call device_Free(this%beta_d)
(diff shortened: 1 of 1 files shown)


src/krylov/bcknd/device/pipecg_device.F90
```diff
@@ -277,9 +291,15 @@ subroutine pipecg_device_free(this)
     end if
     if (c_associated(this%ni_d)) then
        call device_free(this%ni_d)
-    end if    
+    end if
+    if (c_associated(this%alpha_d)) then
+       call device_free(this%alpha_d)
+    end if
+    if (c_associated(this%beta_d)) then
+       call device_Free(this%beta_d)
```
timofeymukha, line 299:
```suggestion
       call device_free(this%beta_d)
```

PR #895, opened 2023-06-18
Misc additions to docstrings

timofeymukha:
* As the subject says
* Provisionally "upgraded" the docstring to our format for one of the routines in speclib. Or is the idea not to touch it? It is a bit said that it is not in doxygen.

### DIFF
--- a/src/sem/speclib.f90
+++ b/src/sem/speclib.f90
@@ -59,19 +59,19 @@
 !
 !==============================================================================
 !
-!     LIBRARY ROUTINES FOR SPECTRAL METHODS
-!
-!     March 1989
-!
-!     For questions, comments or suggestions, please contact:
-!
-!     Einar Malvin Ronquist
-!     Room 3-243
-!     Department of Mechanical Engineering
-!     Massachusetts Institute of Technology
-!     77 Massachusetts Avenue
-!     Cambridge, MA 0299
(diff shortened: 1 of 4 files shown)


njansson:
> * Provisionally "upgraded" the docstring to our format for one of the routines in speclib. Or is the idea not to touch it? It is a bit said that it is not in doxygen.

As long as we preserve some of the origins/history, I don't have any strong feelings that we shouldn't touch `speclib`.

src/sem/speclib.f90
```diff
@@ -150,18 +150,13 @@ module speclib
...
       SUBROUTINE ZWGL (Z,W,NP)
-!--------------------------------------------------------------------
-!
-!     Generate NP Gauss Legendre points (Z) and weights (W)
-!     associated with Jacobi polynomial P(N)(alpha=0,beta=0).
-!     The polynomial degree N=NP-1.
-!     Z and W are in single precision, but all the arithmetic
-!     operations are done in double precision.
-!
-!--------------------------------------------------------------------
```
njansson, line 164: This is actually a good example on where the old docstring is no longer valid. I'm all for history but the documentation has to be correct

njansson (approved):

vbaconnet (approved):

PR #896, opened 2023-06-18
Add a proper user scalar bc module

njansson:
This pr breaks out the user defined scalar bc from user defined inflow 
conditions...


src/bc/usr_scalar.f90
```diff
@@ -0,0 +1,290 @@
...
+    if (c_associated(this%usr_x_d)) then
+       call device_free(this%usr_x_d)
+    end if
+        
+  end subroutine usr_inflow_free
+  
+  !> Scalar apply
+  !! Just imitating inflow for now, but we should look this over
+  !! Applies boundary conditions in eval on x
+  !! @param x The array of values to apply.
```
timofeymukha, line 108:
```suggestion
  !! @param x The field array to apply the boundary condition to.
```

timofeymukha: I think this is the correct meaning, or? We eventually pass the values of x one-by-one to eval, and that sets the value, right?

MartinKarp:
I would instinctively rather rename usr_inflow to usr_bc (or variable_bc). 

In my opinion, I find that the large number of different BC types, for essentially doing very similar things, is a bit redundant and makes things a bit difficult to grasp.

src/bc/usr_scalar.f90
```diff
@@ -0,0 +1,290 @@
...
+     procedure, pass(this) :: apply_vector_dev => usr_scalar_apply_vector_dev
+     procedure, pass(this) :: apply_scalar_dev => usr_scalar_apply_scalar_dev
+  end type usr_scalar_t
+
+  abstract interface
+     !> Abstract interface defining a user defined scalar
+     !! boundary condition (pointwise)
+     !! Just imitating inflow for now, but we should update this
+     !! Probably passing the whole field, params, coef, etcetc would be good
+     !! @param s The scalar's value in this point
```
timofeymukha, line 64:
```suggestion
     !! @param s The scalar in this point, the value of which will be evaluated and set.
```

timofeymukha: Right? We don't pass the value of s here, but rather s is set inside the eval.

njansson:
> I would instinctively rather rename usr_inflow to usr_bc (or variable_bc).
> 
> In my opinion, I find that the large number of different BC types, for essentially doing very similar things, is a bit redundant and makes things a bit difficult to grasp.

Yes and no, the reason for having different instances of similar things is more for error checking. A simple rename here will look very strange in the inheritance  tree

MartinKarp:
My thinking is mainly, both are user-defined bcs, why should there be two types/modules? I think the main issue is that the current name is a bit misleading.

timofeymukha:
The "inflow" is definitely a misnomer, since it is just an inhomogeneous Dirichlet bc applied for various purposes, e.g. driving the lid, rotating a cylinder, and also inflow :-)

njansson:
> My thinking is mainly, both are user-defined bcs, why should there be two types/modules? I think the main issue is that the current name is a bit misleading.

I rather argue that it's too general to simply have one generic user-defined bc module. The scheme now is that we have defined types, let say `A`, then we allow for user defined types `B`. Then we can always check if A is a B and call the correct hooks.

So for scalar it looks a bit strange if we have `dirichlet -> inflow -> user_inflow` ...

njansson:
> The "inflow" is definitely a misnomer, since it is just an inhomogeneous Dirichlet bc applied for various purposes, e.g. driving the lid, rotating a cylinder, and also inflow :-)

Indeed, but that's a more major refactoring atm. So this as mentioned, is to reduce the confusion when trying to figure out where the user provided scalar stuff is applied

timofeymukha (approved):

PR #898, opened 2023-06-19
Add csv files input/output

vbaconnet:
**_Warning: long PR_**

An attempt to add `.csv` file formats for reading and writing data. Also some miscellaneous documentation fixes. I am submitting this for feedback on possible modifications, and I am sure there are a lot of things to improve!

# How to use

The reading and writing of 1D (resp. 2D) arrays is done using wrapper object types, namely `vector_t` (resp. `matrix_t`). I added an example in `examples/user-output`, which is just a slight modification of the 2d lid example `lid2d.f90`, where we compute and write `Ekin` and `enst` every 50th timestep.

Here is an overview of the workflow taken from the example:

```.f90

! 
! Global variables
! 
type(file_t) output_file  ! output file
type(vector_t) :: vec_out ! will store our output data

!
! In [user_initialize_modules]
!
output_file = file_init("output.csv")
call output_file%set_header("t,Ekin,enst")
call vec_out%init(2) ! Initialize our vector with 2 elements (Ekin, enst)

!
! In [user_check]
!
! ... calculations ...
vec_out%x = (/e1, e2/)
call output_file%write(vec_out, t)

!
! In [user_finalize]
!
call vec%free()
call file_free(output_file) ! not sure of this is needed
```

# Some remarks

## Parallel I/O
- The reading and writing are both done on rank 0 (hardcoded in `csv_file.f90`). I am sure there are problems with that approach, but I guess it works fine for simple applications?
- After the data has been read, it is **not** broadcast to other processes. Should it be? Seems like something the user could do themselves if they really need to?

### Header
I added the header attribute `character(len=1024) :: header` and a setter `set_header` in `csv_file_t` and consequently `file_t` as follows:

```.f90
  subroutine file_set_header(this, hd)
    class(file_t), intent(inout) :: this
    character(len=*), intent(in) :: hd

    select type(ft => this%file_type)
    class is (csv_file_t)
       call ft%set_header(hd)
    end select

  end subroutine file_set_header
```

The issue I can see with doing this is it exposes the `header` concept at the `file_t` level, even though it is specific to `csv` files. 

A solution would be to make `header` an attribute of the generic file type `generic_file_t`, with a `set_header` at the `generic_file_t` level, so the user could just call like this:
```.f90
call output_file%file_type%set_header("t,Ekin,enst")
```
 But this still somehow problematic as other file types **do** have headers but not in the same way as `csv` files do, i.e. just a single line, and may not all be represented by a `character`...? 

I think the first option is easier from a user perspective. 

### Format
- By default, the writing is performed with the format descriptor `'(g0)'`, but this can be changed depending on if we want a specific format, like `E` or `F`?
- If the user gives a `vector_t` as input (e.g. `v = (/x1,x2,x3/)`), the data will be written as:
```.csv
header1,header2,header3
x1,x2,x3
```
- If the user gives a `matrix_t` as input for `write`, it will be written as;
```.csv
header1,header2,header3
x11,x12,x13
x21,x22,x23
x31,x32,x33
```
Note the header is by default empty unless set by the `set_header` subroutine.
- If the `t` optional argument is present, it will be added at the beginning of every line:
```.csv
t,header1,header2,header3
t1,x1,x2,x3
```
or
```.csv
t,header1,header2,header3
t1,x11,x12,x13
t1,x21,x22,x23
t1,x31,x32,x33
```
Note that the user has to edit the header themselves if they are also adding time as an argument (see the `user-output` example).

### Detection of header
Right now the way to check if there is a header in an input file is to compare the number of lines in the file with the size of the `vector_t` (length) or `matrix_t` (number of rows), and then just reading the 1st line in the file as a string.
```.f90
if (n_lines .lt. mat%nrows) then
       read (file_unit, '(A)') tmp
       this%header = trim(tmp)
end if
```
Any comments on how to do this in a prettier way?


njansson (approved):

MartinKarp (approved):
As far as I can see this looks pretty spot on.

timofeymukha:
Sorry, I have some question, so I disabled automerge :P

timofeymukha:
First of all, great job, and an amazing PR description, making the rest of us ashamed of ourselves :-). I have a few questions/comments!

* Do we need a new example, or can we just update the existing lid? It seems that what you add is a good upgrade, since we already write stuff out there, or?
* What is the advantage of the multiline matrix IO as opposed to just unrolling it into a single line? If I imagine myself reading it later in Python, I would much rather have 1 line per timestep, so I can read the whole thing into a 2D array with `np.genfromtxt`, and then just reshape stuff. Parsing the multiline output would on the other hand require more effort.
* Could you please add the first line of the docstrings for the TBPs into the type definition? For example, 
`!> Initialise a matrix of size nrows*ncols.`
would go above
`procedure, pass(m) :: init => matrix_init` in the type. This way, one does need to scroll through the file to read what a method does. 
* @njansson  is stuff like `print *, "HEADER: ", d%header` ok? It's just that I've never seen `print` in the code before. Should one use the neko_log stuff?
* A little unfortunate with having to first count lines in the read routines. That can actually take time for large files. As I see, it is only used to check if there is a header. Could we maybe add that as a logical parameter to the routines?

vbaconnet:
- I agree about the example, I think it may be better to just update the existing lid.
- I guess the idea behind the matrix IO was to be able to output, say, some data taken on a 2D plane, and dump it all in a file directly (without the time at the start of each row). However, I agree that adding the time at the start of each line does make it harder to read later in python or something like that. I think for the case **without** the time we could keep the 2D structure of rows and columns, but if we add time as an argument then one can unroll the matrix on one row?
- The "line counting" is indeed annoying. Adding a logical parameter would be much easier, in that case would we change the interface in `generic_file_t`, `generic_file_read`?

Will also add the doc in the type definitions.

timofeymukha:
A little bit scared to change the `generic_file_t` interface, maybe we can live with the line counting for now. I suppose one can argue that this is not meant for large files. I guess if we really want to avoid it later, we can try to read the first line and try to have some logic for figuring out if it is a header or not.

I think the matrix output logic you suggest is good. Just make sure you document it well, because it is not really self-evident that adding the time as an argument results in changing the format. I think, in the end of the day, we have to see how the class is used in practice and then possibly adjust.

vbaconnet:
I guess i will remove the example then and just modify the existing lid example, if everyone is okay with that, and add some documentation for how the vector/matrix outputs are formatted.

njansson (approved):

PR #912, opened 2023-07-05
Fix dangling pointers

njansson:
Add final procedures, nullify and target specifiers to avoid danling pointers

To check, compile with NAG and -C=dangling, and run with NAGFORTRAN_RUNTIME_OPTIONS=show_dangling


timofeymukha:
Is `final` approved for usage now?

njansson:
> Is `final` approved for usage now?

For the iterators it seems to be fine since they are always limited in their scope to one subroutine etc.

timofeymukha (approved):

PR #915, opened 2023-07-19
Store bc_labels in each solver base type

njansson:
Store bc_labels in each solver's base type, such that redudant reads/loads can be avoided, and labels reused at later stage (e.g. init in both fluid_scheme -> fluid_pnpn)


timofeymukha:
Do we need the `bc_labels `attribute to be allocatable?

I am also a bit scared for this argument type
`character(len=*), allocatable, intent(out) :: value(:)`
because I think this is what caused nvfortran to die miserably and why we can't use the array reader in json-fortran.

njansson:
> Do we need the `bc_labels `attribute to be allocatable?

Strictly no, but I wanted the json interface to be as generic as possible to allow for string arrays of unknown size
 
> I am also a bit scared for this argument type `character(len=*), allocatable, intent(out) :: value(:)` because I think this is what caused nvfortran to die miserably and why we can't use the array reader in json-fortran.

This seems to work fine, it can read the labels correctly without SEGV.

timofeymukha:
OK, if nvfortran is happy then it's great.

timofeymukha (approved):

PR #918, opened 2023-07-20
Change case init to allow for either a file or a json object

njansson:
Allow for a case to be created from either a JSON file, or an already loaded JSON object. The intent is to allow for programatic case creation from e.g. pyneko or future endeavours


timofeymukha (approved):

timofeymukha:
This is totally fine, but I would love to get #728 settled so that we can work towards further standardising these things.

Issue #822, opened 2023-04-07
Advection classes improvements


timofeymukha:
* There is a comment by @MartinKarp that the implementation on the GPU is "extremely primitive and unoptimized". Is this something, which we need to address? Also, I am a bit curious what the issue is, the code looks very straightforward.
* Out of curiosity, I wonder why the CPU implementation involves a loop over elements instead of working on the whole array at once like in the device implementation? Is this better? We use extra temporaries, although just element-sized.
* I think the argument `n` in the interface
```  abstract interface
     subroutine apply_adv(this, vx, vy, vz, fx, fy, fz, Xh, coef, n)
```
is superfluous. If we look at `fuid_pnpn` the call is
```fortran
      call this%adv%apply(this%u, this%v, this%w, &
                          f_Xh%u, f_Xh%v, f_Xh%w, &
                          Xh, this%c_Xh, dm_Xh%size())
```
So `n` comes from the dofmap, but the `coef` class, which is also passed, has a pointer to the dofmap and can therefore extract `n`. So it seems we can remove this argument.
* I propose that we rename `apply` to `compute`. It just aligns `adection` with the other classes, like `makeabf/bdf`, `pnpn_res`, etc. They all have `compute`.
```fortran

      call sumab%compute_fluid(u_e, v_e, w_e, u, v, w, &
           ulag, vlag, wlag, ext_bdf%ext%coeffs, ext_bdf%ext%n)
    
.
.
.

      call this%adv%apply(this%u, this%v, this%w, &
                          f_Xh%u, f_Xh%v, f_Xh%w, &
                          Xh, this%c_Xh, dm_Xh%size())

      call makeabf%compute_fluid(temp1, temp2, temp3,&
                           this%abx1, this%aby1, this%abz1,&
                           this%abx2, this%aby2, this%abz2, &
                           f_Xh%u, f_Xh%v, f_Xh%w,&
                           params%rho, ext_bdf%ext%coeffs, n)

      call makebdf%compute_fluid(temp1, temp2, temp3, temp4, temp5, temp6,&
                           ulag, vlag, wlag, f_Xh%u, f_Xh%v, f_Xh%w, &
                           u, v, w, c_Xh%B, params%rho, params%dt, &
                           ext_bdf%bdf%coeffs, ext_bdf%bdf%n, n)
.
.
.
      call prs_res%compute(p, p_res, u, v, w, u_e, v_e, w_e, &
[...]

PR #919, opened 2023-07-24
adv%apply -> adv%compute + docs

timofeymukha:
Renames `apply` to `compute` in advection classes, as discussed in #822. Also adds a bit of comments/docs.

### DIFF
--- a/src/fluid/fluid_pnpn.f90
+++ b/src/fluid/fluid_pnpn.f90
@@ -417,7 +417,7 @@ contains
       call this%bc_apply_vel()
       call this%bc_apply_prs()
 
-      ! C ompute pressure.
+      ! Compute pressure.
       call profiler_start_region('Pressure residual')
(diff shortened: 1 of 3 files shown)


src/fluid/fluid_pnpn.f90
```diff
@@ -378,37 +378,46 @@ subroutine fluid_pnpn_step(this, t, tstep, dt, ext_bdf)
...
       call wlag%update()
+
       !> We assume that no change of boundary conditions 
       !! occurs between elements. I.e. we do not apply gsop here like in Nek5000
       !> Apply dirichlet
       call this%bc_apply_vel()
       call this%bc_apply_prs()
 
-      ! compute pressure
+      ! C ompute pressure.
```
njansson, line 420:
```suggestion
      ! Compute pressure.
```

njansson (changes requested):

njansson (approved):

PR #922, opened 2023-08-04
Put check on whether to execute an action based on time value in a separate class

timofeymukha:
* The `sampler` class implemented the logic for determining whether we need to write or not.
* I put this in a separate class, called `time_based_controller`.
* The reason is that we may want to use this in other contexts. Basically for anything, which should not be executed at each time-step, but rather based on user input. One immediate application is execution of `simcomps`.
* This makes the `sampler` class less heavy as bonus.

Not sure to be honest if the sampler works correctly atm, I just want to commit WIP, therefore, the don't merge label. But it *seems* to work :-), and the code is clean enough to review.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -319,6 +319,7 @@ AC_CONFIG_FILES([tests/stack/Makefile\
                  tests/fast3d/Makefile\
                  tests/time_scheme/Makefile\
                  tests/time_scheme_controller/Makefile\
+                 tests/time_based_controller/Makefile\
                  tests/point_interpolation/Makefile])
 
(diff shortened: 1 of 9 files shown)


njansson (approved):

njansson:
Looks good, I would say that we can merge this, as noted it makes the `sampler` much cleaner.

timofeymukha:
OK, good. But let me add some tests first before we merge.

timofeymukha:
@njansson I think I am ok with merging. As far as I could test, the behaviour is the same as on develop.

PR #928, opened 2023-08-14
Add methods to pass down simulation time and/or time-step to bcs

### LINKED ISSUE #902: Time and time-step in user defined boundary conditions
Certain user defined boundary conditions needs the current simulation time and/or time-step. Therefore we need to extend the boundary condition methods (and user interfaces) to accommodate for this.

This will be a breaking change for most provided examples.

njansson:
This extends the bc definition to allow for (optional) passing down current simulation time and/or time-step, fixes #902.

**Note** this breaks most user provided bcs

### DIFF
--- a/examples/cyl_boundary_layer/cyl_bl.f90
+++ b/examples/cyl_boundary_layer/cyl_bl.f90
@@ -64,7 +64,7 @@ contains
     end do
   end subroutine cylinder_gen_curve
   
-  subroutine user_inflow_eval(u, v, w, x, y, z, nx, ny, nz, ix, iy, iz, ie)
+  subroutine user_inflow_eval(u, v, w, x, y, z, nx, ny, nz, ix, iy, iz, ie, t, tstep)
     real(kind=rp), intent(inout) :: u
     real(kind=rp), intent(inout) :: v
     real(kind=rp), intent(inout) :: w
@@ -78,6 +78,8 @@ contains
     integer, intent(in) :: iy
     integer, intent(in) :: iz
     integer, intent(in) :: ie
+    real(kind=rp), intent(in), optional :: t
+    integer, intent(in), optional :: tstep
     real(kind=rp) ::  u_th,dist,th, yy
     real(kind=rp) ::  arg
 
--- a/examples/lid/lid.f90
+++ b/examples/lid/lid.f90
@@ -26,7 +26,7 @@ module user
   end subroutine user_setup
 
   ! user-defined boundary condition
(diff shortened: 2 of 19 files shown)


timofeymukha:
I don't have time to look at the new code in detail, but conceptually we need this for any kind of transient bc, right? So I would say go ahead regarding the extension of the `bc.f90` decendents. 

But I am not totally convinced we should change the user interface, which I guess is not connected with the changes in bc.f90 stuff programmatically, right? Suppose for a second that we pass the `case` to one of the user routines, `user_check` for example. Then one could create a global variable inside the user module, assign it to `case%dt` in `user_check` at each timestep, and then use that global inside any of the user routines, right? 

My concern here is that this could be a slippery slope in terms of the number of parameters we need to pass. I would like to hope that going forward, more and more "standard" situations will be already in the code, without requiring the user module. By consequence, the user routines will need access to more and more obscure stuff to cover corner cases. Would it be better to just give access to `case` and be done with it? I am actually quite comfortable with that here: the user module is for advanced users, so I think the philosophy of "you can do anything, but you better know what you are doing" is applicable. Another plus is that we will probably have less instances of breaking everybody's user files due to an interface update.

njansson:
> I don't have time to look at the new code in detail, but conceptually we need this for any kind of transient bc, right? So I would say go ahead regarding the extension of the `bc.f90` decendents.
> 
> But I am not totally convinced we should change the user interface, which I guess is not connected with the changes in bc.f90 stuff programmatically, right? Suppose for a second that we pass the `case` to one of the user routines, `user_check` for example. Then one could create a global variable inside the user module, assign it to `case%dt` in `user_check` at each timestep, and then use that global inside any of the user routines, right?
> 
> My concern here is that this could be a slippery slope in terms of the number of parameters we need to pass. I would like to hope that going forward, more and more "standard" situations will be already in the code, without requiring the user module. By consequence, the user routines will need access to more and more obscure stuff to cover corner cases. Would it be better to just give access to `case` and be done with it? I am actually quite comfortable with that here: the user module is for advanced users, so I think the philosophy of "you can do anything, but you better know what you are doing" is applicable. Another plus is that we will probably have less instances of breaking everybody's user files due to an interface update.

Yes, but that could also be rather dangerous, allowing a user to directly modify the solver pipeline. In this case I would rather say that passing down `t` and `tstep`  is needed to make it compatible with the all builtin bcs, be it transient or not.

timofeymukha:
Alright, let's go ahead then and see how things play out later.

PR #941, opened 2023-08-29
added some enhancements to the probe module and modified the example …

adperezm:
In src/io/csv_file.f90 I modified the initialization of the probe module so I can have a custom number of fields

In examples/gslib-findpts/user.case I added a bunch of parameters related to the probes

In examples/gslib-findpts/user.f90 I modified the code to match with what I added

In src/common/probes.f90 i modified a bit how things work.

In general, I added a bit of flexibility with regards to how many fields are "probed". I also made it in such a way that any field in the field registry can be used, and in the json file you can specify which field you want. I do this with the hope that I can easily retrieve temperature or statistic quantities.

I also made it such that the interpolation happens in a loop such that the number of fields is not hard coded. For this I had to modify the input to %interpolate() such that it recieves an array of pointers to fields, rather than the field itself.

Lastly (I think) I added a time based controller that controls how often the sampling is done. This works in the same manner as the normal sampler, i.e, in nstep, nsamples, simulation time modes.

Aditionally I added some memory management functions to copy data from GPU to CPU when interpolating with gslib.

I have tested this with ONE gpu :).

My task now will be to add an improved example, i.e, a plane in RBC


examples/rayleigh-benard-cylinder-probes/rayleigh.f90
```diff
@@ -0,0 +1,242 @@
...
+    call json_get(params, 'case.probes.n_fields', n_fields)
+    call json_get(params, 'case.probes.pts_in_filename', pts_in_filename) 
+    call json_get(params, 'case.probes.pts_out_filename', pts_out_filename) 
+
+    !> Define the input file object
+    input_file = file_t(trim(pts_in_filename))
+
+    !> Read file and initialize the probe object
+    !! Pre initialize the number of fields according to case file
+    pb%n_fields = n_fields
```
vbaconnet, line 149: Maybe we can do this in a way that initializes `n_fields` from the size of the `case.probes.fields` array? That way it eliminates the parameter from the json file.

src/io/csv_file.f90
```diff
@@ -303,7 +303,12 @@ subroutine csv_file_read_probes(f, p)
     n_fields = 4 ! # of fields hardcoded here, u,v,w,p
```
vbaconnet, line 303: Maybe we can change this comment here and/or update the subroutine docstring to explain that if nothing is specified we revert to the Nek5000 default of 4 fields `u,v,w,p`? Or I guess that we can still work on in the `feature/probes` branch before a merge in `develop`.

src/common/probes.f90
```diff
@@ -60,10 +66,17 @@ module probes
...
+     type(time_based_controller_t), allocatable :: controllers(:)
+     ! Fields to be probed
+     type(field_list_t) :: int_fields
+     character(len=20), allocatable  :: which_fields(:)
 
      contains
        !> Initialize probes object.
        procedure, pass(this) :: init => probes_init
+       !> Initialize user defined parameters.
+       procedure, pass(this) :: usr_init => probes_usr_init
```
vbaconnet, line 79: Should the name of this procedure reflect that it is reading values from the json case file, like `init_from_casefile`?? Or whatever we can come up with? Maybe another `usr_init` among everything else will be confusing?

MartinKarp:
I think it looks very nice.

src/common/probes.f90
```diff
@@ -202,49 +268,54 @@ subroutine probes_map(this, coef)
...
+    logical, intent(inout) :: write_output
+
+    !> Supporting variables
+    integer :: il 
+    integer :: n
+    
+    n = this%int_fields%fields(1)%f%dof%size()
+
+    !> Check controller to determine if we must write
+    if (this%controllers(1)%check(t, tstep, .false.)) then
```
vbaconnet, line 291: I have to say the `time_controller` stuff is very cool and I had no idea it was in Neko until now hahah! Only thing I am worried about is to have all of this in the `interpolate` function, whose only purpose is to perform the interpolation. Maybe the output control could be done from the `csv_file` side, like calling `fout%write(pb)`? Or at least something that clearly separates the computation from the output control?

vbaconnet (approved):
I love it! The comments are just for discussion and questions I had, I guess since everything works the way it should we can just merge it!

examples/rayleigh-benard-cylinder-probes/rayleigh.f90
adperezm: I wanted that but I did not find a way to do it right. It might be very easy but I just did not come up with it, let's talk about it!

src/io/csv_file.f90
adperezm: Yeah that sounds good! I think for develop we might need to add tests or so, no? So I guess we need to stay in feature/probes for the moment.

timofeymukha:
Sorry, I've done this in a conventional notepad notes way. Here comes.

- Should probes be a simulation component? It seems to fit the mold, or? We need to snap a time-based controller to the simcomp base-class, that was in fact why I originally created that controller class.
  ...OK, I see nowthat you do everything in the user file now, including parsing the JSON. We could perhaps merge this as is in the spirit of continuous integration :-) and convert to a simcomp in a later PR. But for later stuff, always keep in mind "does this fit the simcomp execution pattern?".
(`probes_usr_init` would correspond to the `init` for simcomps.)

- Probes json:
  * Can `n_fields` be inferred from `fields` at runtime? Search for how `boundary_types` is handled in the code now, or the `neko_simcomps` array init. 
  * A suggestion `pts_in_filename` -> `locations` or `points_file`; `pts_out_filename` -> `output_file`. I don't insist on these in particular, but my beef is with `pts`, I really don't want abbreviations is the user interface, unless it is something very established like "RANS" or whatever.
  * Do we really need a subobject for the `sampler`?
  * Let's use the same parameter names for the sampler everywhere, which is currently `output_control` and `output_value`. I know the last one is not ideal, beer goes to the person who invents a good one, and then we change everywhere at once. Use this also in the code, e.g. in `probe_usr_init`.

- I am annoyned by the fact that we have both - and _ in examples' folder names. Can't be settled in this PR, but we should fix something. I would say _ is kind of a more common thing for folder in linux, or?

- probes.f90: use `only` in all `use` statements.

- probes.f90: default to `private` at module scope.

- probes.f90 `usr` -> `user` everywhere.

- probes_t: Either all attribute doctstrings on the right, or all on top. All on top preferred since length can be arbitrary and multiline if needed. You also use "on top" for procedures.

- probes_t: `int_fields` -> `fields` or `sampled_fields`? The thing is that `int` is ambiguous, which is demonstrated by the fact that your docstring for `out_fields` is "interpolated fields", so that could also be `int`.

- probes_t: why is there an array of time controllers? Init seems to only allocate 1. If we do need them, they should be deallocated in the destructor, which I think is forgotten.

- csv_file.f90: I think the probe reading subroutine does not belong here, and should be moved to probes_t. Let csv_file only implement the stuff needed to work with csv file per se.

- probes_t: Adding some unit tests would be very nice :-). 

- rbc example: please don't use `int_pos.csv` even it is familiar to us FRAMEWORK users, why not simply `probes.csv`? We can of course use whatever name in our own cases, but for examples we should be as clear as possible with everything.

src/common/probes.f90
timofeymukha: The separation might be a good idea, but leave `csv_file` out of it, just splint into subroutines inside the probes type as necessary. The `csv_file` should know nothing about probes or any future classes that use it.

src/io/csv_file.f90
timofeymukha: @adperezm See how  `n_simcomps` is assigned here
https://github.com/ExtremeFLOW/neko/blob/develop/src/simulation_components/simulation_component_global.f90

src/common/probes.f90
vbaconnet: Yes I was wondering if the probes should actually be in the csv file. The reason why I put it there is to avoid reading the probe coordinates with an intermediate `matrix_t` object which `csv_file%read()` requires... So I wanted to read into the file and put the coordinates directly in `probe%xyz`, but maybe there is a nice way to do it? Like declaring `probe%xyz` as type `matrix_t`??

timofeymukha:
OK, I now see that this is a merge into `feature/probes` and not into develop :D. Then I would say merge at your own convenience, but keep my comments in mind for the later merge into develop. I will set my review to green.

timofeymukha (approved):

vbaconnet:
> OK, I now see that this is a merge into `feature/probes` and not into develop :D. Then I would say merge at your own convenience, but keep my comments in mind for the later merge into develop. I will set my review to green.

Hahaha no worries. Niclas suggested the simulation component thing and I agree that it is the nicest way to do it.

In the meantime we will work on the other things you mentioned! Thanks a lot for the feedback

examples/rayleigh-benard-cylinder-probes/rayleigh.f90
timofeymukha: See my wrongly placed comment below :P. https://github.com/ExtremeFLOW/neko/blob/develop/src/simulation_components/simulation_component_global.f90

src/common/probes.f90
timofeymukha: If `probes%xyz` being a `matrix_t` works out, then why not. Otherwise, just convert from the matrix to whatever you need in the probes type. This is a one-time operation, after all, and for a truly huge probe numbers a csv is probably not the way to go anyway.

adperezm:
We need to merge this to develop sometime soon anyway, so we might as well start changing it from now.

njansson:
If it works, I would suggest get this into `develop` more or less directly. After that we can add the simulation components and cleanup with c-interfaces etc.

src/common/probes.f90
adperezm: I was trying to do just that now. But I can't get neko to compile if I use file inside the probe module :')

adperezm:
Hi!

I think I addressed most of the things suggested by tim now.

-  there is now only one init routine in probes_t. It takes the data from Json, reads the probe location file, and allocate arrays.
-  n_fields is now retrieved automatically thanks to the example by tim.
-  the file names have been changed.
- removed the sampler object from the case file
- removed probe_t from the csv_file_t. Now the probe type reads the data it needs using the csv_file_t

Things missing:

- Having a unified comment format. Although I think it is consistent most of the time.
- Unit tests.
- Changing the name of the example.

Regarding the last point. Should we just keep one RBC example? I am not quite sure why we have so many. Even among us the confusion has existed when we talk about the example (it is usually me who is confused :) ).

Reminder that this merge is to the probes branch such that we can delete the probes_ap branch. Then we can push to develop when agreed by all.

njansson:
> 
> Regarding the last point. Should we just keep one RBC example? I am not quite sure why we have so many. Even among us the confusion has existed when we talk about the example (it is usually me who is confused :) ).
> 

Either is fine for me, the only thing I would like to avoid is to pollute the examples folder in `develop` with very specific cases, e.g. someones production case etc, the examples should be rather vanilla

> Reminder that this merge is to the probes branch such that we can delete the probes_ap branch. Then we can push to develop when agreed by all.

Anyone with commit rights should be able to merge this into `probes`, so please go ahead when you think it is ready

PR #950, opened 2023-09-03
Adds output control "never" and throw for non-positive nsamples

timofeymukha:
* Adds `never` as an option for the `time_based_controller`.  Useful for benchmarking or having a simcomp do no output, for example.
* Forces `nsamples` to be positive.

### DIFF
--- a/doc/pages/case-file.md
+++ b/doc/pages/case-file.md
@@ -49,6 +49,7 @@ The three following options are possible.
 2. `tsteps`, then `_value` is the number of time steps between the outputs.
 3. `nsamples`, then `_value` is the total number of outputs that will be
    performed in the course of the simulation.
+4. `never`, then `_value` is ignored and output is never performed.
 
 
 ## The `case` object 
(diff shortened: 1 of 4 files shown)


njansson (approved):
Approved, but one question, will this still allow me to prevent sampling, but perform the simcomps? I can foresee that one would like to skip outputs, but do the components during benchmarking?

timofeymukha:
If I understand you correctly then yes, because computing and outputting have serparate controllers. So you would do something like.
```json
{
"output_control": "never",
"compute_control": "tsteps",
"compute_value": 1
}
```

njansson:
> If I understand you correctly then yes, because computing and outputting have serparate controllers. So you would do something like.
> 
> ```json
> {
> "output_control": "never",
> "compute_control": "tsteps",
> "compute_value": 1
> }
> ```

True, since it's an own time-controller and now global as for some reason thought

timofeymukha:
Rayleigh in reframe has nsamples 0 :P.

timofeymukha:
Added an if-branch in case.f90 to set output_value to 0.0 (a dummy value) when output_control for fluid is "never"

njansson (approved):

PR #957, opened 2023-09-04
Doc/simcopms

timofeymukha:
Adds some doc text for simcomps in a separate page. Should be further populated as we add more simcomps.


njansson (approved):
Consider adding the new simcomps page to the top, (index file)

PR #958, opened 2023-09-04
Added a modified probe object to perform mesh to mesh interpolations

adperezm:
DO NOT MERGE.

I just want this here so I can get early feedback, since I need some ideas.

I modified the probe type to make it "more general" In the sense that it does not need to read a file but instead receives arrays.

Now, the motivation for this is that:
1. I want to be able to restart from other meshes.
2. I want to be able to modify the mesh at runtime and "shift" it to comply with some requirement in the spectral error indicator. 

In both cases, I would need to have the coordinates of both meshes, and then the field values in one of them.

Do you have comments on how you want this done? I do not want to include a bunch of code to then need to change it immediately.

### DIFF
--- a/examples/rayleigh-benard-cylinder-probes/rayleigh.case
+++ b/examples/rayleigh-benard-cylinder-probes/rayleigh.case
@@ -4,7 +4,7 @@
     "mesh_file": "cylinder.nmsh",
     "output_at_end": true,
     "output_boundary": true,
-    "end_time": 250,
+    "end_time": 1,
     "timestep": 2e-3,
     "output_checkpoints": true,
     "checkpoint_control": "simulationtime",
@@ -55,7 +55,7 @@
         "output_file":  "output.csv",
         "fields": ["w","s"],
         "output_control": "simulationtime",
-        "output_value"    : 1,
+        "output_value"    : 0.1,
     }
 }
 }
--- a/examples/rayleigh-benard-cylinder-probes/rayleigh.f90
+++ b/examples/rayleigh-benard-cylinder-probes/rayleigh.f90
@@ -1,5 +1,6 @@
 module user
   use neko
+  use mesh_to_mesh_interpolator, only : mesh_to_mesh_interpolator_t
   implicit none
 
   !> Variables to store the Rayleigh and Prandlt numbers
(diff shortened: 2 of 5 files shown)


examples/rayleigh-benard-cylinder-probes/rayleigh.case
```diff
@@ -55,7 +55,7 @@
         "output_file":  "output.csv",
         "fields": ["w","s"],
         "output_control": "simulationtime",
-        "output_value"    : 1,
+        "output_value"    : 0.1,
```
njansson, line 58: We should take care not to update the examples unless needed, i.e. simply adding the case file via a `git commit -a`

examples/rayleigh-benard-cylinder-probes/rayleigh.f90
```diff
@@ -1,5 +1,6 @@
 module user
   use neko
+  use mesh_to_mesh_interpolator, only : mesh_to_mesh_interpolator_t
```
njansson, line 3: We should put this in the Neko master module.

The idea is that a user only needs to import `Neko` and get everything

njansson (commented):
Looks good, only some minor comments to consider.

timofeymukha:
But wait, this is what we wanted to do with the probes_t in the first place:  decouple all the IO and put into a simcomp, which would use probes_t as backend by giving it the arrays. Then the mesh_to_mesh will just be a thin wrapper around the probes_t.

adperezm:
> But wait, this is what we wanted to do with the probes_t in the first place: decouple all the IO and put into a simcomp, which would use probes_t as backend by giving it the arrays. Then the mesh_to_mesh will just be a thin wrapper around the probes_t.

I suppose this is right. However interpolation is not a simulation component. It is something that should be in the core and be used by others.

So I guess Porbes_t should be in simulation components and use the interpolador type to sample said probes.

timofeymukha:
> > But wait, this is what we wanted to do with the probes_t in the first place: decouple all the IO and put into a simcomp, which would use probes_t as backend by giving it the arrays. Then the mesh_to_mesh will just be a thin wrapper around the probes_t.
> 
> I suppose this is right. However interpolation is not a simulation component. It is something that should be in the core and be used by others.
> 
> So I guess Porbes_t should be in simulation components and use the interpolador type to sample said probes.

I mean sure, I guess it is a matter of which of the type you want to call `probes_t`, the one that does the interpolation or the simcomp. Maybe you are right and `probes_t` should be the simcomp. Then we need a good name for the interpolator. `point_interpolator_t` is taken by the type doing local interpolation in the element. We could refactor that one to `local_point_interpolator_t`, and then use `point_interpolator_t` for the one working with arbitrary points.

PR #959, opened 2023-09-05
Probes with point interpolator

vbaconnet:
Got rid of `findpts_eval` for the interpolation in `probes%interpolate`. Also removed `coef_t` from point_interpolator as it is not using it, and adjusted the RBC example.

### DIFF
--- a/src/common/probes.F90
+++ b/src/common/probes.F90
@@ -56,7 +56,6 @@ module probes
   use point_interpolator
   use point, only: point_t
   use space, only: space_t
-  use dofmap, only: dofmap_t
   implicit none
   private
   
--- a/src/common/probes.F90
+++ b/src/common/probes.F90
@@ -56,6 +56,7 @@ module probes
   use point_interpolator
   use point, only: point_t
   use space, only: space_t
+  use dofmap, only: dofmap_t
   implicit none
   private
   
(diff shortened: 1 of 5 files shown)


njansson (approved):

timofeymukha:
Nice, gslib now only searches for elements and then we use the interpolator you added before? For the space variable, I guess it is  canonical to use `Xh`, with a capital x, as an exception to the general rule. This mimics the math notation in some books.

src/common/probes.F90
```diff
@@ -362,17 +370,18 @@ subroutine probes_interpolate(this, t, tstep, write_output)
...
-                     this%sampled_fields%fields(il)%f%x(:,:,:,this%el_owner(i)))
+                     this%sampled_fields%fields(il)%f%x(:,:,:,this%el_owner(i)+1))
 
                 local_out_field(i) = my_interpolated_field(1)
              end if
           end do
 
           ! Artificial way to gather all values to rank 0
-          call MPI_Reduce(local_out_field, this%out_fields(il,1), this%n_probes, &
+          call MPI_Reduce(local_out_field, this%out_fields(1, il), this%n_probes, &
```
adperezm, line 383: Smart :)

vbaconnet: Yes, and also about out_fields I reshaped the array so we don't need to transpose it for the csv_file output

MartinKarp (approved):
Tremendous work Bacon 1!

PR #960, opened 2023-09-05
Feature lambda2

MartinKarp:
Computes lambda 2 every time step as a simulation component and puts it in the next free field in the fluid output, as Tim talked about we should make it easy to control the frequency and output of the simulation component. Also, now it is technically the last time step and not the current time step that is sampled, because of the ordering of things.


src/simulation_components/simulation_component_global.f90
```diff
@@ -65,13 +65,12 @@ subroutine simcomps_global_init(case)
 
       call case%params%get_core(core)
       call case%params%get('case.simulation_components', simcomp_object, found)
-
+      print *, n_simcomps
```
timofeymukha, line 68: Accidental, right? I guess the whole file is actually unchanged.

njansson, line 68:
```suggestion
```

njansson (changes requested):

timofeymukha:
Very nice! You need to update just a bit to the new simcomp interface now that I've added the execution and output controllers.

njansson (approved):

timofeymukha (approved):

PR #961, opened 2023-09-05
Change order of simulation execution

MartinKarp:
I think this order makes the most sense, but we should also add something so that the simcomps can be executed before the time step even starts. Another thought is that perhaps everything should be in a list of simulation components? Then the ordering would be implied by the order in the list.


njansson (approved):

timofeymukha:
As discussed, I agree witht he simcomp move at least, but I guess we wait for @pschlatt1 to merge.

src/simulation.f90
```diff
@@ -123,18 +123,19 @@ subroutine neko_solve(C)
...
             call neko_simcomps(i)%simcomp%compute(t, tstep)
          end do
        end if
+
+       call C%usr%user_check(t, tstep,&
+            C%fluid%u, C%fluid%v, C%fluid%w, C%fluid%p, C%fluid%c_Xh, C%params)
+
+       call C%q%eval(t, C%dt, tstep)
+       call C%s%sample(t, tstep)
+
```
njansson, line 138:
```suggestion

       call C%q%eval(t, C%dt, tstep)
       call C%s%sample(t, tstep)
         
       call C%usr%user_check(t, tstep,&
            C%fluid%u, C%fluid%v, C%fluid%w, C%fluid%p, C%fluid%c_Xh, C%params)

```

PR #962, opened 2023-09-07
Run-time selectable sources

timofeymukha:
This PR introduces run-time-selectable sources for the fluid (for now), controlled by the JSON, much like the simcomps right now.
## Breaking user changes
* In the case file there is now a possibility to use a `source_terms` array of JSON objects, configured much like the simcomps, e.g. with a `type` keyword.
* Currently 3 types are available: "constant", "user_vector", and "user_pointwise".
* The "constant" source requires an array of reals for the value for each component (3 for fluid, and later 1 for scalar, which is not hooked in now).

## Code design

* The json-driven component is designed just like the simcomps (base class `source_term_t`), there is a deferred constructor, `init`, which takes the json and some other stuff.  Inside the concrete source type, there is a constructor from type components that is called by the json constructor after it has parsed all the necessary stuff from the json. This way we can, in principle, init the source without json if needed.
* The sources get a `field_list`, which should contain the fields for the RHS for the given problem. So for a fluid 3 fields, and for scalar 1, meaning that we should be able to use the same types for both. The sources then directly *add* their contributions directly to these fields. I did this to avoid any copying here. 
* Inside `fluid_scheme` there are now 3 components for the RHS: `f_x`, `f_y`, `f_z` and an array for the sources. In the `step`, there is a loop through the sources, each of them adding there contribution to the RHS.
* The interface for the PnPn residual subroutines has been changed from accepting one `f_Xh` raw array to the 3 fields`f_x`, `f_y`, `f_z`. This is actually more consistent with how we pass `u`, `v`, `w` as 3 fields.
* The user source is fit into the same type structure, the only difference being that it does not support init from json, so that we don't have to pass the user source routine pointers to the factory. It also makes a copy-addition under the hood, from the u, v, w arrays exposed to the user (than can be *assigned* to, like before) to the rhs fields in the `fields` component.


njansson:
> Some reframe template is probably off..

Seems to be related to parsing the json into `real=sp`, restarting usually works.

examples/cyl_boundary_layer/cyl_bl_user.case
```diff
@@ -18,9 +18,6 @@
         "rho": 1,
         "mu": 1,
         "Re": 500,
-        "source_term": {
-            "type": "noforce"
-        },
```
njansson, line 23: If I read things correctly, "noforce" will be the default do nothing approach, and instead of today, by dummy zeroing vectors?

src/fluid/fluid_pnpn.f90
```diff
@@ -360,7 +362,7 @@ subroutine fluid_pnpn_step(this, t, tstep, dt, ext_bdf)
     associate(u => this%u, v => this%v, w => this%w, p => this%p, &
          du => this%du, dv => this%dv, dw => this%dw, dp => this%dp, &
          u_res =>this%u_res, v_res => this%v_res, w_res => this%w_res, &
-         p_res => this%p_res, Ax => this%Ax, f_Xh => this%f_Xh, Xh => this%Xh, &
+         p_res => this%p_res, Ax => this%Ax, user_source_term => this%user_source_term, Xh => this%Xh, &
          c_Xh => this%c_Xh, dm_Xh => this%dm_Xh, gs_Xh => this%gs_Xh, &
          ulag => this%ulag, vlag => this%vlag, wlag => this%wlag, &
          msh => this%msh, prs_res => this%prs_res, &
```
njansson, line 368:
```suggestion
         p_res => this%p_res, Ax => this%Ax, Xh => this%Xh, &
         c_Xh => this%c_Xh, dm_Xh => this%dm_Xh, gs_Xh => this%gs_Xh, &
         ulag => this%ulag, vlag => this%vlag, wlag => this%wlag, &
         msh => this%msh, prs_res => this%prs_res, &
         user_source_term => this%user_source_term
```

Some minor formatting to keep within col 80

src/fluid/fluid_pnpn.f90
```diff
@@ -380,19 +383,41 @@ subroutine fluid_pnpn_step(this, t, tstep, dt, ext_bdf)
...
+         call copy(this%f_y%x, user_source_term%v, n)
+         call copy(this%f_z%x, user_source_term%w, n)
+      end if
+
+      ! Add contribution from all source terms.
+      if (allocated(this%source_terms)) then
+         do i=1, size(this%source_terms)
+            call this%source_terms(i)%source_term%compute(t, tstep)
+         end do
+      end if
```
njansson, line 405: How about encapsulating this into an own module, to keep the step loop rather clean? Then we can also make some opt. that the current setup will prevent, e.g. combined copy loops, merged kernels etc.

src/source_terms/const_source_term.f90
```diff
@@ -0,0 +1,132 @@
...
+
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       do i=1, n_fields
+          call device_cadd(this%fields%fields(i)%f%x_d, this%values(i), n)
+       end do
+    else
+       do i=1, n_fields
+          call cadd(this%fields%fields(i)%f%x, this%values(i), n)
+       end do
+    end if
```
njansson, line 129: Would suggest to (in the future) break out the compute into separate backend folders, such that we can avoid e.g. launching `n_fields` kernels, or having `n_fields` calls to `cadd` which a compiler could fuse if given more information

njansson (commented):
Nice! Some comments, mostly structure related to allow for future optimisations

examples/cyl_boundary_layer/cyl_bl_user.case
timofeymukha: Correct, this object has no effect anymore. If there is no "source_term**s**" at all in the JSON, then the user force term defaults to being zero-valued as previously via "noforce". What is a bit concerning maybe is that the `source_term_t` array in fluid then remains unallocated, which is a bit unsatisfactory. Maybe one should add a single dummy zero-valued source term? But that's some extra operations for nothing. Or maybe we can make the array private to keep away people from touching it.

src/fluid/fluid_pnpn.f90
timofeymukha: Yes good. We can have a thin `fluid_source_term_t` type having the `fluid_user_source_term_t` and the `source_term_t` array as components, and having the above code in a `compute` method. We can put the init logic there as well eventually. Our main types are becoming a bit too large to navigate conveniently, imho, `fluid_scheme_t` being one of them.

src/source_terms/const_source_term.f90
timofeymukha: Yes, good point!

src/fluid/fluid_pnpn.f90
timofeymukha: Now branched out.

src/source_terms/const_source_term.f90
timofeymukha: Now split, but I basically have the same loops in the kernels, so not sure if it helps much.

timofeymukha:
The pipe example works on Dardel-GPU, so I am removing "don't merge"! We should check if something weird crept in during the merge with develop today.

njansson (approved):
Nice!

One minor observation, so we're going to `x,y,z` instead of `u,v,w`?

timofeymukha:
> Nice!
> 
> One minor observation, so we're going to `x,y,z` instead of `u,v,w`?

In `fluid_user_source_term_t` we retain `u, v, w`.
Do you mean `f_x, f_y, f_z` in `fluid_scheme_t`?

njansson:
> > Nice!
> > One minor observation, so we're going to `x,y,z` instead of `u,v,w`?
> 
> In `fluid_user_source_term_t` we retain `u, v, w`. Do you mean `f_x, f_y, f_z` in `fluid_scheme_t`?

True, yes this referred to `f_x` and friends. But we leave it as is now.

timofeymukha:
> > > Nice!
> > > One minor observation, so we're going to `x,y,z` instead of `u,v,w`?
> > 
> > 
> > In `fluid_user_source_term_t` we retain `u, v, w`. Do you mean `f_x, f_y, f_z` in `fluid_scheme_t`?
> 
> True, yes this referred to `f_x` and friends. But we leave it as is now.

OK. I don't have a strong opinion here, since this is in fluid, f_u etc would make sense as well.

examples/rayleigh-benard-cylinder-probes/rayleigh.case
```diff
@@ -20,9 +20,12 @@
         "rho": 1,
         "mu": 1,
         "Re": 1e4,
-        "source_term": {
-            "type": "user_vector"
-        },
+        "source_terms": 
+        [
+            {
+                "type": "user_vector"
```
MartinKarp, line 26: Would it be an idea to do the same for `user_check`in simulation components as well? (And perhaps even for writing the output?). This would also address what #961 . Not immediately related, but I quite like this way of adding the user functions.

MartinKarp: Same applies for the other examples as well.

examples/turbpipe/pipe.f90
```diff
@@ -77,7 +77,7 @@ end function pipe_ic
 
   !> Forcing for if not using vol_flow
   subroutine forcing(f, t)
-    class(source_t), intent(inout) :: f
+    class(fluid_user_source_term_t), intent(inout) :: f
     real(kind=rp), intent(in) :: t
```
MartinKarp, line 81: Would it be an idea to update the .f90 to also use Re_B and Re_tau from the json_file? Or is this hard to incorporate? I see we use a constant forcing now, but I guess it would help to illustrate the power of our new case parameters to use Re_B and Re_tau here? Otherwise it might also be an idea to remove this part

examples/turbpipe/pipe_source.case
```diff
@@ -19,9 +19,13 @@
...
         "mu": 1,
         "Re": 2650,
-        "source_term": {
-            "type": "user_vector"
-        },
+        "source_terms": 
+        [
+            {
+                "type": "constant",
+                "values": [0, 0, 0.0092274831]
```
MartinKarp, line 26: I see, this is the computed value from Re_tau=180 and Re_B=5300 I guess. This is maybe more elegant. I don't feel strongly one way or the other.

src/fluid/fluid_source_term.f90
```diff
@@ -0,0 +1,198 @@
...
+    integer :: n_sources, i
+
+    call this%free()
+
+    this%f_x => f_x
+    this%f_y => f_y
+    this%f_z => f_z
+
+    ! NOTE: the user source procedure pointer is set externally in case.f90 !
+    call this%user_source_term%init(f_x%dof)
```
MartinKarp, line 106: Maybe this runs into circular dependencies, or doesn't align with how we want the code to be, but would it be possible to let the user_source_term be a special case of json_source_terms, similarly to `constant`? I don't immediately see why we want this split between the two.

MartinKarp (commented):
Looks good, Im mostly wondering whether we want this split between user and other source terms. Maybethis would be rather intrusive, but my first thought was that it would be quite nice.

examples/turbpipe/pipe.f90
timofeymukha: Yeah, actually the whole point is that we don't need the forcing function anymore since we use the built-in constant source term. So we should remove it from the f90.

examples/turbpipe/pipe_source.case
timofeymukha: Yeah, the idea was to use this as a demo for the constant source term. I added the formula to compute this value to the readme.

src/fluid/fluid_source_term.f90
timofeymukha: The different software-wise is that the "ordinary" source terms just operate by overloading the `compute` TBP, whereas the user term instead requires two procedure pointers, for `eval` and the `eval_pw`. So the constructor signatures are different, etc. But in principle merging these would be good. Maybe one can have the `compute` as a procedure pointer as well, and then the decendents implement their own compute and a assign the pointer in the constructor? I should give this some thought.

examples/rayleigh-benard-cylinder-probes/rayleigh.case
timofeymukha: I guess the principle question is whether we want to assign meaning to the order of the stuff we have in these types of arrays.

src/fluid/fluid_source_term.f90
timofeymukha: Ok, the issue with the procedure pointer approach is that we cannot do as proposed here
https://fortranwiki.org/fortran/show/Polymorphism+for+procedure+pointers+inside+derived+types

So, if I have a procedure in the base class
````fortran
 type, abstract, public:: source_term_t
     !> The fields to be updated with the source term values
     type(field_list_t) :: fields
     !> Coefficients for the SEM.
     type(coef_t), pointer :: coef => null()
     !> 
     procedure(source_term_compute), pass(this), pointer  :: eval => null()
````
with 
```fortran
  abstract interface
     subroutine source_term_compute(this, t, tstep)  
       import source_term_t, rp
       class(source_term_t), intent(inout) :: this
       integer, intent(in) :: tstep
     end subroutine
  end interface
```
Then in the derived type, `this`, still has to be `class(source_term_t)`, and not the derived type. So then, the sbroutine eval points to loses access to the components of the derived type, which is not acceptable.

I will think a bit more :-)

adperezm: This looks quite nice!

njansson: > Ok, the issue with the procedure pointer approach is that we cannot do as proposed here https://fortranwiki.org/fortran/show/Polymorphism+for+procedure+pointers+inside+derived+types
> 
> So, if I have a procedure pointer in the base class
> 
> ```fortran
>  type, abstract, public:: source_term_t
>      !> The fields to be updated with the source term values
>      type(field_list_t) :: fields
>      !> Coefficients for the SEM.
>      type(coef_t), pointer :: coef => null()
>      !> 
>      procedure(source_term_compute), pass(this), pointer  :: eval => null()
> ```
> 
> with
> 
> ```fortran
>   abstract interface
>      subroutine source_term_compute(this, t, tstep)  
>        import source_term_t, rp
>        class(source_term_t), intent(inout) :: this
>        real(kind=rp), intent(in) :: t
>        integer, intent(in) :: tstep
>      end subroutine
>   end interface
> ```
> 
> Then in the derived type, `this`, still has to be `class(source_term_t)`, and not the derived type. So then, the sbroutine eval points to loses access to the components of the derived type, which is not acceptable.
> 
> I will think a bit more :-)

It should still be possible to have the derived type since `source_term_t` is the base type. But it would then need a `select` block inside `eval` to access the derived type specific data, or maybe I'm missing something?

njansson: This is for example how the `hash` function in `htable.f90` is implemented, base type as a class in the interface,
https://github.com/ExtremeFLOW/neko/blob/b5ae32b5626d96af3bfb42594abc9ddcc2e0ce7c/src/adt/htable.f90#L67-75
 and `select` block inside each specialisation, for example
 https://github.com/ExtremeFLOW/neko/blob/b5ae32b5626d96af3bfb42594abc9ddcc2e0ce7c/src/adt/htable.f90#L1226-1261

njansson: > As preliminary work, I will add `user_t` the `fluid_scheme_t` constructors, so that it can initialize the user source term itself instead of `case_t` having to shove it in after `fluid_scheme_t%init` has already been run.

But we should not forget that `user_t` is not only `fluid` related, it can be rather generic.

timofeymukha: Indeed, the same thing happens e.g. with scalar now, with case setting the source.

timofeymukha:
Alright, nice CI is happy. So, ready for "re-review" I guess. Thanks for the nudge towards merging the user term into the `source_term_t` heirarchy, it was a good decision. There is only one way in which the user terms is special, and that is that it should be initialized directly from components using it's own consturctor and not through the `source_term_factory`, which uses the deferred constructor from JSON. The reason is that I don't want to mess with all the user-defined source procedures inside the factory and have them as part of the shared init interface. If you think it would be better to completely unify things, we can dot that. 

Things are also much cleaner now in the case.f90 since it does need to deal with the user source term.

njansson (commented):
Not sure if I follow, so fluid related source terms are handled by the fluid scheme, but scalar ones are still done with an if block in the case file?

timofeymukha:
> Not sure if I follow, so fluid related source terms are handled by the fluid scheme, but scalar ones are still done with an if block in the case file?

Correct, because this PR does not touch scalar sources at all.

src/fluid/fluid_source_term.f90
```diff
@@ -0,0 +1,209 @@
...
+       rhs_fields%fields(1)%f => f_x
+       rhs_fields%fields(2)%f => f_y
+       rhs_fields%fields(3)%f => f_z
+
+       call json%get_core(core)
+       call json%get('case.fluid.source_terms', source_object, found)
+
+       n_sources = core%count(source_object)
+       allocate(this%source_terms(n_sources))
+       write(*,*) "HI"
```
MartinKarp, line 120:
```suggestion
```

src/fluid/fluid_source_term.f90
```diff
@@ -0,0 +1,209 @@
...
+          if ((trim(type) .eq. "user_vector") .or. &
+              (trim(type) .eq. "user_pointwise")) then
+              
+              call init_user_source(this%source_terms(i)%source_term, &
+                                    rhs_fields, coef, type, user)
+          else 
+              
+              call source_term_factory(this%source_terms(i)%source_term, &
+                                       source_subdict, rhs_fields, coef)
+          end if
```
MartinKarp, line 140: Looks good! One can then add variables used in the user source directly in the json file?

MartinKarp (commented):
I think it looks good, great job Tim!

src/fluid/fluid_source_term.f90
timofeymukha: Classic :-)

src/fluid/fluid_source_term.f90
MartinKarp: Yeah, I guess, but it is still not an argument to user_srouce right?

timofeymukha: That's true, unlike some other user functions, the user source interface does not include `params`.

timofeymukha: But access could be arranged pretty easily, for example by adding the json params as a component in `fluid_user_source_term`. I think we are a still a bit wobbly about what the user subroutines have access to :-)

src/fluid/fluid_scheme.f90
```diff
@@ -640,10 +687,6 @@ subroutine fluid_scheme_validate(this)
        call neko_error('No Krylov solver for pressure defined')
     end if
 
-    if (.not. associated(this%f_Xh%eval)) then
-       call neko_error('No source term defined')
-    end if
-
```
njansson, line 646: I guess it's impossible to validate now when we have a default "do nothing approach" or?

njansson (changes requested):

MartinKarp (approved):

src/fluid/fluid_scheme.f90
njansson: It's Friday, we leave this for later 🍻

njansson (approved):

PR #970, opened 2023-09-08
A draft for programming conventions and patterns

timofeymukha:
This has everything discussed in the dev meetings, as far as I could remember. Also, quite a bit of points, which just put down as a rule something already followed in the code. Plus some additional points from myself, up for discussion.

### DIFF
--- a/doc/pages/dev_patterns.md
+++ b/doc/pages/dev_patterns.md
@@ -14,7 +14,8 @@ base as of now, but we should at least make all new code as clean as possible.
    exceptions take place, e.g. `Re` for the Reynolds number and `Xh` for the
    function space, but these should be kept to a minimum.
 
-2. Follow established naming conventions for certain types. For example, `Xh` for the function space.  
+2. Follow established naming conventions for certain types. For example, `Xh`
+   for the function space.  
 
 3. Use easy-to-remember unambiguous variable names. Avoid abbreviations unless
    they are universally understandable (e.g. RANS).
@@ -35,7 +36,8 @@ base as of now, but we should at least make all new code as clean as possible.
      exceptions.
    * Very important to observe in any code, which is intended as `public`,
      particularly interfaces.
-   * Desirable to observe even for code not derictly exposed to the outside world.
+   * Desirable to observe even for code not derictly exposed to the outside
+     world.
    
 4. All derived types should end with `_t` in the name.
 
@@ -84,8 +86,10 @@ base as of now, but we should at least make all new code as clean as possible.
 ## B. Scope
 
 1. Always use `only` when `using` something from another module.  The `neko`
-   module is an exception and imports everything. This is done so that the
-   `.usr` files need only `use` the `neko` module to get access to everything.
+   module is an exception and imports everything. 
+   
+   This latter is done so that the `.usr` files need only `use` the `neko`
+   module to get access to everything.
 
 2. All modules must have `implicit none` at module level.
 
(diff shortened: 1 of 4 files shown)


doc/pages/dev_patterns.md
```diff
@@ -0,0 +1,209 @@
...
+8. Use `init` for constructor TBPs.
+
+9. Use `free` for destructor TBPs.
+
+## B. Scope
+
+1. Always use `only` when `using` something from another module.  The `neko`
+   module is an exception and imports everything. 
+   
+   This latter is done so that the `.usr` files need only `use` the `neko`
```
njansson, line 91: Maybe change to "user .f90 files" since we don't have .usr files

timofeymukha: Right, this was simply wrong, of course its a .f90 file.

vbaconnet (approved):

njansson (approved):
My only "brasklapp", with long and verbose names, we need to take extra care with line length (column 80)

doc/pages/dev_patterns.md
```diff
@@ -0,0 +1,211 @@
...
+
+The points here are mostly things to consider rather than strict rules.
+Hopefully, they can guide towards better implementation.
+
+1. For types doing have computations, consider creating separate subroutines for
+   each compute backend, and putting them in separate files in the `bcknd`.
+
+   This way nitty-gritty optimisation can happen in `bcknd`, whereas the main
+   code for the type is kept clean.
+
```
MartinKarp, line 176: Regarding adding backend code. I would add something like this:
```suggestion
   
   When adding a function targetting several backends, try to add support for as many backends as possible, the bare minimum should be CPU and one GPU backend. For all unsupported make sure that the code returns an error message.
```

MartinKarp: First try with this suggestion feature, sorry for the formatting.

doc/pages/dev_patterns.md
```diff
@@ -0,0 +1,211 @@
...
+   This point represents the S in SOLID, which is a set of design principles for
+   types. Highly recommended to get acquainted with all of them. 
+
+5. When encountering code repetition, consider whether these two pieces of code
+will likely change at different rates and different reasons in the future. If
+so, code repetition is fine, it will eventually cease to exist naturally. 
+
+   True code duplication is when a change in one place will also necessary a
+   change in the other.
+
```
MartinKarp, line 203: Perhaps we should add something about that code should be reused at the largest extent possible and follow the proposed coding conventions, before possible performance optimizations are performed?

doc/pages/dev_patterns.md
```diff
@@ -0,0 +1,211 @@
...
+
```
MartinKarp, line 204: Perhaps also something about that one should try to use e.g. json_parameters rather than compile time constants to such a large extent as possible? I dunno if this is necessary though.

MartinKarp (commented):
Looks really good I think, I added some thoughts.

doc/pages/dev_patterns.md
adampep: I see a little contradiction with the statement optional marguments are not wellcome. I usuallu use them to avoid code duplication.

timofeymukha:
> My only "brasklapp", with long and verbose names, we need to take extra care with line length (column 80)

Yeah, the consequence is that codes grows "vertically" a bit, with more multiple-line statements. But I have mostly experienced this with the names of TBP bind-names where the type and TBP name get glued. I think it's a worthy sacrifice, though. Names should certainly be of manageable length in general and `associate` can be used when soemthing really needs to get shorter for a code block.

doc/pages/dev_patterns.md
timofeymukha: My only reservation here is that in practice we have been merging CPU only as an initial stage quite a bit, and this might remain practical. On the other hand, it is a form of technical dept..

doc/pages/dev_patterns.md
timofeymukha: @adampep Yeah, I didn't mean that they are completely unwelcome. I should tone down that statement. My point was regarding the situation when one has lots of optional arguments and therefore a lot of `if`s to handle different combinations. That is usually a good sign of redesigning a bit.

doc/pages/dev_patterns.md
MartinKarp: Yeah, perhaps we should just say that adding error messages are important.

njansson: Indeed, bare minimum CPU + error messages? (then one always get two backends for free ;))

timofeymukha:
I've added some small changes to address the comments.

PR #972, opened 2023-09-12
Point interpolator on GPUs (CUDA/HIP)

vbaconnet:
Finally the PR for the interpolation on GPUs! We made a lot of changes in workflow so I will try to be as clear as possible.

# Overview of the work

- Probes now uses `point_interpolator` instead of `findpts_eval`.
- We have a new tensor subroutine [`tnsr3d_el_list`](https://github.com/ExtremeFLOW/neko/blob/c0226176b33b98e22006c22ddf0f9208bd969a9f/src/math/tensor.f90#L186) that computes a triple tensor product using a list of element ids.
- `tnsr3d_el_list` has a CPU and GPU backend (CUDA and HIP).
- Each rank now only does the work on their so-called **"local" probes** (as determined by `proc_owner`, output of `findpts`).
- Each rank now **stores** the weights for each probe (for the GPU).
- The interpolation weights computation is controlled by the `logical :: update_weights` parameter, which is checked [everytime `probes%interpolate` is called](https://github.com/ExtremeFLOW/neko/blob/c0226176b33b98e22006c22ddf0f9208bd969a9f/src/common/probes.F90#L448).
- `point_interpolator` now has a [`compute_weights`](https://github.com/ExtremeFLOW/neko/blob/c0226176b33b98e22006c22ddf0f9208bd969a9f/src/sem/point_interpolator.f90#L115) subroutine.

## `probes` module

The redistribution of the probes on each rank is done according to whatever is in `proc_owner`, which is one of the outputs of `findpts` in the mapping. 

```f90
!
! Count the number of local points
!
if (pe_rank .eq. this%proc_owner(i)) this%n_local_probes = this%n_local_probes+1
```

I guess the biggest change is that each rank does the work on their own version of `this%out_fields`, which is then reduced with an `MPI_SUM` on rank 0. 

```f90
! Artificial way to gather all values to rank 0
if (pe_rank .ne. 0) then
   call MPI_Reduce(this%out_fields(1,1), this%out_fields(1,1), size_outfields, &
        MPI_REAL_PRECISION, MPI_SUM, 0, NEKO_COMM, ierr)
else
   call MPI_Reduce(MPI_IN_PLACE, this%out_fields(1,1), size_outfields, &
         MPI_REAL_PRECISION, MPI_SUM, 0, NEKO_COMM, ierr)
end if
```

The whole thing is built assuming that **EACH PROBE HAS A UNIQUE PROCESS AND ELEMENT OWNER**! I am pretty sure this should be the case most of the time but... I am not 100% sure that `findpts` will never output a non-unique mapping, and more importantly if it happens the interpolated values will be wrong. Maybe @adampep know more about this?

## `point_interpolator` and `tensor` interfaces

I chose to include `point_interpolator_interpolate_fields` in the `interpolate` interface. However this does not match with the decision to not have `tnsr3d_el_list` (or a wrapper like `triple_tensor_product_fields`) in `triple_tensor_product`. Does this matter? Open for discussion.

# Future improvements

- Make the probes a simulation component!
- Don't store the entire probes on each rank.
- When running on GPUs, move the loop over the fields to the GPU to minimize memory movement.
- Add an `fd_weights_full_device` to compute weights on the GPU directly;

### DIFF
--- a/src/sem/point_interpolator.f90
+++ b/src/sem/point_interpolator.f90
@@ -396,9 +396,6 @@ contains
        call tnsr3d_el_list(tmp, 1, sampled_fields_list%fields(i)%f%x, lx, &
             wr, ws, wt, el_owners, n_points)
 
-       ! Unfortunately we need this because we need to use tmp right away
-       call device_sync
-
        ! Bring back tmp_d from the device for the output
        if (NEKO_BCKND_DEVICE .eq. 1) then
           call device_memcpy(tmp, tmp_d, n_points, DEVICE_TO_HOST, sync = .true.)


vbaconnet:
> Regarding the PROBES_TESTCASE, I would prefer not to have in the examples. However, to showcase the functionality, one could add a line of probes to the channel or pipe case. E.g. across z in a channel, so that one compute a spectrum.
> 
> In principle, the right place for testing components are the reframe tests. Maybe one could one for probes there. We don't have to run it in the CI necessarily.

Yes of course, I will remove it once I have finished all of my tests! I didn't plan on leaving it in the examples. I will update the PR description so we are all on the same page.

I agree that some unit tests could be nice, I could look into them maybe next week.

timofeymukha:
Yeah, unit is also very nice to have, but with reframe we actually run a small case, so one *could* add a small case there with 1 probe or something. Not really saying we should, just thowing the possibility out there.

timofeymukha:
>> Yes, also on that note maybe we can merge the rayleigh-benard-cylinder examples together so we only have one in the examples?

Yeah, I'm pro that at least. I think the examples should relatively beefy, more like tutorials, showing what the code can do.

src/sem/point_interpolator.f90
```diff
@@ -313,4 +351,66 @@ function point_interpolator_interpolate_jacobian(this, rst, X,Y,Z) result(jac)
...
+       call tnsr3d_el_list(tmp, 1, sampled_fields_list%fields(i)%f%x, lx, &
+            wr, ws, wt, el_owners, n_points)
+
+       ! Unfortunately we need this because we need to use tmp right away
+       call device_sync
+
+       ! Bring back tmp_d from the device for the output
+       if (NEKO_BCKND_DEVICE .eq. 1) then
+          call device_memcpy(tmp, tmp_d, n_points, DEVICE_TO_HOST, sync = .true.)
+       end if
```
njansson, line 405: With #976 being merged, I think it should be safe to remove the call to `device_sync`.

I guess what happens now is the following:

1. `tensr3d_el_list` performs everything in the stream `glb_cmd_queue`
2. `device_memcpy` does the copy in the default NULL stream 

thus, without syncing the device on line 400, there's no guarantee that `glb_cmd_queue` is ready to be copied. The fixed `device_memcpy` (in #976 ) will default to operate on `glb_cmd_queue` thus preventing this race condition. Furthermore, with everything operating in the same stream it should also be safe to remove the `sync` on line 390 (but **not** on line 404)

vbaconnet: Great! Thank you for the changes! Then I will remove it and give it a try on Dardel.

njansson (approved):

PR #973, opened 2023-09-15
Make mesh private, with type-bound procedures

njansson:
Encapsulate all mesh routines as type bound procedures.

This will make all mesh related routines having the same form with `%init() %free` etc. Furthermore, this will also greatly aid in reducing the amount of `ftn-878` warnings with Cray Fortran, towards resolving (#795 )


timofeymukha:
Very nice! I would prefer the `pass`ed argument to be called `this` as proposed in A.10 in the #970 convention draft. But we could also refactor that later.

njansson:
This also related to #794

PR #974, opened 2023-09-18
Make field private, with type-bound procedures

njansson:
Encapsulate all field routines as type bound procedures, following #973, #795 and #794.


timofeymukha:
You're on a roll! :-)

PR #978, opened 2023-09-20
Fix point_interpolator compiler warnings on Cray

vbaconnet:


njansson (approved):

timofeymukha:
A little annoying that this is needed but oh well.

timofeymukha (approved):

vbaconnet:
yes, I guess another option would be to use `triple_tensor_product_scalar` for each `jac(i,j)` instead of `triple_tensor_product_vector`. Just tried it and it does not throw any warnings either. If it is more elegant, maybe we can use that instead?

```f90
    ! d(x,y,z)/dr
    call triple_tensor_product(jac(1,1), X, lx, hr(:,2), hs(:,1), hr(:,1))
    call triple_tensor_product(jac(1,2), Y, lx, hr(:,2), hs(:,1), hr(:,1))
    call triple_tensor_product(jac(1,3), Z, lx, hr(:,2), hs(:,1), hr(:,1))
```

PR #980, opened 2023-09-25
Throw an error if the re2 file label id is out of range

njansson:
Fixed SEGV if the parsed label id `label` is greather than `NEKO_MSH_MAX_ZLBLS` (or smaller than one)


pschlatt1:
Shouldnt one include the actual value of NEKO_MSH_MAX_ZLBLS in the error string (currently 20)?

timofeymukha (approved):

PR #981, opened 2023-09-26
Fix the gslib installation path

vbaconnet:
I had written the wrong path in the instructions, best to fix it in case someone uses the documentation to link gslib...

### DIFF
--- a/doc/pages/installation.md
+++ b/doc/pages/installation.md
@@ -13,7 +13,7 @@ cd json-fortran && mkdir b && cd b
 cmake -DCMAKE_INSTALL_PREFIX=/path/to/installation -DUSE_GNU_INSTALL_CONVENTION=ON ..
 make install
 ```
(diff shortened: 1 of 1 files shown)


doc/pages/installation.md
```diff
@@ -47,7 +47,7 @@ libgs.a
 Now add the path to gslib to an environment variable `GSLIB`
 
 ``` shell
-export GSLIB=$(pwd)/build
+export GSLIB=$(pwd)/build/lib
```
njansson, line 50: Not sure this is correct. The m4 macro will append `lib` to the path given to `--with-gslib=`

vbaconnet: weird, this is how I have been doing this since the beginning... I always used `build/lib`

vbaconnet:
Both work, so we will keep it as is :)

Issue #983, opened 2023-09-27
User defined Dirichlet boundary conditions


MartinKarp:
With background in previous discussions, I would like to propose an idea of how we can make the user BCs general enough for future developments.  Currently, the user bc is limited only to the velocity field and scalar (there is no support for user-defined pressure conditions), and there are a few problems. If we consider the current inflow eval.
```f90
subroutine usr_inflow_eval(u, v, w, x, y, z, nx, ny, nz, &
                                ix, iy, iz, ie, t, tstep)
       import rp
       real(kind=rp), intent(inout) :: u
       real(kind=rp), intent(inout) :: v
       real(kind=rp), intent(inout) :: w
       real(kind=rp), intent(in) :: x
       real(kind=rp), intent(in) :: y
       real(kind=rp), intent(in) :: z
       real(kind=rp), intent(in) :: nx
       real(kind=rp), intent(in) :: ny
       real(kind=rp), intent(in) :: nz
       integer, intent(in) :: ix
       integer, intent(in) :: iy
       integer, intent(in) :: iz
       integer, intent(in) :: ie
       real(kind=rp), intent(in) :: t
       integer, intent(in) :: tstep
     end subroutine usr_inflow_eval
```
This is constrained in that it is pointwise, meaning the user does not have access to the entire field, does not map well to the GPU, currently does not update at each time step for the GPU, and the user needs to consider which elements and points the BC is to be applied on themselves.

From discussions, a suggestion is to add something like the following that can be specified in the user file
```f90
subroutine user_fluid_dirichlet_update(t, tstep, u_bc, v_bc, w_bc, p_bc, params)
       import rp
       type(field_t), intent(inout) :: u_bc
       type(field_t), intent(inout) :: v_bc
       type(field_t), intent(inout) :: w_bc
       type(field_t), intent(inout) :: p_bc
       real(kind=rp), intent(in) :: t
       integer, intent(in) :: tstep
end subroutine user_fluid_dirichlet_update
```
where the arguments `u_bc, v_bc, w_bc, p_bc` are then used in a new boundary condition, `user_fluid_dirichlet`. `user_fluid_dirichlet_update` would be called only once per time step. In the new bc, `user_fluid_dirichlet` the application of the boundary conditions would then be done something like:
```f90
  subroutine user_fluid_dirichlet_apply(this, u, v, w, p)
    class(user_dirichlet_t), intent(inout), target :: this
[...]

PR #985, opened 2023-09-28
Feature user dirichlet bcs

MartinKarp:
Draft for user Dirichlet bcs (issue #983) that will work better on the device and enable more freedom for the user. My ambition is that the functionality needed from @vbaconnet , @ronithstanly , and @adperezm  should be covered in addition to open up opportunities for even more advanced features (say we add geometric shapes and so on). 

I have tested it with the cyl_bl case and it seems to work at first glance. I added a case file `cyl_bl_user_bc_test.case` that should be removed/changed before this is merged. This new case illustrates how this functionality can be used together with an updated `cyl_bl.f90`. This is a rather rough draft that I would appreciate help from @vbaconnet and @adperezm to make into a finished product. There are a few points that needs to be fixed.

- [x] What arguments do we want to pass to user_dirichlet_update? I think we might need coef and perhaps other things too. 
- [x]  We should implement a field_dirichlet_vector that can be added to the bc list for velocity and make the application of BCs for the different velocity components in one go.
- [ ] We need to check whether the zones marked with dirichlet bcs for parts of the velocity component should be added to the sym and pressure surfaces in fluid_pnpn. This needs to be tested extensively.
- [ ] When we have settled these points we also need to test it further.
- [x] Extend the GPU support for HIP
- [x] Extend the GPU support for OpenCL.
- [x] The bc labels are currently very minimalistic `du,dv,...` perhaps this should be more verbose. This would also mean that we extend the length of each bc_label by more than 20 characters.
- [x] Add the same functionality to the scalar and add it to the bc list in case.f90.
- [ ] split_string needs to be cleaned, it is not pretty.

I expect there to be a few bugs, but I think it gives a good indication of which parts of the framework we need to look at to create a good solution for BCs. 

The changes proposed in this PR should be completely optional in that it does not impact any case that does not make use of these new features.


adperezm:
I will start checking this next week (or by the end of this one).  Just an update

vbaconnet:
Starting to look at gather scatter for shared points whose owning elements have different boundary conditions, as done in Nek5000 ([here](https://kth-nek5000.github.io/KTH_Framework/plan4_8f.html#aedd42d0105ac2c4eac995ac75c027a4a) and [here](https://kth-nek5000.github.io/KTH_Framework/dssum_8f.html#ab0865b005738b43cd2bcc73e7bac86f7)).

vbaconnet:
Just a thought I had while playing around with one of my cases: Since the dirichlet user bc will be used to set boundary conditions (not only, but mostly), would it make sense to expose BC masks to the user? 

At the moment, if one wants to set a non-uniform boundary condition, I think the only way is to loop over each `labeled_zones(i)`'s `facet_el` and select the GLL points based on the facet index, etc. Thoughts?

src/fluid/fluid_scheme.f90
```diff
@@ -736,8 +781,43 @@ subroutine fluid_scheme_bc_apply_vel(this, t, tstep)
...
+       call this%bc_field_u%apply_scalar_dev(this%u%x_d, t, tstep)
+       call this%bc_field_v%apply_scalar_dev(this%v%x_d, t, tstep)
+       call this%bc_field_w%apply_scalar_dev(this%w%x_d, t, tstep)
+    else
+       call this%bc_field_u%apply_scalar(this%u%x, this%dm_Xh%size(), t, tstep)
+       call this%bc_field_v%apply_scalar(this%v%x, this%dm_Xh%size(), t, tstep)
+       call this%bc_field_w%apply_scalar(this%w%x, this%dm_Xh%size(), t, tstep)
+    end if
+
+    !call this%gs_Xh%op(this%u, GS_OP_MAX)
```
MartinKarp, line 799: I think this is necessary as we accommodate shared points along bc interfaces. Also I think the parts that are commented here are probably the simplest way to ensure that the dirichlet is actually set on both sides of the element interface.

src/fluid/fluid_scheme.f90
```diff
@@ -746,8 +826,17 @@ subroutine fluid_scheme_bc_apply_prs(this, t, tstep)
...
+
     call bc_list_apply_scalar(this%bclst_prs, this%p%x, &
                               this%p%dof%size(), t, tstep)
+
+    !call this%gs_Xh%op(this%p, GS_OP_MAX)
+    !
+    !call bc_list_apply_scalar(this%bclst_prs, this%p%x, &
+    !                          this%p%dof%size(), t, tstep)
+    !
+    !call this%gs_Xh%op(this%p, GS_OP_MIN)
```
MartinKarp, line 838: Same here, something to ensure that the correct bc is applied on both sides o the element is necessary.

src/fluid/fluid_pnpn.f90
```diff
@@ -204,13 +215,32 @@ subroutine fluid_pnpn_init(this, msh, lx, params, user, material_properties)
...
+    call this%bc_dv%mark_zones_from_list(msh%labeled_zones, 'dv', &
+                                         this%bc_labels)
+    call this%bc_dv%finalize()
+    call this%bc_dv%set_g(0.0_rp)
+    call this%bc_dw%init(this%dm_Xh)
+    call this%bc_dw%mark_zones_from_list(msh%labeled_zones, 'dw', &
+                                         this%bc_labels)
+    call this%bc_dw%finalize()
+    call this%bc_dw%set_g(0.0_rp)
+
```
MartinKarp, line 243: Here is where I think we need to expand the finalize for the boundary conditions because of the interface between different elements issue. 

This could be done by something like this
``` fortran90
call this%bc_dw%finalize()
temp_array = 1.0
call this%bc_dw%apply(temp_array) !sets some values to 0
call gs%op(temp_array,GS_OP_MIN) !might change values on element boundaries.
new_mask = get_index(temp_array .ne. 1.0) !Check what values are affaected
call bc_dw%update_bc_mask(new_mask) !Update to new (correct) mask that takes care of element interfaces.
```
Something like that I think. An alternative would be to pass down gs when finalizing the bc and do something similar. However, we really need to make sure that we can have that the bc changes between element and not only in unique points. 

For some background and to illustrate the problem I am talking about I have the small picture below. When we create the mask in a bc for BC2 this will be applied on the face between p1 and p2, but p1 will also have the condition BC1. If these are mixed Neumann/Dirichlet we have an issue as p1 might then have two values. The mask for the Dirichlet condition must as such be carried out on both sides of the interface and this can be checked with a gs%op when setting up the mask. The issue does not exist for BC2 and BC3 since p3 is unique and only exists on one element. This case is the most common and this is why this bug has not been addressed until now.

Element 1 Element 2
   BC1-------BC2
px -----  p1 ------ p3
| interior | interior |   BC3
| interior | interior | 
px ----- p2 ------- p4

MartinKarp:
Added some comments on how I think these things can be resolved.

src/fluid/fluid_scheme.f90
vbaconnet: This will be taken care of in another PR

src/fluid/fluid_scheme.f90
vbaconnet: This will be taken care of in another PR.

src/fluid/fluid_pnpn.f90
vbaconnet: This will be taken care of in another PR

src/scalar/scalar_pnpn.f90
```diff
@@ -184,10 +185,20 @@ subroutine scalar_pnpn_init(this, msh, coef, gs, params, user, &
...
+    ! Check for field dirichlet bcs
+    call this%bc_ds%init(this%dm_Xh)
+    call this%bc_ds%mark_zones_from_list(msh%labeled_zones, 'd_s', &
+                                         this%bc_labels)
+    call this%bc_ds%finalize()
+    call this%bc_ds%set_g(0.0_rp)
+
     call bc_list_init(this%bclst_ds)
     call bc_list_add(this%bclst_ds, this%bc_res)
+    call bc_list_add(this%bclst_ds, this%bc_ds)
```
timofeymukha, line 201: Perhaps confused, but do you need `bc_ds` at all? Can't you mark the `d_s` zones in the already existing `bc_res`?

src/fluid/fluid_pnpn.f90
```diff
@@ -73,6 +73,9 @@ module fluid_pnpn
      type(facet_normal_t) :: bc_sym_surface !< Surface term in pressure rhs
      type(dirichlet_t) :: bc_vel_res   !< Dirichlet condition vel. res.
      type(dirichlet_t) :: bc_dp   !< Dirichlet condition vel. res.
+     type(dirichlet_t) :: bc_du   !< Dirichlet condition vel. res.
+     type(dirichlet_t) :: bc_dv   !< Dirichlet condition vel. res.
+     type(dirichlet_t) :: bc_dw   !< Dirichlet condition vel. res.
      type(non_normal_t) :: bc_vel_res_non_normal   !< Dirichlet condition vel. res.
```
timofeymukha, line 79: Both here and in the scalar, the naming du, dv, dw, ds will cause confusion, because we also use that to refere to $\Delta u$ etc, i.e. the increment. But here the d refers to dirichlet, or?

src/simulation.f90
```diff
@@ -99,6 +99,10 @@ subroutine neko_solve(C)
        tstep = tstep + 1
        start_time = MPI_WTIME()
        cfl = C%fluid%compute_cfl(C%dt)
+
+       call C%usr%user_dirichlet_update(C%dirichlet_bc_field_list, &
+            C%dirichlet_bc_bc_list, C%fluid%c_Xh, t, tstep)
+
```
timofeymukha, line 105: I am wondering if we can "decentralize" and avoid putting the bc stuff into the case. What if scalar and fluid call this routine themselves. The user is passed to scalar and fluid constructors, so they could store a procedure pointer. Then, inside the user code one would make an "if" based on the field names to decide what to do, I guess.

The reason I want this is because if e.g. the scalar becomes a simcomp, then it's the simcomp that has access to the case but not vice versa,so the relationship between these types has to be inversed in the code.

timofeymukha (changes requested):

src/scalar/scalar_pnpn.f90
timofeymukha: Yeah, I didn't get it either at first. The idea is that this forces 0 for the change in the solution, wherever we prescribe some dirichlet condition.

src/fluid/fluid_pnpn.f90
vbaconnet: Yes I believe d refers to dirichlet. Perhaps another name is needed indeed :)

src/fluid/fluid_scheme.f90
```diff
@@ -373,8 +380,40 @@ subroutine fluid_scheme_init_common(this, msh, lx, params, scheme, user, &
...
+    call this%bc_field_u%mark_zones_from_list(msh%labeled_zones,&
+                        'd_vel', this%bc_labels)
+    call this%bc_field_u%finalize()
+
+    call MPI_Allreduce(this%bc_field_u%msk(0), integer_val, 1, &
+         MPI_INTEGER, MPI_SUM, NEKO_COMM, ierr)
+    if (integer_val .gt. 0)  call this%bc_field_u%init_field('d_u')
+
+    ! dirichlet for v-velocity
+    call this%bc_field_v%init(this%dm_Xh)
```
MartinKarp, line 396: is the one for v,w necessary anymore? If we check for d_vel, we can just init bc_field_v and bc_field_w directly

src/fluid/fluid_scheme.f90
```diff
@@ -739,13 +790,50 @@ subroutine fluid_scheme_validate(this)
...
     real(kind=rp), intent(in) :: t
     integer, intent(in) :: tstep
+
     call bc_list_apply_vector(this%bclst_vel,&
          this%u%x, this%v%x, this%w%x, this%dm_Xh%size(), t, tstep)
+
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call this%bc_field_u%apply_scalar_dev(this%u%x_d, t, tstep)
+       call this%bc_field_v%apply_scalar_dev(this%v%x_d, t, tstep)
+       call this%bc_field_w%apply_scalar_dev(this%w%x_d, t, tstep)
```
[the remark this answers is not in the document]
MartinKarp: Yes, but tbh I guess there should simply be something like field_list instead of vector/scalar or what do you think?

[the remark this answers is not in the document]
MartinKarp: I think now thoough as we moved away from individual du, dv, dw, this oculd be done pretty easily

MartinKarp (commented):
Might not be entirely correct, but I think maybe we could merge a lot of the stuff for u,v,w individually since we have the vel_dir condition now instead

vbaconnet:
Maybe an update on this, after discussions with Martin it seems that separating the field dirichlet into individual "u,v,w" components offered a bit more possibilities, so I reverted back to that. Will adress the other points now.

src/fluid/fluid_scheme.f90
vbaconnet: Moved back to uvw individual components so I guess this is not applicable anymore right?

src/math/bcknd/device/device_math.F90
```diff
@@ -919,10 +936,25 @@ subroutine device_copy(a_d, b_d, n)
...
+    integer :: n, m
+#ifdef HAVE_HIP
+    call hip_masked_copy(a_d, b_d, mask_d, n, m)
+#elif HAVE_CUDA
+    call cuda_masked_copy(a_d, b_d, mask_d, n, m)
+#elif HAVE_OPENCL
+    call neko_error('no OpenCL backend')
+#else
+    call neko_error('no device backend configured')
+#endif
```
njansson, line 954: @njansson  add an OpenCL kernel

src/bc/field_dirichlet.f90
```diff
@@ -0,0 +1,178 @@
...
+  end type field_dirichlet_t
+
+  abstract interface
+     !> Abstract interface defining a dirichlet condition on a list of fields.
+     !! @param field_bc_list List of fields that are used to extract values for field_dirichlet.
+     !! @param dirichlet_bc_list List of BCs containing field_dirichlet_t BCs only.
+     !! @param coef Coef object.
+     !! @param t Current time.
+     !! @param tstep Current time step.
+     subroutine field_dirichlet_update(dirichlet_field_list, dirichlet_bc_list, coef, t, tstep)
```
njansson, line 76:
```suggestion
  !> Abstract interface defining a dirichlet condition on a list of fields.
  !! @param field_bc_list List of fields that are used to extract values for field_dirichlet.
  !! @param dirichlet_bc_list List of BCs containing field_dirichlet_t BCs only.
  !! @param coef Coef object.
  !! @param t Current time.
  !! @param tstep Current time step.
  abstract interface
     subroutine field_dirichlet_update(dirichlet_field_list, dirichlet_bc_list, coef, t, tstep)
```

src/fluid/fluid_scheme.f90
```diff
@@ -101,6 +104,10 @@ module fluid_scheme
      integer :: pr_projection_dim          !< Size of the projection space for ksp_pr
      type(no_slip_wall_t) :: bc_wall           !< No-slip wall for velocity
      class(inflow_t), allocatable :: bc_inflow !< Dirichlet inflow for velocity
+     type(field_dirichlet_vector_t) :: bc_field_vel   !< Dirichlet pressure condition
+     !type(field_dirichlet_t) :: bc_field_v   !< Dirichlet pressure condition
+     !type(field_dirichlet_t) :: bc_field_w   !< Dirichlet pressure condition
```
njansson, line 109: remove?

njansson (changes requested):

src/math/bcknd/device/device_math.F90
njansson: added

src/simulation.f90
timofeymukha: @vbaconnet As far as I can see things are still in `case`, right? Was that too difficult to change?

vbaconnet: Thinking about this right now, haven't had time to do much lately :( But I am having trouble understanding what you mean with this, would `fluid` and `scalar` each have a procedure pointer to `user_dirichlet_update`?

timofeymukha: No problem, just asking. Yes, that was what I was thinking. They would store a pointer to this routine and call it in `step` at an appropriate location. They would then pass their own field lists and masks. And then the user has to figure out what to do depending on what is passed, probably based on the names of the fields.

vbaconnet: Sorry didn't answer this, I just pushed something that should implement what you described :)

src/fluid/fluid_pnpn.f90
```diff
@@ -186,13 +189,21 @@ subroutine fluid_pnpn_init(this, msh, lx, params, user, material_properties)
...
+                                                 'd_vel', this%bc_labels)
     call this%bc_prs_surface%finalize()
     call this%bc_prs_surface%set_coef(this%c_Xh)
     ! Initialize symmetry surface terms in pressure rhs
     call this%bc_sym_surface%init(this%dm_Xh)
     call this%bc_sym_surface%mark_zone(msh%sympln)
     call this%bc_sym_surface%mark_zones_from_list(msh%labeled_zones,&
                                                  'sym', this%bc_labels)
+    ! Same here, should du, dv, dw be marked here?
     call this%bc_sym_surface%finalize()
```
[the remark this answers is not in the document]
vbaconnet: Not sure if resolved but everything I have ran so far with these BCs have been working as expected. So a priori I would say yes? :)

examples/cyl_boundary_layer/cyl_bl_user_bc_test.case
```diff
@@ -0,0 +1,64 @@
+{
```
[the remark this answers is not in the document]
vbaconnet: Updated, however I don't know what to do with this example. It showcases how to use the field BCs, but that's pretty much it. Perhaps we can keep it to have an example of usage somewhere but in the future perhaps it should have its own example?

src/bc/field_dirichlet.f90
```diff
@@ -0,0 +1,182 @@
...
+       import rp
+       import field_list_t
+       import bc_list_t
+       import coef_t
+       type(field_list_t), intent(inout) :: dirichlet_field_list
+       type(bc_list_t), intent(inout) :: dirichlet_bc_list
+       type(coef_t), intent(inout) :: coef
+       real(kind=rp), intent(in) :: t
+       integer, intent(in) :: tstep
+       character(len=*), intent(in) :: which_solver
```
vbaconnet, line 87: The choice of the naming here was kind of arbitrary...

timofeymukha:
Amazing work. Is there something we need to update in the user-file documentation?

vbaconnet:
> Amazing work. Is there something we need to update in the user-file documentation?

Yes I forgot to mention that, I will add a small section in the user file doc about this! Then it should be ready (finally!) for final revisions from everyone :)

src/simulation.f90
vbaconnet: marking this as resolved!

vbaconnet:
Pushing some documentation but is not finished yet.

vbaconnet:
Okay I think this is the best I can do for now :) I think this version is ready for final reviews. Documentation is finished but can be rewritten at will for clarifications or improvements. 

The user interface could probably be improved in the way the fields and bcs are handled but that would require a lot more time that I unfortunately don't have :(

src/fluid/fluid_pnpn.f90
```diff
@@ -617,6 +654,17 @@ subroutine fluid_pnpn_step(this, t, tstep, dt, ext_bdf, dt_controller)
...
+      !We should implement a bc that takes three field_bcs and implements vector_apply
+      if (NEKO_BCKND_DEVICE .eq. 1) then
+         call this%bc_field_dirichlet_u%apply_scalar_dev(u_res%x_d, t, tstep)
+         call this%bc_field_dirichlet_v%apply_scalar_dev(v_res%x_d, t, tstep)
+         call this%bc_field_dirichlet_w%apply_scalar_dev(w_res%x_d, t, tstep)
+    else 
+       call this%bc_field_dirichlet_u%apply_scalar(u_res%x, this%dm_Xh%size(), t, tstep)
+       call this%bc_field_dirichlet_v%apply_scalar(v_res%x, this%dm_Xh%size(), t, tstep)
+       call this%bc_field_dirichlet_w%apply_scalar(w_res%x, this%dm_Xh%size(), t, tstep)
+    end if
```
njansson, line 666:
```suggestion
      if (NEKO_BCKND_DEVICE .eq. 1) then
         call this%bc_field_dirichlet_u%apply_scalar_dev(u_res%x_d, t, tstep)
         call this%bc_field_dirichlet_v%apply_scalar_dev(v_res%x_d, t, tstep)
         call this%bc_field_dirichlet_w%apply_scalar_dev(w_res%x_d, t, tstep)
      else 
         call this%bc_field_dirichlet_u%apply_scalar(u_res%x, this%dm_Xh%size(), t, tstep)
         call this%bc_field_dirichlet_v%apply_scalar(v_res%x, this%dm_Xh%size(), t, tstep)
         call this%bc_field_dirichlet_w%apply_scalar(w_res%x, this%dm_Xh%size(), t, tstep)
      end if
```

src/fluid/fluid_scheme.f90
```diff
@@ -387,8 +399,79 @@ subroutine fluid_scheme_init_common(this, msh, lx, params, scheme, user, &
     call this%bc_wall%finalize()
     call bc_list_add(this%bclst_vel, this%bc_wall)
 
+    ! Setup field dirichlet bc for v-velocity
```
njansson, line 402:
```suggestion
    ! Setup field dirichlet bc for u-velocity
```

src/fluid/fluid_scheme.f90
```diff
@@ -525,8 +607,10 @@ subroutine fluid_scheme_init_all(this, msh, lx, params, kspv_init, kspp_init,&
     logical :: kspp_init
     character(len=*), intent(in) :: scheme
     real(kind=rp) :: real_val, dong_delta, dong_uchar
+    real(kind=rp), allocatable :: real_vec(:)
+    integer :: integer_val, ierr
     character(len=:), allocatable :: string_val1, string_val2
-    integer :: integer_val
+    !integer :: integer_val
```
njansson, line 613:
```suggestion
```

njansson (approved):
Great stuff!

I would suggest that we merge it now, and cleanup the interface later

MartinKarp (commented):
I also think this looks very good now, if @timofeymukha thinks its also ok I would suggest we merge. Maybe before #1215 ? Regardless I guess there will be a bunch of conflicts between them.

njansson:
> I also think this looks very good now, if @timofeymukha thinks its also ok I would suggest we merge. Maybe before #1215 ? Regardless I guess there will be a bunch of conflicts between them.

Great! Let's get it in an we deal with the cleanup later 🐈‍⬛

timofeymukha:
Guys I am a little overwhelmed atm, and I've reviewed this thoroughly at the last iteration, with all the stuff addressed it seems. So if two more pair of eyes thing this looks good -- please merge.

MartinKarp:
Its only I cannot approve, since I am the initial author of the PR.

timofeymukha (approved):

PR #994, opened 2023-10-09
Dimensional solvers, new material properties input and user hook

timofeymukha:
This is a preview for switching to a fully dimensional solver with an additional user hook to set material properties.

Current organization is as follows:
- `Re` and `Pr` are removed from the `fluid_scheme` and `scalar_scheme` types.
-  Instead, `mu`, `lambda` and `cp` are added, for dynamic viscosity, conductivity, and specific heat (`mu` was already in the type but unused). 
- All the residual subroutines are updated in the same way.
- A `user_t%material_properties` hook, that allows to set `rho`, `mu`, `lambda`, `cp`. This runs just after the fluid scalar are initialized *and then at the end of each time step*.

So, the solver is unchanged in the sense that it still operates on dynamic quantities, the equations are not divided by `rho`, etc.

In the case file, the input is either the dimensional quantities listed above or just `Re` for the fluid and `Pe` for the scalar. If one tries to prescribe both the dimensional and non-dimensional parameters, an error is thrown.  If the input is non-dimensional, `rho` and `cp` are set to one, and `lambda = 1/Pe`, `mu = 1/Re`. None of this logic is run if the `user_t%material_properties` is set, and the user is responsible for setting all the properties there. So, one can set arbitrary parameters to the case file and then read them inside `user_t%material_properties`, see e.g. the RBC example, where the case file contains only Ra and Pr for the material properties.


timofeymukha:
Updated as per discussions!

timofeymukha:
Compiled on Dardel-GPU also, just in case there is some issue with the procedure pointer.

src/simulation.f90
```diff
@@ -137,6 +137,10 @@ subroutine neko_solve(C)
             call neko_simcomps(i)%simcomp%compute(t, tstep)
          end do
        end if
+
+       ! Update material properties
+       call C%usr%material_properties(t, tstep, C%fluid%rho, C%fluid%mu, &
+                                      C%scalar%cp, C%scalar%lambda, C%params)
```
njansson, line 143: Should we have this before initialising fluid/scalar schemes? Also how are we sure that these changes propagates down to the local copies of e.g. `mu` and friends in the schemes?

timofeymukha: This is run also in case.f90 with t=0 and tstep=0, right *after* initializing the scalar and fluid.

Another alternative is to run this inside the init for both the fluid_scheme and scalar_scheme, since both types have access to C%usr there. Then, fluid_scheme will call with fluid%rho, fluid%mu and then just two dummies for the scalar stuff, and the scalar_scheme would do the opposite. So, an extra call, but fully encapsulated in the constructors. 

You are right about the copy! This concerns only `rho`, though, it is the only one which is both in fluid in scalar. Maybe I should try to make the `rho` in scalar a pointer to the one in fluid?

njansson: Can't we just have it before fluid/scalar is setup, then it should be correct.


For example let it run after the mesh deform

timofeymukha: But how will that solve the issue? The problem is that scalar%rho is simply not set in the usr%material_properties, right? One way or another we need to pin it to whatever is set/updated in fluid%rho.

njansson: No that doesn't solve `scalar%rho` I was more thinking that setting up the properties as early as possible would be best. If for example a preconditioned or something uses it

src/fluid/fluid_scheme.f90
```diff
@@ -208,17 +211,48 @@ subroutine fluid_scheme_init_common(this, msh, lx, params, scheme, user)
...
+    if (associated(user%material_properties, user_mp_ptr)) then
+
+       ! Incorrect user input
+       if (params%valid_path('case.fluid.Re') .and. &
+           (params%valid_path('case.fluid.mu') .or. &
+            params%valid_path('case.fluid.rho'))) then
+           call neko_error("Set Re OR mu and rho in the case file, not both")
+
+       ! Non-dimensional case
+       else if (params%valid_path('case.fluid.Re')) then
```
MartinKarp, line 230: Maybe we should print a more extensive message here about how the dimensions work here. I imagine that will be helpful for people when debugging their runs. Perhaps always printing mu, rho, even in the non-dimensional case.

MartinKarp (approved):
Looks good I think!

src/fluid/fluid_scheme.f90
njansson: Yes, but I would suggest to have these debug output on the second message level, such that we can turn it on/off via the command line.

src/simulation.f90
njansson: But then it will be called multiple times, sounds a bit scary (if e.g. a user changes rho in the json)

src/fluid/fluid_scheme.f90
njansson: There's an optional last argument to message for the level, default is to only display level zero , the enable more check the appendix for the correct environment variable

src/simulation.f90
timofeymukha: Yes, but this is just for the case when we need a single init for both types based on the json. So in this case the user should not be able to manipulate anything.

A more "radical" approach would be to have a small type collecting all the material properties as a component inside `case`, and then point to its components inside fluid and scalar.

njansson: Like parameters.f90 ;)

timofeymukha: The latter would actually solve both the init, and the updating of the properties in all types that use them. Also give a separate place in the log to put all the warnings/explanations. I'm starting to like this idea.

timofeymukha:
Update with new material_properties_t type, which stores the properties, and gets update by the user routine. The scalar and fluid types now use it in init and have their material properties as pointers.

Additionally, forces non-dimensional input for scalar, if Re is set for fluid. The following output in the log for non-dimensional input.

```
   -----Material properties------  
   Read non-dimensional values:
   Re         : 1.750000E+03
   Pe         : 7.100000E-01
   Set dimensional values:
   rho        : 1.000000E+00
   mu         : 5.714286E-04
   cp         : 1.000000E+00
   lambda     : 1.408451E+00
```

src/fluid/fluid_pnpn.f90
```diff
@@ -124,7 +126,7 @@ subroutine fluid_pnpn_init(this, msh, lx, params, user)
     call this%free()
     
     ! Initialize base class
-    call this%scheme_init(msh, lx, params, .true., .true., scheme, user)
+    call this%scheme_init(msh, lx, params, .true., .true., scheme, user, material_properties)
```
njansson, line 129: Col 80 
```suggestion
    call this%scheme_init(msh, lx, params, .true., .true., scheme, user, &
                          material_properties)
```

src/common/material_properties.f90
```diff
@@ -0,0 +1,201 @@
...
+  !> Contains all the material properties necessary in the simulation.
+  type, public :: material_properties_t
+     !> Density
+     real(kind=rp) :: rho
+     !> Dynamic viscosity.
+     real(kind=rp) :: mu
+     !> Scalar conductivity
+     real(kind=rp) :: lambda
+     !> Scalar specific heat capacity.
+     real(kind=rp) :: cp
```
njansson, line 55: How about adding some math symbols to the docstring?

src/common/material_properties.f90
```diff
@@ -0,0 +1,201 @@
...
+  implicit none
+  private
+  
+  !> Contains all the material properties necessary in the simulation.
+  type, public :: material_properties_t
+     !> Density
+     real(kind=rp) :: rho
+     !> Dynamic viscosity.
+     real(kind=rp) :: mu
+     !> Scalar conductivity
```
timofeymukha, line 52:
```suggestion
     !> Scalar conductivity, f$\lambda \f$
```

src/common/material_properties.f90
```diff
@@ -0,0 +1,201 @@
...
+  use user_intf, only : user_t, dummy_user_material_properties, &
+                        user_material_properties
+  use utils, only : neko_warning, neko_error
+  use comm, only : pe_rank
+  implicit none
+  private
+  
+  !> Contains all the material properties necessary in the simulation.
+  type, public :: material_properties_t
+     !> Density, f$\rho \f$
```
timofeymukha, line 48:
```suggestion
     !> Density, f$\rho \f$.
```

njansson (approved):

PR #998, opened 2023-10-11
Merge rbc cylinder with probes and expose tensor in neko.f90

vbaconnet:
Merged the probes example into rbc-cylinder. 

Also exposed the tensor module in neko.f90, seemed small enough to include it in this PR instead of having its own.


MartinKarp (approved):
Looks good to me

timofeymukha (approved):

PR #999, opened 2023-10-11
Time interpolator

adperezm:
Added a module to perform time interpolation on a field given two others.

This is really just linear interpolation of order dt**2, which is a weighted average. But the idea is that if higher order interpolation is to be added, it can be done here.

Ideally, whenever one wants to write a field at a given time, this is used, however I guess we will not. The tool will be here, however.


timofeymukha:
- Type docstring is incorrect (copy-pasted :-) ).
- An issue I see with the interpolate function is that for orders > 2, you will need to pass not only t_past, f_past, but also t_past_past, etc., right? So as it is it cannot serve as a common interface for interpolations of various orders. So, one either has to rethink what we pass, or settle for not having a common interface and for now just putting the contents of interpolate_order_2 inside the interpolate subroutine .

It is a bit difficult to make the decision now that the type is not actually used. Things tend to become more apparent at that point. So I would go for the simplest solution, which is to keep this a purely linear interpolator for now.

timofeymukha:
>Ideally, whenever one wants to write a field at a given time, this is used, however I guess we will not.

I guess the situation when you have a variable time-step and you want to write at specific times could make use of this one, right? Unless one auto-adjust the time-step to hit the exact writing time.

adperezm:
> > Ideally, whenever one wants to write a field at a given time, this is used, however I guess we will not.
> 
> I guess the situation when you have a variable time-step and you want to write at specific times could make use of this one, right? Unless one auto-adjust the time-step to hit the exact writing time.

Yes, that would be the idea.

adperezm:
> * Type docstring is incorrect (copy-pasted :-) ).
> * An issue I see with the interpolate function is that for orders > 2, you will need to pass not only t_past, f_past, but also t_past_past, etc., right? So as it is it cannot serve as a common interface for interpolations of various orders. So, one either has to rethink what we pass, or settle for not having a common interface and for now just putting the contents of interpolate_order_2 inside the interpolate subroutine .
> 
> It is a bit difficult to make the decision now that the type is not actually used. Things tend to become more apparent at that point. So I would go for the simplest solution, which is to keep this a purely linear interpolator for now.

I thought this would be the pull request were the docstring would be perfect but I failed! :(

I considered what you say regarding the inputs. At this stage it is implicitly hard coded to order 2. But since I did not intend to add higher orders, I could not anticipate what type of input would be needed.

src/common/time_interpolator.f90
```diff
@@ -0,0 +1,117 @@
...
+  end subroutine time_interpolator_free
+
+  !> Interpolate a field at time t from fields at time t-dt and t+dt
+  !! @param t time to get interpolated field
+  !! @param f interpolated field
+  !! @param t_past time in the past for interpolation
+  !! @param f_past field in the past for interpolation
+  !! @param t_future time in future for interpolation
+  !! @param f_future time in future for interpolation
+  subroutine time_interpolator_interpolate(this,t,f, t_past, f_past, t_future,f_future)
```
timofeymukha, line 91:
```suggestion
  subroutine time_interpolator_interpolate(this, t, f, t_past, f_past, t_future, f_future)
```

src/common/time_interpolator.f90
```diff
@@ -0,0 +1,117 @@
...
+    class(time_interpolator_t), intent(inout) :: this
+    real(kind=rp), intent(inout) :: t,t_past,t_future
```
timofeymukha, line 93:
```suggestion
    real(kind=rp), intent(inout) :: t, t_past, t_future 
```

src/common/time_interpolator.f90
```diff
@@ -0,0 +1,117 @@
...
+  !! @param t time to get interpolated field
+  !! @param f interpolated field
+  !! @param t_past time in the past for interpolation
+  !! @param f_past field in the past for interpolation
+  !! @param t_future time in future for interpolation
+  !! @param f_future time in future for interpolation
+  subroutine time_interpolator_interpolate(this,t,f, t_past, f_past, t_future,f_future)
+    class(time_interpolator_t), intent(inout) :: this
+    real(kind=rp), intent(inout) :: t,t_past,t_future 
+    type(field_t), intent(inout) :: f,f_past,f_future
```
timofeymukha, line 94:
```suggestion
    type(field_t), intent(inout) :: f, f_past, f_future
```

src/common/time_interpolator.f90
```diff
@@ -0,0 +1,117 @@
...
+    real(kind=rp) :: w_past,w_future !Weights for the interpolation
```
timofeymukha, line 95:
```suggestion
    real(kind=rp) :: w_past, w_future !Weights for the interpolation
```

src/common/time_interpolator.f90
```diff
@@ -0,0 +1,117 @@
...
+
+      n = f%dof%size()
+      w_past   = ( t_future - t ) / ( t_future - t_past )
+      w_future = ( t - t_past ) / ( t_future - t_past )
+
+      if (NEKO_BCKND_DEVICE .eq. 1) then
+        call device_add3s2(f%x_d, f_past%x_d, f_future%x_d, &
+        w_past, w_future ,n) 
+      else
+        call add3s2(f%x, f_past%x, f_future%x, w_past, w_future ,n)
```
timofeymukha, line 108:
```suggestion
        call add3s2(f%x, f_past%x, f_future%x, w_past, w_future, n)
```

src/common/time_interpolator.f90
```diff
@@ -0,0 +1,117 @@
...
+
+    if (this%order .eq. 2) then
+
+      n = f%dof%size()
+      w_past   = ( t_future - t ) / ( t_future - t_past )
+      w_future = ( t - t_past ) / ( t_future - t_past )
+
+      if (NEKO_BCKND_DEVICE .eq. 1) then
+        call device_add3s2(f%x_d, f_past%x_d, f_future%x_d, &
+        w_past, w_future ,n)
```
timofeymukha, line 106:
```suggestion
        w_past, w_future, n) 
```

src/common/time_interpolator.f90
```diff
@@ -0,0 +1,117 @@
...
+!
+!> Implements type time_interpolator_t.
+module time_interpolator
+  use num_types, only : rp
+  use field, only : field_t
+  use neko_config
+  use device_math, only : device_add3s2
+  use math, only : add3s2
+  use utils, only : neko_error
+  !use device, only: DEVICE_TO_HOST, device_memcpy
```
timofeymukha, line 41:
```suggestion
```

src/common/time_interpolator.f90
```diff
@@ -0,0 +1,117 @@
...
+  !use utils, only: NEKO_FNAME_LEN
```
timofeymukha, line 42:
```suggestion
```

timofeymukha:
An alternative would be to have a smaller type, which only returns the interpolation weights given an array with time-values. The order is then defined automatically by the size of the passed array, e.g. 2 time values -> linear interpolation. Then the actual interpolation has to be done by the user given the weights.  This is more or less how the time integration schemes are implemented.

But again, I think we can decide when the interpolation is used in the code. For now we can maybe just add this file in the spirit of continuous integration :-).

timofeymukha:
> I thought this would be the pull request were the docstring would be perfect but I failed! :(
You were very close :-). I feel a bit bad about complaining about these things. My ideal vision is also to treat all the text in the docstrings as real sentences, starting with a capital letter and ending with a period, for ultimate visual appeal. But I decidied not to annoy people too much ^^.

adperezm:
> > I thought this would be the pull request were the docstring would be perfect but I failed! :(
> 
> You were very close :-). I feel a bit bad about complaining about these things. My ideal vision is also to treat all the text in the docstrings as real sentences, starting with a capital letter and ending with a period, for ultimate visual appeal. But I decidied not to annoy people too much ^^.

It is good. It helps to also build the habit. This time I was thinking in the documentation from the moment I started writing, which I typically do not do. So at some points it becomes second nature.

timofeymukha (approved):

adperezm:
How does one tipically fix these "Reframe" checks? This stuff compiles on the platforms I have available.

timofeymukha:
> How does one tipically fix these "Reframe" checks? This stuff compiles on the platforms I have available.

With enough effort you can setup reframe tests to run on your own computer. But here it was pretty clear that it's some random crap acting up, because there is now way an unused type could affect the code. The fact that sp succeeded and dp failed is also a giveaway, that one should just restart the CI.

If a test does fail for a legit reason, one can usually see what happened in the log, but not always. What we need to do is have the CI put up the logs as artifacts, like for the unit tests. But no one got around to setting it up it yet.

njansson (approved):

PR #1000, opened 2023-10-11
Global interpolation

MartinKarp:
Implements the global_interpolation type that is supposed to be a base class for any functionality that needs to evaluate the value in a number of points in the simulation.  Currently uses gslib, but the dependency will grow less and less strong as we progress. Will incorporate point interpolator at a later point for the evaluation.

Incorporates global_interpolation in probes.f90, but leaves the use of point_interpolator as is for now. The functionality is the same, only change is that we do not have any gslib calls in probes, but this is instead moved to global_interpolation.

Next step is to use it for restarting from files with different meshes.


adperezm:
Pull request #1000! :)

njansson (approved):

src/common/global_interpolation.F90
```diff
@@ -0,0 +1,311 @@
...
+  use logger, only: neko_log, LOG_SIZE
+  use utils, only: neko_error, neko_warning
+  use comm
+  use math, only: copy
+  use mpi_types
+  use, intrinsic :: iso_c_binding
+  implicit none
+  private
+  
+  type, public :: global_interpolation_t
```
timofeymukha, line 52: We typically put the dosctring on the type, leaving the module dosctring just saying that the module implements said type.

src/common/global_interpolation.F90
```diff
@@ -0,0 +1,311 @@
...
+     !! Sets up correct values to be able to evalute the points
+     procedure, pass(this) :: find_points => global_interpolation_find
+     !> Evaluate the value of the field in each point.
+     procedure, pass(this) :: evaluate => global_interpolation_evaluate     
+  end type global_interpolation_t
+  
+contains
+
+  !> Initialize user defined variables.
+  !! @dof dofmap on which the interpolation is to be carried out.
```
timofeymukha, line 100:
```suggestion
  !! @param dof Dofmap on which the interpolation is to be carried out.
```

src/common/global_interpolation.F90
```diff
@@ -0,0 +1,311 @@
...
+  end subroutine global_interpolation_free_points
+
+  !> Finds the corresponding r,s,t coordinates 
+  !! in the correct global element as well as which process that owns the point.
+  !! After this the values at these points can be evaluated.
+  !! If the locations of the points change this must be called again.
+  !! - `error_code`: returns `0` if point found, `1` if closest point on a border
+  !! (check dist2), `2` if not found
+  !! - `dist2`: distance squared (used to compare the points found by each
+  !! processor)
```
timofeymukha, line 175:
```suggestion
  !! processor)
  !! @param xyz The coordinates of the points.
  !! @param n_points The number of points.
```

src/common/global_interpolation.F90
```diff
@@ -0,0 +1,311 @@
...
+
+#else
+    call neko_error('Neko needs to be built with GSLIB support')
+#endif
+
+  end subroutine global_interpolation_find
+
+
+  !> Evalute the interpolated value in the points given a field on the dofmap
+  !! @param intrp_points array of values in the given points.
```
timofeymukha, line 290:
```suggestion
  !! @param intrp_points Array of values in the given points.
```

src/common/global_interpolation.F90
```diff
@@ -0,0 +1,311 @@
...
+  !! @param field array of values used for interpolation.
```
timofeymukha, line 291:
```suggestion
  !! @param field Array of values used for interpolation.
```

src/common/global_interpolation.F90
timofeymukha: When we abbreviate "interpolation", can we do it consistently to `interp`, like you do with `global_interp` in probes.f90? Also, I think a better name here would be interp_values, "points" make me think that this is point locations.

vbaconnet (approved):
I really like that `probes` has been stripped down of gslib! Excited to see the mesh-to-mesh stuff now :)

timofeymukha (approved):

PR #1001, opened 2023-10-12
Mesh to mesh restarts

MartinKarp:
Intitial mesh to mesh interpolation for restarts. 

The major issue is mostly regarding gslib identifying points along the boundary sometimes to be outside the mesh. This seems to happen quite often, but seems to be extremely case dependent how important this is. I guess for restarts it is mostly whether the simulation will blow up or not. The default behavior for these points is that they are set to 0.

Also, I simply could not get the warnings for the points when they do not match to be displayed properly through the logger. If someone else can have ago that would be much appreciated. It is important that they are output in the log in one go as otherwise it might conflict with other processes when they print.

I expect us to run into issues as the meshes grow larger, but we will take one thing at the time.


timofeymukha:
Since gslib is temporary, there is probably no use fighting it? Maybe one should the possibility to set a default fill-in value for the missing boundary values for now?

It seems the necessity to interpolate is now detected automatically, which is nice.

MartinKarp:
Addresses the issues with points on boundaries by moving all points in the mesh by a factor equal to the tolerance of the Newton iteration closer to the center of the element. This appears to solve all the problems, and the tolerance can be very small (say 1e-13).

src/common/global_interpolation.F90
```diff
@@ -79,33 +79,39 @@ module global_interpolation
...
+       !> Finds the process owner, global element number,
+       !! and local rst coordinates for each point.
+       !! Sets up correct values to be able to evalute the points
+       procedure, pass(this) :: find_points_coords => global_interpolation_find_coords
+       procedure, pass(this) :: find_points_xyz => global_interpolation_find_xyz
+       generic :: find_points => find_points_xyz, find_points_coords
+       !> Evaluate the value of the field in each point.
+       procedure, pass(this) :: evaluate => global_interpolation_evaluate
+
+    end type global_interpolation_t
```
njansson, line 98:
```suggestion
   contains
     !> Initialize the global interpolation object on a dofmap.
     procedure, pass(this) :: init => global_interpolation_init
     !> Destructor
     procedure, pass(this) :: free => global_interpolation_free
     !> Destructor for arrays related to evaluation points
     procedure, pass(this) :: free_points => global_interpolation_free_points
     !> Finds the process owner, global element number,
     !! and local rst coordinates for each point.
     !! Sets up correct values to be able to evalute the points
     procedure, pass(this) :: find_points_coords => global_interpolation_find_coords
     procedure, pass(this) :: find_points_xyz => global_interpolation_find_xyz
     generic :: find_points => find_points_xyz, find_points_coords
     !> Evaluate the value of the field in each point.
     procedure, pass(this) :: evaluate => global_interpolation_evaluate
  end type global_interpolation_t
```

njansson: Formatting

src/common/global_interpolation.F90
```diff
@@ -259,20 +249,20 @@ subroutine global_interpolation_find(this, xyz, n_points)
...
-                xyz(:,i),&
-                ' Differ from interpolated coords: ',&
+          write(*,*) 'Points with coords: ',&
+                this%xyz(1,i),this%xyz(2,i),this%xyz(3,i), &
+                'Differ from interpolated coords: ',&
                 x_check(i), y_check(i), z_check(i),&
-                'Absolute differences(x,y,z): ', xdiff, ydiff, zdiff
-          call neko_warning(log_buf)
+                'Distance squared: ',&
+                xdiff, ydiff, zdiff
```
njansson, line 265: I guess `neko_warning` isn't suitable since it filters the output to rank 0.

Maybe one should have a global version that outputs `*** warning (RANK ###) ...` or something

MartinKarp: Okay, I thought I changed everything to a conventional write(*,*) as I couldn't get warning/neko_log to work properly. If one gets this error formatting is not really the major issue probably

njansson (approved):

timofeymukha (approved):

PR #1004, opened 2023-10-16
Enhancement Probes

MartinKarp:
Starting to become OK now in my eyes. Probably some documentation and garbage collecting needed still, but have a look. The changes in the PR are the following:

- We no longer duplicate data for the probes across all ranks, meaning that if we add a parallel file format for the probes this should not impact the probes module as we do no longer assume that each rank has global knowledge.
- We redistribute the probes so that each rank only has its own local points (redist in global_interpolation).
- Probes are now a simulation component.
- Introduces "local_interpolation" which has similar functionality to point_interpolator, but stores the interpolation weights and only works on arrays of points. 

Things left that we should know about in the future:

- CSV files are not very scalable, we should add some parallel format as we approach very many probes. Right now we do all reading and writing on rank 0. We should perhaps also consider storing matrices transposed as writing to CSV right now is quite irritating.
- The redistribution of the points is also not really optimal, I would anticipate this to take quite some time during init if we have very many probes and very many ranks.  I think this can be done nicer with MPI_AlltoAllv instead of the loop nest here: https://github.com/ExtremeFLOW/neko/blob/34b8ae29331e71fb3909509805b6d42557a9fc4e/src/common/global_interpolation.F90#L478-L498


src/simulation_components/probes.F90
```diff
@@ -0,0 +1,412 @@
...
+    call this%init_from_attributes(case%fluid%dm_Xh, output_file)
+    if(allocated(xyz)) deallocate(xyz)
+
+  end subroutine probes_init_from_json
+
+
+  !> Initialize user defined variables.
+  !! @param t Current simulation time.
+  !! @param params case file.
+  !! @coef Xh Function space.
```
timofeymukha, line 154: Seems incorrect.

src/sem/local_interpolation.f90
```diff
@@ -0,0 +1,185 @@
...
+! COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+! INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+! BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+! LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+!
+!> Routines to obtain interpolated values on a set of points with known rst coordinates in elements local to this process
```
njansson, line 33:
```suggestion
!> Routines to obtain interpolated values on a set of points with known 
!! rst coordinates in elements local to this process.
```

njansson (changes requested):
Very nice!

As pointed out by @timofeymukha, would be great if this can be documented as well, e.g. list Probes in the list of provided simulations components (in the manual)

src/sem/local_interpolation.f90
```diff
@@ -0,0 +1,186 @@
...
+     type(c_ptr) :: weights_r_d = c_null_ptr
+     type(c_ptr) :: weights_s_d = c_null_ptr
+     type(c_ptr) :: weights_t_d = c_null_ptr
+   contains
+     !> Constructor.
+     procedure, pass(this) :: init => local_interpolator_init
+     !> Destructor.
+     procedure, pass(this) :: free => local_interpolator_free
+     !> Interpolates the scalar field \f$ X \f$ on the specified coordinates
+     procedure, pass(this) :: evaluate => local_interpolator_evaluate_values
```
vbaconnet, line 70: How about this?
```suggestion
     procedure, pass(this) :: evaluate => local_interpolator_evaluate
```

MartinKarp: I changed it now, but I anticipate we will have many evaluates with different number of fields etc. in the future.

MartinKarp:
> Very nice. So probes are just a simcomp wrapping together IO and the global interpolation, right? And the global interpolation does the work, using under the hood local_interpolation?
> 
> Could you elaborate a bit more on the local_interpolation vs point_interpolator? When is one to use which? It may be worth adding this info to the type docstring of both types, since anyone who would go looking for the functionality will wonder about the 2 similar types.

Regarding local_interpolator vs point_interpolator, they provide very similar functionality. IMO using local_interpolation is the way to go unless one has a bunch of type(point_t) points. I am updating it so that point_interpolator only works with the typ(point_t) type while local_interpolation only handles arrays. point_interpolator also provides functionality to directly get the jacobian and such, so I did not want to remove it.

vbaconnet:
Curious about the behaviour of probes as simcomp now, does this mean that one could add two probes simcomps with different input files? Like this for example:

```json
{
   "type": "probes",
   "compute_control": "simulationtime",
   "compute_value"    : 1,
   "points_file":  "probes1.csv",
   "output_file":  "output1.csv",
   "fields": ["w","s"]
 }
 {
   "type": "probes",
   "compute_control": "simulationtime",
   "compute_value"    : 1,
   "points_file":  "probes2.csv",
   "output_file":  "output2.csv",
   "fields": ["u","p"]
 }
```

vbaconnet:
This kind of makes me question the need for `point_interpolator`. If I understood correctly it is not being used at all right now correct? The idea behind it when I implemented it was to write some functions (with the Jacobian etc) to be used for replacing gslib in the local mapping inside elements, but maybe this can now go into `local interpolation`. Might be out of the scope of this PR but noteworthy I think!

vbaconnet (approved):
I think it's really nice to have everything properly organized into global and local interpolation. Really like it! And the simcomps stuff opens up a lot of possibilities.

doc/pages/simcomps.md
```diff
@@ -39,11 +39,31 @@ If no parameters for the `output_` parametersare provided, they areset to be the
...
  ## List of simulation components
 
- ## Vorticity
+ ## vorticity
  Computes the vorticity field an stores in the field registry as `omega_x`,
  `omega_y` and `omega_z`.
  Currently produces no output.
 
+ ## lambda2
+ Computes Lambda2 for the velocity field and stores it in the normal output files as the first unused field.
```
timofeymukha, line 48:
```suggestion
 Computes \f$ \lambda_2 \f$ for the velocity field and stores it in the normal output files as the first unused field.
```

doc/pages/simcomps.md
```diff
@@ -39,11 +39,31 @@ If no parameters for the `output_` parametersare provided, they areset to be the
...
+ This means that lambda2 can be found in the temeprature field in then fld files if running without a scalar
```
timofeymukha, line 49:
```suggestion
 This means that \f$ \lambda_2 \f$ can be found in the temeprature field in then fld files if running without a scalar
```

doc/pages/simcomps.md
```diff
@@ -39,11 +39,31 @@ If no parameters for the `output_` parametersare provided, they areset to be the
...
+ and s1 if neko is run with one scalar.
+ 
+ ## probes
+ Probes the fluid field in the points given by an input file. Example usage:
```
timofeymukha, line 53:
```suggestion
 Probes selected solution fields at the points given inside an input file. Example usage:
```

njansson (approved):

timofeymukha (approved):

Issue #965, opened 2023-09-07
Geometry subset selection


timofeymukha:
We need to eventually add support for defining named subsets of geometry to be to localize the effects of sources or relevant simcomps. Like a localized heat source and that sort of stuff. One can think of what type of entities are relevant. An obvious one is a list of dofs, but perhaps a list of faces and a list of elements can also be relevant in some contexts. We may want to create a way for defining the entities from some primitive shapes, e.g. a box, a sphere, etc. Not sure if it is reasonable to do the selection always at runtime or if it should be done prior to the run with definitions saved to some file.

This is probably a more long-term project, given that there is more urgent stuff, but good to have in the list, so we can have it in the back of our heads.

njansson:
Can't we use `zone_t` for the marking, or do we need more functionality for the markers (besides adding a name field in the zone type)?

timofeymukha:
`zone_t` is currently for selecting faces only, right? So we need something for more or less arbitrary volumes as well.

njansson:
> `zone_t` is currently for selecting faces only, right? So we need something for more or less arbitrary volumes as well.

Yes that's how it's used. But the design is to store tuples, currently `(facet, el)` but it can be arbitrary, and we can even extend it to hold arbitrary mapping

timofeymukha:
yes, that is probably the right direction for the "backend".

timfelle:
Since the geometry subsets PR is done is this issue done as well?

njansson:
Indeed, isn't this kind of done now with the about mention PR merged?

timofeymukha:
Yes, I think we can close!

PR #1012, opened 2023-10-19
Add support for geometry subsets

vbaconnet:
TLDR: 
- missing docstring
-  probably some stuff to take care about related to pointers, and abstract types but so far is working.
- `zone_t` is now renamed `facet_zone_t`
- `point_zone_t` is used to describe a set of GLL points
- The definition of default point zones (box & sphere) can be done in the case file as demonstrated in the test case example.
- Do we need a parent class for `facet_zone` and `point_zone`? If so how?

# Geometry subsets and point_zones

The idea of this feature is to be able to "select" zones of the mesh corresponding to some predefined shapes like a box or a sphere. The ultimate goal would be to apply source terms to these specified subsets, as discussed in #965.

# Design

Important additions in this PR:
-  introduction of the concept of geometry subset through the `point_zone_t` type,
- implementation of 2 predefined shapes, box and sphere,
- creation of a point_zone registry, almost an exact copy of `field_registry` and works in the exact same way.

## `point_zone`

The new type `point_zone_t` is based on the already existing `zone_t`. The differences with `zone_t` are as follows:
- `point_zone_t` is an abstract type.
- `point_zone_t` has an additional attribute `name`, which is used to retrieve a `point_zone` from the `point_zone` registry.
- `point_zone_t` has an attribute `mask`, which functions similarly to a `bc_t` mask. It stores linear indices of GLL points.
- `point_zone_t` also has additional subroutines, namely `map` and the abstract interface `criterion`.

## Predefined shapes

The abstract type `point_zone_t` is the parent class of two predefined types, `box_point_zone_t` and `sphere_point_zone_t`. These are predefined shapes that allows for the selection of GLL points within those predefined shapes.

Each of these predefined shape types implements their own version of `criterion`. This function is meant to return `.true.` if a given GLL point with `x,y,z` coordinates is to be selected as part of the predefined shape.

Note that `criterion` is used in the base class by the subroutine `point_zone_map`, where we go through all dofmap and only add GLL points if they satisfy `criterion`. 

Each predefined shape also has their own parameters to describe them. As I have done it now, a `box` is defined by its `x,y,z` bounds `xmin, xmax, ... etc`. A `sphere` is defined by its `center` and its `radius`. Of course this is simplistic and I am anticipating that we would want to expand on how a shape is characterized.

The allocation of predefined `point_zones` is handled by the `point_zone_factory`.

## Point zone registry

The point zone registry is basically a copy of the field registry and has the same intended use. The reason why I chose to use a registry is to be able to refer to point zones anywhere, namely in `fluid_source_term` and `probes`.

## Example of usage

I have provided a small test case to showcase the use of the point zone registry. Of course this is meant to be deleted before merging the PR, and is just here for testing purposes :)

## Changes

- Previously named `zones_t` like `inlet`, `wall`, etc... are now `facet_zones_t`.


src/mesh/point_zone.f90
```diff
@@ -0,0 +1,217 @@
...
+     integer, allocatable :: mask(:)
+     type(c_ptr) :: mask_d
+     type(stack_i4_t), private :: scratch
+     integer :: size = 0
+     logical, private :: finalized = .false.
+     character(len=80) :: name
+   contains
+     procedure, pass(this) :: init_base => point_zone_init_base
+     procedure, pass(this) :: free_base => point_zone_free_base
+     procedure, pass(this) :: finalize => point_zone_finalize
```
timofeymukha, line 56: The `finalize` thing is something I don't like, because after I run `init` on an object, I generally expect it to be constructed and ready for use. Is it impossible to somehow have things in a way that `init` does the work inside `finalize`? I know that we use the finalize approach in other parts as well, but maybe with this new type we can be cleaner?

vbaconnet: The way I see it there are two solutions to your comment:
1. Add the `finalize` stuff into `map`, or
2. Add the `map` and `finalize` stuff into `init`.

Personally I would rather go for option 1 and call `map` and `init` separately. However I don't see any scenario were `map` wouldn't immediately follow `init`, so ultimately... why not option 2.

I am also thinking about the fact that `finalize` kind of "locks in" the points that were added in `scratch` into the `mask`. Right now having `map` and `finalize` separately still leaves some flexibility to not "lock in" the points. But maybe that is not needed? Just a thought.

vbaconnet:
> > The point zone registry is basically a copy of the field registry and has the same intended use.
> 
> Now, this is a bit unfortunate because I am pretty sure that type is broken :-). Can you try to see if expanding the size works?

Just tried it and it seems to work! What seemed to be the problem with the field registry?

src/mesh/point_zone.f90
MartinKarp: For the zones I guess it is mostly that they are read from a mesh and it is not known beforehand how many there will be. I think this is the original idea at least, I guess @njansson also has some thoughts about it.

examples/test-point-zones/README.md
njansson, line 1: Given the title of the readme, should this still be in the pr?

examples/test-point-zones/state.py
njansson, line 1: Also here, I don't mind having the preview file in the repo if it make sense.

njansson:
Do we like this to be in the 0.7 release or not? (@timofeymukha @vbaconnet )

examples/test-point-zones/README.md
vbaconnet: No it shouldn't :) Or at least it wasn't meant to. I will remove it.

vbaconnet:
> Do we like this to be in the 0.7 release or not? (@timofeymukha @vbaconnet )
I would lean more towards no. Actually I think it is missing an example with a proper usage of `point_zone`, and also missing some sentences in the case file documentation. Also maybe some tests?

njansson:
> > Do we like this to be in the 0.7 release or not? (@timofeymukha @vbaconnet )
> 
> I would lean more towards no. Actually I think it is missing an example with a proper usage of `point_zone`, and also missing some sentences in the case file documentation. Also maybe some tests?

Ok, although even if we said feature freeze 31/10, I would say that this is ready enough to be part of it, given that examples etc are easy enough to add during the two weeks of release engineering fixes. 

Think about it at least

vbaconnet:
Actually this could be ready. I just cannot think of any good examples with just this feature. I've added an example in the documentation that shows how point zones can be used (mainly for source term stuff). Maybe we could add something to probes so that we can probe GLL points in a point_zone (without interpolation), then we can add that to an existing example?

pschlatt1:
Could one have an example with a pressure normalisation in a specific region for instance?

vbaconnet:
> Could one have an example with a pressure normalisation in a specific region for instance?

Yes, why not, would that be on an existing example?

PR #1013, opened 2023-10-23
Feature/variable time step

adperezm:
Added logic for variable time stepping to simulation.

The way it works now is that the time step that one sets in the case file is the *maximun* time step allowed. One then adds the parameter "constant_cfl" to the case to activate the variable time stepping.

In the code we calculate the average of cfl and if it is off from the set cfl value by 20% then the change in tstep is done. There are also some limiters to make sure that the dt is not change too drastically over time.

@njansson @timofeymukha @MartinKarp There are a number of things that need to be done and your input is valuable, partly because I started changing a bunch of files in the source code and I realized that I should ask such that I do not need to delete everything later on :)

Details can be found below, but I think the main point I want to make is: Can you think of nice and simple ways to expose the time lags and the u/s lags in the code? In some manner similar to the field registry?

Some details of the things I think need to be adressed are shown now:

- [ ] Adjust the samplers:
   - In the sampling process, only the sampler knows when it is time to interpolate, therefore:
       - should the sampler be the one that issues the time interpolation to the desire time stamp to be sampled?
       - should the fluid_output type take care of its own interpolation?
      
   - Time interpolation in general requires that the lags are exposed somehow, but here some other challenges come:
       - the lags are only present currently in the pnpn_t types and not in the fluid/scalar_scheme_t and this was giving me 
          trouble 
         in the initial prototype of this that I had. I think it should kind of be assumed that any scheme would have lags, no?
       - The lags are not field_t rather field_series_t, therefore I do not know how one would expose them nicely. I was thinking of 
         using the field registry, but that is not an option given the type mismatch
       - The way I am thinking thinks should work would also require that the time lags are exposed. The type of this in CASE is 
         just an array, so I would not know how to expose that either.
 
- [ ] Adjust Statistics:
   - Since I decided to stop adjusting the samplers, I thought to go to the statistics, however before we do anything there: 
      should this be moved to be a simulation component instead? If so, then I guess that would need to be done first.

- [ ] Adjust projections:
   - I think once we have a way to expose tlag and perhaps dtlag to the code, then a simple check can be included to the 
     projections to just discard everything when the time step changes.


timofeymukha:
Hi Adalberto! Regarding the time lag arrays: the simplest way is to just add an array pointer to the _scheme types, and then pass the dt and dtlag from the case to the _scheme constructor and associate (=>) the poiners to those arrays. This way the values can be updated inside the case (like it is now, right?) and the _scheme types will point to them.

I think it is OK to promote the lag fields to the base type. It is hard to imagine a solver that does not need them. Them, the type that interpolate should probably have access to them via a pointer, similar to what I described for the dt and stuff.

Not sure about the type structure /responisbility, but would it be reasonable to put the interpolator type as a component in output_t?

MartinKarp:
Playing devils advocate maybe, but say if one would do a solver that does not use a multistep, but multistage like Runge-Kutta, then there would be no lag arrays.

timofeymukha:
> Playing devils advocate maybe, but say if one would do a solver that does not use a multistep, but multistage like Runge-Kutta, then there would be no lag arrays.

That's true. Maybe one can think where the lags should be stored then. Because for output interpolation I guess they are always needed.

pschlatt1:
indeed, multi-stage schemes I was also thinking about. However, there are some fundamental issues with those in conjunction with pressure correction schemes, so we do not use them for the fluid part (the scalar part in principle is not a problem). Thus, I would say that most relevant solvers for neko would need the lag arrays, or at least could/should have access.

MartinKarp:
One thing I am also concerned with duplicating all data and maintaining two time steps, perhaps if we limit it it to only the u,v,w,p,scalar fields that ok. Then we would however need some logic so that the sampler knows whether the smpled field has a lag array or not.

pschlatt1:
I guess the last point is now more towards samplers, rather than variable time step. Interpolation for statistics can also happen if you have a constant time step.

Shiyu-Sandy-Du:
simulation.f90 is updated by using a running average to evaluate cfl_avg

PR #1014, opened 2023-10-25
Lift and drag and 1D GLL point mapping

MartinKarp:
In this PR I add some functionality to calculate the drag and torque around an object and functionality to map a dofmap to a 1D grid. I also change the average field in space script so it does not need to know the number of elements beforehand.

As I was interested of lift as a function of y I added map_1d.f90 which takes a dofmap and creates a mapping (i,j,k,e) -> (idx) where idx is the global index in a 1D-GLL point grid in some specified direction. Essentially it is mapping to a certain plane in the mesh. This is quite useful when one wants to for example do averaging or similar. 

I illustrate one example of how map_1d can be used in the new contrib script which calculates the lift for each plane in some direction. Observe that the lift calculation currently is only implemented on the CPU.


src/qoi/drag_torque.f90
```diff
@@ -0,0 +1,318 @@
...
+      !call gop(dragvx,w1,'+  ',2)
+      !call gop(dragvy,w1,'+  ',2)
+      !call gop(dragvz,w1,'+  ',2)
+!
+      !call gop(torqpx,w1,'+  ',2)
+      !call gop(torqpy,w1,'+  ',2)
+      !call gop(torqpz,w1,'+  ',2)
+      !call gop(torqvx,w1,'+  ',2)
+      !call gop(torqvy,w1,'+  ',2)
+      !call gop(torqvz,w1,'+  ',2)
```
njansson, line 109:
```suggestion
```

njansson (changes requested):

njansson (approved):

contrib/calc_lift_from_field/calc_lift_from_field.f90
```diff
@@ -0,0 +1,199 @@
+!> Program to calculate the lift generated along one zone.
+!! Can also calculate the torque around a point if one changes the value center (defaults to 0,0,0)
+!! Computes both the toal drag/lift and outputs in stdout as well as the drag lift along a specified direction (x,y,z)
```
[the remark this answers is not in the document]
timofeymukha: I am confused here a bit. So it is both drag and lift, not just lift as it says in the 1st row? Also, a bit confused by the direction thing. I guess lift and drag are just force projections on some two axis, right? Drag for the streamwise and lift for the one normal to that. So I don't quite understand what is meant by line 3.

src/sem/map_1d.f90
```diff
@@ -0,0 +1,188 @@
...
+    allocate(this%pt_lvl(lx, lx, lx, nelv))
+
+    do i = 1, nelv
+       !store which direction r,s,t corresponds to speciifed direction, x,y,z
+       !we assume elements are stacked on eachother...
+       ! Check which one of the normalized vectors are closest to dir
+       ! If we want to incorporate other directions, we should look here
+       el_dim(1,:) = abs(this%msh%elements(i)%e%pts(1)%p%x-this%msh%elements(i)%e%pts(2)%p%x)
+       el_dim(1,:) = el_dim(1,:)/norm2(el_dim(1,:))
+       el_dim(2,:) = abs(this%msh%elements(i)%e%pts(1)%p%x-this%msh%elements(i)%e%pts(3)%p%x)
```
timofeymukha, line 90:
```suggestion
       el_dim(2,:) = abs(this%msh%elements(i)%e%pts(1)%p%x - this%msh%elements(i)%e%pts(3)%p%x)
```

src/sem/map_1d.f90
```diff
@@ -0,0 +1,188 @@
...
+       el_dim(2,:) = el_dim(2,:)/norm2(el_dim(2,:))
+       el_dim(3,:) = abs(this%msh%elements(i)%e%pts(1)%p%x-this%msh%elements(i)%e%pts(5)%p%x)
```
timofeymukha, line 92:
```suggestion
       el_dim(3,:) = abs(this%msh%elements(i)%e%pts(1)%p%x - this%msh%elements(i)%e%pts(5)%p%x)
```

src/sem/map_1d.f90
```diff
@@ -0,0 +1,188 @@
...
+    end do
+    glb_min =  glmin(line,n)
+    glb_max =  glmax(line,n)
+
+    i = 1
+    this%el_lvl = -1
+    do e = 1, nelv
+       el_min = minval(line(:,:,:,e))
+       min_vals(:,:,:,e) = el_min
+       if (relcmp(el_min,glb_min,this%tol)) then
```
timofeymukha, line 106:
```suggestion
       if (relcmp(el_min, glb_min, this%tol)) then
```

src/sem/map_1d.f90
```diff
@@ -0,0 +1,188 @@
...
+       ! Checks which directions in rst the xyz corresponds to
+       ! 1 corresponds to r, 2 to s, 3 to t and are stored in dir_el
+       this%dir_el(i) = maxloc(el_dim(:,this%dir),dim=1)
+    end do
+    glb_min =  glmin(line,n)
+    glb_max =  glmax(line,n)
+
+    i = 1
+    this%el_lvl = -1
+    do e = 1, nelv
```
timofeymukha, line 103: Martin, we have basically 3 do loops here. Maybe just a short comment line above each, telling what is done inside?

contrib/calc_lift_from_field/calc_lift_from_field.f90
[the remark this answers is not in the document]
MartinKarp: This computes the "drag" in each direction x,y,z. Whether this is a lifting force or not is kind of up to the user to interpret based on what is the spanwise, streamwise etc direction in their case.

src/qoi/drag_torque.f90
```diff
@@ -0,0 +1,305 @@
...
+    SKPDAT(4,5)=1
+    SKPDAT(5,5)=NY
+    SKPDAT(6,5)=1
+
+    SKPDAT(1,6)=1           + NX*NY*(NZ-1)
+    SKPDAT(2,6)=NX          + NX*NY*(NZ-1)
+    SKPDAT(3,6)=1
+    SKPDAT(4,6)=1
+    SKPDAT(5,6)=NY
+    SKPDAT(6,6)=1
```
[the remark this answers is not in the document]
MartinKarp: Yes, but this si more or less exaclt yhow it is done in Nek5000 in many places. I guess @njansson might have some opinion here too. IN most places where we do things like this we check what face something is on instead (*look at for example calcuating the normals in coef.f90), but this is quite nice in that one can then do a simple double loop.

src/sem/map_1d.f90
MartinKarp: I added some doc

MartinKarp:
Another note is the license, I think this is quite similar to the Nek5000 code and should perhaps have the Argonne thing in it right?

contrib/calc_lift_from_field/calc_lift_from_field.f90
```diff
@@ -0,0 +1,199 @@
+!> Program to calculate the lift generated along one zone.
+!! Can also calculate the torque around a point if one changes the value center (defaults to 0,0,0)
+!! Computes both the total drag/lift and outputs in stdout as well as the drag lift along a specified direction (x,y,z)
+!! Martin Karp 17/01-24
```
timofeymukha, line 4:
```suggestion
!> Program to calculate the force acting on a single boundary zone.
!! Can also calculate the torque around a point if one changes the value of center (defaults to 0,0,0).
!! Computes and outputs both the total force as well as its projection on a specified direction (x,y,z),
!! which can be used to compute lift or drag.
!! Martin Karp 17/01-24
```

contrib/calc_lift_from_field/calc_lift_from_field.f90
```diff
@@ -0,0 +1,199 @@
...
+     dir = 1
+     line => dof%x
+  else if (trim(hom_dir) .eq. 'y') then
+     dir = 2
+     line => dof%y
+  else if (trim(hom_dir) .eq. 'z') then
+     dir = 3
+     line => dof%z
+  else 
+     call neko_error('homogenous direction not supported')
```
timofeymukha, line 113:
```suggestion
     call neko_error('The homogeneous direction should be "x", "y"or "z"')
```

contrib/calc_lift_from_field/calc_lift_from_field.f90
```diff
@@ -0,0 +1,199 @@
...
+  call s22%init(dof)
+  call s33%init(dof)
+  call s12%init(dof)
+  call s13%init(dof)
+  call s23%init(dof)
+  glb_n_gll_pts = map_1d%n_el_lvls*lx
+  call drag_torq%init(glb_n_gll_pts, 14)
+  if (pe_rank .eq. 0) call output_file%set_header('time, coord, dragpx, dragpy, &
+             & dragpz, dragvx, dragvy, dragvz, torqpx, torqpy, &
+             & torqpz, torqvx, torqvy, torqvz')
```
timofeymukha, line 132:
```suggestion
  if (pe_rank .eq. 0) call output_file%set_header('time, coord, forcepx, forcepy, &
             & forcepz, forcevx, forcevy, forcevz, torqpx, torqpy, &
             & torqpz, torqvx, torqvy, torqvz')
```

contrib/calc_lift_from_field/calc_lift_from_field.f90
```diff
@@ -0,0 +1,199 @@
...
+  integer :: argc, i, j, k, n, lx, e, zone_id, dir, f, glb_n_gll_pts, ierr, mem, t
+  real(kind=rp) :: visc, nv(3), dgtq(12)
+  
+  argc = command_argument_count()
+
+  if ((argc .lt. 6) .or. (argc .gt. 6)) then
+     if (pe_rank .eq. 0) then
+        write(*,*) 'Usage: ./calc_lift_from_field mesh.nmsh field.fld zone_number viscosity function_of_coord output.csv' 
+        write(*,*) 'Example command: ./calc_lift_from_field mesh.nmsh fieldblabla.fld 5 0.04 y out.csv'
+        write(*,*) 'Computes the lift at each level in the mesh at coord y'
```
timofeymukha, line 35:
```suggestion
        write(*,*) 'Outputs the total force and torque on zone 5 using velocity values from fieldblabla.fld'
        write(*,*)  'as well as writes the distribution of the force and torque across y to output.csv'
```

contrib/calc_lift_from_field/calc_lift_from_field.f90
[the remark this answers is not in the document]
timofeymukha: OK, but then I would not call this drag or lift, but simply force? The way it is now, the utility name and the first line of the description are very specific that it is lift, so I think that will confuse more people than myself :-). I added a suggestion for a revised docstring. I realize now that the "drag" routines we lifted from nek here are also just forces and not necessarily drag, but maybe we can live with that, at least for now.

I would suggest maybe renaming the utility itself to something like "calc_zone_forces" or similar. 

Sorry about being picky with this! I think this utility will be used a lot so I wanted to polish the use experience a bit :-)!

njansson:
> Another note is the license, I think this is quite similar to the Nek5000 code and should perhaps have the Argonne thing in it right? The drag_torque.f90 that is.

Agree

src/qoi/drag_torque.f90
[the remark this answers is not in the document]
timofeymukha: Pinging @njansson

MartinKarp:
I have mostly worked on the thesis, so I have not had it very high on my priority list. I might work on it this week though as I have a somewhat reasonable draft for the thesis.

timofeymukha (approved):

PR #1017, opened 2023-10-30
Add gmsh2nek under contrib

timofeymukha:
I didn't try to add it to our build system and simply patched together a compile script that contains the stuff from Nek5000. I'm pretty sure I'd spend hours trying, and perhaps for this temporary tools this solution is OK. All the deps are also copied directly to the gmsh2nek folder. 

This version of gmsh2nek is taken from @adampep private repo and has been adjusted by @MartinKarp to work better.

Added licence headers to all .f90 files.

### DIFF
--- a/contrib/gmsh2nek/README.md
+++ b/contrib/gmsh2nek/README.md
@@ -1,0 +1,13 @@
+This gmsh2nek utility is taken from Nek5000 as a temporary way of converting gmsh meshes to Neko. To compile the utility run
+the `compile.sh`. 
+
+Reads a GMSH file and generates a .re2 file
+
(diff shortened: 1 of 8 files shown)


njansson (commented):
Can't we use speclib and mxm from Neko?

Also a license file is also needed I would say

timofeymukha:
> Can't we use speclib and mxm from Neko?
> 
> Also a license file is also needed I would say

I am starting to get the feeling that it is best to leave this in full nek5000 mode. This code does crazy stuff like passing reals to a char dummy argument, whereas our mxm and speclib rely on sound typing. For the same reason, it won't build properly with our build system. To me, it seems like fixing this will probably introduce a lot of hard-to-catch bugs.

PR #1039, opened 2023-11-22
Scalar Initial Condition

timfelle:
- Added support for initial conditions in the scalar field
- Updated documentation with changes in the case file
- Default behavior is retained as uniform field with value 0


njansson (commented):
Very nice!
How about updating the RBC example and/or the reframe test case to use scalar ic instead of setting the initial scalar values via `fluid_user_ic`?

timfelle:
Sure I can take a look, didn't think any of the examples used it since it was not implemented.

njansson:
> Sure I can take a look, didn't think any of the examples used it since it was not implemented.

Great thanks!

reframe/src/rayleigh.f90
```diff
@@ -50,27 +50,18 @@ subroutine scalar_bc(s, x, y, z, nx, ny, nz, ix, iy, iz, ie, t, tstep)
...
+    type(field_t), intent(inout) :: s
     type(json_file), intent(inout) :: params
-    type(field_t), pointer :: s
     integer :: i, e, k, j
     real(kind=rp) :: rand, z
-    s => neko_field_registry%get_field('s')
 
-    call rzero(u%x,u%dof%size())
-    call rzero(v%x,v%dof%size())
-    call rzero(w%x,w%dof%size())
```
njansson, line 72: We need to retain this (I think), thus having both a `fluid_user_ic` and a `scalar_user_ic`

timfelle: I changed the case file to just use a uniform initial condition for the fluid, so it should be good.

src/scalar/scalar_ic.f90
```diff
@@ -0,0 +1,127 @@
...
+       call device_memcpy(s%x, s%x_d, s%dof%size(), HOST_TO_DEVICE)
+    end if
+
+    ! Ensure continuity across elements for initial conditions
+    call gs%op(s%x, s%dof%size(), GS_OP_ADD)
+
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_col2(s%x_d, coef%mult_d, s%dof%size())
+    else
+       call col2(s%x, coef%mult, s%dof%size())
```
timofeymukha, line 109: Just a question, what does this do? I looked in coef and mult seems to be set to 1.

src/scalar/scalar_ic.f90
```diff
@@ -0,0 +1,127 @@
...
+module scalar_ic
+  use gather_scatter
+  use neko_config
+  use num_types
+  use device_math
+  use device
+  use field
+  use utils
+  use coefs
+  use math
```
timofeymukha, line 43: Please use `only` and in these as well as much as possible.

src/scalar/scalar_ic.f90
```diff
@@ -0,0 +1,127 @@
...
+    type(coef_t), intent(in) :: coef
+    type(gs_t), intent(inout) :: gs
+    character(len=*) :: type
+    type(json_file), intent(inout) :: params
+    ! Variables for retrieving json parameters
+    real(kind=rp) :: ic_value
+
+    if (trim(type) .eq. 'uniform') then
+       call json_get_or_default(params, 'case.scalar.initial_condition.value', &
+                                ic_value, 0.0_rp)
```
timofeymukha, line 70: I would say, no need for a default here. A value of zero is not really special in any way to default to it.

reframe/src/rayleigh.f90
njansson: Ok great!

src/case.f90
```diff
@@ -276,6 +277,18 @@ subroutine case_init_common(C)
             C%fluid%c_Xh, C%fluid%gs_Xh, C%usr%fluid_user_ic, C%params)
     end if
 
+    if (scalar) then
+       call json_get_or_default(C%params, 'case.scalar.initial_condition.type',&
+                  string_val, 'uniform')
```
timofeymukha, line 282: Whether to have a default here is also a question. We have no default for fluid, and I think it is best to be consistent. I think initial conditions is such an important thing that it is best to have it completely explicit.

src/scalar/scalar_ic.f90
```diff
@@ -0,0 +1,127 @@
...
+       call set_scalar_ic_uniform(s, ic_value)
+    else
+       call neko_error('Invalid initial condition')
+    end if
+
+    call set_scalar_ic_common(s, coef, gs)
+
+  end subroutine set_scalar_ic_int
+
+  !> Set scalar intial condition (user defined)
```
timofeymukha, line 80: Please add `@param` doctstrings for the dummy arguments here and in the other subroutines.

njansson (approved):

timofeymukha (changes requested):

timofeymukha:
Note: If we agree to not drop back to uniform 0 as a default, as I suggest, we need to update the case files of all the examples using `scalar`.

src/scalar/scalar_ic.f90
timfelle: > Just a question, what does this do? I looked in coef and mult seems to be set to 1.

I have no idea. 
It is copied directly from flow_ic.f90, so I figured it was important.

timfelle:
> Note: If we agree to not drop back to uniform 0 as a default, as I suggest, we need to update the case files of all the examples using `scalar`.

Yes, which is exactly why i thought it might be useful to have the default. It is the current behavior and would avoid breaking examples other people might be working with. But yes, it is a disconnect from the fluid way of doing initial conditions.

src/scalar/scalar_ic.f90
timfelle: Done, will wait with a push till we have resolved the other comments as well.

src/scalar/scalar_ic.f90
timfelle: Done, will wait with a push till we have resolved the other comments as well.

src/case.f90
timfelle: Lets keep it consistent. It makes the most sense, and I guess its not going to be that much of a bother for people.

timofeymukha: Yes. @adperezm since you are the main scalar user, here is a heads up!

src/scalar/scalar_ic.f90
timfelle: with the default initial condition type removed, this is also removed.

PR #1041, opened 2023-11-27
Expanded use of logging levels

timfelle:
- Global variables introduced to simplify assignment of labels -- neko_log_quiet = 0
-- neko_log_info = 1
-- neko_log_verbose = 2
-- neko_log_debug = 10
- Documentation appendix updated to describe the update.
- A few places trailing whitespace have been removed.

### DIFF
--- a/doc/pages/appendix.md
+++ b/doc/pages/appendix.md
@@ -7,3 +7,12 @@ Name                    | Description                                           
 `NEKO_AUTOTUNE`         | Force Ax auto-tuning strategy (``'1D'``,``'KSTEP'``)          | Unset
 `NEKO_LOG_LEVEL`        | Log verbosity level (integer > 0, default: 1)                 | Unset
 `NEKO_GS_STRTGY`        | Gather-scatter device MPI sync. strategy (0 < integer < 5 )   | Unset
+
+### Logging level details
+
+A number of logging levels are supported.
+
+- `NEKO_LOG_LEVEL=0`   : Quiet mode, minimal logging during execution.
+- `NEKO_LOG_LEVEL=1`   : Default information mode, adding step informations.
+- `NEKO_LOG_LEVEL=2`   : Verbose mode, logging extra details.
+- `NEKO_LOG_LEVEL=10`  : Debug mode.
--- a/contrib/pyneko/src/neko_intf.f90
+++ b/contrib/pyneko/src/neko_intf.f90
@@ -47,11 +47,11 @@ contains
 
     call neko_init()
 
-    call date_and_time(time=time, date=date)           
-    call neko_log%section("Session Information")       
+    call date_and_time(time=time, date=date)
+    call neko_log%section("Session Information")
     write(log_buf, '(A,A,A,A,1x,A,1x,A,A,A,A,A)') 'Start time: ',&
          time(1:2),':',time(3:4), '/', date(1:4),'-', date(5:6),'-',date(7:8)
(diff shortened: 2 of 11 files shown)


doc/pages/appendix.md
```diff
@@ -7,3 +7,12 @@ Name                    | Description
...
 `NEKO_GS_STRTGY`        | Gather-scatter device MPI sync. strategy (0 < integer < 5 )   | Unset
+
+### Logging level details
+
+A number of logging levels are supported.
+
+- `NEKO_LOG_LEVEL=0`   : Quiet mode, minimal logging during execution.
+- `NEKO_LOG_LEVEL=1`   : Default information mode, adding step informations.
+- `NEKO_LOG_LEVEL=2`   : Verbose mode, logging extra details.
+- `NEKO_LOG_LEVEL=10`  : Debug mode.
```
timofeymukha, line 18: I wonder if we need to use an environmental variable instead of an entry in .case file. I would say the latter is generally preferable from a user perspective? However, one would have to set the log level of `neko_log` after having read the case file, I guess in the init of the `case_t`, so you wouldn't be able to control the verbosity of anything executed prior to that. But I'm not sure if that is a big concern.

timfelle: Well I think that would be a very good way of doing it. Currently i was going for minimal changes.

All the information printed before the case file is read is device information and stuff which i personally prefer to always be available. It makes it very clear what example was run on what hardware.

Additionally, if the new format for solver statistics get approved, we might even pipe that into a separate log file based on the user input in the case file.

timofeymukha: Yes, I agree with you. @MartinKarp @njansson , what do you think, should we set the log level in the .case?

njansson: I'm more in favour of an env. variable, such that we disconnect logging levels from a particular case. Rationale is if we start to build "solver" pipelines, e.g. via pyneko, where the case might just be different json blobs thrown in

timofeymukha: I don't feel too strongly about this since one can set the env variable in a run script for slurm in practice. One option could be reading the env variable, but if something is set in the json, it overrides it. I guess this model is not so uncommon, it is a bit akin having several config files for say spack, once central one, but also a user one, which takes priority.

timfelle: I think you are probably right that it is a good idea to have the environment variable. 
I guess we look into built-in logfile setup controled by the case file with some extended convergence information in a form useful for statistics and such. This might go well with some of the proposals at the dev meetings.

timofeymukha (approved):

doc/pages/appendix.md
timofeymukha: I think we merge this, it is surely a step forward. And then we can build on this PR based on further discussion and user input.

njansson: Not with the intent to delay this further, but what about the constants, in the rest of the code we always use capital letters when used as arguments etc.

timofeymukha: Should one also add `parameter` to the definition of the log levels?

timfelle: @njansson Yes it makes sense, also just to highlight the connection to the environment variable.
@timofeymukha Sure, I am still rather new at fortran so i didn't think of that one. But yes I have added it in the lates commit.

njansson (approved):

PR #1047, opened 2023-11-29
Restart simulation fixes

MartinKarp:
Adds various arrays to the restart such that a restart no longer starts the time integration with time integration of order 1, but rather exactly progresses the simulation exactly as if the simulation did not restart but simply ran for longer. Should be backwards compatible, but could use further testing.

Additions to the checkpoint file:
- Lagged arrays for the scalar
- Lagged extrapolated contributions (abx,aby,abz etc.) for both the scalar and fluid
- Dt and tlag arrays

To be improved:
- checkpoint should be moved out of fluid and perhaps restructured entirely.
- each solver should add what it needs to restart into the checkpoint upon init, currently, I do this in the case file.
- Feel like I did a lot of code cuplication
- I did a test for debug message in the scalar solver, can we do this better?

To be noted:
There are two ways to restart in some sense, one is to save everything (like I do in this PR), and another is to recompute certain values from the field values at earlier stages (abx, aby, etc. can technically be calculated from the lagged u,v,w,s fields). However, due to the coupling with the forcing to calculate the abX arrays I chose to simply save everything. I feel this avoids a bunch of issues if someone would change the forcing or similar. We should just enforce that all new solvers add exactly what they need to the checkpoint upon init. The good thing about the recomputation part would be that one could restart easily from field files, but I think we can add some kind of reader for that as well in the future if it is necessary.


src/fluid/fluid_pnpn.f90
```diff
@@ -271,6 +278,138 @@ subroutine fluid_pnpn_init(this, msh, lx, params, user, material_properties)
...
+    !end if
+
+    !call this%adv%compute(u_temp, v_temp, w_temp, &
+    !                      this%f_x%x, this%f_y%x, this%f_z%x, &
+    !                      this%Xh, this%c_Xh, this%dm_Xh%size())
+    !this%abx1 = this%f_x
+    !this%aby1 = this%f_y
+    !this%abz1 = this%f_z
+
+  end subroutine fluid_pnpn_restart
```
njansson, line 412:
```suggestion
  end subroutine fluid_pnpn_restart
  
```

src/scalar/scalar_pnpn.f90
```diff
@@ -204,6 +208,42 @@ subroutine scalar_pnpn_init(this, msh, coef, gs, params, user, &
     call advection_factory(this%adv, this%c_Xh, logical_val, integer_val + 1)
 
   end subroutine scalar_pnpn_init
+  !> I envision the arguments to this func might need to be expanded
+  subroutine scalar_pnpn_restart(this,dtlag, tlag)
```
njansson, line 212:
```suggestion

  !> I envision the arguments to this func might need to be expanded
  subroutine scalar_pnpn_restart(this,dtlag, tlag)
```

src/scalar/scalar_pnpn.f90
```diff
@@ -274,6 +315,17 @@ subroutine scalar_pnpn_step(this, t, tstep, dt, ext_bdf)
...
+         write(log_buf,'(A,A,E15.7,A,E15.7,A,E15.7)') 'Scalar debug',&
+              ' l2norm s', glsc2(this%s%x,this%s%x,n),&
+              ' slag1', glsc2(this%slag%lf(1)%x,this%slag%lf(1)%x,n),&
+              ' slag2', glsc2(this%slag%lf(2)%x,this%slag%lf(2)%x,n)
+         call neko_log%message(log_buf)
+         write(log_buf,'(A,A,E15.7,A,E15.7)') 'Scalar debug2',&
+              ' l2norm abx1', glsc2(this%abx1%x,this%abx1%x,n),&
+              ' abx2', glsc2(this%abx2%x,this%abx2%x,n)
+         call neko_log%message(log_buf)
+      end if
```
njansson, line 328:
```suggestion
      end if
      
```

src/scalar/scalar_scheme.f90
```diff
@@ -122,6 +123,16 @@ subroutine scalar_scheme_init_intrf(this, msh, coef, gs, params, user,&
...
+  !> Abstract interface to restart a scalar formulation
+  abstract interface
+     subroutine scalar_scheme_restart_intrf(this,dtlag, tlag)
+       import scalar_scheme_t
+       import chkp_t
+       import rp
+       class(scalar_scheme_t), target, intent(inout) :: this
+       real(kind=rp) :: dtlag(10), tlag(10)
+     end subroutine scalar_scheme_restart_intrf
+  end interface
```
njansson, line 135:
```suggestion
  end interface
  
```

src/simulation.f90
```diff
@@ -38,7 +38,10 @@ module simulation
   use file
   use math
   use logger
+  use device
+  use device_math
   use jobctrl
+  use field, only: field_t
```
njansson, line 44:
```suggestion
  use field, only : field_t
```

src/simulation.f90
```diff
@@ -279,7 +258,7 @@ subroutine simulation_restart(C, t)
     call neko_log%message(log_buf)
     call neko_log%end_section()
 
-
+
```
njansson, line 261:
```suggestion

```

njansson (commented):
Great stuff!  🍻 
I'll cleanup the rest and also backport it to `release/0.7`

njansson (approved):

src/scalar/scalar_pnpn.f90
```diff
@@ -205,6 +209,43 @@ subroutine scalar_pnpn_init(this, msh, coef, gs, params, user, &
 
   end subroutine scalar_pnpn_init
 
+  !> I envision the arguments to this func might need to be expanded
+  subroutine scalar_pnpn_restart(this,dtlag, tlag)
```
timofeymukha, line 213:
```suggestion
  subroutine scalar_pnpn_restart(this, dtlag, tlag)
```

src/scalar/scalar_pnpn.f90
```diff
@@ -205,6 +209,43 @@ subroutine scalar_pnpn_init(this, msh, coef, gs, params, user, &
...
+    class(scalar_pnpn_t), target, intent(inout) :: this
+    real(kind=rp) :: dtlag(10), tlag(10)
+    integer :: n
+
+
+    n = this%s%dof%size()
+
+    call col2(this%s%x,this%c_Xh%mult, n) 
+    call col2(this%abx1%x,this%c_Xh%mult, n)
```
timofeymukha, line 222:
```suggestion
    call col2(this%abx1%x, this%c_Xh%mult, n) 
```

src/scalar/scalar_pnpn.f90
```diff
@@ -205,6 +209,43 @@ subroutine scalar_pnpn_init(this, msh, coef, gs, params, user, &
...
+    class(scalar_pnpn_t), target, intent(inout) :: this
+    real(kind=rp) :: dtlag(10), tlag(10)
+    integer :: n
+
+
+    n = this%s%dof%size()
+
+    call col2(this%s%x,this%c_Xh%mult, n) 
+    call col2(this%abx1%x,this%c_Xh%mult, n) 
+    call col2(this%abx2%x,this%c_Xh%mult, n)
```
timofeymukha, line 223:
```suggestion
    call col2(this%abx2%x, this%c_Xh%mult, n) 
```

src/scalar/scalar_pnpn.f90
```diff
@@ -205,6 +209,43 @@ subroutine scalar_pnpn_init(this, msh, coef, gs, params, user, &
...
+    real(kind=rp) :: dtlag(10), tlag(10)
+    integer :: n
+
+
+    n = this%s%dof%size()
+
+    call col2(this%s%x,this%c_Xh%mult, n) 
+    call col2(this%abx1%x,this%c_Xh%mult, n) 
+    call col2(this%abx2%x,this%c_Xh%mult, n) 
+    call col2(this%slag%lf(1)%x,this%c_Xh%mult, n)
```
timofeymukha, line 224:
```suggestion
    call col2(this%slag%lf(1)%x, this%c_Xh%mult, n) 
```

src/scalar/scalar_pnpn.f90
```diff
@@ -205,6 +209,43 @@ subroutine scalar_pnpn_init(this, msh, coef, gs, params, user, &
...
+    integer :: n
+
+
+    n = this%s%dof%size()
+
+    call col2(this%s%x,this%c_Xh%mult, n) 
+    call col2(this%abx1%x,this%c_Xh%mult, n) 
+    call col2(this%abx2%x,this%c_Xh%mult, n) 
+    call col2(this%slag%lf(1)%x,this%c_Xh%mult, n) 
+    call col2(this%slag%lf(2)%x,this%c_Xh%mult, n)
```
timofeymukha, line 225:
```suggestion
    call col2(this%slag%lf(2)%x, this%c_Xh%mult, n) 
```

src/scalar/scalar_scheme.f90
```diff
@@ -122,6 +123,17 @@ subroutine scalar_scheme_init_intrf(this, msh, coef, gs, params, user,&
      end subroutine scalar_scheme_init_intrf
   end interface
 
+  !> Abstract interface to restart a scalar formulation
+  abstract interface
+     subroutine scalar_scheme_restart_intrf(this,dtlag, tlag)
```
timofeymukha, line 128:
```suggestion
     subroutine scalar_scheme_restart_intrf(this, dtlag, tlag)
```

src/case.f90
```diff
@@ -206,6 +206,17 @@ subroutine case_init_common(C)
...
+    C%fluid%chkp%tlag => C%tlag
+    C%fluid%chkp%dtlag => C%dtlag
+    select type(f => C%fluid)
+    type is(fluid_pnpn_t)
+      f%chkp%abx1 => f%abx1
+      f%chkp%abx2 => f%abx2
+      f%chkp%aby1 => f%aby1
+      f%chkp%aby2 => f%aby2
+      f%chkp%abz1 => f%abz1
+      f%chkp%abz2 => f%abz2
```
timofeymukha, line 218: High jacking this piece of code to start a threaded comment. I think we should merge this asap because we simply have to fix the restarts, so great job Martin!

In the longer run, I see two issues here.
1. The proliferation of somewhat unfortunate naming coming from Nek5000. I had to actually open the code to remind myself that `abx1` etc are lag arrays for the explicit RHS. 
2. I think we need some interface for objects to inject whatever fields they need to the checkpoint file. This will mostly be `scheme` types, but potentially also simcomps.
3. I think the above would make it pretty difficult to rely on write order to know what is where the file, so we way want to think of using something more "rich", such as hdf5 or similar. With checkpoints, we are much more free to choose whatever compared to files intended for visualization, so we should give it a thought.

njansson: Yes, for later 🍻

PR #1049, opened 2023-12-01
Fix/whitespace and formatting

timfelle:
This is more a suggestion of how to handle formatting in a unified way.

findent was used to reformat all files according to the code style guide.

findent can be integrated directly into Vim, Emacs and GEdit.
In VSCode the ModernFortran package allow the use of "findent"

```
export FINDENT_FLAGS="-i2 -d3 -f3 -s3 -w3 -t3 -j3 -k- -Rr -c3"
find src/ -name "*.f90" -exec bash -c 'findent < {} > {}.tmp; mv -f {}.tmp {}' \;
```


timfelle:
The force push was a rebase onto the newest develop branch.

timofeymukha:
Nice!

njansson (commented):
Very nice, however not sure what todo with the legacy `speclib` file, which the linter has a hard time with

njansson (approved):

timfelle:
Please note that i have added a Formatting CI rule. 
This now checks if the format described in our code style guide is followed for pull requests.
This is something we might want to change to be just committed instead of throwing an error. However, I do not have wide enough permissions to setup git access tokens and such in the secrets. 

(I also have never tried doing something like that, but i found a few stack overflow pages mentioning it.)

PR #1059, opened 2023-12-04
Make the sync. argument non optional for device_memcpy

njansson:


timofeymukha:
So, the `sync` argument is no longer optional at all, right? I like it.

njansson:
> So, the `sync` argument is no longer optional at all, right? I like it.

Correct, not optional.

timofeymukha (approved):

MartinKarp (changes requested):
Looks good! I think there are two places where we might need a sync?

src/math/schwarz.f90
```diff
@@ -201,9 +204,11 @@ subroutine schwarz_setup_wt(this)
       ! endif
       
       if (NEKO_BCKND_DEVICE .eq. 1) then
-         call device_memcpy(work1, this%work1_d, n, HOST_TO_DEVICE)
+         call device_memcpy(work1, this%work1_d, n, &
+                            HOST_TO_DEVICE, sync=.false.)
          call this%gs_h%op(work1, n, GS_OP_ADD) 
-         call device_memcpy(work1, this%work1_d, n, DEVICE_TO_HOST)
+         call device_memcpy(work1, this%work1_d, n, &
+                            DEVICE_TO_HOST, sync=.false.)
```
njansson, line 211:
```suggestion
                            DEVICE_TO_HOST, sync=.true.)
```

PR #1061, opened 2023-12-05
Feature/Formatter-CI: Automated formatting for pull requests

timfelle:
Development of an automated formatting tool


timfelle:
This CI step, currently runs the formatter and uses git diff to report if the format is satisfied.
I think it would be better if the formatting was committed to the branch and pushed, however, that requires setting up secrets and stuff in the repository.

timfelle:
Ill move this to the clean up request, since this will always fail on its own

PR #1065, opened 2023-12-07
Fix formatter action

njansson:
- Comment unsupported keyword
- Case insensitive search


timfelle:
A note on the openmp directives "!$ ", it can be turned off by setting "--openmp=0" if we wish to. This way the '!$ nthreads " will be aligned with the "!$omp" commands.
Alternatively we could add the space for the "!$omp" commands to let them align according to the rules as well.

njansson:
> A note on the openmp directives "!$ ", it can be turned off by setting "--openmp=0" if we wish to. This way the '!$ nthreads " will be aligned with the "!$omp" commands. Alternatively we could add the space for the "!$omp" commands to let them align according to the rules as well.

Excellent! exactly what I was looking for

MartinKarp:
Should this be merged?

njansson:
> Should this be merged?

The workflow script needs some more love

timfelle (approved):
@njansson I think the formatter is now useful to us. I did some of the same things we did with the linter. So now we only check the files modified by the current PR and use that to determine if we allow the check to succeed.

njansson:
should we merge this then?

MartinKarp:
Just to check, how will this impact someone making a PR? 

Maybe we should add a link here under codestyle to make it very clear that linting and findent are important. https://neko.cfd/docs/develop/d1/d5a/contributing.html

timfelle:
> Just to check, how will this impact someone making a PR?
> 
> Maybe we should add a link here under codestyle to make it very clear that linting and findent are important. [neko.cfd/docs/develop/d1/d5a/contributing.html](https://neko.cfd/docs/develop/d1/d5a/contributing.html)

The impact will be that the files someone changes in a PR will have to follow the style defined in the guidelines. Unfortunately we don't have a way to create a slightly lax check like the linter since here a files is formatted or not.

Yes that might be useful, maybe also a link from the github action to the documentation in case of a failed check.

MartinKarp:
What would be really great is of course some kind of copy-paste install and execute command that one could copy there as well.

timfelle:
> What would be really great is of course some kind of copy-paste install and execute command that one could copy there as well. I am myself not sure what versions of what to get exactly and such.

I will add that to the documentation.

timfelle:
@MartinKarp Added

MartinKarp (approved):
Youre a champ!

PR #1067, opened 2023-12-08
Coarse-grained runtime statistics

njansson:
Add optional collection of coarse-grained runtime statistics for a simulation.

Enabled via the case file, e.g.
```json
  "runtime_statistics": {
    "enabled": true,
    "start_step": 100,
  },
```


src/common/runtime_statistics.f90
```diff
@@ -0,0 +1,150 @@
...
+                             this%enabled_, .false.)
+    call json_get_or_default(params, 'case.runtime_statistics.start_step', &
+                             this%start_step_, 1)
+
+    
+    if (this%enabled_) then
+       call json_get(params, 'case.timestep', dt)
+       call json_get(params, 'case.end_time', end_time)
+
+       this%total_steps_ =  ceiling(end_time / dt)
```
timofeymukha, line 82: Assumes constant `dt`, right? We should probably avoid this in new code.

src/common/runtime_statistics.f90
```diff
@@ -0,0 +1,150 @@
...
+
+       allocate(this%data(this%total_steps_, size(RT_STATS_ID)))
```
timofeymukha, line 84: I guess to avoid constant `dt` we would need `data` to a sort of dynamic array which gets expanded on demand. `end_time / dt` would be a good initial size, I guess.

src/common/runtime_statistics.f90
```diff
@@ -0,0 +1,150 @@
...
+     logical, private :: enabled_
+     real(kind=dp), private, allocatable :: elapsed_time_(:)
+     integer, private, allocatable :: nsamples_(:)
+     integer, private :: start_step_
+     integer, private :: total_steps_
+     real(kind=dp), allocatable :: data(:,:)
+   contains
+     procedure, pass(this) :: init => runtime_stats_init
+     procedure, pass(this) :: record => runtime_stats_record
+     procedure, pass(this) :: report => runtime_stats_report
```
timofeymukha, line 60: Docstring for each component and TBP, please :-). Also, param-doctrings for the TBPs further below.

src/simulation.f90
```diff
@@ -120,6 +126,7 @@ subroutine neko_solve(C)
             'Elapsed time (s):', end_time-start_time_org, ' Step time:', &
             end_time-start_time
        call neko_log%end_section(log_buf)
+       call rt_stats%record(RT_STATS_FLUID, end_time-start_time, t, tstep)
```
timofeymukha, line 129: Would it be better to have the objects in question own their own rt_stats and update them? That would clean the main time-loop and allow for more granularity, e.g. one could enable for fluid and disable for scalar. We would only need an extra subroutine to pull in the stuff from all the objects and print the summary.

njansson: Great idea! I'll restructure it

PR #1070, opened 2023-12-12
Restart fixes - second round

MartinKarp:
Further fixes of the restarts... Also found a data race for using the device for user bcs due to the asynchronous memcpy.

I anticipate that both Adalberto and Siavash should observe more similar results now. 

We should perhaps go further with regards to async memcpy and try to limit async to things that happen in the iterative solvers to avoid things like this.


timofeymukha:
Looks very reasonable. Let's wait for the tests people are doing in case we need to amend something.

MartinKarp:
I have not looked at fixing it for the first time step

njansson (changes requested):

njansson (approved):

PR #1074, opened 2023-12-19
Use the global scratch manager for scalar rhs

timofeymukha:
* Follows what is already done for fluid, and seems to have raised no complaints. For scalars, the residuals subroutines don't use temporaries, only the rhs makers do. Note that on the device, no temporaries are used at all, so there this simply deletes unnecessary arguments. 

* The scalar scheme types now `use ...., only`. 

Also, some improvements to docstrings here and there.

### DIFF
--- a/src/scalar/bcknd/cpu/scalar_residual_cpu.f90
+++ b/src/scalar/bcknd/cpu/scalar_residual_cpu.f90
@@ -16,7 +16,7 @@ module scalar_residual_cpu
 contains
 
   !> Compute the residual.
-  !! @param Ax The Helmholz operator.
+  !! @param Ax The Helmholtz operator.
   !! @param s The values of the scalar.
   !! @param s_res The values of the scalar residual.
   !! @param f_xH The right hand side.
--- a/src/scalar/bcknd/cpu/scalar_residual_cpu.f90
+++ b/src/scalar/bcknd/cpu/scalar_residual_cpu.f90
@@ -1,4 +1,4 @@
-!> Residuals in the Pn-Pn formulation (CPU version)
+!> Residuals in the scalar equation (CPU version).
 module scalar_residual_cpu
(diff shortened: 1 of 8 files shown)


src/scalar/scalar_residual.f90
```diff
@@ -50,6 +50,19 @@ module scalar_residual
   end type scalar_residual_t
 
   abstract interface
+     !> Interface for computing the residual of a scalar transport equation.
+     !! @param Ax The Helmholz operator.
```
njansson, line 54:
```suggestion
     !! @param Ax The Helmholtz operator.
```

src/scalar/bcknd/cpu/scalar_residual_cpu.f90
```diff
@@ -1,18 +1,33 @@
...
   type, public, extends(scalar_residual_t) :: scalar_residual_cpu_t
    contains
+     !> Compute the residual.
      procedure, nopass :: compute => scalar_residual_cpu_compute
   end type scalar_residual_cpu_t
 
 contains
 
+  !> Compute the residual.
+  !! @param Ax The Helmholz operator.
```
njansson, line 19:
```suggestion
  !! @param Ax The Helmholtz operator.
```

njansson (commented):
Looks good!

njansson (approved):

src/scalar/bcknd/cpu/scalar_residual_cpu.f90
timofeymukha: I keep misspelling this guy!

MartinKarp (approved):
Yes, looks good! We should just continue to keep in mind with the async stuff that the scratch arrays could easily become exposed to race conditions. But this issue we have talked about previously as well.

PR #1075, opened 2023-12-20
Add possibility to prescribe execution order to simcomps

timofeymukha:
The main novelty is the `order` JSON parameter for simcomps, which is used to determine the order of execution of all the subroutines in the global simcomp array. Lowest order executes first. By default, the order of a simcomp is equal to its position in the JSON object array in the case file.

Also adds a type called `simcomp_executor`, which stores the simcomps and provides an interface to call e.g. compute and restart on all of them, in the correct order. This cleans up the main time-loop a bit, and makes things more convenient now that one has to handle order. This exchanges the `simulation_component_global` module, and the `neko_simcomps`  global is now `simcomp_executor` instead of just an array of simcomp wrappers.


njansson (approved):

MartinKarp (approved):
Looks good to me, do we have a test with different order simcomps?

njansson:
> Looks good to me, do we have a test with different order simcomps?

I think we need some more flags like dependency etc for that

PR #1076, opened 2023-12-20
Add JSON sources to scalar

### LINKED ISSUE #1010: Get scalar_scheme up to speed with fluid_scheme
The `scalar_scheme_t` type is now a bit behind the enhancements introduced to `fluid_scheme_t`. In particular:

- [ ] It does not use a scratch manager. If we are happy with how that performed for the fluid, this can be added to the scalar.
- [ ] Source terms are still in the old format.
- [ ] No `only` in `use`. 

Aiming for post-0.7 here.

timofeymukha:
This depends #1074 and therefore is in don't merge mode util then. I also need to find a good test case because we don't have an example yet. I mainly just want it to be here to run the CI and make it easy to pickup later after Christmas.

This adds the same model for source terms as already implemented in the fluid, so a `source_terms` object in case file for the scalar, etc.

Closes #1010.


njansson:
With #1074 merged, can we remove the "don't merge" from this?

timofeymukha:
Yes!

timofeymukha:
But I should add an example that tests this before merging.

timofeymukha:
The commit is now fully ready for the review.

I've added a new example, which uses the method of manufactured solutions to verify the implementation of the scalar. The solution is currently trivial, and the idea is to make it more interesting later. But the relevance for this PR is that the source terms clearly work.

njansson (approved):

src/scalar/scalar_source_term.f90
```diff
@@ -0,0 +1,196 @@
+
+! Copyright (c) 2023, The Neko Authors
```
timfelle, line 2: Maybe the copyright should say 2024 now?

timfelle (approved):

src/scalar/scalar_source_term.f90
timofeymukha: Fair point :-)

PR #1077, opened 2023-12-20
Add max_iter as component to the krylov type

timofeymukha:
The `max_iter` is now part of `ksp_t`. This lets us remove the associated components in the `scheme` types. Also fixes a bug in the GMRES solver, which did not handle the optional `niter` argument to `solve()`. 

My impression is that relative tolerances are simply not used in the code now at all because the factory looks like this
`subroutine krylov_solver_factory(ksp, n, solver, max_iter, abstol, M)`

I think we should maybe get rid of all the `optional` qualifiers, at least for the tolerances. The only thing that happens is that the base class uses these default values for whatever is missing:
```fortran
  real(kind=rp), public, parameter :: KSP_ABS_TOL = 1d-9 !< Absolut tolerance
  real(kind=rp), public, parameter :: KSP_REL_TOL = 1d-9 !< Relative tolerance
```
Nothing stops importing and using them elsewhere whenever a filler value is needed. The if clauses handling the possible combinations of 3 optionals is painful to look at and also prone to bugs. The GMRES thing is 1 thing, but I think there were also several if clauses trying to pass abs_tol, whin it was not `present`. I didn't dare to make the `max_iter` optional for this reason.


timofeymukha:
> Some how I feel that maybe iterations and tolerances might be good as optional I'm solve (in addition to init) such that can create a type and later use it with relaxed tolerances without having to create a new type. For example during the initial transient.

No need for a new type, right? You can just change the values inside the existing type. Or we can do as with max_iter now, that we optionally pass it directly to solve (I guess this is what you meant).

MartinKarp (approved):
I like this, in principal I would also be in support of making most things in the init of the ksp type non-optional. This would also clean up the factory a fair bit. As for relative tolerances, I think as we look over what tolerances make the most sense, this will be come more important.

timofeymukha:
@njansson Can we merge, or what do you think?

njansson (approved):

PR #1085, opened 2023-12-22
Add support for double precision `fld` files output

vbaconnet:
1. There is a new parameter, `output_precision`, in the case file that can now be set to `double` to force double precision output of the `fld` files.

  - For this I added an additional optional parameter `precision` in the `file_t` constructor, to be able to use `fld_file_set_precision`. 

  - We now read `output_precision` in `case.f90` and initialize `fluid_output_t` with an additional parameter `precision`. I chose to not make `precision` an optional parameter to force us to be explicit about what type of precision we want, and avoid confusions. 

  - Also we print the precision to the console at the beginning of the simulation so there is a trace of it in the log file.

```
      -----Adding write output------  
      File name        : field.fld
      Write control    : nsamples
      Output precision : double
      Total samples:            20
      Writes per time unit (Freq.):  1.000000E+01
      Time between writes:  1.000000E-01

```

2. The default precision is still single precision


njansson (approved):
Much needed 🍻

timofeymukha:
God jul! A bit unsure regarding moving the output parameters. Just before the cold struck me down, I've been looking at depriviliging scalar to a simcomp with the idea of @MartinKarp 's "everything is a simcomp" dream, and it looked quite doable. Output is one big thing that is not really there for simcomps yet, but I can't imagine any other solution but that each of them has its own output_ parameters. It would be strange to have fluid and scalar be different in this way.

So I think this may require more thinking. Maybe cherrypick it to another branch for now, so we can pull it in if we become 100% sure?

vbaconnet:
Hmm indeed it does not make sense to remove `output` from `case.field` then. Since it is such a small change in this PR I can remove it so we only add the double precision output stuff here.

vbaconnet:
Sorry for the mess here... :) Changed the name of the PR to reflect that the only change we are doing is adding support for precision selection in the case file.

MartinKarp (approved):
Tiptop!

PR #1086, opened 2023-12-23
Fix of device masks.

timfelle:
Masks on the device was off by one in point zones.

@njansson Is all the device codes C based so this works across them all or did i miss something?

### DIFF
--- a/src/mesh/point_zone.f90
+++ b/src/mesh/point_zone.f90
@@ -189,9 +189,11 @@ contains
        call this%scratch%clear()
 
        if (NEKO_BCKND_DEVICE .eq. 1) then
+          this%mask = this%mask -1 !< Devices are 0-based
           call device_map(this%mask, this%mask_d, this%size)
           call device_memcpy(this%mask, this%mask_d, this%size, &
                              HOST_TO_DEVICE, sync=.false.)
+          this%mask = this%mask +1 !< Restore 1-based indexing
        end if
 
        this%finalized = .true.


njansson:
Yes, currently all device code is C based

vbaconnet:
I really hope that it didn't take you too long to find this issue... If it did I apologize for the oversight haha :) thanks for the fix!!

vbaconnet (dismissed):

vbaconnet:
Sorry, I will retract my approval for this just because I maybe went over it too fast :)

I think you have raised an interesting discussion with this PR (for me at least). It seems that currently most integer arrays are duplicates and have the same index base (0 or 1-based) on both the host and the device. We then offset the array by +1 or -1 when necessary in any subroutines that use these arrays. 

I was looking at the boundary condition masks, to illustrate this type of behaviour. Both `msk` and `msk_d` are 1-based, but we correct this in the device implementations by hardcoding an offset of `-1` where necessary. 

Here an example with `inflow_apply_vector` on the GPU ([source](https://github.com/ExtremeFLOW/neko/blob/d549a465b5755ac853104bb84a1e2507ca4e6c72/src/bc/bcknd/device/cuda/inflow_kernel.h#L51-L56)):

```c++
  for (int i = idx ; i < m; i += str) {
    const int k = (msk[i+1] - 1);
//                          ^
//                        here
    x[k] = gx;
    y[k] = gy;
    z[k] = gz;
  }
```

When the CPU version is ([source](https://github.com/ExtremeFLOW/neko/blob/d549a465b5755ac853104bb84a1e2507ca4e6c72/src/bc/inflow.f90#L83-L89)):

```.f90
    do i = 1, m
       k = this%msk(i)
       x(k) = this%x(1)
       y(k) = this%x(2)
       z(k) = this%x(3)
    end do
```

We do the same with `global_interpolation`, where the element owner arrays `el_owner` and `el_owner_d` are both 0-based (since they come from the `gslib` C interface). The CPU implementation of `global_interpolation_evaluate` uses `tensr3d_el_list`, where we increment the `el_owner` by +1 to be 1-based ([source](https://github.com/ExtremeFLOW/neko/blob/d549a465b5755ac853104bb84a1e2507ca4e6c72/src/math/tensor.f90#L211-L214)):

```.f90
       do i = 1, n_pt
          !       Note the use of el_list(i) + 1, because of the gslib C interface
          call tnsr3d_el_cpu(v(1,i), nv, u(1,el_list(i)+1), nu, A(1,1,i), Bt(1,1,i), Ct(1,1,i))
          !                                            ^
          !                                           here
       end do
``` 

I guess the point zone stuff is intended to be visible for the user, so having it like you made it in this PR seems more user-friendly. At least as a user I would not want to have to think about indices on the GPU, but then it depends on how one uses the GPU routines? I think it may be worth a quick discussion to determine how we should deal with it in the future, for those of us who are going to play around with device backends?

timfelle:
This was exactly the stuff i wanted to highlight by setting it as a draft PR first. I just added the don't merge so that we can have the discussion before we pull anything in.

In general I would consider internal usage, some of the most important to disambiguate. So internally i would always prefer to have the correct indices stored in the list. So is this something we want to go through and fix? 
I don't think i can come up with a usecase where it would be useful to do the correction on use rather than on initialization. But please let me know if you know of a case.

timfelle:
Additionally i would suggest to implement at least some of the math operators for integers as well as for reals. This could reduce the code significantly and avoid temporary changes to the CPU arrays before copying them to the device.

timfelle:
We will move this PR onto the main repository.

PR #1098, opened 2024-01-03
Feature/metadata in fldfile

adperezm:
Metadata mentioned by @siavash has been included into the files. It seems to work fine although I have not tested if it fixed the visualization problems yet.


adperezm:
I also fixed the rbc example. I should have created a new request but that was inconvenient. I hope that is okay.

njansson (approved):
Nice! Hope this resolves the issues with large files

src/io/fld_file.f90
```diff
@@ -379,6 +379,86 @@ subroutine fld_file_write(this, data, t)
...
+    end if
+
+
+    do i = 1, n_scalar_fields
+       !Without this redundant if statement, Cray optimizes this loop to Oblivion
+       if (i .eq. 2) then
+          mpi_offset = int(mpi_offset,i8) + &
+                       int(1_i8*glb_nelv, i8) * &
+                       int(2, i8) * &
+                       int(MPI_REAL_SIZE, i8)
```
MartinKarp, line 447: This is not correct I think, there will be a gap after the first metadata for the scalar. Look at how we do for the fields, there we save one temp_offset before the loop.

src/io/fld_file.f90
```diff
@@ -402,6 +482,89 @@ subroutine fld_file_write(this, data, t)
...
+  
+     call MPI_File_write_at_all(fh, byte_offset, buffer, nout, &
+            MPI_REAL, status, ierr)
+
+  end subroutine fld_file_write_metadata_scalar
+
+  function vlmin(vec,n) result(vecmin)
+    integer, intent(in) :: n
+    real(kind=rp), intent(in) ::  vec(n)
+    real(kind=rp) :: vecmin
```
MartinKarp, line 543: vlmin and vlmax should probably be moved to math.

MartinKarp (changes requested):
Looks good! This will be really good I think. The only thing is to double-check the offset for the scalar metadata and move vlmin and vlmax. Also, if you have had time to check if it helps with the slicing that would be great :)

adperezm:
Hej! I corrected the suggestions. When you have time check again.

MartinKarp (approved):
Looks good!

PR #1100, opened 2024-01-05
Add optional stream/queue argument for all device math routines

njansson:
This adds the feature of running device math routines in optional streams


timofeymukha (approved):

timofeymukha:
With this kind of stuff, one really wishes one had CI on the device. I didn't check every function of course, and this refactoring is somehow prone to copy-paste mistakes. Did you run some examples on CUDA and HIP to see that things are still fine?

njansson:
> With this kind of stuff, one really wishes one had CI on the device. I didn't check every function of course, and this refactoring is somehow prone to copy-paste mistakes. Did you run some examples on CUDA and HIP to see that things are still fine?

Very soon we will have proper accelerator based CI runners 🦫

PR #1101, opened 2024-01-06
Fused CG for CUDA/HIP

njansson:
This adds the fused CG formulation for accelerators (currently limited to CUDA/HIP), with minimal data movement H2D during iterations.


timofeymukha:
A bit confused: is this a new type of solver, or is this somehow a fusion of codes for 2 types of accelerators into one?

njansson:
> A bit confused: is this a new type of solver, or is this somehow a fusion of codes for 2 types of accelerators into one?

Same solver, just another version with a different optimisation strategy, kernel fusing (fewer large ones), more similar to our current CPU implementation, and more accelerator friendly. However, we should probably not replace our current `cg_device`, under certain circumstances, the many small kernels, can easier be overlapped in `hsmg`

In short, see it as yet an alternative CG, as pipelined, enabled by `fusedcg` in the case file

timofeymukha:
I see, great stuff. So it is selectable from the case file then? If so, please add it to the case-file.md docs! Ideally, with some minimal guideline on when to select it.

njansson:
> I see, great stuff. So it is selectable from the case file then? If so, please add it to the case-file.md docs! Ideally, with some minimal guideline on when to select it.

Fair enough, always forgets about the documentation (T_T)

timofeymukha (approved):

MartinKarp (approved):

PR #1102, opened 2024-01-09
Fix ROCm detection on Cray EX platforms for none HPE provided modules

njansson:
This fixes an issue with detecting ROCm on CRAY EX platforms when using the wrappers (CC,ftn) but a none HPE provided rocm module

### DIFF
--- a/m4/ax_cray.m4
+++ b/m4/ax_cray.m4
@@ -195,6 +195,12 @@ AC_DEFUN([AX_CRAY_ROCM],[
 	    hip_bcknd="1"
 	  else
 	    AC_MSG_RESULT([no])
+            if test "${ROCM_PATH}"; then
+              AX_HIP
+            else
+              AC_MSG_ERROR([Cray ROCm Toolkit not found])
+              have_hip="no"
+            fi
             AC_MSG_ERROR([Cray ROCm Toolkit not found])
 	    have_hip="no"
 	  fi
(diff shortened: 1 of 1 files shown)


yanbosmu:
Is that OK for me to use a GPU mi250x taken from a hpe Cray ex235a machine now? I tried in ubuntu And it doesn't works, Seems like lacking driver?

njansson:
> Is that OK for me to use a GPU mi250x taken from a hpe Cray ex235a machine now? I tried in ubuntu And it doesn't works, Seems like lacking driver?

This PR only relates to an issue with the CCE wrappers together with none HPE modules of ROCm.

Neko is known to work on Ubuntu with mi250x, but feel free to open a new issue describing the problem.

m4/ax_cray.m4
```diff
@@ -195,6 +195,12 @@ AC_DEFUN([AX_CRAY_ROCM],[
...
 	  else
 	    AC_MSG_RESULT([no])
+            if test "${ROCM_PATH}"; then
+              AX_HIP
+            else
+              AC_MSG_ERROR([Cray ROCm Toolkit not found])
+              have_hip="no"
+            fi
             AC_MSG_ERROR([Cray ROCm Toolkit not found])
 	    have_hip="no"
```
njansson, line 205:
```suggestion
```

njansson: Left over typo

PR #1106, opened 2024-01-17
Added missing include guards

timfelle:


timofeymukha (approved):

njansson (commented):
In principle ok, the style our current style for guards `__X_Y_Z` but that's a minor thing.

However, I would like to test this with OpenCL before say ok.

njansson (approved):

PR #1109, opened 2024-01-20
Interpolation onto equidistant points in elements

Shiyu-Sandy-Du:
In this PR, a post-processing program is written to interpolate field from GLL points into equidistant points within elements in the specified direction. E.g. If the original field is on GLL points on x,y and z direction, and user specifies the interpolation direction (hom_dir) to be 'x', then the field will be interpolated onto grid with GLL points on y and z direction but equidistant points on x direction (locally) within each element. Note that the positions of elements' vertices remain the same. Hence, this program will interpolate fields onto equidistant mesh only when the elements' sizes on the specified homogeneous direction are the same.


Shiyu-Sandy-Du:
Maybe someone could help with the problem of function "TRANSPOSE" in the compilation using ifort and ifx?

contrib/oneD_equidistant_field/oneD_equidistant_field.f90
```diff
@@ -0,0 +1,124 @@
+!> Program to interpolate the field from GLL points onto equidistant points in elements in a homogenous direction
+!> The number of points inside an element of the input and output field file is supposed to be the same
+!> Shiyu Du 16/01-24 note: the program is designed for Cartesian Box meshes with equidistant elements on the hom-dir,
+!                          the version for rotational coordinates is not yet been thought about.
+!> Shiyu Du 16/01-24 note: the program is currently tested on 1 core and MPI, GPU is not tested yet.
```
timofeymukha, line 5: Could we test on >1 core before merging? It looks like it should just work, or?

contrib/oneD_equidistant_field/oneD_equidistant_field.f90
```diff
@@ -0,0 +1,124 @@
...
+
+  ! redistribute the coordinates to be equidistant within elements
+  if (trim(hom_dir) .eq. 'x') then
+    call tnsr1_3d(field_data%x%x, lx, lx, wt, ident, ident, msh%nelv)
+  else if (trim(hom_dir) .eq. 'y') then
+    call tnsr1_3d(field_data%y%x, lx, lx, ident, wtt, ident, msh%nelv)
+  else if (trim(hom_dir) .eq. 'z') then
+    call tnsr1_3d(field_data%z%x, lx, lx, ident, ident, wtt, msh%nelv)
+  else 
+    call neko_error('homogenous direction not supported')
```
timofeymukha, line 78: Test this once, above, right after you read the command line args. This way the user won't have to wait for reading the mesh etc to get the error. And you won't need the `else` clauses below.

timofeymukha, line 78:
```suggestion
    call neko_error('The homogenous direction should be "x", "y" or "z"')
```

contrib/oneD_equidistant_field/oneD_equidistant_field.f90
```diff
@@ -0,0 +1,124 @@
...
+  ! construct the weight matrix
+  allocate(wtt(lx,lx))
+  allocate(ident(lx,lx))
+  do i = 1, lx
+    x_equid = -1.0_rp + (i-1) * 2.0_rp/(lx-1)
+    call fd_weights_full(x_equid, Xh%zg(:,1), lx-1, 0, wtt(:,i))
+    ident(i,i) = 1.0_rp
+  end do
+  allocate(wt(lx,lx))
+  wt = TRANSPOSE(wtt)
```
timofeymukha, line 68: I suggest you fill `wt` inside the loop, row by row, once you get the column of `wtt`. This way you don't have to deal with the intrinsic.

contrib/oneD_equidistant_field/oneD_equidistant_field.f90
```diff
@@ -0,0 +1,124 @@
+!> Program to interpolate the field from GLL points onto equidistant points in elements in a homogenous direction
+!> The number of points inside an element of the input and output field file is supposed to be the same
+!> Shiyu Du 16/01-24 note: the program is designed for Cartesian Box meshes with equidistant elements on the hom-dir,
+!                          the version for rotational coordinates is not yet been thought about.
+!> Shiyu Du 16/01-24 note: the program is currently tested on 1 core and MPI, GPU is not tested yet.
+!! Shiyu Du 16/01-24
+program oneD_equidistant_field
```
timofeymukha, line 7: Maybe "map_to_equidistant_1d" ? I kind of want to avoid the capital D.

contrib/oneD_equidistant_field/oneD_equidistant_field.f90
```diff
@@ -0,0 +1,124 @@
+!> Program to interpolate the field from GLL points onto equidistant points in elements in a homogenous direction
```
timofeymukha, line 1: Out of curiosity, what is the use case here?

Shiyu-Sandy-Du: To perform FFT on one direction. Case could be channel flow or TBL etc with box meshes.

contrib/oneD_equidistant_field/oneD_equidistant_field.f90
Shiyu-Sandy-Du: Sure, name is not a problem

contrib/oneD_equidistant_field/oneD_equidistant_field.f90
Shiyu-Sandy-Du: tests on multiple CPUs works. Now the statement has been changed

contrib/oneD_equidistant_field/oneD_equidistant_field.f90
Shiyu-Sandy-Du: fixed

timofeymukha (approved):

njansson (approved):

PR #1110, opened 2024-01-21
Add advecting cone example

timofeymukha:
This is an example used both P. Fischer and by the Nektar++ folks to demonstrate the benefit of high order. A cone is placed in a vortex and advected a single revolution. At the end, we look at the form of the cone and see if it is preserved.


examples/rotating_cone/rotating_cone.f90
```diff
@@ -0,0 +1,79 @@
...
+       y = u%dof%y(i,1,1,1)
+
+       ! Angular velocity is pi, giving a full rotation in 2 sec
+       u%x(i,1,1,1) = -y*pi
+       v%x(i,1,1,1) = x*pi
+       w%x(i,1,1,1) = 0
+    end do
+
+    if ((NEKO_BCKND_DEVICE .eq. 1) .or. (NEKO_BCKND_HIP .eq. 1) &
+       .or. (NEKO_BCKND_OPENCL .eq. 1)) then
```
njansson, line 69:
```suggestion
    if (NEKO_BCKND_DEVICE .eq. 1) then
```
Device covers CUDA/hip and opencl

examples/rotating_cone/rotating_cone.f90
```diff
@@ -0,0 +1,79 @@
...
+       ! Check if the point is inside the cone's base
+       if (r > cone_radius) then
+         s%x(i,1,1,1) = 0.0
+       else
+         s%x(i,1,1,1) = 1.0 - r / cone_radius
+       endif
+    end do
+
+    if ((NEKO_BCKND_DEVICE .eq. 1) .or. (NEKO_BCKND_HIP .eq. 1) &
+       .or. (NEKO_BCKND_OPENCL .eq. 1)) then
```
njansson, line 42:
```suggestion
    if (NEKO_BCKND_DEVICE .eq. 1) then
```
Device covers all cases

njansson (changes requested):

pschlatt1:
Excellent example! We could use this nicely in teaching! One simple comment: shoukdnt we call it convecting cone?

timofeymukha:
> Excellent example! We could use this nicely in teaching! One simple comment: shoukdnt we call it convecting cone?

Thanks, Philipp! Maybe `advected_cone` since we call it advection in the code?

pschlatt1:
yes. advecting (or advected) is in principle better. Gottlieb and Orszag (1977) where the problem originated, called it convecting cone. So perhaps then advecting cone? Knowing that advected would be better.

timofeymukha:
> yes. advecting (or advected) is in principle better. Gottlieb and Orszag (1977) where the problem originated, called it convecting cone. So perhaps then advecting cone? Knowing that advected would be better.

Changed to advecting cone!

njansson (approved):

PR #1112, opened 2024-01-21
Update map_to_equidistant_1d.f90

Shiyu-Sandy-Du:
Assign zero entries in the identity matrix explicitly.


timofeymukha:
it's a tiny matrix, right? So you can just add `ident = 0.0_rp` right after allocation, which will assign everything to 0 from the beginning (or use `rzero` from `math.f90`, but I think it is kind of overkill). Then you can skip the do-loop and ifs you add here.

Shiyu-Sandy-Du:
> it's a tiny matrix, right? So you can just add `ident = 0.0_rp` right after allocation, which will assign everything to 0 from the beginning (or use `rzero` from `math.f90`, but I think it is kind of overkill). Then you can skip the do-loop and ifs you add here.

Sure it's a better way. Code is changed now.

timofeymukha (approved):

Shiyu-Sandy-Du:
The output had a bit bug and fixed now.

MartinKarp (approved):
Looks good

PR #1117, opened 2024-01-23
Generic fld_output, global output_control and simcomp file output

timofeymukha:
This PR introduces:

- `output_control: "global"` for simcomps, which syncs `output_control` and `output_value` with the fluid. Useful to control output of multiple thing by a single set of parameters.
- `fld_output_t`, which is much luck `fluid_output` but does not add the fluid fields to the output. Can be used to output any number of arrays to a new fld.
- Working on outputing simcomp fields to files. What I have now is that if you don't specify `output_filename`, the field (e.g. vorticity) gets added to the `fluid_output` in the case. Otherwise, a new fld file is created. Opinions welcome.
- A small refactor to the `output_t` constructor, so that one can use `init` in descendents as is now standard.

When working on this I realized that e.g. the probes simcom just manages its own output, without adding an `output` to the `sampler`. I'm not sure what is best and how things will go forwards, so here I focused on making it easier to use the `sampler` by adding `fld_output_t`.

### DIFF
--- a/src/io/fld_output.f90
+++ b/src/io/fld_output.f90
@@ -1,0 +1,107 @@
+! Copyright (c) 2020-2023, The Neko Authors
+! All rights reserved.
+!
+! Redistribution and use in source and binary forms, with or without
+! modification, are permitted provided that the following conditions
+! are met:
+!
+!   * Redistributions of source code must retain the above copyright
+!     notice, this list of conditions and the following disclaimer.
+!
+!   * Redistributions in binary form must reproduce the above
+!     copyright notice, this list of conditions and the following
+!     disclaimer in the documentation and/or other materials provided
+!     with the distribution.
+!
+!   * Neither the name of the authors nor the names of its
+!     contributors may be used to endorse or promote products derived
+!     from this software without specific prior written permission.
+!
+! THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+! "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+! LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+! FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+! COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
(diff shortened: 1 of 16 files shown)


timofeymukha:
TODO: handle precision, add the same to lambda2.

njansson (approved):

njansson:
🍻

timofeymukha:
@MartinKarp Whenever you have time :-).

src/io/fld_output.f90
```diff
@@ -0,0 +1,107 @@
...
+! COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+! INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+! BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+! LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+!
+!> Implements `fld_output_t`.
```
MartinKarp, line 33: fld_list_output? Sorry for nitpick, but thought it might be easy to confuse fluid_output and fld_output.

MartinKarp (changes requested):
Looks good I think, good that it is now easier to add things to the "sampler" (which we should not forget to rename at some point). I do not really see the reason for why we have som many output types after this change though, from my side, I think as much as possible should then use fld_(list)_output unless one does some special operation before.

src/io/fld_output.f90
timofeymukha: Yes, I think, in particular the fluid_output could be quite easily removed, which would incidentally address your second comment :-). Here by `fld` I meant the .fld file suffix, not an abbreviation of field or so. Maybe `fld_file_foutput_t` then?

MartinKarp: Sounds good!

timofeymukha: Done

MartinKarp (approved):

PR #1124, opened 2024-02-05
Variable time step

Shiyu-Sandy-Du:
change the cfl evaluation to running avg


timofeymukha:
Something went sideways here, right? The update for some reason contains diffs from a whole bunch of commits.

Shiyu-Sandy-Du:
> Something went sideways here, right? The update for some reason contains diffs from a whole bunch of commits.

Yes, because the variable dt should be compatible with the current version, so I upload the entire package I used for test here. (Or do I only need to update the simulation.f90 and make the sideways merge at later steps?

Shiyu-Sandy-Du:
> > Something went sideways here, right? The update for some reason contains diffs from a whole bunch of commits.
> 
> Yes, because the variable dt should be compatible with the current version, so I upload the entire package I used for test here. (Or do I only need to update the simulation.f90 and make the sideways merge at later steps?

or maybe the better workflow is simply get the push access to the branch and manipulate the branch?

PR #1136, opened 2024-02-13
Check if the file exists

timfelle:
Ensure calls to file readers actually check if the file they are supposed to read exists.

### DIFF
--- a/src/io/generic_file.f90
+++ b/src/io/generic_file.f90
@@ -111,7 +111,6 @@ contains
     integer :: neko_mpi_ierr
 
     file_exists = .false.
-    send_buf = 0
 
     if (pe_rank .eq. 0) then
        ! Stop if the file does not exist
--- a/src/io/generic_file.f90
+++ b/src/io/generic_file.f90
@@ -102,14 +102,27 @@ contains
   !> check if the file exists
   subroutine generic_file_check_exists(this)
     use utils, only: neko_error
+    use comm, only: pe_rank, NEKO_COMM
+    use mpi_f08
     implicit none
 
     class(generic_file_t), intent(in) :: this
     logical :: file_exists
+    integer :: neko_mpi_ierr
+    integer :: send_buf
 
-    ! Stop if the file does not exist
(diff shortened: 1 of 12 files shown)


njansson (changes requested):
To avoid potential badness on e.g. lustre filesystem at scale, I would suggest to do the inquire on one rank, and bcast the logical.




issues on lustre filesystems, I would

timofeymukha:
Would it be better to add a `check_exists` TBP to `generic_file_t` with
```fortran
    inquire(file=this%fname, exist=file_exists)
    if (.not. file_exists) then
       call neko_error('File does not exist: '//trim(this%fname))
    end if
```
and only have `call check_exists()` in the subtypes?

src/io/generic_file.f90
```diff
@@ -97,5 +99,34 @@ subroutine generic_file_set_start_counter(this, n)
...
+       ! Stop if the file does not exist
+       inquire(file=this%fname, exist=file_exists)
+       if (file_exists) send_buf = 1
+
+    end if
+    call MPI_Bcast(send_buf, 1, MPI_INT, 0, NEKO_COMM, neko_mpi_ierr)
+
+    if (send_buf .eq. 0) then
+       call neko_error('File does not exist: '//trim(this%fname))
+    end if
```
njansson, line 127: A bit nit picky, but how about using `MPI_LOGICAL` directly?
```suggestion

    file_exists = .false.
    send_buf = 0

    if (pe_rank .eq. 0) then
       ! Stop if the file does not exist
       inquire(file=this%fname, exist=file_exists)
    end if
    call MPI_Bcast(file_exists, 1, MPI_LOGICAL, 0, NEKO_COMM, neko_mpi_ierr)

    if (.not. file_exists) then
       call neko_error('File does not exist: '//trim(this%fname))
    end if
```

timfelle: It is just a bad habit comming from C++ where the boolean MPI data type does not exist.

njansson (approved):

PR #1137, opened 2024-02-13
Adds Neumann boundary condition for scalars.

timofeymukha:
First step towards Neumann / flux boundary conditions.

* Currently only works and CPUs and scalar fields.
* Selected via `n= 'value'` in the JSON, similar to the approach for Dirichlet bcs.
* The scalar_mms is updated to use a Neumann bc consistent with the desired solution, and it seems to work well. Clearly shows errors if one sets the incorrect flux as well.

Not sure if it works when the axes are not aligned with the faces, due to the following statement in the Nek5000 docs regarding scalar bcs. "If the normal, tangent, and bitangent vectors are not aligned with the mesh the stress formulation has to be used." I don't see why this is the case for the scalar, though.

There a couple of questions regarding the implementation of bcs in the scalar types, which I would like to clear up in this PR review as well, I will leave some comments below for threaded discussions.


src/scalar/scalar_scheme.f90
```diff
@@ -97,9 +98,14 @@ module scalar_scheme
...
+     !> Neumann conditions.
+     type(neumann_t) :: neumann_bcs(NEKO_MSH_MAX_ZLBLS)
      !> User Dirichlet conditions.
      type(usr_scalar_t) :: user_bc
+     !> Number of Dirichlet bcs.
      integer :: n_dir_bcs = 0
+     !> Number of Neumann bcs.
+     integer :: n_neumann_bcs = 0
      !> List of boundary conditions.
      type(bc_list_t) :: bclst
```
timofeymukha, line 110: This list is never used, it seems. Do we need it? Maybe doesn't hurt to have it...

src/scalar/scalar_scheme.f90
```diff
@@ -201,30 +207,39 @@ subroutine scalar_scheme_add_bcs(this, zones, bc_labels)
...
+!          do j = 1, i-1
+!             if (bc_label .eq. bc_labels(j)) then
+!                bc_exists = .true.
+!                bc_idx = j
+!             end if
+!          end do
+
+!          if (bc_exists) then
+!             call this%dir_bcs(j)%mark_zone(zones(i))
+!          else
```
timofeymukha, line 227: @MartinKarp As you see, I've commented out this stuff and not used it for the Neumann bc either. What is the idea here? Are bcs with the same Dirichlet values somehow merged into 1? The reason I removed it is that I seem to get a crash if I set 2 bcs to the same value.

src/scalar/scalar_pnpn.f90
```diff
@@ -186,11 +191,18 @@ subroutine scalar_pnpn_init(this, msh, coef, gs, params, user, &
     end if
     call this%bc_res%finalize()
     call this%bc_res%set_g(0.0_rp)
+
```
timofeymukha, line 194: @MartinKarp Could you say what is the going on ehre with the `bc_res`? I believe that one creates one Dirichlet bc, that has all the facets of the existing Dirichlet bcs, and then sets the value to 0. Then we put this bc into a list, which is used in the step function as

```fortran
     ! Apply Dirichlet boundary conditions
      call bc_list_apply_scalar(this%bclst_ds, s_res%x, dm_Xh%size())
```
Why can we not create a list with all the Dirichlet conditions directly, as I do with Neumann? And why does it work out to apply a single condition, which has g set to 0? Thanks!!

src/scalar/scalar_scheme.f90
MartinKarp: I think maybe this list should be the one you use? Neumann conditions will not pnpn specific or?

src/scalar/scalar_pnpn.f90
MartinKarp: I think Neumann should be in the scheme right? Because those a BCs that make "physical" sense. The BCs in bclst_ds are for ds, e.g. Delta s, so a dirichlet condition on s means setting 0 for delta s along those boundaries, so we need to create new BCs for that. The BCs for s should be specified in the scalar scheme base class as that is universal regardless of the exact scheme we are using. We could technically implement a version of pnpn that does not operate on delta s and then this part would of course not make much sense.

src/scalar/scalar_scheme.f90
MartinKarp: Yeah, the idea I think was to merge them into one bc I think. I guess it doesn't matter too much, but it would be a bit more perfromant to merge them since we would avoid some kernel launches.

MartinKarp (commented):
Some comments, I hope they help. I think Neumann should be in the base class, rather than pnpn.

src/scalar/scalar_pnpn.f90
timofeymukha: Yes, sounds good, I can move the Neumann up. I missed that the second argument was s_res, and not s. Follow up question then, where are the Dirichlet bcs for `s` applied then? It is not in `step` then?

src/scalar/scalar_scheme.f90
timofeymukha: OK, I get the idea.

src/scalar/scalar_scheme.f90
timofeymukha: That's a possibility, but should the dirichlet bcs be excluded from it then, because now everything is added here.

src/scalar/scalar_pnpn.f90
MartinKarp: It is called every step I am pretty sure, lets see... line 363 I think it is in scalar_pnpn.

src/scalar/scalar_scheme.f90
MartinKarp: Depends on how you wanna do with the scalar_bc subroutine which applies the dirichlet bc each step I guess. If you wanna do Neumann there too or only dirichlet

timofeymukha: Yes, now I see it, so the list is used in fact. Unfortunately, I think we will have to keep separate lists, because we can't do a single call to the `bclist_apply` anyway. For Dirichlet one passes the scalar field and for Neumann the rhs field, so we need two calls.

Regarding this%bc_apply, it is a question of when we need to apply the different bcs. For Neumann, we have to do it before makeext. What about dirichlet, can it be moved up to the same place or does it need to be applied where it is done now?

timofeymukha: I think for now i will remove bc_apply and just do the 2 calls bclist_apply calls for neuman and dirichlet separately. Better be on the safe side.

timofeymukha:
OK, now I am happy, so ready for review :-).

src/scalar/scalar_scheme.f90
```diff
@@ -201,30 +209,42 @@ subroutine scalar_scheme_add_bcs(this, zones, bc_labels)
...
```
njansson, line 232: Should these comments be kept?

src/bc/neumann.f90
```diff
@@ -0,0 +1,153 @@
+! Copyright (c) 2020-2021, The Neko Authors
```
njansson, line 1:
```suggestion
! Copyright (c) 2024, The Neko Authors
```

src/scalar/scalar_scheme.f90
timofeymukha: Yes, I'm the one who commented this out and added the comment. So it's not new code, but rather a soft, perhaps temporary, removal.

njansson (approved):

MartinKarp (approved):

PR #1138, opened 2024-02-13
Add documentation for the user file

vbaconnet:
I am creating the PR now even though I am not done yet, in case people have strong opinions on the wording, the structure of the document or how certain things should be explained. 

Will finish the other half tomorrow. 

Kind of wondering if the main page `index.md` should be made prettier... maybe not a priority though.

### DIFF
--- a/doc/pages/user-file.md
+++ b/doc/pages/user-file.md
@@ -1,0 +1,234 @@
+# User File {#user-file}
+
+## High-level structure
+
+The current high-level structure of the user file is shown below.
+
+```fortran
+module user
+  use neko
+  implicit none
+
+contains
+
+  ! Register user defined functions here (see user_intf.f90)
+  subroutine user_setup(u)
+    type(user_t), intent(inout) :: u
+
+    
+  end subroutine user_setup
+ 
+end module user
+
+```
+
+The user file implements the `user` module. The `user` modules contains a subroutine named `user_setup`, which we use to interface the internal procedures defined in `src/common/user_intf.f90` with the subroutines that you will implement in your user file. Each user subroutine should be implemented under the `contains` statement, below `user_setup`.
+
+A basic working example is demonstrated [below](#basic-working-example).
+
+@note The above code snippet is the most basic code structure for the user file. Compiling it and running it would be equivalent to running the "vanilla" neko executable `bin/neko` in your local neko installation folder.
+
+## Default user functions
+
+The following user functions, if defined in the user file, will always be executed, regardless of what is set in the case file: 
+
+- [user_init_modules](#user-file_init-and-final): For initializing user variables and objects
+- [user_finalize_modules](#user-file_init-and-final): For finalizing, e.g freeing variables and terminating processes
+- [user_check](#user-file_user-check): Executed at the end of every time step, for e.g. computing and/or outputting user defined quantities.
+- [material_properties](#user-file_mat-prop): For computing and setting material properties such as `rho`, `mu`, `cp` and `lambda`.
+- [user_mesh_setup](#user-file_user-mesh-setup): For applying any deformation to the element nodes of the mesh specified in the case file.
+- [scalar_user_bc](#user-file_scalar-bc): For applying boundary conditions to the scalar, on all zones that are not already specified with uniform dirichlet values e.g. `d=1`. For more information on the scalar, see the [relevant section of the case file](#case-file_scalar).
+
+### Initializing and finalizing {#user-file_init-and-final}
+
+These two subroutines may be used to initialize/finalize any user defined variables, external objects, or processes. They are respectively executed right before/after the simulation time loop.
+
+```.f90
+
+  ! Initialize user variables or external objects
+  subroutine initialize(t, u, v, w, p, coef, params)
+    real(kind=rp) :: t
+    type(field_t), intent(inout) :: u
+    type(field_t), intent(inout) :: v
+    type(field_t), intent(inout) :: w
+    type(field_t), intent(inout) :: p
+    type(coef_t), intent(inout) :: coef
+    type(json_file), intent(inout) :: params
+
+    ! insert your initialization code here
+
+  end subroutine initialize
+
+  ! Finalize user variables or external objects
+  subroutine finalize(t, params)
+    real(kind=rp) :: t
+    type(json_file), intent(inout) :: params
+
+    ! insert your code here
+
+  end subroutine initialize
+```
+
+In the example above, the subroutines `initialize` and `finalize` contains the actual implementations, and both need to be interfaced to the internal procedures `user_init_modules` and `user_finalize_modules` in `user_setup`:
+
+```.f90
+
+  ! Register user defined functions (see user_intf.f90)
+  subroutine user_setup(u)
+    type(user_t), intent(inout) :: u
+
+    u%user_init_modules => initialize 
+    u%user_finalize_modules => finalize 
+
+  end subroutine user_setup
+
+```
+
+@note `user_init_modules` and `user_finalize_modules` are independent of each other. Using one does not require the use of the other.
+
+### Computing at every time step {#user-file_user-check}
+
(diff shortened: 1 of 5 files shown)


njansson:
Great stuff!

doc/pages/user-file.md
```diff
@@ -0,0 +1,234 @@
...
+
+And of course not forgetting to register our function in `user_setup` by adding the following line:
+
+```.f90
+u%material_properties => set_material_properties
+```
+
+### Runtime mesh deformation {#user-file_user-mesh-setup}
+
+This user function allows for the modification of the mesh at runtime, by acting on the element nodes of the mesh specified in the case file. The example below is taken from the [compression example](https://github.com/ExtremeFLOW/neko/blob/a0613606360240e5059e65d6d98f4a57cf73e237/examples/tgv/tgv.f90#L27).
```
timofeymukha, line 151: Here one should maybe make clear that this is a one-time thing done in the beginning, and *not* an actual moving mesh capability.

vbaconnet:
I think this is more or less finished! I guess things to discuss could be how things are explained, the wording, nomenclatures, and/or vocabulary, etc.

doc/pages/user-file.md
```diff
@@ -0,0 +1,588 @@
...
+### Registries {#user-file_tips_registries}
+
+Neko uses the concept of `registry` as a practical way to retrieve fields and point zones anywhere in the user file. 
+
+The field registry `neko_field_registry` is often used in user functions where certain fields are not directly accessible as arguments. One can retrieve any field in the registry by its `name` with `neko_field_registry%%get_field(name)`. Default fields that are added to the registry are `u,v,w,p` and `s` if running with the scalar enabled. For a practical example of usage, see the [rayleigh benard example](https://github.com/ExtremeFLOW/neko/blob/49925b7a04a638259db3b1ddd54349ca57f5d207/examples/rayleigh-benard/rayleigh.f90#L102-L105)
+
+Other fields will be added the registry in the following cases:
+- If running with `simulation_components.vorticity` enabled, the fields `omega_x, omega_y, omega_z` will be accessible in the registry.
+- If running with `simulation_components.lambda2` enabled, the field `lambda2` will be accessible in the registry.
+
```
vbaconnet, line 585: @timofeymukha I did not include the fields added to the registry when using the LES stuff [(here)](https://github.com/ExtremeFLOW/neko/blob/3732f904fc6ee0926979a15c076050271a33fcf8/src/les/les_model.f90#L116-L121) as I did not completely understand it... if you would like to add a few lines here to explain it that would be nice!

timofeymukha: I think having a definitive list here is a bit beyond the scope. Rather, the doc for each simcomp should discuss its interaction with the registry. Here, it is perhaps sufficient to give an example or to, as you have done.

timofeymukha: But to answer your question, for the les, you select a name for the field that will hold the computed SGS viscosity, otherwise the default name is "nut", if I remember correctly :-).

vbaconnet:
Removed the `don't merge` label :)

doc/pages/user-file.md
```diff
@@ -0,0 +1,726 @@
...
+u%material_properties => set_material_properties
+```
+
+### Runtime mesh deformation {#user-file_user-mesh-setup}
+
+This user function allows for the modification of the mesh at runtime, by acting
+on the element nodes of the mesh specified in the case file. This function is
+only called once before the simulation time loop. The example below is taken
+from the [compression
+example](https://github.com/ExtremeFLOW/neko/blob/a0613606360240e5059e65d6d98f4a57cf73e237/examples/tgv/tgv.f90#L27-L42).
```
timofeymukha, line 221: Is that one up to date? It has been lagging behind quite a bit before. Maybe safer to just ref the ordinary TGV.

doc/pages/user-file.md
```diff
@@ -0,0 +1,726 @@
...
+As explained in the [case file](case-file.md) page, certain components of the
+simulation can be set to be user defined. These components and their associated
+user functions are:
+
+| Description                     | User function                                              | JSON Object in the case file     |
+|---------------------------------|------------------------------------------------------------|----------------------------------|
+| Fluid initial condition         | [fluid_user_ic](#user-file_user-ic)                        | `case.fluid.initial_condition`   |
+| Scalar initial condition        | [scalar_user_ic](#user-file_user-ic)                       | `case.scalar.initial_condition`  |
+| Fluid inflow boundary condition | [fluid_user_if](#user-file_fluid-user-if)                  | `case.fluid.inflow_condition`    |
+| Scalar boundary conditions      | [scalar_user_bc](#user-file_scalar-bc)                     | (user function is always called) |
```
timofeymukha, line 303: So maybe the doc for this one should be in this section, not above?

doc/pages/user-file.md
```diff
@@ -0,0 +1,726 @@
...
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_memcpy(s%x, s%x_d, s%dof%size(), &
+                          HOST_TO_DEVICE, sync=.false.)
+    end if
+
+  end subroutine set_s_ic
+
+```
+
+@note Notice the use of the `NEKO_BCKND_DEVICE` flag, which will be set to 1 if
```
timofeymukha, line 419: Repetition, can be removed.

doc/pages/user-file.md
timofeymukha: Referring to line 303

timofeymukha:
Great job, just a few suggestions. I also added a mention of the Neuman bc to the scalar_bc user function, a bit ahead of time :-).

doc/pages/user-file.md
vbaconnet: Sure!

doc/pages/user-file.md
vbaconnet: Yea so I added a comment below on line 307 about this, but maybe we can move it if you think it is clearer for the user

doc/pages/user-file.md
timofeymukha: I see. It's just that there were 2 really close by here, so felt unnecessary. But it's not a big deal.

doc/pages/user-file.md
vbaconnet: oh i just realized the links points to tgv but the code snippet is compression

doc/pages/accelerators.md
```diff
@@ -82,11 +82,9 @@ To copy data between host and device (and device to use) use the routine device:
   call device_memcpy_cptr(y_d, x_d, s, DEVICE_TO_DEVICE)
 ~~~~~~~~~~~~~~~
 
-@attention device::device_memcpy_cptr defaults to asynchronous data transfer. The optional boolean argument `sync` must be true if synchronous transfers are needed.
```
njansson, line 85: I think this should still be there as a more developer info

timofeymukha (approved):

njansson (changes requested):

doc/pages/accelerators.md
timofeymukha: But the `sync` is no longer optional, right? So some modification is required.

vbaconnet: Not sure how it should be changed now that the argument is non-optional anymore. Any specific instructions?

doc/pages/accelerators.md
```diff
@@ -82,11 +82,9 @@ To copy data between host and device (and device to use) use the routine device:
   call device_memcpy_cptr(y_d, x_d, s, DEVICE_TO_DEVICE)
 ~~~~~~~~~~~~~~~
```
njansson, line 84:
```suggestion

@attention device::device_memcpy_cptr defaults to asynchronous data transfer. The optional boolean argument `sync` must be true if synchronous transfers are needed.
```

njansson (approved):

PR #1139, opened 2024-02-14
Docstrings, and "only" for operators

timofeymukha:
* Adds `only` to imports in operators.f90
* Adds docstrings to the operators. I am unsure about one parameter in the `ortho`, please help :-).
* Renamed `c_Xh` to `coef` in `curl`, because all the other operators use `coef`.

### DIFF
--- a/src/math/operators.f90
+++ b/src/math/operators.f90
@@ -180,7 +180,7 @@ contains
   !! @param vy The y component of the advecting velocity.
   !! @param vz The z component of the advecting velocity.
   !! @param Xh The function space for the fields involved.
-  !! @param coef The SEM coeffients.
+  !! @param coef The SEM coefficients.
   !! @param es Starting element index, defaults to 1.
   !! @param ee Last element index, defaults to mesh size.
   subroutine conv1(du, u, vx, vy, vz, Xh, coef, es, ee)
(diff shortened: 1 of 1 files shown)


njansson (commented):
I think that with dropping `Xh` we also dropped the connection/indication that the given coefficient is related to a certain function space

timofeymukha:
> I think that with dropping `Xh` we also dropped the connection/indication that the given coefficient is related to a certain function space

I know you like the `Xh`, but here there was just 1 operator that used this notation, so I decided to homogenize.

njansson:
> > I think that with dropping `Xh` we also dropped the connection/indication that the given coefficient is related to a certain function space
> 
> I know you like the `Xh`, but here there was just 1 operator that used this notation, so I decided to homogenize.

fair enough, also given that what `Xh` means in this context is unclear

njansson (approved):

src/math/operators.f90
```diff
@@ -108,20 +130,30 @@ subroutine opgrad(ux,uy,uz,u,coef, es, ee) ! weak form of grad
   end subroutine opgrad
 
   !> Othogonalize with regard to vector (1,1,1,1,1,1...,1)^T.
-  subroutine ortho(x,n ,glb_n)
+  !! @param x The vector to orthogonolize.
+  !! @param n The size of `x`.
+  !! @param glb_n ?
```
MartinKarp, line 135:
```suggestion
  !! @param glb_n The global number of elements of 'x' across all MPI ranks. Be careful with overflow!
```

MartinKarp (changes requested):
Looks good!

timofeymukha:
@MartinKarp You forgot to approve :-)

PR #1142, opened 2024-02-16
Feature/brinkman

timfelle:
Introduction of a Brinkman forcing term.

The idea of this is to enforce regions of low permeability to block flow in certain regions.
Included is an example with a Stanford bunny placed in a square duct as a block of resistance.

Still some todo points, but wanted to open it up so I could start to get some feedback

- [x] Documentation of the example and the source term in general, including an image or two.
- [ ] Device acceleration for computation of the brinkman term.
- [ ] Add open mesh example.
- [ ] Trim aabb_tree, triangles outside the fluid domain. -> Postponed, there are some situations which require this to be done with care.
- [ ] Expand to utilize point zones. -> moved to sepparate PR

PR #1143, opened 2024-02-16
Extending the brinkman force to support Point Zone objects

timfelle:
Dependent on: #1142 

Extension of the brinkman forcing term to also allow point zones to be inserted as areas of low permeability.

- [ ] Implement point zone brinkman forcing.
- [ ] Setup example with block and sphere.
- [ ] Ensure multiple zones can be added together.


timfelle:
I am currently making sure the default case files in the example folder actually functions as well as they should.

njansson (approved):

timfelle:
@timofeymukha, could you have a look at this PR. I believe it contains all it is ready to go now that the multiple objects system have been implemented correctly.

PR #1148, opened 2024-02-20
Feature/variable dt

Shiyu-Sandy-Du:
This PR is re-issued following PR #1013. The branch in PR #1013 is too old and hence full of conflicts. This branch is built on most up-to-date develop-branch. In this branch, two things are modified:

1. variable time stepping uses a running average instead of an arithmetic average (to cope with sudden explosion of the cfl number)
2. lag fields are promoted to _scheme type (not just for pnpn)

Special acknowledge to @adperezm for setting up the architecture in PR#1013.

### DIFF
--- a/src/simulation.f90
+++ b/src/simulation.f90
@@ -282,7 +282,7 @@ contains
   !> Set new dt based on cfl if requested
   !! @param dt current time step
   !! @param params json case file
-  !! @param cfl courant number of current iteration
+  !! @param cfl Courant number of current iteration.
   !! @param cfl_avrg average courant number
   !! @param dt_last_change time step since last dt change
   subroutine simulation_setdt(dt, params, cfl, cfl_avrg, dt_last_change, tstep)
--- a/src/simulation.f90
+++ b/src/simulation.f90
@@ -66,14 +66,21 @@ contains
     ! for variable_time_steping
     integer :: dt_last_change = 0
     real(kind=rp) :: cfl_avrg = 0_rp
+    real(kind=rp) :: set_cfl
 
     t = 0d0
     tstep = 0
     call neko_log%section('Starting simulation')
     write(log_buf,'(A, E15.7,A,E15.7,A)') 'T  : [', 0d0,',',C%end_time,')'
     call neko_log%message(log_buf)
-    write(log_buf,'(A, E15.7)') 'dt :  ', C%dt
-    call neko_log%message(log_buf)
+    call C%params%get('case.constant_cfl', set_cfl, found)
+    if (.not. found) then
+       write(log_buf,'(A, E15.7)') 'dt :  ', C%dt
+       call neko_log%message(log_buf)
+    else
+       write(log_buf,'(A, E15.7)') 'cfl :  ', set_cfl
+       call neko_log%message(log_buf)
+    end if
 
     call C%params%get('case.restart_file', restart_file, found)
     if (found .and. len_trim(restart_file) .gt. 0) then
@@ -102,8 +109,12 @@ contains
        call profiler_start_region('Time-Step')
        tstep = tstep + 1
        start_time = MPI_WTIME()
-       cfl = C%fluid%compute_cfl(C%dt)
+       if (tstep .eq. 1) cfl = C%fluid%compute_cfl(C%dt)
        if (dt_last_change .eq. 0) cfl_avrg = cfl
+       call simulation_setdt(C%dt, C%params, cfl, cfl_avrg, dt_last_change, tstep)
+       !calculate the cfl after the possibly varied dt
+       cfl = C%fluid%compute_cfl(C%dt)
+
        call neko_log%status(t, C%end_time)
        write(log_buf, '(A,I6)') 'Time-step: ', tstep
        call neko_log%message(log_buf)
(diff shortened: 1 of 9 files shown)


Shiyu-Sandy-Du:
restart is plausible but turns out to be wrong here. Work on time step info in chkp file is on-going.

src/simulation.f90
```diff
@@ -267,5 +281,72 @@ subroutine simulation_joblimit_chkp(C, t)
 
   end subroutine simulation_joblimit_chkp
 
+    !> Set new dt based on cfl if requested
+  !! @param dt current time step
```
njansson, line 285:
```suggestion
  !> Set new dt based on cfl if requested
  !! @param dt current time step
```

src/simulation.f90
```diff
@@ -134,6 +150,8 @@ subroutine neko_solve(C)
        call neko_simcomps%compute(t, tstep)
 
        call C%q%eval(t, C%dt, tstep)
+       !> Sample is called after fluid step. Therefore, C%ulag(1) has 
+       !! u for the last time step
```
njansson, line 154:
```suggestion
```

src/simulation.f90
```diff
@@ -94,19 +104,25 @@ subroutine neko_solve(C)
...
+
        call neko_log%status(t, C%end_time)
        write(log_buf, '(A,I6)') 'Time-step: ', tstep
        call neko_log%message(log_buf)
        call neko_log%begin()
 
        write(log_buf, '(A,E15.7,1x,A,E15.7)') 'CFL:', cfl, 'dt:', C%dt
        call neko_log%message(log_buf)
-
+
```
njansson, line 125:
```suggestion

```

src/simulation.f90
```diff
@@ -63,14 +63,24 @@ subroutine neko_solve(C)
...
     write(log_buf,'(A, E15.7,A,E15.7,A)') 'T  : [', 0d0,',',C%end_time,')'
     call neko_log%message(log_buf)
-    write(log_buf,'(A, E15.7)') 'dt :  ', C%dt
-    call neko_log%message(log_buf)
+    call C%params%get('case.constant_cfl', set_cfl, found)
+    if (.not. found) then
+       write(log_buf,'(A, E15.7)') 'dt :  ', C%dt
+       call neko_log%message(log_buf)
+    else
+       write(log_buf,'(A, E15.7)') 'cfl :  ', set_cfl
```
njansson, line 81:
```suggestion
       write(log_buf,'(A, E15.7)') 'CFL :  ', set_cfl
```

src/simulation.f90
```diff
@@ -94,19 +104,25 @@ subroutine neko_solve(C)
...
     do while (t .lt. C%end_time .and. (.not. jobctrl_time_limit()))
        call profiler_start_region('Time-Step')
        tstep = tstep + 1
        start_time = MPI_WTIME()
+       if (tstep .eq. 1) cfl = C%fluid%compute_cfl(C%dt)
+       if (dt_last_change .eq. 0) cfl_avrg = cfl
+       call simulation_setdt(C%dt, C%params, cfl, cfl_avrg, dt_last_change, tstep)
+       !calculate the cfl after the possibly varied dt
        cfl = C%fluid%compute_cfl(C%dt)
+
```
njansson, line 117:
```suggestion
    call profiler_start
    cfl = C%fluid%compute_cfl(C%dt)
    start_time_org = MPI_WTIME()

    do while (t .lt. C%end_time .and. (.not. jobctrl_time_limit()))
       call profiler_start_region('Time-Step')
       tstep = tstep + 1
       start_time = MPI_WTIME()
       if (dt_last_change .eq. 0) then
          cfl_avrg = cfl
       end
       call simulation_setdt(C%dt, C%params, cfl, cfl_avrg, dt_last_change, tstep)
       !calculate the cfl after the possibly varied dt
       cfl = C%fluid%compute_cfl(C%dt)

```

njansson (changes requested):

njansson (approved):

timofeymukha:
Please update case-file.md with the new options.

src/simulation.f90
```diff
@@ -267,5 +279,72 @@ subroutine simulation_joblimit_chkp(C, t)
...
+    call params%get('case.timestep', max_dt, found)
+    call params%get('case.constant_cfl', set_cfl, found)
+
+    if (found .eqv. .true.) then
+       if (tstep .eq. 1) then ! set the first dt for desired cfl
+          dt = min(set_cfl/cfl*dt, max_dt)
+       else
+         call json_get_or_default(params, 'case.cfl_min_update_frequency',&
+                                    min_update_frequency, 1)
```
timofeymukha, line 310: I'm perhaps confused, but isn't this the maximum frequency? This is expressed in the number of timesteps, right? So if I put to, it will update at a max once per 2 timesteps and not with a higher frequency.

timofeymukha (changes requested):

src/simulation.f90
Shiyu-Sandy-Du: You point out the problem here. So it should be named "min_update_time_steps" instead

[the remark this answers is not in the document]
Shiyu-Sandy-Du: would be a good idea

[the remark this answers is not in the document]
Shiyu-Sandy-Du: shall I issue it in this PR?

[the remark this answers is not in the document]
timofeymukha: yes, just update the branch and the PR will use the latest commit automatically.

timofeymukha: These are always a horror to name. I guess one can contract time_steps to tsteps, because we use that for "output_control" as well.

timfelle:
It seems you are missing an update to the .depends, which is why the tests are failing.

Shiyu-Sandy-Du:
> It seems you are missing an update to the .depends, which is why the tests are failing.

Yes it's the issue. now it's fixed.

Shiyu-Sandy-Du:
> Please update case-file.md with the new options.

Done

timofeymukha (approved):

PR #1149, opened 2024-02-21
Updated point zone initialization

timfelle:
Point zones are now initialized before the physics. This opens the door for zone wise material parameters, source terms etc.


vbaconnet (approved):
Great, thanks for the fix!

PR #1152, opened 2024-02-21
Addition of new point zone type: Cylinder.

timfelle:
The cylinder point zone is constructed from its end points and radius. This allow the user to define the cylinder in arbitrary orientations and sizes.


src/mesh/point_zones/cylinder_point_zone.f90
```diff
@@ -0,0 +1,184 @@
+! Copyright (c) 2019-2021, The Neko Authors
```
timofeymukha, line 1:
```suggestion
! Copyright (c) 2024, The Neko Authors
```

njansson (approved):

vbaconnet (approved):
love to see the point zones growing!!

PR #1153, opened 2024-02-22
Add possibility to provide default string for json string array reader

timofeymukha:
Useful to not get garbage in the read array. For example, for bc_labels, we can default to "not" as we do now outside of JSON, or some similar dummy value.

### DIFF
--- a/src/common/json_utils.f90
+++ b/src/common/json_utils.f90
@@ -179,11 +179,13 @@ contains
   !> Retrieves a string array parameter by name or throws an error
   !! @param json The json to retrieve the parameter from.
   !! @param name The full path to the parameter.
-  !! @value value The variable to be populated with the retrieved parameter.
-  subroutine json_get_string_array(json, name, value)
+  !! @param value The variable to be populated with the retrieved parameter.
+  !! @param filler The default string to fill empty array items with.
+  subroutine json_get_string_array(json, name, value, filler)
     type(json_file), intent(inout) :: json
     character(len=*), intent(in) :: name
     character(len=*), allocatable, intent(out) :: value(:)
+    character(len=*), optional, intent(in) :: filler
     logical :: found
     type(json_value), pointer :: json_val, val_ptr
     type(json_core) :: core
@@ -211,6 +213,8 @@ contains
 
        if (len(string_value) .gt. 0) then
           value(i) = string_value
+       else if(present(filler)) then
+          value(i) = filler
        end if
     end do
 


njansson (approved):

timfelle:
It is really sad bounds checking is not really a thing in fortran...

timfelle (approved):

njansson:
> It is really sad bounds checking is not really a thing in fortran...

Hey, bounds checking **is really a thing** in fortran compared to C ;) (e.g. `-fbounds-check` in gfortran)

timfelle:
> > It is really sad bounds checking is not really a thing in fortran...
> 
> Hey, bounds checking **is really a thing** in fortran compared to C ;) (e.g. `-fbounds-check` in gfortran)

Ah right. Why do we not have a CI that does that then?

njansson:
> > > It is really sad bounds checking is not really a thing in fortran...
> > 
> > 
> > Hey, bounds checking **is really a thing** in fortran compared to C ;) (e.g. `-fbounds-check` in gfortran)
> 
> Ah right. Why do we not have a CI that does that then?

Since that would fail due to our "creative" use of indexing (i.e `x(i,1,1,1)`) to flatten some arrays, and get nice vectorisation

timfelle:
> > > > It is really sad bounds checking is not really a thing in fortran...
> > > 
> > > 
> > > Hey, bounds checking **is really a thing** in fortran compared to C ;) (e.g. `-fbounds-check` in gfortran)
> > 
> > 
> > Ah right. Why do we not have a CI that does that then?
> 
> Since that would fail due to our "creative" use of indexing (i.e `x(i,1,1,1)`) to flatten some arrays, and get nice vectorisation

Is that used in places where an elemental function could not do the trick instead?

njansson:
> > > > > It is really sad bounds checking is not really a thing in fortran...
> 
> > > > 
> 
> > > > 
> 
> > > > Hey, bounds checking **is really a thing** in fortran compared to C ;) (e.g. `-fbounds-check` in gfortran)
> 
> > > 
> 
> > > 
> 
> > > Ah right. Why do we not have a CI that does that then?
> 
> > 
> 
> > Since that would fail due to our "creative" use of indexing (i.e `x(i,1,1,1)`) to flatten some arrays, and get nice vectorisation
> 
> 
> 
> Is that used in places where an elemental function could not do the trick instead? 

Yes and no, the main issue with elemental is the uncertainty on how well certain compilers implement it.

PR #1154, opened 2024-02-25
Adds a Boussinesq source term for boyancy

timofeymukha:
Formulated as  $\beta (T - T_{ref}) g$, where $\beta$ defaults to $1/T_{ref}$.
Reads the following entries:
   - `scalar_field`: The name of the scalar that drives the source term, defaults to "s".
   - `reference_value`: The reference value of the scalar.
   - `g`: The gravity vector.
   - `beta`: The thermal expansion coefficient, defaults to the inverse of `ref_value`.

I modify the RBC example to use the source term. In the example, the forcing is just equal to $T$ in the $z$ direction, which is achieved by manipulating the constants in the source term. Also, it seems possible to increase the timestep by a factor of 100, @adperezm can you confirm?


njansson (approved):

MartinKarp:
Have you run the full RBC e.g. (to T=250)? The time step can be really large in the beginning as you don't have any convection, but will explode after if its not low. I think the transition to turbulence is around T=30 in this case.

examples/rayleigh-benard-cylinder/rayleigh.case
```diff
@@ -20,7 +20,10 @@
             "Ra": 1e8,
             "source_terms": [
                 {
-                    "type": "user_vector"
+                    "type": "boussinesq",
+                    "g" : [0, 0, 1],
+                    "reference_value": 0,
+                    "beta": 1
```
MartinKarp, line 26: If you do another normalization, could one do an expression based on Ra and Nu here?

adperezm: As I understood, one can not have formulas in json. Right? Or did I misunderstand the question?

MartinKarp: I guess, I am mostly wondering if one can use this easily if one plays around with the normalization. Otherwise I guess one would go for the user forcing for other normalizations

timofeymukha: As long as it fits into the formula somehow than yes. But unfortunately one would have to punch in the numbers, we can't directly refer to other json parameters, just like @adperezm says.

timofeymukha:
I can run up to 250, but I guess it does not look very turbulent. This is the velocity in y at t=250

image

timofeymukha:
I seem to get the same on develop though. @adperezm could you comment on how the solution should look :-)?

MartinKarp:
It should undergo a transition at t=30, at least that was the case when I set it up orignally ;) For me that still happens though, here is the velocity mag for me when I run on develop:
Screenshot from 2024-03-06 16-05-08

timofeymukha:
@MartinKarp @adperezm So, it turns out RBC is tricky, with a big time step it just doesn't transition, and you seem to need 2e-3 to keep it alive once it does. So,, I removed the dt change for the case. It takes at least an hour to run the whole case, so I ran for 10s with develop and the new source term and the solution is identical. So I think we are good

MartinKarp:
I guess, if the time step is too big the initial disturbances are probably just smoothed out rather quickly. Of course, one can increase the noise in the ic, but yeah there is a trade-off to be made.

MartinKarp:
Yeah, go for it

timofeymukha:
> Yeah, go for it

You need to approve :)

njansson:
> > Yeah, go for it
> 
> You need to approve :)

@MartinKarp, go, no-go?

timfelle (approved):

PR #1159, opened 2024-02-28
TS wave test examples

Shiyu-Sandy-Du:
This case is used as an example case for variable dt. Please merge it only after feature/variable_dt is merged.


examples/TS_channel/README.md
```diff
@@ -0,0 +1,3 @@
+#Tollmien-Schlichting (TS) wave test in channel flow.
```
njansson, line 1:
```suggestion
## Tollmien-Schlichting (TS) wave test in channel flow.
```

njansson (approved):

MartinKarp (changes requested):
Really nice! Check for indentation and also I think it might be good with a few comments in the code.

examples/TS_channel/TS_channel.f90
```diff
@@ -0,0 +1,312 @@
...
+                        ur_2D,ui_2D,vr_2D,vi_2D, &
+                        ur_3D,ui_3D,vr_3D,vi_3D,wr_3D,wi_3D)
+       u%x(i,1,1,1) = uvw(1)
+       v%x(i,1,1,1) = uvw(2)
+       w%x(i,1,1,1) = uvw(3)
+    end do
+
+  end subroutine user_ic
+
+  function inArray(y,y_list) result(isIn)
```
timofeymukha, line 132: use _ in names instead of camelCase

MartinKarp:
@Shiyu-Sandy-Du do you think you will get a review for @pschlatt1? Otherwise I think we can merge it in.

njansson:
> @Shiyu-Sandy-Du do you think you will get a review from @pschlatt1? Otherwise I think we can merge it in.

We have discussed thus offline, so I think we can merge

PR #1162, opened 2024-03-01
Add built-in profiler

timfelle:
Addition of a built in profiler which will report the global cpu-time and evaluation count.

The new profiler is disabled by default and can be turned on by setting `--enable-profiling` when configuring Neko.


timofeymukha:
Is this related to #1067 ?

timfelle:
> Is this related to #1067 ?

I had no idea that PR existed. Just just wanted to measure time so i expanded the existing profiler to run without a separate backend.

timfelle:
@timofeymukha and @njansson Should we close this or do you think some of this approach could be useful in the other PR?

timfelle:
Should we kill this PR @njansson?

timfelle:
Superseded by PR #1467

PR #1163, opened 2024-03-02
Add support for hex objects

njansson:
This is the first step towards using the aabb search tree for interpolation


src/mesh/search_tree/aabb.f90
```diff
@@ -206,6 +210,29 @@ function get_aabb_triangle(triangle) result(aabb)
     call aabb%init(box_min, box_max)
   end function get_aabb_triangle
 
+  !> @brief Get the aabb of a hexahedron.
+  !! @details This function calculates the aabb of a hexahedron. The
+  !! padding is a multiple of the diameter of the aabb, and is used to
+  !! avoid numerical issues when the hexahedron itself it axis
```
timofeymukha, line 216:
```suggestion
  !! avoid numerical issues when the hexahedron itself is axis
```

src/mesh/search_tree/aabb.f90
```diff
@@ -206,6 +210,29 @@ function get_aabb_triangle(triangle) result(aabb)
     call aabb%init(box_min, box_max)
   end function get_aabb_triangle
 
+  !> @brief Get the aabb of a hexahedron.
+  !! @details This function calculates the aabb of a hexahedron. The
+  !! padding is a multiple of the diameter of the aabb, and is used to
+  !! avoid numerical issues when the hexahedron itself is axis
+  !! aligned.
```
timfelle, line 217: Padding was moved to the get_aabb subroutine. The comment lingered it seems. Probably should also remove it from the triangle type.
```suggestion
  !! @details This function calculates the aabb of a hexahedron.
```

src/mesh/search_tree/aabb.f90
```diff
@@ -172,8 +174,10 @@ function get_aabb(object, padding) result(box)
       type is (aabb_t)
        box = object
       type is (tri_t)
-       box = get_aabb_triangle(object)
-
+         box = get_aabb_triangle(object)
+      type is (hex_t)
+         box = get_aabb_hexahedron(object)
+
```
timfelle, line 180: Formatting
```suggestion
       box = get_aabb_triangle(object)
      type is (hex_t)
       box = get_aabb_hexahedron(object)

```

timfelle (changes requested):

timfelle (commented):
Could this be implemented for the generic element_t instead to avoid having to implement it for all variants?

njansson:
> Could this be implemented for the generic element_t instead to avoid having to implement it for all variants?

Not sure, max can't handle array ranges afaik, which could make it difficult for generic element

timfelle (commented):
> > Could this be implemented for the generic element_t instead to avoid having to implement it for all variants?

> Not sure, max can't handle array ranges afaik, which could make it difficult for generic element  

Right, but how would it be different? would it not just be doable in a loop through the entries of element_t%pts like we both did with tri_t and hex_t? Except we both hardcoded the loops, and used the trick that max does accept abitrary numbers of inputs and does the looping for you.

njansson:
> > > Could this be implemented for the generic element_t instead to avoid having to implement it for all variants?
> 
> > Not sure, max can't handle array ranges afaik, which could make it difficult for generic element
> 
> Right, but how would it be different? would it not just be doable in a loop through the entries of element_t%pts like we both did with tri_t and hex_t? Except we both hardcoded the loops, and used the trick that max does accept abitrary numbers of inputs and does the looping for you.

True, we could probably do a loop, over the pts array, and do several max/min checks

timfelle:
> > > > Could this be implemented for the generic element_t instead to avoid having to implement it for all variants?
> > 
> > 
> > > Not sure, max can't handle array ranges afaik, which could make it difficult for generic element
> > 
> > 
> > Right, but how would it be different? would it not just be doable in a loop through the entries of element_t%pts like we both did with tri_t and hex_t? Except we both hardcoded the loops, and used the trick that max does accept abitrary numbers of inputs and does the looping for you.
> 
> True, we could probably do a loop, over the pts array, and do several max/min checks

I am not exactly sure how the overload resolution would function in fortran in that case though. Would tri_t and hex_t trigger the element_t function since they are extending it or how would that work?

njansson:
> > > > > Could this be implemented for the generic element_t instead to avoid having to implement it for all variants?
> > > 
> > > 
> > > > Not sure, max can't handle array ranges afaik, which could make it difficult for generic element
> > > 
> > > 
> > > Right, but how would it be different? would it not just be doable in a loop through the entries of element_t%pts like we both did with tri_t and hex_t? Except we both hardcoded the loops, and used the trick that max does accept abitrary numbers of inputs and does the looping for you.
> > 
> > 
> > True, we could probably do a loop, over the pts array, and do several max/min checks
> 
> I am not exactly sure how the overload resolution would function in fortran in that case though. Would tri_t and hex_t trigger the element_t function since they are extending it or how would that work?

Yes, as long as you access parts define din the base type it works without a select block

timfelle:
So, the point_t is based on strictly double precision while the aabb_t uses the rp type to define the precision. I guess I need to be more considerate when setting up the precision of stuff. Do we intend to only use the rp one to define fields and phycis but in general keeping goemetric objects on full double precision?

timofeymukha:
> So, the point_t is based on strictly double precision while the aabb_t uses the rp type to define the precision. I guess I need to be more considerate when setting up the precision of stuff. Do we intend to only use the rp one to define fields and phycis but in general keeping goemetric objects on full double precision?

Could be wrong, but I think that was the idea.

njansson:
> > So, the point_t is based on strictly double precision while the aabb_t uses the rp type to define the precision. I guess I need to be more considerate when setting up the precision of stuff. Do we intend to only use the rp one to define fields and phycis but in general keeping goemetric objects on full double precision?
> 
> Could be wrong, but I think that was the idea.

The original idea was to make sure that the mesh is always in double, hence point being double. 

For the aabb tree, don't we need the higher precision as well?

timfelle:
> > > So, the point_t is based on strictly double precision while the aabb_t uses the rp type to define the precision. I guess I need to be more considerate when setting up the precision of stuff. Do we intend to only use the rp one to define fields and phycis but in general keeping goemetric objects on full double precision?
> > 
> > 
> > Could be wrong, but I think that was the idea.
> 
> The original idea was to make sure that the mesh is always in double, hence point being double.
> 
> For the aabb tree, don't we need the higher precision as well?

Yes we do. I had just been lazy and set everything to kind=rp, since locally i always run in double precision. I will make sure to think more about it going forward.

timfelle (approved):

src/common/json_utils.f90
```diff
@@ -73,6 +74,23 @@ subroutine json_get_real(json, name, value)
        call neko_error("Parameter "//name//" missing from the case file")
     end if
   end subroutine json_get_real
+  
+  !> Retrieves a double precision real parameter by name or throws an error
+  !! @param json The json to retrieve the parameter from.
+  !! @param name The full path to the parameter.
+  !! @value value The variable to be populated with the retrieved parameter.
```
timofeymukha, line 81: is `@value` something that exists, or did I accidentally introduce it, and now it proliferates by copy-paste :-)?

timfelle: It does not. Ill fix it

PR #1167, opened 2024-03-04
Fix periodic

MartinKarp:
Seems like there was an issue that the facets belonging to a periodic facet could be changed by a neighboring element, but then not added to the table correctly when generating connectivity. 

This PR changes this. For future developments, we should note though that if the entire side is not periodic but a mesh would change from periodic to some other BC along a boundary, this is not necessarily well defined if the element it changes on is not guaranteed to be on the same rank.

### DIFF
--- a/src/mesh/mesh.f90
+++ b/src/mesh/mesh.f90
@@ -1400,11 +1400,6 @@ contains
     ep => this%elements(el)%e
     el_glb_idx = el + this%offset_el
 
-    do i = 1, NEKO_QUAD_NPTS
-       p_local_idx = this%get_local(this%points(p(i)))
-       call this%point_neigh(p_local_idx)%push(el_glb_idx)
-    end do
-
     select type(ep)
     type is (quad_t)
        call ep%init(el_glb_idx, &
--- a/src/mesh/mesh.f90
+++ b/src/mesh/mesh.f90
@@ -380,55 +380,12 @@ contains
 
   subroutine mesh_finalize(this)
     class(mesh_t), target, intent(inout) :: this
-    integer :: i, id, el
+    integer :: i
     class(element_t), pointer :: ep
     type(tuple_i4_t) :: e
     type(tuple4_i4_t) :: f
     integer :: p_local_idx, res
     
-    !If we generate connectivity, we do that here.
-    if (this%lgenc) then
-       do el = 1, this%nelv
-          ep => this%elements(el)%e
-          select type(ep)
-          type is (hex_t)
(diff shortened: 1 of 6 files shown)


MartinKarp:
Victor, please try out this PR for your turbine mesh

MartinKarp:
I dunno what we want as intentional behavior here, but as things stands it is not valid to create a field and dofmap without doing generate_conn at least in the mesh

MartinKarp:
I think the issue of connecting the topological and geometric data really shows now in the tests. A nicer solution might be that if one inits a of map, it checks the mesh and requires that the mesh has created its connectivity data. Right now one gets obscure issues instead as the dofmap assumes information about facets, edges and how they are connected.

timofeymukha:
I've never looked into this stuff, so will be difficult to review. @adampep , could you take a look?

src/mesh/mesh.f90
```diff
@@ -1902,6 +1876,7 @@ function mesh_get_local_facet(this, f) result(local_id)
     integer :: local_id
 
     if (this%htf%get(f, local_id) .gt. 0) then
+       print *, 'error', f
        call neko_error('Invalid global id (local facet)')
```
njansson, line 1880:
```suggestion
       call neko_error('Invalid global id (local facet) ' // f)
```

src/mesh/mesh.f90
```diff
@@ -381,8 +381,11 @@ end subroutine mesh_free
   subroutine mesh_finalize(this)
     class(mesh_t), target, intent(inout) :: this
     integer :: i
-
-
+    class(element_t), pointer :: ep
+    type(tuple_i4_t) :: e
+    type(tuple4_i4_t) :: f
+    integer :: p_local_idx, res
+
```
njansson, line 388: are these some left overs? Doesn't seems to be used

vbaconnet:
I am running it now and it seems to work, I will approve this once I have poked around a bit more :)

src/mesh/mesh.f90
MartinKarp: Will check

src/mesh/mesh.f90
```diff
@@ -558,6 +605,7 @@ subroutine mesh_generate_conn(this)
     !
     ! Find all external (between PEs) boundaries
     !
+    print *, pe_size
```
vbaconnet, line 608:
```suggestion
```

vbaconnet (approved):

njansson (approved):

PR #1171, opened 2024-03-10
Refactor pc_jacobi (CPU) to allow for vectorisation

njansson:


MartinKarp:
Nice with vectorization. However, this function call is not necessary unless G has changed from the previous time step. 

Hence I guess adding a check for that instead would probably be good.

src/krylov/bcknd/cpu/pc_jacobi.f90
```diff
@@ -91,88 +91,1294 @@ subroutine jacobi_solve(this, z, r, n)
     call col3(z,r,this%d,n)
   end subroutine jacobi_solve
 
+  !> Update Jacobi preconditioner
   subroutine jacobi_update(this)
```
MartinKarp, line 95: Does not update anything unless G has changed.

MartinKarp (changes requested):

njansson:
> Nice with vectorization. However, this function call is not necessary unless G has changed from the previous time step.
> 
> Hence I guess adding a check for that instead would probably be good.

True, but should we have this check here, or somewhere else? I guess we can have a flag in `coef` to flag if `G` has been updated, and do a quick return if the flag is `.false.`

src/krylov/bcknd/cpu/pc_jacobi.f90
```diff
@@ -91,88 +91,1294 @@ subroutine jacobi_solve(this, z, r, n)
     call col3(z,r,this%d,n)
   end subroutine jacobi_solve
 
+  !> Update Jacobi preconditioner
```
njansson, line 94:
```suggestion
  !> Update Jacobi preconditioner if the geometry G has changed
```

MartinKarp:
Pehaps, another option would b eto point %update to some do_nothing function and only call this one upon init (if we wanna assume that G does not change).

njansson:
> Pehaps, another option would b eto point %update to some do_nothing function and only call this one upon init (if we wanna assume that G does not change).


If the geometry is changing between stationary and moving, I think the best option would be an if statement instead of swapping function pointers all the time?

timofeymukha (approved):

timofeymukha:
I'll let you guys decide on the logic. The refactor per se is perfectly fine I guess.

njansson:
My suggestion is to merge this, and work out the other points in another pr, I've some ideas how it can be done

MartinKarp (approved):

PR #1173, opened 2024-03-11
Update projection.f90

Shiyu-Sandy-Du:
Fix bugs for projections for fixed dt


njansson (approved):

MartinKarp (approved):
Top! So the issues were primarily the update so we only did every second time step and the reorthogonalization with CPU as well discussed. I also like that we do the same thing now on the CPU and the GPU.

PR #1174, opened 2024-03-12
Projection

Shiyu-Sandy-Du:
1. Projection is accommodated for variable dt:
- When dt changes, clear the projection space anddo not perform projection for the next 5 time steps
- When projection space is empty, do not add the projection info into log file because at that time step, only proj_back is activated but info log recorded at proj_on
2. Move dt_last_change into type time_step_controller


MartinKarp:
Is there any special reason for not restarting directly, but waiting 5 time steps?

pschlatt1:
I guess you want to wait to stabilize the simulation a bit before taking new snapshots. In fact i would perhaps even wait more than 5, but that should be tested (or even done adaptively)

Shiyu-Sandy-Du:
> Is there any special reason for not restarting directly, but waiting 5 time steps?

The reason is mentioned by @pschlatt1. And in my test, the simulation explodes when I restart the projection right after dt changes.

Shiyu-Sandy-Du:
Now this branch is ready to be reviewed and merged

njansson:
@Shiyu-Sandy-Du @MartinKarp , Should we include this in 0.8, or wait for 0.9?

src/common/projection.f90
```diff
@@ -585,14 +593,30 @@ subroutine print_proj_info(this,string)
...
+
+    this%m = 0
+    this%proj_m = 0
+    do i = 1, this%L
+       call rzero(this%xx(1,i),n)
+       call rzero(this%bb(1,i),n)
+    end do
+
+  end subroutine bcknd_clear
+
```
njansson, line 621: Isn't there a missing code path for `NEKO_DEVICE_BCKND` here, calling `device_rzero`?

Shiyu-Sandy-Du:
> @Shiyu-Sandy-Du @MartinKarp , Should we include this in 0.8, or wait for 0.9?

We could include this in 0.8, otherewise there is a bug when applying projection as well as variable dt

njansson:
> > @Shiyu-Sandy-Du @MartinKarp , Should we include this in 0.8, or wait for 0.9?
> 
> We could include this in 0.8, otherewise there is a bug when applying projection together with variable dt

Ok good, then we have to add it

src/case.f90
```diff
@@ -176,7 +176,8 @@ subroutine case_init_common(C)
     !
     ! Time step
     !
-    call json_get(C%params, 'case.timestep', C%dt)
+    call C%params%get('case.variable_timestep', logical_val, found)
+    if (.not. logical_val) call json_get(C%params, 'case.timestep', C%dt)
```
njansson, line 180:
```suggestion
    call C%params%get('case.variable_timestep', logical_val, found)
    if (.not. logical_val) then
        call json_get(C%params, 'case.timestep', C%dt)
   end if
```

njansson: To be consistent

src/fluid/fluid_scheme.f90
```diff
@@ -166,7 +168,7 @@ end subroutine fluid_scheme_free_intrf
 
   !> Abstract interface to compute a time-step
   abstract interface
-     subroutine fluid_scheme_step_intrf(this, t, tstep, dt, ext_bdf)
+     subroutine fluid_scheme_step_intrf(this, t, tstep, dt, ext_bdf, if_variable_dt, dt_last_change)
```
njansson, line 171: Could we make this a bit nicer, e.g. encapsulate these new argument in some module that we can keep inside `fluid_scheme` (or the time step scheme). Rational, I envision that these arguments would be rather fixed for most steps

njansson (commented):
Looks good, some comments on the design.

Maybe @timofeymukha and/or @MartinKarp  has some more comments?

src/fluid/fluid_scheme.f90
Shiyu-Sandy-Du: The most naive idea is to pass type time step controller to there. but it will lead to a cross-dependency issue since module case is included in time step controller module

njansson: True, but why do we need the entire case in the time step controller? As far as I can see, what we need is only the parameter Json blob, and some of the parameters. Or maybe I'm missing something, @timofeymukha ?

timofeymukha: Sorry, what type are you referring to?

doc/pages/case-file.md
```diff
@@ -71,9 +71,14 @@ but also defines several parameters that pertain to the simulation as a whole.
...
 | `restart_file`       | checkpoint to use for a restart from previous data                                                    | Strings ending with `.chkp`                     | -             |
-| `constant_cfl`       | The desired CFL number                                                                                    | Positive real | - |
-| `cfl_max_update_frequency` | The minimum interval between two time-step-updating steps in terms of time steps                     | Integer | `1` |
-| `time_step`          | Time-step size if `constant_cfl` is not specified; maximum time-step size if `constant_cfl` is specified.                                             | Positive reals                                  | -             |
+| `time_step`          | Time-step size | Positive reals                                  | -             |
+| `variable_timestep`  | Whether to use variable dt |       `true` or `false`        | `false` |
+| `max_timestep`          | Maximum time-step size when variable time step is activated | Positive reals   | -             |
```
timofeymukha, line 76: either `time_step` everywhere or `timestep` everywhere.

src/common/time_step_controller.f90
```diff
@@ -59,34 +62,41 @@ module time_step_controller
...
 
-    call C%params%get('case.timestep', this%max_dt, found)
-    call C%params%get('case.constant_cfl', this%set_cfl, this%if_variable_dt)
+    this%dt_last_change = 0
+    call json_get_or_default(C%params, 'case.variable_timestep',&
+                                    this%if_variable_dt, .false.)
+    call json_get_or_default(C%params, 'case.target_cfl',&
+                                    this%set_cfl, 0.4_rp)
+    call json_get_or_default(C%params, 'case.max_timestep',&
+                                    this%max_dt, 99999999.0_rp)
```
timofeymukha, line 72: can use `huge`

src/fluid/fluid_scheme.f90
timofeymukha: OK, I found it now. Yeah, it seems like it only needs the params!

src/fluid/fluid_pnpn.f90
```diff
@@ -555,9 +562,20 @@ subroutine fluid_pnpn_step(this, t, tstep, dt, ext_bdf)
...
+            else if (dt_last_change .gt. this%proj_prs%activ_step - 1) then
+               ! activate projection some steps after dt is changed
+               ! note that dt_last_change start from 0
+               call this%proj_prs%project_on(p_res%x, c_Xh, n)
+               call this%proj_prs%log_info('Pressure')
+            end if
+         else
+            call this%proj_prs%project_on(p_res%x, c_Xh, n)
+            call this%proj_prs%log_info('Pressure')
+         end if
```
timofeymukha, line 578: I guess similar to Niclas comment, could this logic be hidden somewhere? Maybe a `compute` subroutine in  `projection_t`?

Shiyu-Sandy-Du: do you mean the entire if structure starting from if (tstep .gt. ...)? Or we only capsulate the structure of if (if_variable_dt) then ...?

timofeymukha: I guess as much as possible. The whole thing looks like the inner workings of the projection, right? But I haven't looked at how hard it would be to do that in detail :-).

Shiyu-Sandy-Du: It is not hard at all to encapsulate all of them

doc/pages/case-file.md
Shiyu-Sandy-Du: set to timestep (the only error is in the documentation). Now it's fixed

src/common/time_step_controller.f90
Shiyu-Sandy-Du: changed to huge(0.0_rp)

src/fluid/fluid_pnpn.f90
Shiyu-Sandy-Du: wrapped up now

timofeymukha (approved):

njansson (approved):

PR #1178, opened 2024-03-16
Customisation of variable dt feature

Shiyu-Sandy-Du:
In this PR, customisation about variable dt and related documentation is added now.

### DIFF
--- a/doc/pages/case-file.md
+++ b/doc/pages/case-file.md
@@ -78,7 +78,7 @@ but also defines several parameters that pertain to the simulation as a whole.
 | `job_timelimit`      | The maximum wall clock duration of the simulation.                                                    | String formatted as HH:MM:SS                    | No limit      |
 
 ### Variable time stepping
-Variable time stepping is activated by setting `constant_cfl` and `time_step`, which refer to the target CFL number and the maximum time step limit, respectively. The time stepping is adjusted when the average CFL number (which is evaluated by running-averaging) deviates from the target CFL number by `20%` by default. If you want to customise more about the variable time stepping, you can go to `/path/to/neko/src/common/time_step_controller.f90` to customise the parameters.
+Variable time stepping is activated by setting `constant_cfl` and `time_step`, which refer to the target CFL number and the maximum time step limit, respectively. The time stepping is adjusted when the average CFL number (which is evaluated by running-averaging) deviates from the target CFL number by `20%` by default. If you want to customise more about the variable time stepping, you can go to `/path/to/neko/src/common/time_step_controller.f90` to customise the parameters in `subroutine time_step_controller_set_dt`.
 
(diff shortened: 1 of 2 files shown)


doc/pages/case-file.md
```diff
@@ -77,6 +77,9 @@ but also defines several parameters that pertain to the simulation as a whole.
 | `end_time`           | Final time at which the simulation is stopped.                                                        | Positive reals                                  | -             |
 | `job_timelimit`      | The maximum wall clock duration of the simulation.                                                    | String formatted as HH:MM:SS                    | No limit      |
 
+### Variable time stepping
+Variable time stepping is activated by setting `constant_cfl` and `time_step`, which refer to the target CFL number and the maximum time step limit, respectively. The time stepping is adjusted when the average CFL number (which is evaluated by running-averaging) deviates from the target CFL number by `20%` by default. If you want to customise more about the variable time stepping, you can go to `/path/to/neko/src/common/time_step_controller.f90` to customise the parameters in `subroutine time_step_controller_set_dt`.
+
```
njansson, line 82: Shouldn't we make these tuneable via the parameter file? We shouldn't assume that the end user have the source, they might only have a pre built module on their machines

src/common/time_step_controller.f90
```diff
@@ -87,6 +87,9 @@ subroutine time_step_controller_set_dt(this, C, cfl, cfl_avrg, dt_last_change, t
     real(kind=rp) :: dt_old, scaling_factor
     character(len=LOG_SIZE) :: log_buf    
     real(kind=rp) :: alpha = 0.5_rp !coefficient of running average
+    real(kind=rp) :: cfl_deviation = 0.2_rp
+    real(kind=rp) :: scaling_factor_max = 1.2_rp ! upper bound for the scaling factor
+    real(kind=rp) :: scaling_factor_min = 0.8_rp ! lower bound for the scaling factor
```
njansson, line 92: If these aren't adjustable from the outside, we should make them `parameters`. But given the comment above, these should probable be set via the param file, and use "if found else" json parsing

doc/pages/case-file.md
Shiyu-Sandy-Du: Do you mean case file? Sure it could be done there easily.

pschlatt1: In guess through the json file

njansson: Indeed, meant the json file

Shiyu-Sandy-Du: Could be easily done, I will make it soon...

Shiyu-Sandy-Du:
Considering there is some conflicts in this branch with the most recent projection branch, I will make the changes after projection branch is merged with a new PR.

PR #1182, opened 2024-03-19
Updates to the gitignore

timfelle:
We now ignore all files with no extension. Except an explicit list.

Additionally added some temporary file types to the ignored list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,13 @@
+# Ignore everything which does not have an extension or are a directory
+*
+!*.*
+!*/
+
+# Exceptions to the no-extension rule
+!AUTHORS
+!COPYING
+
+# General list of files to ignore
 *.mod
 *.o
 *.dirstamp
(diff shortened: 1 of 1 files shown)


timofeymukha (approved):

njansson (commented):
My feeling is that this is growing and growing, while instead (for those who want to silence things) can simply add a

```shell
[status]
        showuntrackedfiles = no
``` 

to their own gitconfig

timfelle:
> My feeling is that this is growing and growing, while instead (for those who want to silence things) can simply add a
> 
> ```shell
> [status]
>         showuntrackedfiles = no
> ```
> 
> to their own gitconfig

Well it was more that we see pull requests containing files that should not be there.

njansson:
> > My feeling is that this is growing and growing, while instead (for those who want to silence things) can simply add a
> 
> > 
> 
> > ```shell
> 
> > [status]
> 
> >         showuntrackedfiles = no
> 
> > ```
> 
> > 
> 
> > to their own gitconfig
> 
> 
> 
> Well it was more that we see pull requests containing files that should not be there.

Fair enough

njansson (approved):

PR #1183, opened 2024-03-19
Average fields in two directions

MartinKarp:
Adds support for computing average in 2 spatial directions directly if there is more than one homogenous direction (for example channel flow).

We also update map_1d so it does not break if compiled with GPU backend, it now gives a warning instead. Still, the mapping is only available on the CPU currently.

### DIFF
--- a/contrib/average_field_in_space/average_field_in_space.f90
+++ b/contrib/average_field_in_space/average_field_in_space.f90
@@ -2,7 +2,6 @@
 !! Martin Karp 17/02-23
 program average_field_in_space
   use neko
-  use mean_flow
   implicit none
   
   character(len=NEKO_FNAME_LEN) :: inputchar, mesh_fname, field_fname, hom_dir, output_fname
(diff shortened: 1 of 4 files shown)


contrib/average_field_in_space/average_field_in_space.f90
```diff
@@ -18,15 +17,27 @@ program average_field_in_space
...
-        write(*,*) 'Usage: ./average_field_in_space mesh.nmsh field.fld dir(x, y, z)  outfield.fld' 
-        write(*,*) 'Example command: ./average_field_in_space mesh.nmsh fieldblabla.fld x  outfield.fld'
-        write(*,*) 'Computes the spatial average in x of the field fieldblabla.nek5000 and stores in outfield.fld'
+        write(*,*) 'Usage: ./average_field_in_space mesh.nmsh field.fld dir(x, y, z, xz, xy, yz)  outfield.(fld,csv)' 
+        write(*,*) '----'
+        write(*,*) 'Example command for avg in 1 direction: ./average_field_in_space mesh.nmsh fieldblabla.fld x  outfield.fld'
+        write(*,*) 'Computes spatial average in 1 direction and saves it in outfield.fld'
+        write(*,*) '----'
+        write(*,*) 'Example command: ./average_field_in_space mesh.nmsh fieldblabla.fld xy  out.csv'
+        write(*,*) 'Computes the spatial average in 2 directions diretly of the field fieldblabla.nek5000 and stores in out.csv'
```
timofeymukha, line 36:
```suggestion
        write(*,*) 'Computes the spatial average in 2 directions directly of the field fieldblabla.nek5000 and stores in out.csv'
```

contrib/average_field_in_space/average_field_in_space.f90
```diff
@@ -18,15 +17,27 @@ program average_field_in_space
...
+        write(*,*) '----'
+        write(*,*) 'In out.csv the firsct col are the coords of the gll points'
```
timofeymukha, line 38:
```suggestion
        write(*,*) 'In out.csv the first col are the coords of the GLL points'
```

timofeymukha:
Very nice. @MartinKarp , I assume you have a channel case where you test this or something. How about adding a run.sh script to the turb_channel example, where one puts in all the commands necessary to do that case from start to finish, e.g. starting with a genmeshbox command and up to the averaging of statistical fields with this new utility. I think that would be very useful to just have that sequence of commands recorded and then one can adapt to one's needs.

MartinKarp:
I added some comments in the README about the postprocess, the exact usage I feel will vary depending on what time people start to gather stats and so on.

timofeymukha (approved):

njansson (approved):

PR #1185, opened 2024-03-19
User-created simcomps

timofeymukha:
This is a proof of concept of how user simcomps can be incorporated into the `neko_simcomps` object so that they are executed just like the ordinary ones. As a reminder, the issue is that the simcomp factory cannot know about the new simcomp types, so the init has to happen in the user file. Here I use the user initialization routine to that end, so comments are provided there.
To make this work, I had to make the simcomp array inside the `simcomp_exectutor_t` to be larger than the number of simcomps in the JSON file. A better solution is to add a proper `add` routine.

I modify the tgv example to demonstrate things, `test.f90` is just a copy of the vorticity simcomp with a new type name. The JSON for the user simcomp is added inside a "user_simcomps" array, but of course any configuration can be used, since it is the user code that parses the JSON.

I think this is in principle a pretty ok solution, which could facilitate users to encapsulate things in simcomps. I will put some more thought in how to make this as streamlined as possible.


timofeymukha:
A limitation right now is that the user simcomps do not respect the "order" keyword.

njansson (commented):
Nice! But I would suggest to create a new example for this instead of diluting tgv further ;)

timofeymukha:
> Nice! But I would suggest to create a new example for this instead of diluting tgv further ;)

Yeah, this is just for demonstration.

adperezm:
Nice!

I like that this would provide a nice framework to create "user modules". Since the controllers would also be initialized and all that.

I have a question:  How does one compile the user_simcomp? does makeneko pick it up?

I still do not see a lot of advantages over a user just creating a module and using it in their user file. I guess in that sense I am an opponent of "everything as a simcomp" vision that a lot of you have.

But I also trust the process. So far all your suggestions have made the code better :)

timofeymukha:
> Nice!
> 
> I like that this would provide a nice framework to create "user modules". Since the controllers would also be initialized and all that.
> 
> I have a question: How does one compile the user_simcomp? does makeneko pick it up?
> 
> I still do not see a lot of advantages over a user just creating a module and using it in their user file. I guess in that sense I am an opponent of "everything as a simcomp" vision that a lot of you have.
> 
> But I also trust the process. So far all your suggestions have made the code better :)

There is not really a contradiction I think, because it's exactly as you say, you create a new module and compile it with makeneko alongside your user file. So here it will be `makeneko tgv.f90 test.f90`. And then I use the simcomp type inside test.f90 in the user file tgv.f90. So, if what you are adding fits into the simcomp concept, you can implement a simcomp in your module and plug it into the simulation pipleline. I think an argument for that contra putting things in user_check etc is that the simcomp encapsulates everything, so if you put your simcomp online somewhere, it is pretty clear how I add it to my case. Another discussion was that this would make it easier to integrate code developed by users, since things are already put into the correct framework from the beginning, so to speak.

MartinKarp:
Another aspect I guess is controlling when the user functionality occurs. user_check is always at the end fo the time step, which might not always be what one wants (foe example if one is changing the field in some way).

src/simulation_components/simcomp_executor.f90
```diff
@@ -42,6 +42,8 @@ module simcomp_executor
   implicit none
   private
 
+  integer :: NEKO_MAX_USER_SIMCOMPS = 10
```
timfelle, line 45: Is this something we could move over to the neko_config and have it be possible
to configure on compilation like the working precision etc. It seem arbitrary to
have it fixed, but i know you want to know the value at compiletime to avoid
deepcopying the simulation components.

timfelle:
@timofeymukha and @njansson I have done a number of updates so it should now be more useful. 

Instead of creating an example case as we have done so far, I decided to utilize the doxygen feature to create examples. Since what is needed here is more a code snippet of how to do it, and not an actual case.

doc/examples/simulation_components/user_simcomp.f90
```diff
@@ -0,0 +1,124 @@
...
+!! COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+!! INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+!! BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+!! LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+!! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+!! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+!! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+!! POSSIBILITY OF SUCH DAMAGE.
+
+! Implements the `user_simcomp_t` type.
```
njansson, line 38: Don't we need the `!>` here for the doctoring?

src/simulation_components/probes.F90
```diff
@@ -82,7 +82,7 @@ module probes
      integer, allocatable :: n_local_probes_tot_offset(:)
      integer, allocatable :: n_local_probes_tot(:)
      !>  For output on rank 0
-     logical :: seq_io
+     logical :: seq_io = .true.
```
njansson, line 85: Is the implicit save for `seq_io` intended here?

src/simulation_components/probes.F90
```diff
@@ -640,6 +634,8 @@ subroutine probes_evaluate_and_write(this, t, tstep)
     integer, intent(in) :: tstep
     integer :: i, ierr
 
+    write(*,*) 'Probes evaluate and write'
```
njansson, line 637: can't we pipe this via the log system?

njansson (commented):
Looks good, some minor questions, and unsure if we want to have the copyright statement with doxygen tags (strongly leaning towards a no there)

doc/examples/simulation_components/user_simcomp.f90
timfelle: Well examples in doxygen are a bit different. They do not use the docstring stuff, it actually gets removed. Try to run the make html and look in the examples section. Then you can see how its handled by doxygen.

src/simulation_components/probes.F90
timfelle: No I missed it when I cleaned up the code. I was not supposed to modify that file, but it for some reason didnt want to run. without.

src/simulation_components/probes.F90
timfelle: Again only there for debugging when i needed to test the order of execution.

timfelle:
> Looks good, some minor questions, and unsure if we want to have the copyright statement with doxygen tags (strongly leaning towards a no there)

Well it's more a suggestion. It will format it directly in our documentation in a more useful way rather than it only being in the source code. Try to run `make html` and look at the `simcomp_executor.f90` file reference: `/neko/doc/html/d8/d65/simcomp__executor_8f90.html`

timfelle:
@njansson & @timofeymukha I reverted the doxygen stuff. Are we happy to merge? I guess Tim cannot officially review since its his PR and i feel i shouldn't due to the changes i made. But i can do an accept if Tim agrees.

njansson (approved):

timfelle:
Sorry @njansson I did a bunch of changes to the executor that need a look before merge.

njansson (approved):

timofeymukha:
I ran out of time :(. But on Sunday!

src/common/user_intf.f90
```diff
@@ -426,4 +444,56 @@ subroutine dummy_user_material_properties(t, tstep, rho, mu, cp, lambda,&
...
+
+  end function simulation_component_user_settings
+
+
+  !> @example simulation_components/user_simcomp.f90
+  !! @brief User defined simulation components.
+  !! @details
+  !! Example of how to use the simcomp_executor to add a user defined
+  !! simulation component to the list.
+  !! @include simulation_components/user_simcomp.case
```
timofeymukha, line 498: Cool, so this is a way to link to an "example" from anywhere in the code?

src/simulation_components/simcomp_executor.f90
```diff
@@ -74,81 +79,251 @@ module simcomp_executor
...
+    integer :: i, position
+
+    ! Find the first empty position
+    position = 0
+    do i = 1, this%n_simcomps
+       if (.not. allocated(this%simcomps(i)%simcomp)) then
+          position = i
+          exit
+       end if
+    end do
```
timofeymukha, line 222: Do I understand correctly that the idea is that the factory will leave an unallocated simcomp in the executors array when a user simcomp is defined, and then at this point it is filled in?

timofeymukha (commented):
Very nice, Tim. I would have spent a lot of time on that ordering thing :D.

src/common/user_intf.f90
timfelle: Well not quite anywhere, but yes. You set an "example_path" in the doxygen configuration file which is then used to look up the file you are linking to.
You can even add multiple files in the example, I think you need to do `!! @include file2.whatever` in the description.

src/simulation_components/simcomp_executor.f90
timfelle: Yes that is the idea.
Actually i think it would be better to leave all components uninitialized, and just store an array of the json settings. Because, even though you might specify the user component to be called first, it will be initialized after the builtin ones. And they are also initialized in the order they are added in code, not by the order specified in the json if you have multiples.

src/common/user_intf.f90
```diff
@@ -426,4 +444,56 @@ subroutine dummy_user_material_properties(t, tstep, rho, mu, cp, lambda,&
...
+    integer :: i
+    logical :: found, is_user
+
+    call params%get_core(core)
+    call params%get(simcomp_object)
+    call params%info('', n_children=n_simcomps)
+
+    ! We need a separate loop to figure out the order, so that we can
+    ! apply the order to the initialization as well.
+    found = .false.
```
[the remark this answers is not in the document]
timfelle: That is true indeed.

```suggestion
    found = .false.
```

doc/examples/simulation_components/user_simcomp.f90
```diff
@@ -0,0 +1,157 @@
...
+contains
+
+  ! Constructor from json.
+  subroutine simcomp_test_init_from_json(this, json, case)
+    class(user_simcomp_t), intent(inout) :: this
+    type(json_file), intent(inout) :: json
+    class(case_t), intent(inout), target :: case
+
+    call this%init_from_attributes()
+    call this%init_base(json, case)
```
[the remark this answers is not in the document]
timfelle: Yes that would probably be better, but that would require either changing the interface of the attribute initializer, or to save the json settings. Since this is how all the current simcomps are doing it.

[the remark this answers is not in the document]
timofeymukha: Oh, right, the base one actually needs the JSON. But then probably just switch the order of 66 and 67, or? Not that it matters much for this example of course.

src/common/user_intf.f90
timofeymukha: I look forward to see how it look!

src/simulation_components/simcomp_executor.f90
timofeymukha: Not sure I completely understood, I guess at some point you must initialize. Do you mean to do it in the finalize routine for all of them? But anyway, I think this is an ok solution for what we have to work with in Fortran. I dislike a bit the idea that a factory can give you back an unallocated pointer, but we do check that everything is allocated in `finalize` so at least we know things are good after the constructor is run.

timofeymukha:
@timfelle feel free to "approve" and merge!

doc/examples/simulation_components/user_simcomp.f90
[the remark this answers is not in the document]
timfelle: One question is, why is it split into 3 at all?
I get the base_init, but why do we create an `init_from_attributes` at all for this example? It's not part of the interface to do so.

timfelle (approved):

PR #1186, opened 2024-03-19
User statistics simulation component - Draft

adperezm:
Hi,

I wanted to have a very drafty version of user statistics simcomp. I do this PR Just to collect your ideas.

I think that something like this would provide a first framework to easily add user statistics.

I have not made sure that anything compiles, hence this is very drafty. But technically, we should put in the json file the name of the field that should be averaged and then the running average should be calculated based on the field registry.

The reason I make this now, is because now that @timofeymukha added the draft for user_simcomp, perhaps this is a natural progression. A user simcomp can calculate the custom fields and add them to the field registry, and the with this simcomp one averages them. 

Since this implementation can easily be considered redundant, I did not make sure that it runs (in fact it does not, since it is not added to the build system yet.)

So when you have time, please check. And if you think that something like this could be good to have, then I develop it further.


timofeymukha:
yeah, something along these lines would be great. What would be totally awesome is to use Martin's 1D mapping stuff to add the option of averaging across one or two homogeneous directions on the fly. But that could be a later PR.

adperezm:
> yeah, something along these lines would be great. What would be totally awesome is to use Martin's 1D mapping stuff to add the option of averaging across one or two homogeneous directions on the fly. But that could be a later PR.

Okay! I can look into that.

MartinKarp:
I think adding something like this should be a natural next step after having the prebaked statistics. My thinking is perhaps to add an argument to the prebaked ones to get the total (all things needed for budgets and so on) and perhaps only the reynolds stresses. Any more advanced statistics would use this kind of simcomp, I guess we mostly should also add a simcomp to generate a field that say is the multiplication of two current fields or similar. This is of course a next topic.

timofeymukha:
I think the only thing really needed apart from the fluid "package" is a simcomp the computes the mean of a list of fields provided by the user by name. (Co)variances and more advanced stuff are anyway obtained a posteriori in our methodology by using the mean of products. Initially, the fields are to be computed inside the user check / user simcomp. However, as Martin says it may be a good idea to add some simple simcomps for simple field arithmetics.

MartinKarp:
Updated a fair bit, it seems to compile and run.

adperezm:
Good stuff! thanks. I have not been at finishing a bunch of open pull requests.

timofeymukha:
Closing, since we now merged in Martin's version of this.

PR #1192, opened 2024-03-21
Add field_writer simcomp

timofeymukha:
This PR adds a new simcomp called `field_writer`, which basically dumps a given list of fields from the registry to an .fld file. Either a separate fld, or adding new fields to the fluid output. Intended use is custom fields from the user file. Quite useful for debugging as well. Also, I think I will try to have it as a component in other simcomps that output stuff, like vorticity and lambda2, to avoid code duplication.

Additionally, gives field_registry a face lift and adds the possibility to just do nothing when attempting to add a field, which already exists. Very often we have a construct which first checks that a field does not exists, and then adds it. This way, we can do it in 1 line.


vbaconnet:
Amazing! Looking forward to using this asap :)

njansson (approved):
Very nice! 🍻

timofeymukha:
Nice bonk from gfortran-10. The issue seems to be the intent in the field list appending.
```fortran
  !> Append a field to the list.
  !! @param field The field to append.
  subroutine field_list_append(this, field)
    class(field_list_t), intent(inout) :: this
    class(field_t), intent(inout), target :: field
    type(field_ptr_t), allocatable :: tmp(:)
    integer :: len

    len = size(this%fields)

    allocate(tmp(len+1))
    tmp(1:len) = this%fields
    call move_alloc(tmp, this%fields)
    this%fields(len+1)%f => field

```

Can we change this to `in` for the `field` argument?

njansson:
> Nice bonk from gfortran-10. The issue seems to be the intent in the field list appending.
> 
> ```fortran
>   !> Append a field to the list.
>   !! @param field The field to append.
>   subroutine field_list_append(this, field)
>     class(field_list_t), intent(inout) :: this
>     class(field_t), intent(inout), target :: field
>     type(field_ptr_t), allocatable :: tmp(:)
>     integer :: len
> 
>     len = size(this%fields)
> 
>     allocate(tmp(len+1))
>     tmp(1:len) = this%fields
>     call move_alloc(tmp, this%fields)
>     this%fields(len+1)%f => field
> 
>   end subroutine field_list_append
> ```
> 
> Can we change this to `in` for the `field` argument?

Should be safe I would say

timfelle (dismissed):

doc/pages/simcomps.md
```diff
@@ -16,25 +16,32 @@ of objects called `simulation_components`, which resides directly under the
...
+ provided, they are set to be the same as for `compute_`.
+
+For simcomps that compute 3D fields, the output can be either added to the main
+`.fld` file, containing velocity and pressure, or saved to a separate file. For
+the latter, the `output_filename` keyword should be provided. One can
+additionally provide the `precision` keyword, which can be set to either
+`single` or `double` to control the precision of the written data.
+
+For example, in the `tgv` example case the `vorticity` component is executed
+once per 50 time steps. The `output_` paramaters are synced to that, and the
```
timfelle, line 36:
```suggestion
once per 50 time steps. The `output_` parameters are synced to that, and the
```

doc/pages/simcomps.md
```diff
@@ -47,7 +54,7 @@ If no parameters for the `output_` parameters are provided, they are set to be t
  Computes \f$ \lambda_2 \f$ for the velocity field and stores it in the normal output files as the first unused field.
  This means that \f$ \lambda_2 \f$ can be found in the temeprature field in then fld files if running without a scalar
```
timfelle, line 55:
```suggestion
 This means that \f$ \lambda_2 \f$ can be found in the temperature field in then fld files if running without a scalar
```
Might as well update this now we are working on this file anyway

src/simulation_components/field_writer.f90
```diff
@@ -0,0 +1,161 @@
...
+
+contains
+
+  !> Constructor from json.
+  !> @param json JSON object with the parameters.
+  !! @param case The case object.
+  subroutine field_writer_init_from_json(this, json, case)
+    class(field_writer_t), intent(inout) :: this
+    type(json_file), intent(inout) :: json
+    class(case_t), intent(inout), target :: case
```
timfelle, line 74:
```suggestion
    class(case_t), intent(inout), target :: neko_case
```
Might be good to avoid masking variables

timfelle (changes requested):

src/simulation_components/field_writer.f90
timofeymukha: Good opportunity to ask, coz I kind of ignored these messages. What is  masked" here, `this%case`? But it is still accessable I guess?

njansson: Isn't it due to our module named `case` ?

timfelle: Nope, the entire module 'case'

src/field/field_list.f90
```diff
@@ -18,7 +18,7 @@ module field_list
   !! @param field The field to append.
   subroutine field_list_append(this, field)
     class(field_list_t), intent(inout) :: this
-    class(field_t), intent(inout), target :: field
+    class(field_t), intent(in), target :: field
```
timfelle, line 21: We also have a masked variable here.

src/simulation_components/field_writer.f90
timofeymukha: OK, so the name should be the same as in the abstract interface, since this is an implementation of a deferred routine. So changing to something different from `case` would require ploughing through all simcomps. Can we maybe forgive this one for now?

timfelle (approved):

PR #1195, opened 2024-03-22
F2py coupling

adperezm:
Hi.

With this PR I refactored what I had written before for data streaming.

I now made it so that we do not always need to stream 4 fields at a time. This provides much more flexibility on which data we move around. Unfortunately I think that it might be slower. 

I removed a bunch of annoying extra folders we had to use. This limits the runtime configuration of the streaming via ADIOS.

I removed a bunch of code that I was going to use for compression. I think that the cleaner way is that I create a compressor type instead of reusing this one.

While I called this F2py coupling. We can really couple with any other code that would have an ADIOS2 reader/writer. Potentially we could couple 2 nekos if that is of interest.

We can now write and receive data in the stream. So this is a 2 way coupling.

In the coming days I will make my other code available to all, so you can see what type of thing this could be useful for.


njansson (approved):

timofeymukha:
I guess not being locked to 4 fields is very good!

timofeymukha (approved):

PR #1198, opened 2024-03-22
debug on projection

Shiyu-Sandy-Du:
debug projection to make simulation work without projection and remove some dummy stuffs

### DIFF
--- a/src/fluid/fluid_pnpn.f90
+++ b/src/fluid/fluid_pnpn.f90
@@ -483,9 +483,6 @@ contains
     integer :: temp_indices(3)
     ! Counter
     integer :: i
-    ! time step controller's component
-    logical :: if_variable_dt
-    integer :: dt_last_change ! 0 at the step where dt changes
 
     if (this%freeze) return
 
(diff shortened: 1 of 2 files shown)


timofeymukha:
A little bit sad to have the ifs back in the fluid step. Would it be a good trick to init the projections with activ_step=huge(0) when the space size is 0, or something like that, so that they just effectively never run?

Shiyu-Sandy-Du:
> A little bit sad to have the ifs back in the fluid step. Would it be a good trick to init the projections with activ_step=huge(0) when the space size is 0, or something like that, so that they just effectively never run?

I have no idea, it sounds work, but those ifs are quite simple, right?

njansson:
> > A little bit sad to have the ifs back in the fluid step. Would it be a good trick to init the projections with activ_step=huge(0) when the space size is 0, or something like that, so that they just effectively never run?
> 
> 
> 
> I have no idea, it sounds work, but those ifs are quite simple, right?

I tend to agree with @timofeymukha comments, nice to keep the step loop as clean as possible and hide the logic somewhere else

Shiyu-Sandy-Du:
Ohhh actually we don't need to do anything, but remove the "if" in the constructor of objects. In pre and post-solving projection subroutine, all processing start only if this%L .gt. 0

njansson (approved):

njansson:
We need to fix this to address currently failed CI runs

PR #1204, opened 2024-03-26
Explicitly list all types in factories for friendly error messages

timofeymukha:
This is a suggestion to enhance all our json-based factories with a KNOWN_TYPES list with the names of all the types the factory can create. We can then output this list when an unknown type is provided by the user in the case file.

This has proven to be a nice way of exploring what is available in the code in e.g. OpenFOAM. One intentionally puts a jibberish string in the "type" just to see the options. The minus is a bit of extra burden on the developers, but I think it is tolerable.

So far "don't merge" as I will extend it to all factories upon approval.

Question: is the string concatenation routine ok? Will it just reallocate the result to the necessary size as we are concating in new items to it?


src/simulation_components/simulation_component_factory.f90
```diff
@@ -42,12 +42,21 @@ module simulation_component_fctry
...
   public :: simulation_component_factory
 
+  ! List of all possible types created by the factory routine
+  character(len=20) :: KNOWN_TYPES(5) = [character(len=20) :: &
+     "vorticity", &
+     "lambda2", &
+     "probes", &
+     "les_model", &
+     "field_writer"]
+
```
njansson, line 59: Is the extra `[character...` necessary?

In other cases it works without, for example:  https://gitlab.com/ExtremeFLOW/neko_opr/-/blob/master/driver.f90?ref_type=heads#L19

```suggestion
  character(len=20) :: KNOWN_TYPES(5) = [&
     "vorticity", &
     "lambda2", &
     "probes", &
     "les_model", &
     "field_writer"]

```

timofeymukha: Interesting, in my case it would crash saying the strings are of different lengths. Can it be that the example you link to works because you pad all the stings to the same length with spaces?

timofeymukha:
Besides for the `character` thing, are we in agreement for extending this to other types?

timfelle (commented):
> Besides for the `character` thing, are we in agreement for extending this to other types? 

It quite a good idea to have this. One think I thougfht of was, what is the maximal string length and could that be an issue?

timofeymukha:
@njansson @timfelle @MartinKarp The PR is now open for review. In addtion to adding the KNOWN_TYPES, in indroduced the following name convention:
- The name of the variable for the object being allocated is `object`
- The name of the variable for the name of the type, (extracted from the json or passed to the factory) is `type_name`

njansson:
There's several changes in the BSD license text, probably from some refactoring, which needs to be reverted

njansson (changes requested):
BSD license text shall not be altered

timofeymukha:
> BSD license text shall not be altered

I knew using %s/ will bite me in the butt :-)

src/krylov/krylov_fctry.f90
```diff
@@ -46,213 +46,236 @@ module krylov_fctry
...
-          call kp%free()
+          call obj%free()
        type is(gmres_device_t)
-          call kp%free()
+          call obj%free()
        type is(bicgstab_t)
-          call kp%free()
+          call obj%free()
        end select
     end if
```
timfelle, line 281: Why are we not just calling `obejct%free()`

src/krylov/krylov_fctry.f90
```diff
@@ -46,213 +46,236 @@ module krylov_fctry
...
-          call kp%init(n, max_iter)
+          call obj%init(n, max_iter)
        type is(gmres_device_t)
-          call kp%init(n, max_iter)
+          call obj%init(n, max_iter)
        type is(bicgstab_t)
-          call kp%init(n, max_iter)
+          call obj%init(n, max_iter)
        end select
     end if
```
timfelle, line 245: Is there any good reason we do this massive init case stuff. Can't we just do
init, without selecting the type? The type is detected through the allocation
stuff above.

src/krylov/krylov_fctry.f90
timofeymukha: @njansson ?

src/krylov/precon_fctry.f90
```diff
@@ -41,37 +41,52 @@ module precon_fctry
...
   !> Create a preconditioner
-  subroutine precon_factory(pc, pctype)
+  subroutine precon_factory(pc, type_name)
     class(pc_t), target, allocatable, intent(inout) :: pc
-    character(len=*), intent(in) :: pctype
+    character(len=*), intent(in) :: type_name
+    character(len=:), allocatable :: type_string
+
+    type_string =  concat_string_array(KNOWN_TYPES, NEW_LINE('A') // "-  ", &
+                                       .true.)
```
timfelle, line 64: Maybe move this to the else statement, since we dont use it before then

timofeymukha: True

src/krylov/krylov_fctry.f90
timfelle: We do it in stuff like the simcomps

src/les/les_model_fctry.f90
```diff
@@ -14,7 +14,7 @@
 !     disclaimer in the documentation and/or other materials provided
 !     with the distribution.
 !
-!   * Neither the name of the authors nor the names of its
+!   * Neither the type_name of the authors nor the names of its
```
timfelle, line 17:
```suggestion
!   * Neither the name of the authors nor the names of its
```

src/krylov/krylov_fctry.f90
njansson: Should work, but we had issues with compilers in the past

src/krylov/krylov_fctry.f90
njansson: Same here, should work, but we had issues with compilers in the past.

src/krylov/krylov_fctry.f90
timfelle: Interesting, which compilers did you try out, and did you find out what went wrong? From what i understand it should be functional in terms of the Fortran language. So is it just missing features in some of the compilers?

src/krylov/krylov_fctry.f90
njansson: Fully agree, but given the experience, not too keen on introducing these PITA compiler issues at this time

timofeymukha:
@njansson NB: ifx just got fixed by commenting out the `private` statements I added to the precon and ksp factory modules. I think we should try to clean up the `use` statements with respect to these types, there seems to be a lot of implicit propagation going on via just `using` the factory module.

src/krylov/krylov_fctry.f90
timofeymukha: Let us at this point add a comment for why it is there, and that we should at some point see if we can get away with removing this stuff.

njansson:
> @njansson NB: ifx just got fixed by commenting out the `private` statements I added to the precon and ksp factory modules. I think we should try to clean up the `use` statements with respect to these types, there seems to be a lot of implicit propagation going on via just `using` the factory module.

Great, but what a mess (sigh). I through that we had cleaned these in the past.

src/krylov/krylov_fctry.f90
timfelle: Sure, I fully support attempting that kind of change in a separate PR with rigorous testing if we already know there might be issues. Its not relevant for the scope of this PR anyway. 

Maybe we should add an issue to track this?

src/krylov/krylov_fctry.f90
timfelle: > I think more like bugs than missing features :-)

Oh the joy 😹

timofeymukha:
> > @njansson NB: ifx just got fixed by commenting out the `private` statements I added to the precon and ksp factory modules. I think we should try to clean up the `use` statements with respect to these types, there seems to be a lot of implicit propagation going on via just `using` the factory module.
> 
> Great, but what a mess (sigh). I through that we had cleaned these in the past.

We did to some extent, but not thoroughly enough perhaps :-). Although, for individual ksp types the `use` seems to be pretty damn clean, almost everything behind `only.`

src/krylov/krylov_fctry.f90
timofeymukha: I mean the death of ifx here is a good example :-). Myabe something on our side, but then why don't all the other compilers complain. And anyway, ifx just segfaulted instead of complaining about the code ^^

src/krylov/krylov_fctry.f90
timofeymukha: Done

timfelle (approved):

src/.depends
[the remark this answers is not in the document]
timfelle: I updated it

njansson (approved):

PR #1205, opened 2024-03-26
Add probe generators for simple geometries

### LINKED ISSUE #967: Simcomps for sampling based on geometric entities
With probes coming into shape, the next step could be to allow some convenient sampling simcomps that pre-arrange the probes on some common geometric entities.
* A line connecting two points in space divided by a selected number of points.
* A plane, defined in one way or another. 
* An arbitrary surface with point locations read from, e.g., an STL file. 
* More?

The same simcomps should also allow sampling directly from gll points, without interpolation, also based on appropriately defined entities as discussed in #965.

We can obviously have the I/O behind an interface so that we can dump to different formats. We have CSV, but we should think of alternatives, preferably without reinventing the wheel. HDF5 comes to mind, but I'm not sure if it's good for the situation when you have to output the data bit by bit. Maybe there is something cool in the ML universe for this?

timfelle:
This pull request adds the functionality to generate probe points on simple geometric objects. The objects are assigned as a "point" list in the JSON specification for the simcomp.

- [x] Reading from a file.
- [x] Directly listing points.
- [x] Linearly distribute across a line
- [x] Distribute on a circle.
- [ ] Distribute points on a plane.
- [ ] Distribute across an STL file.
- [x] Probe all points defined by a point zone.

The old specification of a points_file is preserved, but the new specification can also read from csv files.

The code has also been refactored to move the adding of new points to a separate function and remove unused variables.

Plane and STL files have been postponed

Closes #967


timfelle (commented):
Do anyone have additional ideas of objects we should add to the list here?
Additionally I hope i didn't make too much a mess of the code, but i tried to clean it up a bit as well.

timofeymukha:
Great stuff. I think STL would indeed be nice to support, but not sure if we need it explcitly here of it is enough to be able to create an STL-based point zone, and then define the probes based on that. 

I would not call the input via file "legacy". We had people introduce hundreds of thousands of probes for some cases, so one would not have that in the case file I think. Then again, it should probably not be a .csv either, but still a file.

Regarding geometries, I don't think we need much more, for special stuff one should just write a Python script to dump whatever one needs to a file. I think even circle is a bit borderline, but I guess for pipes it is nice, so we can for sure keep it.

How were you thinking about the plane? Particularly, its bounds.

timfelle:
> Great stuff. I think STL would indeed be nice to support, but not sure if we need it explcitly here of it is enough to be able to create an STL-based point zone, and then define the probes based on that.

Hmm well the STL can be done in a couple of different way. You could use the vertices of the mesh, centers of faces, or the entire space that is enclosed by it. In general I think any 1D or 2D manifold whould be candidates for this way of doing things. For 3D objects I would definitely say that point zones are the way forward. However, currently point zones are only defined based on the GLL points. We could probably extend that with some other sampling algorithms.

> I would not call the input via file "legacy". We had people introduce hundreds of thousands of probes for some cases, so one would not have that in the case file I think. Then again, it should probably not be a .csv either, but still a file.

Oh, I just meant that specifying the file it through the `points_file` rather than the `points` array with a `type: "file"`. It seemed silly to only allow a single input type for the probes, so I figured I'd move the file specification to the same interface.

> Regarding geometries, I don't think we need much more, for special stuff one should just write a Python script to dump whatever one needs to a file. I think even circle is a bit borderline, but I guess for pipes it is nice, so we can for sure keep it.

Sure, that is definitely also what should be done if you need a complex system. I just personally needed a circle for verifying the vortex shedding for our brinkman terms so i might as well add it while i was working on the other stuff instead of making a new file in a different language ;P

> How were you thinking about the plane? Particularly, its bounds.

I were thinking intersecting the plane with the bounding box and then sample that way. I should be able to determine the 4 parametric corners corners of the plane and then in the parametric space distribute points wither with some sampling algorithm such as Poisson Disk Sampling or just in a linear grid in parametric space.

timofeymukha:
Right, the point zone is just a collection of GLLs. Then I guess it makes sense to treat the STL directly with probes. I think just sampling at each node in the triangulation is probably good enough for most cases.

timfelle:
> Right, the point zone is just a collection of GLLs. Then I guess it makes sense to treat the STL directly with probes. I think just sampling at each node in the triangulation is probably good enough for most cases.

Yea, that would atleast be the starting point i would go for. If we want some random sapling at some point it would be possible to choose a random face and sample a point randomly on that through barycentric interpolation. But i dont think that will be relevant any time soon. Atleast for the probing, maybe for other purposes?

src/simulation_components/probes.F90
```diff
@@ -107,46 +117,301 @@ subroutine probes_init_from_json(this, json, case)
...
+
+    ! Setup the required arrays and initialize variables.
+    this%n_local_probes = 0
+    this%n_global_probes = 0
+
+    ! Read the legacy point specification from the points file.
+    if (json%valid_path('points_file')) then
+
+       ! Todo: We should add a deprecation notice here
+       call json_get(json, 'points_file', input_file)
```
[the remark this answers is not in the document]
timfelle: Well the point of this was that we do not need the `point_file` setting any longer. Since you would now specify it as a file entry in the `points` array. An example using most of the types would look like bellow.

The reason `points_file` still exsist is to have backward compatibility. The idea would be to remove that entirely later on when people have had time to change to the new syntax.

```json
"simulation_components": [
    {
        "type": "probes",
        "fields": [ "u", "v" ],
        "output_file": "out_probes.csv",
        "points": [
            {
                "type": "file",
                "file_name": "probes.csv"
            },
            {
                "type": "point",
                "coordinates": [
                    0.1, 0.1, 0.0,
                    0.5, 0.0, 0.0,
                    1.0, 0.0, 0.0,
                    0.0, 0.0, 0.0
                ],
            },
            {
                "type": "line",
                "start": [ 0.0, 0.0, 0.0 ],
                "end": [ 1.0, 0.0, 0.0 ],
                "amount": 3
            },
            {
                "type": "line",
                "start": [ 0.0, 0.0, 0.0 ],
                "end": [ 1.0, 0.0, 0.0 ],
                "amount": 3
            },
            {
                "type": "circle",
                "center": [ 0.0, 0.0, 0.0 ],
                "normal": [ 0.0, 0.01, 1.0 ],
                "axis": "x",
                "radius": 0.5,
                "amount": 4
            }
        ]
    }
],
```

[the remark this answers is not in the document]
vbaconnet: Aha okay my bad, just realized you already explained it in your other comments... Sorry!!

njansson (approved):

src/simulation_components/probes.F90
```diff
@@ -107,53 +125,349 @@ subroutine probes_init_from_json(this, json, case)
...
+    logical :: found
+
+    ! Ensure only rank 0 reads the coordinates.
+    if (pe_rank .ne. 0) return
+    call json%get('coordinates', rp_list_reader, found)
+
+    ! Check if the coordinates were found and were valid
+    if (.not. found) then
+       call neko_error('No coordinates found.')
+    end if
```
timofeymukha, line 253: Can we use json_get from the json_utils here?

src/simulation_components/probes.F90
```diff
@@ -107,53 +125,349 @@ subroutine probes_init_from_json(this, json, case)
...
+
+    if (mod(size(rp_list_reader), 3) /= 0) then
+       call neko_error('Invalid number of coordinates.')
+    end if
+
+    ! Allocate list of points and reshape the input array
+    allocate(point_list(3, size(rp_list_reader)/3))
+    point_list = reshape(rp_list_reader, [3, size(rp_list_reader)/3])
```
timofeymukha, line 261: @njansson is `reshape` ok? Just checking since we usually avoid most intrinsics.

timofeymukha, line 261: @njansson is `reshape` ok? Just checking, since we usually avoid most intrinsics.

timofeymukha (commented):
@timfelle Very nice stuff indeed! Would it be possible to add the new functionality to the documentation, particularly since this one is to go into 0.8?

I think the circle in particular could make use of an example, I didn't quite get how that axis parameter works :-).

In a later PR we should see if some of the examples can make use of this new stuff. I think the scalar_mms example uses probes along a line for example.

src/simulation_components/probes.F90
timfelle: I think the point on avoiding reshape was to preserve vectorisation in critical sections of the code. I do not expect this part of the code to be critical as it is reading a list of points from a file.

src/simulation_components/probes.F90
timfelle: Done

src/simulation_components/probes.F90
timfelle: @njansson Can we get your input here? it is the last thing needed for the PR

njansson: reshape would imply a copy, can't it be avoided?

timfelle: Sure, we could go through and construct some convoluted `bounds remapping`
through a wrapper function.
However, this is a list manually typed into the json file.
I do not expect that people would wish to list enough points manually in the
json for the copy to be relevant in any circumstance.

If someone wish to list that many points, then they should do so in a sepparate
file.

PR #1207, opened 2024-03-27
Harmonize factory module filenames

timofeymukha:
Most modules were named `_fctry` but some rogue ones (mostly coded be my) were `_factory`. Better to have the same naming, so I rounded off everything to `_fctry`.

Edit: not module names but rather the file names for the modules.


njansson (commented):
To save space in the declaration, I'm in favour of `fctry`

timofeymukha:
> To save space in the declaration, I'm in favour of `fctry`

Yeah, everything is fctry now. But my message was confusing, this only changes filenames to _fctry.f90, the module names were already correct.

timfelle:
I guess we should be consistent on the contract or not across the different pars of the code. But yes the fctry is probably fine.
In any case it is generally good to have `module xyz` in file `xyz.f90`. I have to open up files relatively often to figure out where the module i am looking for is ;P

timofeymukha:
> In any case it is generally good to have `module xyz` in file `xyz.f90`
Yeah, this is exactly what I address here, 3 file names changed to be consistent with module name.

timofeymukha:
@timfelle Just to add, I think the consistency between file and module name is more or less policy, so one can issue PRs to address inconsistencies, like this one. One biggie is `coef.f90`, with the module name `coefs`. But things are a bit more difficult now with the name shadowing taken into account. The coef_t variables are called coef in a lot of routines. So maybe we wait for _m module names for that one, I don't know.

timfelle:
> @timfelle Just to add, I think the consistency between file and module name is more or less policy, so one can issue PRs to address inconsistencies, like this one. One biggie is `coef.f90`, with the module name `coefs`. But things are a bit more difficult now with the name shadowing taken into account. The coef_t variables are called coef in a lot of routines. So maybe we wait for _m module names for that one, I don't know.

Yea, question is wether we should use a `module xyz_m` for file `xyz.f90`. Even though i can see someone like pFUnit does not work if the module and file names are identical.

timofeymukha:
> Even though i can see someone like pFUnit does not work if the module and file names are identical.

If that is the case, I suppose it forces our hand.

timfelle:
> > Even though i can see someone like pFUnit does not work if the module and file names are identical.
> 
> If that is the case, I suppose it forces our hand.

Well it's for the `.pf` files, where it is struct, but thats also for their preprocessor thing.

timofeymukha:
The CI somehow really got cursed here.

timfelle (approved):

timofeymukha:
CI finally blessed the PR, I think we can merge.

PR #1210, opened 2024-03-27
Refactor and extend field_list

timofeymukha:
Here, I start implementing a convention for helper types.
- `_list_t` types have its component called `items`
- `_ptr_t` types have its component called `ptr`

I add a lot of convenience subroutines for the `field_list_t`, so that one can avoid expressions like `this%fields%items(i)%ptr%x_d` etc as much as possible. I would say quite a lot of code looks more elegant now. I will add a routine to return `x` for an item as well, I just need to look up how to return allocatable arrays in a good way (maybe it is trivial).

The next step will be to add a similar list type for vector_ptr_t and array_ptr_t.

### DIFF
--- a/src/field/field_list.f90
+++ b/src/field/field_list.f90
@@ -1,19 +1,63 @@
 module field_list
   use field, only : field_ptr_t, field_t
+  use iso_c_binding, only : c_ptr
   implicit none
   private
 
   !> field_list_t, To be able to group fields together
   type, public :: field_list_t
-     type(field_ptr_t), allocatable :: fields(:)
+     type(field_ptr_t), allocatable :: items(:)
    contains
+     !> Constructor. Allocates array and pointers.
+     procedure, pass(this) :: init => field_list_init
+     !> Destructor.
+     procedure, pass(this) :: free => field_list_free
      !> Append a field to the list.
      procedure, pass(this) :: append => field_list_append
-     !> Destructor
-     procedure, pass(this) :: free => field_list_free
+     !> Get an item pointer by array index
+     procedure, pass(this) :: get => field_list_get
+     !> Point item at given index.
+     generic :: set => set_to_ptr, set_to_field_ptr
+     procedure, pass(this) :: set_to_ptr => field_list_set_to_ptr
+     procedure, pass(this) :: set_to_field_ptr => field_list_set_to_field_ptr
+     procedure, pass(this) :: set_to_field => field_list_set_to_field
+
+     !> Get device pointer for a given index
+     procedure, pass(this) :: x_d => field_list_x_d
+
+     !> Get number of items in the list.
+     procedure, pass(this) :: size => field_list_size
   end type field_list_t
 
 contains
+  !> Constructor. Just allocates the array.
+  !! @param size The size of the list to preallocate
+  subroutine field_list_init(this, size)
+    class(field_list_t), intent(inout) :: this
+    integer, intent(in) :: size
+    integer :: i
+
+    call this%free()
+
+    allocate(this%items(size))
+  end subroutine field_list_init
+
+  !> Get number of items in the list.
+  pure function field_list_size(this) result(n)
+    class(field_list_t), intent(in) :: this
+    integer :: n
+    n = size(this%items)
+  end function field_list_size
+
+  !> Get an item pointer by array index
+  !! @param i The index of the item.
+  function field_list_get(this, i) result(f)
+    class(field_list_t), intent(inout) :: this
+    type(field_t), pointer :: f
+    integer :: i
+    f = this%items(i)%ptr
+  end function field_list_get
+
   !> Append a field to the list.
   !! @param f The field to append.
   subroutine field_list_append(this, f)
@@ -22,12 +66,12 @@ contains
     type(field_ptr_t), allocatable :: tmp(:)
     integer :: len
 
-    len = size(this%fields)
+    len = size(this%items)
 
     allocate(tmp(len+1))
-    tmp(1:len) = this%fields
-    call move_alloc(tmp, this%fields)
-    this%fields(len+1)%f => f
+    tmp(1:len) = this%items
+    call move_alloc(tmp, this%items)
+    this%items(len+1)%ptr => f
 
   end subroutine field_list_append
 
@@ -36,16 +80,58 @@ contains
     class(field_list_t), intent(inout) :: this
     integer :: i, n_fields
 
-    if (allocated(this%fields)) then
-       n_fields = size(this%fields)
+    if (allocated(this%items)) then
+       n_fields = this%size()
        do i=1, n_fields
-          call this%fields(i)%f%free()
-          nullify(this%fields(i)%f)
+          call this%items(i)%ptr%free()
+          nullify(this%items(i)%ptr)
        end do
-       deallocate(this%fields)
+       deallocate(this%items)
     end if
 
   end subroutine field_list_free
+
+  !> Get device pointer for a given index
+  !! @param i The index of the item.
+  function field_list_x_d(this, i) result(ptr)
+    class(field_list_t), intent(inout) :: this
+    type(c_ptr), pointer :: ptr
+    integer :: i
+    ptr = this%items(i)%ptr%x_d
+  end function field_list_x_d
+
+  !> Point item at a given index.
+  !! @param i The index of the item.
+  !! @param ptr A field pointer to point the item to.
+  subroutine field_list_set_to_ptr(this, i, ptr)
+    class(field_list_t), intent(inout) :: this
+    integer, intent(in) :: i
+    type(field_t), pointer, intent(in) :: ptr
+
+    this%items(i)%ptr => ptr
+  end subroutine field_list_set_to_ptr
+
+  !> Point item at a given index.
+  !! @param i The index of the item.
+  !! @param ptr An encapsulated field pointer to point the item to.
+  subroutine field_list_set_to_field_ptr(this, i, ptr)
+    class(field_list_t), intent(inout) :: this
+    integer, intent(in) :: i
+    type(field_ptr_t), intent(in) :: ptr
+
+    this%items(i)%ptr => ptr%ptr
+  end subroutine field_list_set_to_field_ptr
+
+  !> Point item at a given index.
+  !! @param i The index of the item.
+  !! @param field A field to point the item to.
+  subroutine field_list_set_to_field(this, i, fld)
+    class(field_list_t), intent(inout) :: this
+    integer, intent(in) :: i
+    type(field_t), target, intent(in) :: fld
+
+    this%items(i)%ptr => fld
+  end subroutine field_list_set_to_field
 
 
 
--- a/src/field/scratch_registry.f90
+++ b/src/field/scratch_registry.f90
@@ -99,7 +99,7 @@ contains
     if (present(size)) then
        allocate (this%fields(size))
        do i= 1, size
-          allocate(this%fields(i)%f)
+          allocate(this%fields(i)%ptr)
        end do
        allocate (this%inuse(size))
     else
@@ -125,8 +125,8 @@ contains
 
     if (allocated(this%fields)) then
        do i=1, this%nfields
-          call this%fields(i)%f%free()
-          deallocate(this%fields(i)%f)
+          call this%fields(i)%ptr%free()
+          deallocate(this%fields(i)%ptr)
        end do
 
        deallocate(this%fields)
@@ -182,7 +182,7 @@ contains
     temp(1:this%nfields) = this%fields(1:this%nfields)
 
     do i=this%nfields +1, size(temp)
-       allocate(temp(i)%f)
+       allocate(temp(i)%ptr)
     enddo
 
     call move_alloc(temp, this%fields)
@@ -208,11 +208,11 @@ contains
          if (this%inuse(index) .eqv. .false.) then
             write (name, "(A3,I0.3)") "wrk", index
 
-            if (.not. allocated(this%fields(index)%f%x)) then
-               call this%fields(index)%f%init(this%dof, trim(name))
+            if (.not. allocated(this%fields(index)%ptr%x)) then
+               call this%fields(index)%ptr%init(this%dof, trim(name))
                nfields = nfields + 1
             end if
-            f => this%fields(index)%f
+            f => this%fields(index)%ptr
             this%inuse(index) = .true.
             this%nfields_inuse = this%nfields_inuse + 1
             return
@@ -225,8 +225,8 @@ contains
       nfields_inuse = nfields_inuse + 1
       this%inuse(nfields) = .true.
       write (name, "(A3,I0.3)") "wrk", index
-      call this%fields(nfields)%f%init(this%dof, trim(name))
-      f => this%fields(nfields)%f
+      call this%fields(nfields)%ptr%init(this%dof, trim(name))
+      f => this%fields(nfields)%ptr
 
     end associate
   end subroutine request_field
--- a/src/fluid/fluid_stats.f90
+++ b/src/fluid/fluid_stats.f90
@@ -227,49 +227,49 @@ contains
     call this%e13%init(this%stats_work, 'e13')
     call this%e23%init(this%stats_work, 'e23')
 
-    allocate(this%stat_fields%fields(this%n_stats))
+    allocate(this%stat_fields%items(this%n_stats))
 
-    this%stat_fields%fields(1)%f => this%pp%mf
-    this%stat_fields%fields(2)%f => this%uu%mf
-    this%stat_fields%fields(3)%f => this%vv%mf
-    this%stat_fields%fields(4)%f => this%ww%mf
-    this%stat_fields%fields(5)%f => this%uv%mf
-    this%stat_fields%fields(6)%f => this%uw%mf
-    this%stat_fields%fields(7)%f => this%vw%mf
-    this%stat_fields%fields(8)%f => this%uuu%mf !< <uuu>
-    this%stat_fields%fields(9)%f => this%vvv%mf !< <vvv>
-    this%stat_fields%fields(10)%f => this%www%mf !< <www>
-    this%stat_fields%fields(11)%f => this%uuv%mf !< <uuv>
-    this%stat_fields%fields(12)%f => this%uuw%mf !< <uuw>
-    this%stat_fields%fields(13)%f => this%uvv%mf !< <uvv>
-    this%stat_fields%fields(14)%f => this%uvw%mf !< <uvv>
-    this%stat_fields%fields(15)%f => this%vvw%mf !< <vvw>
-    this%stat_fields%fields(16)%f => this%uww%mf !< <uww>
-    this%stat_fields%fields(17)%f => this%vww%mf !< <vww>
-    this%stat_fields%fields(18)%f => this%uuuu%mf !< <uuuu>
-    this%stat_fields%fields(19)%f => this%vvvv%mf !< <vvvv>
-    this%stat_fields%fields(20)%f => this%wwww%mf !< <wwww>
-    this%stat_fields%fields(21)%f => this%ppp%mf
-    this%stat_fields%fields(22)%f => this%pppp%mf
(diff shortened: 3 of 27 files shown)


timofeymukha:
Quite happy now. A little bit scared if the %x and %x_d routines in field_list_t are correctly written though.

src/field/field_list.f90
```diff
@@ -1,19 +1,67 @@
...
+     !> Destructor.
+     procedure, pass(this) :: free => field_list_free
      !> Append a field to the list.
      procedure, pass(this) :: append => field_list_append
-     !> Destructor
-     procedure, pass(this) :: free => field_list_free
+     !> Get an item pointer by array index
+     procedure, pass(this) :: get => field_list_get
+     !> Point item at given index.
+     generic :: set => set_to_ptr, set_to_field_ptr
```
timfelle, line 21: Howabout using `assign` instead of `set`. That might be clearer for what we are
actually doing.

src/field/scratch_registry.f90
```diff
@@ -125,8 +125,8 @@ subroutine scratch_registry_free(this)
 
     if (allocated(this%fields)) then
        do i=1, this%nfields
-          call this%fields(i)%f%free()
-          deallocate(this%fields(i)%f)
+          call this%fields(i)%ptr%free()
+          deallocate(this%fields(i)%ptr)
        end do
```
timfelle, line 130:
```suggestion
    if (allocated(this%fields)) then
       call this%fields%free()
```

Shouldnt we use the shining new `free` function instead.

src/field/scratch_registry.f90
```diff
@@ -182,7 +182,7 @@ subroutine expand(this)
     temp(1:this%nfields) = this%fields(1:this%nfields)
 
     do i=this%nfields +1, size(temp)
-       allocate(temp(i)%f)
+       allocate(temp(i)%ptr)
     enddo
```
timfelle, line 186: Are we allocating an empty pointer here?
I dont think thats what we wanted.

src/field/field_list.f90
```diff
@@ -1,19 +1,63 @@
 module field_list
   use field, only : field_ptr_t, field_t
+  use iso_c_binding, only : c_ptr
   implicit none
   private
 
   !> field_list_t, To be able to group fields together
   type, public :: field_list_t
-     type(field_ptr_t), allocatable :: fields(:)
+     type(field_ptr_t), allocatable :: items(:)
```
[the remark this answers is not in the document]
timofeymukha: That would in principle be very nice. Then we would have to have wrappers for all the components of `field_t`. And I think things may get a bit complicated. Consider this statment: `n = fields%items(1)%ptr%dof%size()`. Let's day we have a `dof(i)` wrapper function. I don't think one can write `n = fields%dof(i)%size()`, right? I.e. one cannot chain the call to the function with `%`. If that is the case, we would need to have an extra variable, which is a pointer to a `dofmap_t`, point that to `fields%dof%(i)` and then call `size`. This kind of cripples the convenience of using the wrapper. What do you think?

src/field/field_list.f90
timofeymukha: Pondered on this as well. Somehow assign in my head associated with making a copy, but perhaps it is a missjudgment. I was also thinking of `point` to clarify that this is pointer assignment.

src/field/scratch_registry.f90
timofeymukha: Have to be a bit careful here, because `fields` here is not a `field_list_t`, so we have to go manual I believe. Now, maybe it *should* be a `field_list_t` :-). But maybe in a future PR.

src/field/scratch_registry.f90
timofeymukha: So I guess this allocates an unititialized `field_t`, and then the `init` is called in the `request` subroutine.

src/field/field_list.f90
timfelle: Well what we do is pointer assignment and since the user of the interface should understand that this is a pointer based list then it is clear that we are working with an assignment here.

src/field/field_list.f90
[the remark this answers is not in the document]
timfelle: Yes, the lack of chaining for function statements is kinda annoying in Fortran. But then again, is the size of the `dof` different than the size of the `field` itself. And could one not simply use `fields%size(i)` to access the size of the given field?

src/field/scratch_registry.f90
timfelle: Ahh i see, yes that should be a future PR then.

src/field/scratch_registry.f90
timfelle: Hmm why not just allocate in the request then. I don't see the point of allocating something with no guarantee that it is ever initialized.

It is also not done for the default case where size is not supplied, which make me think it was actually just something we forgot to delete earlier.

src/fluid/fluid_stats.f90
```diff
@@ -648,35 +648,35 @@ subroutine fluid_stats_post_process(this, mean, reynolds, pressure_flatness,&
     integer :: n
 
     if (present(mean)) then
-       n = mean%fields(1)%f%dof%size()
-       call copy(mean%fields(1)%f%x,this%u_mean%x,n)
-       call copy(mean%fields(2)%f%x,this%v_mean%x,n)
-       call copy(mean%fields(3)%f%x,this%w_mean%x,n)
-       call copy(mean%fields(4)%f%x,this%p_mean%x,n)
+       n = mean%items(1)%ptr%dof%size()
+       call copy(mean%items(1)%ptr%x,this%u_mean%x,n)
```
timofeymukha, line 652: Would ideally like to use the new %x() wrapper in statements. It does return an array pointer though.

src/field/field_list.f90
timofeymukha: Yeah, you are right, assign should be perfectly fine.

src/field/scratch_registry.f90
timofeymukha: Regarding the last thing, you mean in `init`? I think we just preallocate 10 ptrs. But I think you are right, it would be better to do it in request.

How about I open an issue for that and we can treat that separately?

src/fluid/fluid_stats.f90
njansson: should work, if you do an assignment like `x => apa%x()`

src/field/field_list.f90
```diff
@@ -36,17 +84,68 @@ subroutine field_list_free(this)
...
+    x => this%items(i)%ptr%x
+  end function field_list_x
+
+  !> Point item at a given index.
+  !! @param i The index of the item.
+  !! @param ptr A field pointer to point the item to.
+  subroutine field_list_set_to_ptr(this, i, ptr)
+    class(field_list_t), intent(inout) :: this
+    integer, intent(in) :: i
+    type(field_t), pointer, intent(in) :: ptr
```
njansson, line 122:
```suggestion
    type(field_t), pointer, target, intent(in) :: ptr
```

src/field/field_list.f90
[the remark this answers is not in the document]
timofeymukha: I think we can certainly write more wrappers, although `field_t` does not have a `size` component itself (would probably be nice to add it). But I fear that there will always be some new special case where one would want to pull out something new from the `items`, which will be impossible if we make it private, or at least make things inconvenient.

Like, look at this stuff in fld_file.f90
```
p%ptr => data%items(1)%ptr%x(:,1,1,1)
```
We have a wrapper for `x`, but to do this weird reshape I guess one would again need an extra variable.

How about we call it `items_` as a soft `private` indicator?

src/fluid/fluid_user_source_term.f90
```diff
@@ -224,16 +224,16 @@ subroutine fluid_user_source_term_compute(this, t, tstep)
...
+       call device_add2(this%fields%x_d(1), this%u_d, n)
+       call device_add2(this%fields%x_d(2), this%v_d, n)
+       call device_add2(this%fields%x_d(3), this%w_d, n)
     else
-       call add2(this%fields%fields(1)%f%x, this%u, n)
-       call add2(this%fields%fields(2)%f%x, this%v, n)
-       call add2(this%fields%fields(3)%f%x, this%w, n)
+       call add2(this%fields%items(1)%ptr%x, this%u, n)
+       call add2(this%fields%items(2)%ptr%x, this%v, n)
+       call add2(this%fields%items(3)%ptr%x, this%w, n)
```
njansson, line 236: Not sure if I understand the new `x_d` procedure, but wouldn't also the `x` procedure work in the same way, thus removing the extreme `items( )%ptr%x` ?

timofeymukha: Yes, that is the idea. I basically wanted you to check if the x and x_d procedures are well-written before I go ahead. I got sheepish with the pointer stuff after recent discussions :-) I guess for x_d I just copy the pointer, because that is just an ordinary C pointer, right? And for x I return a Fortran pointer like `ptr => this%items(i)%ptr%x`.

src/field/field_list.f90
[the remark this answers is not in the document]
timfelle: The solution to the example you point out here would be to do the reshape whenever you use the p%ptr instead of when you assign it to x.

I don't think we should code up stuff for special new cases. If people wish to access more complex parts of the system then they will do so. And should they not access that stuff by referring to the direct field instead, `neko_field_registry%get_field()`?

As for the size, if the field_t have the size of the dof it points to, then `field%size()` should return the same as `field%dof%size()`, in which case the `field_list_t%size(i)`would just refer to the dof associated with it.

timofeymukha:
Interesting, it seems like some compilers won't accept passing an array pointer to a subroutine which taken an array as an in/out argument. Am I interpreting this correctly, @njansson ? Then I guess the usefulness of the `x()` wrapper is quite limited.

timofeymukha:
> > Interesting, it seems like some compilers won't accept passing an array pointer to a subroutine which takes an array as an in/out argument. Am I interpreting this correctly, @njansson ? Then I guess the usefulness of the `x()` wrapper is quite limited.
> 
> True, the standard would (rightfully so) reject that ;) Ok then I guess we drop `x()` for now

Yes.. And with that unfrotunately also the dream of `private` `items`.

timfelle (commented):
>>> > > Interesting, it seems like some compilers won't accept passing an array pointer to a subroutine which takes an array as an in/out argument. Am I interpreting this correctly, @njansson ? Then I guess the usefulness of the `x()` wrapper is quite limited.
>> 
>> True, the standard would (rightfully so) reject that ;) Ok then I guess we drop `x()` for now

> Yes.. And with that unfrotunately also the dream of `private` `items`. 

Could we atleast add the `x` and `x_d` pointers to `field_ptr_t` so we don't have to do `list(i)%ptr%x` but can do `list(i)%x`?

timfelle (commented):
>>> > > Interesting, it seems like some compilers won't accept passing an array pointer to a subroutine which takes an array as an in/out argument. Am I interpreting this correctly, @njansson ? Then I guess the usefulness of the `x()` wrapper is quite limited.
>> 
>> True, the standard would (rightfully so) reject that ;) Ok then I guess we drop `x()` for now

> Yes.. And with that unfrotunately also the dream of `private` `items`. 

Could we atleast add the `x` and `x_d` pointers to `field_ptr_t` so we don't have to do `list(i)%ptr%x` but can do `list(i)%x`?

timofeymukha:
> > > > > > Interesting, it seems like some compilers won't accept passing an array pointer to a subroutine which takes an array as an in/out argument. Am I interpreting this correctly, @njansson ? Then I guess the usefulness of the `x()` wrapper is quite limited.
> > > 
> > > 
> > > True, the standard would (rightfully so) reject that ;) Ok then I guess we drop `x()` for now
> 
> > Yes.. And with that unfrotunately also the dream of `private` `items`.
> 
> Could we atleast add the `x` and `x_d` pointers to `field_ptr_t` so we don't have to do `list(i)%ptr%x` but can do `list(i)%x`?

`x_d` we still have in the list since that does not seem to cause trouble, as we copy a c_ptr. Regarding `x` I honestly have no idea :-). I guess since it is a pointer it will have the same issue? Or if we store the %x pointer as a component, it would be different compared to returning a pointer from a function?

src/fluid/fluid_user_source_term.f90
njansson: Maybe it would work by returning a 4d pointer? e.g. by declaring the return pointer as `ptr(:,:,:,:)`

timofeymukha: @njansson 
This is what we had
```
  function field_list_x(this, i) result(x)
    class(field_list_t), target, intent(inout) :: this
    real(kind=rp), pointer :: x(:,:,:,:)
    integer, intent(in) :: i
    x => this%items(i)%ptr%x
  end function field_list_x
```

timfelle: Did it not work if you set the intent of `this` to strictly `in`?

timofeymukha: Yeah this is what I am trying now :-). Maybe that was the issue

timfelle: It worked for me 🎆

timofeymukha: I think last time it was only nvfortran that blew up :-). Let's see.

timofeymukha: @timfelle Seems like GCC won't have it :(

timfelle: Are we certain the problem is not in how `copy` is defined to accept a 1D array, and we pass it a pointer to a 4D array?

src/field/field_list.f90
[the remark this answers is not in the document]
MartinKarp: I quite like the idea of having the `field%size()`. As for the shape and such of the field, the iteration of `field%x(i,1,1,1)` is used in a lot of places where we just want to iterate from `i = 1:n` but where the field is of dimension `N+1,N+1,N+1,nel`. Is there a simple way to reshape it to a 1d array, also when we do not use the ptr, but when we pass down the field itself?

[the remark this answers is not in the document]
timfelle: > I quite like the idea of having the `field%size()`. As for the shape and such of the field, the iteration of `field%x(i,1,1,1)` is used in a lot of places where we just want to iterate from `i = 1:n` but where the field is of dimension `N+1,N+1,N+1,nel`. Is there a simple way to reshape it to a 1d array, also when we do not use the ptr, but when we pass down the field itself?

Well the intrinsic reshape does exist, but i am not exacly sure of the performance characteristics of that one when used as deeply in the code as we might want to.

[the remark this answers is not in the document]
MartinKarp: Allright, maybe that would be interesting then :)

One way I thought of was to let field_t rely on vector_t under the hood maybe, that one is a 1d array, but I'm not sure that would be very pretty either. Then `field%x` would be a 4d pointer to that array or something.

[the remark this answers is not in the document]
njansson: > > I quite like the idea of having the `field%size()`. As for the shape and such of the field, the iteration of `field%x(i,1,1,1)` is used in a lot of places where we just want to iterate from `i = 1:n` but where the field is of dimension `N+1,N+1,N+1,nel`. Is there a simple way to reshape it to a 1d array, also when we do not use the ptr, but when we pass down the field itself?
> 
> Well the intrinsic `reshape` does exist, but i am not exacly sure of the performance characteristics of that one when used as deeply in the code as we might want to.

`reshape` will return a copy, so not really what we would need at these places ;)

src/io/fld_file.f90
```diff
@@ -147,65 +147,65 @@ subroutine fld_file_write(this, data, t)
...
+          tem%ptr => data%items(2)%ptr%x(:,1,1,1)
           write_pressure = .true.
           write_temperature = .true.
        case (3)
-          u%x => data%fields(1)%f%x(:,1,1,1)
-          v%x => data%fields(2)%f%x(:,1,1,1)
-          w%x => data%fields(3)%f%x(:,1,1,1)
+          u%ptr => data%items(1)%ptr%x(:,1,1,1)
+          v%ptr => data%items(2)%ptr%x(:,1,1,1)
+          w%ptr => data%items(3)%ptr%x(:,1,1,1)
```
adperezm, line 168: I am not very familiar with what ptr is in this case, so the following comment might be irrelevant. But while perhaps not ideal, I did like the idea in the previous case that when you reached the %x in the chain, then you start thinking about the shape. Now it seems that it is mixed depending on the type? or am I misunderstanding?

With this new definition in this particular example: To access the data that u%ptr points to, would you do so by u%ptr%x(:,:,:,:) or u%ptr(:,:,:,:)? Because I would find it confusing if ptr meant different things depending on the context.

Sorry if this comment made no sense.

timofeymukha: No, I think what you say is valid. The variables in this code snippet are `type(array_ptr_t)`.  My rationale was that if you have a wrapper type like that, which essentially is just one component, which is a pointer to something, the name of the component is always `ptr`. Usually, the pointer would be to a derived type, like `field_t`. However, `array_ptr_t` is kind of a corner case, because here the pointer is actually to a raw array, meaning that one can operate on it directly as if it was data. So one could argue for an exception here and call it `x`. But then maybe one forgets it is a pointer, so I don't know :-).

adperezm: Oh okay I see. Yes, maybe it is better to not make exceptions and leave it as you propose in the PR. then at least one knows what one has.

timofeymukha:
How do we feel about merging this in now that we have the 0.8 branched out?

njansson (approved):

src/field/field_list.f90
```diff
@@ -1,19 +1,78 @@
...
+    call this%free()
+
+    allocate(this%items(size))
+  end subroutine field_list_init
+
+  !> Get number of items in the list.
+  pure function field_list_size(this) result(n)
+    class(field_list_t), intent(in) :: this
+    integer :: n
+    n = size(this%items)
```
MartinKarp, line 64: Is it the intended behavior to return a number even if no fields have been added to the list?

src/field/field_list.f90
```diff
@@ -22,12 +81,12 @@ subroutine field_list_append(this, f)
...
 
     allocate(tmp(len+1))
-    tmp(1:len) = this%fields
-    call move_alloc(tmp, this%fields)
-    this%fields(len+1)%f => f
+    tmp(1:len) = this%items
+    call move_alloc(tmp, this%items)
+    this%items(len+1)%ptr => f
 
   end subroutine field_list_append
```
MartinKarp, line 91: So we are somehat backwards compatible with how it work before or is the idea to always use append?

src/fluid/fluid_scheme.f90
```diff
@@ -450,15 +450,15 @@ subroutine fluid_scheme_init_common(this, msh, lx, params, scheme, user, &
     !
     ! Initialize field list and bc list for user_dirichlet_update
     !
-    allocate(this%field_dirichlet_fields%fields(4))
+    allocate(this%field_dirichlet_fields%items(4))
 
-    this%field_dirichlet_fields%fields(1)%f => &
+    this%field_dirichlet_fields%items(1)%ptr => &
```
MartinKarp, line 455: Okay, I see, we keep the original functionality as well

MartinKarp (approved):
I think this is pretty great, the only major thing I'm thinking about is that it is somewhere between a list datatype and an array of pointers. It is not necessarily something wrong in this and I am partially responsible since I named it field_list to begin with (when in practice it was just an array of field pointers). It is mostly something I would like to note, and kinda relates to the meaning of size() as well as it now returns the size of the array, not the size of the list kind of. Im very happy with this overall though so I think it can be approved.

src/fluid/fluid_scheme.f90
timofeymukha: Yeah exactly, I did not change any of this logic here.

src/field/field_list.f90
timofeymukha: I guess at the moment yes.

PR #1213, opened 2024-03-31
Modernize advection types to use json

timofeymukha:
This makes some improvements to the advection types.

The factory is moved to a separate module, as with the other factories. Furthermore, it now accepts the parameter file and does the figuring out of which type to create based on that. This paves the way to have the advection type JSON-selectable, if someone wants to implement other forms of the operators (conservative, skew-symmetric etc). 

Added mandatory destructor to advection types, which is called in the factory. In the future we may want to make a mandatory JSON-based constructor as per usual, but for now we can live without it.


MartinKarp (approved):
Looks good to me, I guess having something similar for operators will maybe also make sense in the future.

njansson (approved):

PR #1216, opened 2024-04-01
Remove u, v, w fron Dong outflow constructor

timofeymukha:
Uses the field registry instead. This is in the same direction as PR #1215 , working towards a unified constructor interface.

If we ever need other fields rather than "u", "v", "w", they can be made JSON-selectable.


njansson (approved):

njansson:
Do we want to have this in 0.8 or later?

timofeymukha:
This is so small that it probably doesn't matter.

timfelle (approved):

PR #1222, opened 2024-04-03
Introduce probe header and update probe documentation

timfelle:
This pull request introduces a probe header that contains the number of probes, number of fields, and a list of the fields probed in order. 

It also updates the documentation of probes to include the output format of the probed information.


njansson (approved):

MartinKarp (approved):
Very nice! @vbaconnet is there anything you would like to add?

PR #1226, opened 2024-04-06
More only statements

timofeymukha:
I avoided touching device related stuff here, because I compile for a CPU backend and some misses are only detected at link type. So those should be tackled in an upcoming sweep.


timofeymukha:
Never seizes to amaze me how inconsistent things are across compilers :-)

njansson (commented):
Not sure what the best way to deal with small modules, for example adding`only` for `gs_op.f90` is a bit overkill since the module only provides what we then add in the only statement.

timofeymukha:
> Not sure what the best way to deal with small modules, for example adding`only` for `gs_op.f90` is a bit overkill since the module only provides what we then add in the only statement.

Yeah, that's true. Still, I'd rather minimize Cray's warnings when possible :-). I guess there aren't many modules that fall into this category though.

njansson:
> > Not sure what the best way to deal with small modules, for example adding`only` for `gs_op.f90` is a bit overkill since the module only provides what we then add in the only statement.
> 
> Yeah, that's true. Still, I'd rather minimize Cray's warnings when possible :-). I guess there aren't many modules that fall into this category though.

I've kind of admit defeat with recent Cray Fortran (some of these warnings can't be removed, e.g. things related to `use mpi_f08` or `iso_c_bindings` sigh), and now use `-M878,1753` to silence these warnings (^-^)v

(Ps. seems like `only` opens up other interesting way of breaking stuff, see for example https://github.com/ExtremeFLOW/neko/pull/1179)

timofeymukha:
> > > Not sure what the best way to deal with small modules, for example adding`only` for `gs_op.f90` is a bit overkill since the module only provides what we then add in the only statement.
> > 
> > 
> > Yeah, that's true. Still, I'd rather minimize Cray's warnings when possible :-). I guess there aren't many modules that fall into this category though.
> 
> I've kind of admit defeat with recent Cray Fortran (some of these warnings can't be removed, e.g. things related to `use mpi_f08` or `iso_c_bindings` sigh), and now use `-M878,1753` to silence these warnings (^-^)v
> 
> (Ps. seems like `only` opens up other interesting way of breaking stuff, see for example #1179)

Yeah, I saw that one, so I avoided `comm` here :-). I still think it's nice with only since one sees where stuff comes from, but as usual with Fortran there will be some exceptions to any rule.

timofeymukha:
Merge?

timfelle (approved):
From what I can check from here, everythings looks reasonable.

src/gs/gs_mpi.f90
```diff
@@ -32,10 +32,9 @@
 !
 !> Defines MPI gather-scatter communication
 module gs_mpi
-  use neko_config
-  use num_types
-  use gs_comm
-  use gs_ops
+  use num_types, only : rp
+  use gs_comm, only : gs_comm_t, GS_COMM_MPI, GS_COMM_MPIGPU
+  use gs_ops, only : GS_OP_ADD, GS_OP_MAX, GS_OP_MIN, GS_OP_MUL
```
njansson, line 37: Follow up on the prev discussion, I would allow the only for these kinds of bare bones modules only providing a set of parameters

njansson: This will simply not scale, especially now when we are adding more backends

MartinKarp (approved):
Looks good to me, really nice you took the time to add all these.

src/gs/gs_mpi.f90
timfelle: @njansson How is this discussion going. Is it something we are happy with? It is the only obstacle to a merge at the moment.

njansson: I would prefer not to use `only` for these bare bones modules

timofeymukha: Yeah I am ok with changing that for these cases, just haven't had the time. Most of the time it is only rp imported from num_types and GS_OP_ADD for the gs_ops, in those cases I would keep the `only`.

timfelle: Interesting, do if any of those modules change down the line, we would have to track down all the places where we did not do only's and then add them? 
Is there a technical reason to not add them, or is it purely philosophical (Not that my argument is not, or better for that matter).

timfelle: And should we update our [Dev-Patterns](https://neko.cfd/docs/release/d0/d47/dev_patterns.html) to accommodate importing the small modules without using `only`?

njansson: I'm just a bit burnt by #1179, so a bit reluctant of "unnecessary" `only`.

timfelle: Sure, I am not that much against it. Maybe we should just add a list with that kind of "gotchas". We know the `comm` module should not be imported using `only` so lets add that as an anti-pattern in the documentation. I know I have added that one with `only` in quite a bunch of my pieces of code.

PR #1229, opened 2024-04-08
NEKO v0.8.0

njansson:


timfelle:
I guess the new CI is not added in here. Is that the plan?

njansson:
> I guess the new CI is not added in here. Is that the plan?

Good question. I guess it would be rather painful to cherry pick all changes

timfelle:
> Good question. I guess it would be rather painful to cherry pick all changes

We can also just update the roccm version in a separate PR directly to the release and then wait for 0.8.1 with the CI

njansson:
> > Good question. I guess it would be rather painful to cherry pick all changes
> 
> We can also just update the roccm version in a separate PR directly to the release and then wait for 0.8.1 with the CI

Sounds like an excellent compromise ;)

MartinKarp:
How are we looking on this one? To me it looks good, there are of course things getting added all the time, but I think this is in a relatively good state to get merged and us having a more recent release.

MartinKarp (approved):

njansson:
> How are we looking on this one? To me it looks good, there are of course things getting added all the time, but I think this is in a relatively good state to get merged and us having a more recent release.

Agree, I think this was also the conclusion at the latest dev meeting.

Let me fix the version number etc , and we focus on 0.9 instead

PR #1240, opened 2024-04-17
Boundary condition type updates

timofeymukha:
This continues the series of PRs on cleaning up `bc_t` and descendants.

- A deferred destructor is added to `bc_t`  and implemented in all types.
- The constructor and destructor in `bc_t` are called `init_base` and `free_base` to let `init` and `base` be deferred. This is similar to other hierarchies we have.
- The hierarchy is completely flattened, so everything descends from `bc_t` now. This is perhaps the easiest if we will have a common factory for all types. So a clean slate in that sense, and then we see if we can combine/remove some bcs or introduce and extra layer of inheritance.
- In symmetry, dong and non_normal we now have `init` instead of the individual constructor-like routine names. Hopefully to spread to all types under one interface.
- Similarly, the Dong bc uses JSON in the constructor as a preview for what will probably be the case for all types. Cleans up json parsing in fluid_scheme a bit. 
- Rouge use of a local `coef` component was removed, using now the one in `bc_t`.


njansson (commented):
A bit nitpick, but looks a bit strange that all bc's are calling `init_base`. Shouldn't the logic rather be `init()` in the type, and that init might call `init()` (can be call `init_base`) in the base type

timofeymukha:
> A bit nitpick, but looks a bit strange that all bc's are calling `init_base`. Shouldn't the logic rather be `init()` in the type, and that init might call `init()` (can be call `init_base`) in the base type

So what I am aiming for eventually is that `init` is deferred in the base time and follows an `abstract interface`. Since you can't have a default implementation for a deferred routine, the name of the constructor for `bc_t` has to be called something else. We used `init_base` in a few hierarchies, so I picked it up here as well. So, then `init` will call `init_base` in the beginning to setup the base type. The deferred `init` is not quite there yet, but this PR is kind of prepping for that :-).

njansson (approved):

timofeymukha:
@MartinKarp Whenever you have time :-). A new batch of updates is coming.

src/bc/blasius.f90
```diff
@@ -35,18 +35,18 @@ module blasius
...
   use, intrinsic :: iso_fortran_env
   use, intrinsic :: iso_c_binding
+  use bc, only : bc_t
   implicit none
   private
 
   !> Blasius profile for inlet (vector valued)
-  type, public, extends(inflow_t) :: blasius_t
-     type(coef_t), pointer :: c => null()
+  type, public, extends(bc_t) :: blasius_t
```
MartinKarp, line 48: Have we fixed Blasius for elements that are rotated? Otherwise I think maybe we should issue some kind of warning or so while we are at it as it does not work in general.

src/bc/blasius.f90
```diff
@@ -121,19 +123,19 @@ subroutine blasius_apply_vector(this, x, y, z, n, t, tstep)
          select case(facet)
          case(1,2)
             x(k) = this%bla(zc(idx(1), idx(2), idx(3), idx(4)), &
-                 this%delta, this%x(1))
+                 this%delta, this%uinf(1))
```
MartinKarp, line 126: I guess this is the part that is problematic, since facet 1,2 are not necessarily aligned with x, we could fix this by doing a check similar to symmetry to check what direction is actually aligned with each facet. Perhaps this is out of scope for this PR though, I just want to note this down somewhere.

MartinKarp (approved):
I think is is really nice, I guess the things I mentioned about Blasius is kind of out of scope and so is the whole thing about renaming inflow to something reasonable (and perhaps usr to user). But I think this is a good way in the right direction.

src/bc/blasius.f90
timofeymukha: Haven't fixed. I have in fact so far not touched the actual conditions much at all. Working mainly developing a shared interface and moving the construction to a factory. I will add a warning and a docstring in the next PR

MartinKarp: Tiptop, sounds good

src/bc/blasius.f90
timofeymukha: I add a TODO to the code.

PR #1244, opened 2024-04-18
Dissipation

Shiyu-Sandy-Du:
In this PR, dissipation tensor, as a post-processing statistics, is completed with the velocity gradient part. However, the resulting field should be multiplied by 2*nu to get the correct tensor, which I am not sure whether should we put it here.
p.s. the formulation is $\epsilon_{i, j} = 2*\nu \left< \frac{\partial u_i}{\partial u_k}\frac{\partial u_j}{\partial u_k}\right>$ and the current output is $\left<\frac{\partial u_i}{\partial u_k}\frac{\partial u_j}{\partial u_k}\right>$


adperezm:
I am a bit confused as to what you call the dissipation tensor. 

I am thinking of the dissipation term in the kinetic energy equation, which is a scalar equation and will give you a scalar output:

image

Could you clarify?

adperezm:
Hi,

My opinion would be that if we are providing dissipation, then it should be dissipation. i.e., multiply it by what it needs to be multiplied by.

Otherwise we get into a situation like with the jac variable in the code. Which in my opinion is named wrongly.

As another note, are you sure you are not missing a 1/2 scaling somewhere? I typically get the dissipation from the strain rate, which is what I guess you did here, but you write it in a different way that I am not familiar with :) (I am very bad with Einstein notation)

Just for reference:
image

image

timofeymukha:
I guess the scalar eps is what one typically means when one says "dissipation", e.g. a term in the equation for the TKE. But perhaps the tensor is also relevant.

pschlatt1:
The dissipation tensor relates to the disspation terms that appear in the transport equations for the individual Reynolds stress components. The scalar dissipation (pseudo-dissipation) is the trace of that tensor. In the end, we should make sure that we have the complete Reynolds stress transport terms in the statistics. For now, I would in fact recommend to just directly compute the TKE dissipation which is the only one needed to compute the isotropic Kolmogorov units. The rest can come later.

Regarding the form of the terms, one should clearly include nu (or 1/Re) in it. The factor of two is a bit unclear, as if you do the dissipation for the ui uj terms, you typically do not have the 1/2, but if you do it for TKE, then you do (since TKE is 1/2 u_i u_i).

timofeymukha:
Should one have viscosity as a command line argument to the post-processing utility, maybe? Or should it just read the case file?

Shiyu-Sandy-Du:
> Should one have viscosity as a command line argument to the post-processing utility, maybe? Or should it just read the case file?

Maybe it's better to read it from the case file, because the nu value is not always directly there (if users solve the dimensional eqn then it's mu/rho; if dimensionless then it's 1/Re).

Shiyu-Sandy-Du:
> The dissipation tensor relates to the disspation terms that appear in the transport equations for the individual Reynolds stress components. The scalar dissipation (pseudo-dissipation) is the trace of that tensor. In the end, we should make sure that we have the complete Reynolds stress transport terms in the statistics. For now, I would in fact recommend to just directly compute the TKE dissipation which is the only one needed to compute the isotropic Kolmogorov units. The rest can come later.
> 
> Regarding the form of the terms, one should clearly include nu (or 1/Re) in it. The factor of two is a bit unclear, as if you do the dissipation for the ui uj terms, you typically do not have the 1/2, but if you do it for TKE, then you do (since TKE is 1/2 u_i u_i).

I could write out another term of scalar epsilon simply as 1/2 of the trace of the tensor (and each term of the tensor is multiplied by 2nu).

pschlatt1:
The dissipation is not a free parameter, but directly connected to the simulation. So it should be part of the statistics file. And as such it is not only relevant for dissipation, but for all terms in the various transport equations. Thus my suggestion is to not just focus in this discussion on dissipation, but all relevant terms in the transport equation of TKE, including visous transport, turbulent transport, production etc.

MartinKarp:
For computing the lift I added it as a command line argument, but perhaps one could read the json as an argument instead and get it from there.

Shiyu-Sandy-Du:
> The dissipation is not a free parameter, but directly connected to the simulation. So it should be part of the statistics file. And as such it is not only relevant for dissipation, but for all terms in the various transport equations. Thus my suggestion is to not just focus in this discussion on dissipation, but all relevant terms in the transport equation of TKE, including visous transport, turbulent transport, production etc.

In dissipation, we could of course include nu in <nu du/dx du/dx> (suppose u = <u>+u').  But when we do the subtraction nu <du'/dx du'/dx> =  <nu du'/dx du'/dx> - nu d<u>/dx d<u>/dx , we however need nu specified since the second term is calculated from the mean field, And I don't think it's wise to store nu d<u>/dx as a field ... So maybe reading nu from json file is a better choice?

PR #1250, opened 2024-04-23
Feature/stress formulation, smagorinsky model

Shiyu-Sandy-Du:
Classical Smagorinsky model is added.

### DIFF
--- a/src/les/bcknd/cpu/smagorinsky_cpu.f90
+++ b/src/les/bcknd/cpu/smagorinsky_cpu.f90
@@ -109,9 +109,8 @@ contains
           s13 = 0.5_rp * (a13%x(i,1,1,e) + a13%x(i,1,1,e))
           s23 = 0.5_rp * (a23%x(i,1,1,e) + a23%x(i,1,1,e))
 
-          s_abs = s11*s11 + s22*s22 + s33*s33 + &
-                  s12*s12 + s13*s13 + s23*s23
-          s_abs = sqrt(2*s_abs)
+          s_abs = sqrt(2 * (s11*s11 + s22*s22 + s33*s33) + &
+                  4 * (s12*s12 + s13*s13 + s23*s23))
 
           nut%x(i,1,1,e) = c_s**2 * delta%x(i,1,1,e)**2 * s_abs
        end do
--- a/src/les/bcknd/cpu/smagorinsky_cpu.f90
+++ b/src/les/bcknd/cpu/smagorinsky_cpu.f90
@@ -1,0 +1,127 @@
+! Copyright (c) 2023, The Neko Authors
+! All rights reserved.
+!
+! Redistribution and use in source and binary forms, with or without
+! modification, are permitted provided that the following conditions
+! are met:
+!
+!   * Redistributions of source code must retain the above copyright
+!     notice, this list of conditions and the following disclaimer.
+!
+!   * Redistributions in binary form must reproduce the above
+!     copyright notice, this list of conditions and the following
+!     disclaimer in the documentation and/or other materials provided
+!     with the distribution.
+!
+!   * Neither the name of the authors nor the names of its
+!     contributors may be used to endorse or promote products derived
+!     from this software without specific prior written permission.
+!
+! THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+! "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+! LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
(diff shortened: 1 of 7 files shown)


src/wall_models/wall_model_fctry.f90
```diff
@@ -34,6 +34,7 @@
 module les_model_fctry
   use les_model, only : les_model_t
   use vreman, only : vreman_t
+  use smagorinsky, only : smagorinsky_t
   use dofmap, only : dofmap_t
   use coefs, only : coef_t
   use json_module, only : json_file
```
timofeymukha, line 40: This file should not be modified for this. I guess it's contents are wrong now, as it is a copy for the les factory, instead of a wall model factory. I probably forgot to push an update for it, sorry.

src/les/bcknd/cpu/smagorinsky_cpu.f90
```diff
@@ -0,0 +1,126 @@
...
+    call dudxyz (a12%x, u%x, coef%drdy, coef%dsdy, coef%dtdy, coef)
+    call dudxyz (a13%x, u%x, coef%drdz, coef%dsdz, coef%dtdz, coef)
+
+    call dudxyz (a21%x, v%x, coef%drdx, coef%dsdx, coef%dtdx, coef)
+    call dudxyz (a22%x, v%x, coef%drdy, coef%dsdy, coef%dtdy, coef)
+    call dudxyz (a23%x, v%x, coef%drdz, coef%dsdz, coef%dtdz, coef)
+
+    call dudxyz (a31%x, w%x, coef%drdx, coef%dsdx, coef%dtdx, coef)
+    call dudxyz (a32%x, w%x, coef%drdy, coef%dsdy, coef%dtdy, coef)
+    call dudxyz (a33%x, w%x, coef%drdz, coef%dsdz, coef%dtdz, coef)
```
timofeymukha, line 100: I think we can use the operator for computing the strain rate in operators.f90?

src/les/bcknd/cpu/vreman_cpu.f90
```diff
@@ -130,7 +130,8 @@ subroutine vreman_compute_cpu(t, tstep, coef, nut, delta, c)
           ! alpha_ij alpha_ij
           aijaij = beta11 + beta22 + beta33
 
-          nut%x(i,1,1,e) = c*delta%x(i,1,1,e) * sqrt(b_beta/(aijaij + NEKO_EPS))
+          nut%x(i,1,1,e) = c*delta%x(i,1,1,e)*delta%x(i,1,1,e) &
+                            * sqrt(b_beta/(aijaij + NEKO_EPS))
        end do
```
timofeymukha, line 135: Yeah, I see it now again, looking at the paper, good catch!

timofeymukha (changes requested):

src/les/bcknd/cpu/smagorinsky_cpu.f90
Shiyu-Sandy-Du: Yes maybe better. The code would look concise and we only need to allocate 6 fields instead of 9

src/wall_models/wall_model_fctry.f90
Shiyu-Sandy-Du: Oh yes exactly. Then how should I deal with this wrong file here?

timofeymukha: Just undo all changes to it so it is no longer in the PR.

src/les/bcknd/cpu/smagorinsky_cpu.f90
```diff
@@ -0,0 +1,107 @@
...
+    real(kind=rp), intent(in) :: t
+    integer, intent(in) :: tstep
+    type(coef_t), intent(in) :: coef
+    type(field_t), intent(inout) :: nut
+    type(field_t), intent(in) :: delta
+    real(kind=rp), intent(in) :: c_s
+    type(field_t), pointer :: u, v, w
+    ! double of the strain rate tensor
+    type(field_t), pointer :: s11_2, s22_2, s33_2, s12_2, s13_2, s23_2
+    real(kind=rp) :: s_abs
```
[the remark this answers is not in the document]
Shiyu-Sandy-Du: the value is double of the tensor here, so s11 etc. might be a bit confusing.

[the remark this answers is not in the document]
timofeymukha: Oh, I see what you mean, that's a bit tricky. We are not missing an extra 0.5 in the calculation of s_abs (since we are squaring the s_ij componenta), are we?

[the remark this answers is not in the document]
Shiyu-Sandy-Du: I make the compensation when calculating s_abs, so it's correct now.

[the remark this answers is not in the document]
timofeymukha: Alright I trust you :D

PR #1252, opened 2024-04-24
Expand ax_helm to allow a coupled compute

timofeymukha:
So here is an initial proposition for adding an interface to ax, which works on 3 vectors in a coupled manner. After our discussions, this follows the easiest path, which does not directly generalize to a different number of vectors, but perhaps this is OK for no to get the stress formulation in. The ax_helm factory now takes a parameter "formulation", which can be either "simple", meaning the simplified stress tensor, for "full", for the full tensor. Naturally, the existing fluid scheme only uses the simple one. 

The idea with this PR is mainly to settling on how to introduce the coupled ax_helm, so that we can start putting in GPU kernels into develop.


timofeymukha:
Is it possible to figure out what flint is complaining about?

src/math/ax.f90
```diff
@@ -43,6 +43,7 @@ module ax_product
   type, public, abstract :: ax_t
    contains
      procedure(ax_compute), nopass, deferred :: compute
+     procedure(ax_compute3), pass(this), deferred :: compute3
```
njansson, line 46: How about adding a generic compute here, such that we can keep it consistent within solvers etc?

timofeymukha: @njansson There is an issue. We can have a generic interface, but generic resolution relies on the shape of the arrays of the dummy arguments. So, we cannot do the usual trick of sending a 1D array to a dummy argument which is 4D. The compiler will say that there is no implementation of the generic found for the type of arguments passed.

timofeymukha: (and the solver types do use the trick)

njansson (commented):
Nice, but can we find a better name instead of `simple` ?

src/math/ax.f90
```diff
@@ -68,4 +69,33 @@ subroutine ax_compute(w, u, coef, msh, Xh)
...
+       type(mesh_t), intent(inout) :: msh
+       type(coef_t), intent(inout) :: coef
+       real(kind=rp), intent(inout) :: au(Xh%lx, Xh%ly, Xh%lz, msh%nelv)
+       real(kind=rp), intent(inout) :: av(Xh%lx, Xh%ly, Xh%lz, msh%nelv)
+       real(kind=rp), intent(inout) :: aw(Xh%lx, Xh%ly, Xh%lz, msh%nelv)
+       real(kind=rp), intent(inout) :: u(Xh%lx, Xh%ly, Xh%lz, msh%nelv)
+       real(kind=rp), intent(inout) :: v(Xh%lx, Xh%ly, Xh%lz, msh%nelv)
+       real(kind=rp), intent(inout) :: w(Xh%lx, Xh%ly, Xh%lz, msh%nelv)
+     end subroutine ax_compute3
+  end interface
```
njansson, line 99: I would suggest we call this `_vector` to be consistent with similar interface in Neko

src/math/bcknd/cpu/ax_helm_cpu.f90
```diff
@@ -1,4 +1,4 @@
-! Copyright (c) 2021, The Neko Authors
+! Copyright (c) 2024, The Neko Authors
```
njansson, line 1:
```suggestion
! Copyright (c) 2021-2024, The Neko Authors
```
There isn't enough changes I would say that warrant to see the file as new

src/math/ax_helm_full.f90
njansson, line 1: If this is the file from the stress branch, it's a bit bad to loose the commit history for it by trying to reinvent it here.

flinter_rc.yml
```diff
@@ -266,7 +266,6 @@ fortran-syntax:
       - 'while'
 
   ponctuations:
-    - ','
```
njansson, line 269: I  would argue that we still want to check ',' as a ponctuation

src/math/ax_helm_fctry.f90
```diff
@@ -32,36 +32,45 @@
...
     end if
 
-    if (NEKO_BCKND_SX .eq. 1) then
-       allocate(ax_helm_sx_t::Ax)
-    else if (NEKO_BCKND_XSMM .eq. 1) then
-       allocate(ax_helm_xsmm_t::Ax)
-    else if (NEKO_BCKND_DEVICE .eq. 1)then
-       allocate(ax_helm_device_t::Ax)
+    if (full_formulation) then
+      allocate(ax_helm_full_cpu_t::Ax)
```
njansson, line 63: Were need to thrown an error here unless we're on a backend that can use the CPU part

flinter_rc.yml
timofeymukha: We can do that, but then we need to go through the codebase and make changes to comply with that, mainly in array indexing. I can start with the stuff in axhelm_full here, but it is the tip of an iceberg.

src/math/ax_helm_full.f90
timofeymukha: Yeah, it is. I'm not sure if there is a way to make git see the connection, though. Do you know how one would do that?

flinter_rc.yml
timofeymukha: Not entirely convinced spaces actually help readability. Somehow, I find it clearer to see the mathematical expression when the stuff the math operators work on is condensed to a list of symbols with no spaces. But could be expression-dependent @timfelle what do you reckon?

```fortran
          u1 = wur(i, 1, 1) * drdx(i, 1, 1, e) + wus(i, 1, 1) * dsdx(i, 1, 1, e) &
                                               + wut(i, 1, 1) * dtdx(i, 1, 1, e)
          u2 = wur(i, 1, 1) * drdy(i, 1, 1, e) + wus(i, 1, 1) * dsdy(i, 1, 1, e) &
                                               + wut(i, 1, 1) * dtdy(i, 1, 1, e)
          u3 = wur(i, 1, 1) * drdz(i, 1, 1, e) + wus(i, 1, 1) * dsdz(i, 1, 1, e) &
                                               + wut(i, 1, 1) * dtdz(i, 1, 1, e)
```
contra
```fortran
          w1 = wwr(i,1,1) * drdx(i,1,1,e) + wws(i,1,1) * dsdx(i,1,1,e) &
                                          + wwt(i,1,1) * dtdx(i,1,1,e)
          w2 = wwr(i,1,1) * drdy(i,1,1,e) + wws(i,1,1) * dsdy(i,1,1,e) &
                                          + wwt(i,1,1) * dtdy(i,1,1,e)
          w3 = wwr(i,1,1) * drdz(i,1,1,e) + wws(i,1,1) * dsdz(i,1,1,e) &
                                          + wwt(i,1,1) * dtdz(i,1,1,e)
```

njansson: True, I tend to agree (and not be consistent my self) with having it condense for these kind of operations, and have it "fluffy" for function calls and other definitions.

timofeymukha: Can we tweak the flint rule to ignore the specific case of array indexing?

dauptaia: I am afraid It will be difficult within Flint current implementation, (not aware of the array indexing) sorry.

I will keep this feature in mind for our current project (tucan) which will replace flinter in the long. term and is able to contextualize the analysis.

timofeymukha: Thank you for the prompt reply!

timfelle: Yes it is really a tough one. In my opinion its not great either way. 4 way indexing is never pretty and as dauptaia mentioned, there is no distinction between function calls and array indexing. And i am still not a fan of using the indexing directly to flatten the arrays, but that have got performance implications so its important.

So yes as @njansson mention, there are no good solution, so I only plan on being consistent.

src/math/ax_helm_full.f90
njansson: git will not see the connection since it's a new file. Either we just have a place holder in this pr, merge the stress branch and move that formulation here (`git mv`) to preserve history

timfelle: One solution could be to start a branch on the stress branch, and revert all the changes not related to axhelm and then merge that branch into the branch of this PR. that way the history is preserved. 

But you have a commit reverting a lot of changes. The PR should sort that out as it looks for files with differences between the base and the PR not the accumulated files changes by the commits.

flinter_rc.yml
njansson: I think that will break flint?

src/fluid/bcknd/cpu/pnpn_res_cpu.f90
```diff
@@ -155,6 +155,8 @@ subroutine pnpn_vel_res_cpu_compute(Ax, u, v, w, u_res, v_res, w_res, &
     call Ax%compute(v_res%x, v%x, c_Xh, msh, Xh)
     call Ax%compute(w_res%x, w%x, c_Xh, msh, Xh)
 
+!    call Ax%compute3(u_res%x, v_res%x, w_res%x, u%x, v%x, w%x, c_Xh, msh, Xh)
+
```
timfelle, line 159:
```suggestion
```

src/fluid/fluid_pnpn.f90
```diff
@@ -150,7 +150,7 @@ subroutine fluid_pnpn_init(this, msh, lx, params, user, material_properties)
                           material_properties)
 
     ! Setup backend dependent Ax routines
-    call ax_helm_factory(this%ax)
+    call ax_helm_factory(this%ax, full_formulation=.false.)
```
timfelle, line 153:
```suggestion
    call ax_helm_factory(this%ax, full_formulation = .false.)
```

src/krylov/pc_hsmg.f90
```diff
@@ -183,7 +183,7 @@ subroutine hsmg_init(this, msh, Xh, coef, dof, gs_h, bclst, crs_pctype)
     call this%c_mg%init(this%gs_mg)
 
     ! Create backend specific Ax operator
-    call ax_helm_factory(this%ax)
+    call ax_helm_factory(this%ax, full_formulation=.false.)
```
timfelle, line 186:
```suggestion
    call ax_helm_factory(this%ax, full_formulation = .false.)
```

src/math/ax.f90
```diff
@@ -68,4 +69,33 @@ subroutine ax_compute(w, u, coef, msh, Xh)
      end subroutine ax_compute
   end interface
 
+  !> Abstract interface for computing\f$ Ax \f$ inside a Krylov method,
+  !! taking 3 vectors at the same time.
+  !! @param w Vector of size @a (lx,ly,lz,nelv).
+  !! @param u Vector of size @a (lx,ly,lz,nelv).
+  !! @param coef Coefficients.
+  !! @param msh Mesh.
+  !! @param Xh Function space \f$ X_h \f$.
```
timfelle, line 78: We are missing some parameter descriptions here.

flinter_rc.yml
timfelle: It should be fine, its just a list of regex rules being checked through python. I'll try it out and make a PR if it works.

src/math/ax_helm.f90
```diff
@@ -0,0 +1,68 @@
...
+
+  !> Matrix-vector product for a Helmholtz problem.
+  type, public, abstract, extends(ax_t) :: ax_helm_t
+   contains
+     !> Compute the product for 3 fields.
+     procedure, pass(this) :: compute_vector => ax_helm_compute_vector
+  end type ax_helm_t
+
+contains
+  subroutine ax_helm_compute_vector(this, au, av, aw, u, v, w, coef, msh, Xh)
```
timfelle, line 51: Add doxygen string

src/scalar/scalar_pnpn.f90
```diff
@@ -149,7 +149,7 @@ subroutine scalar_pnpn_init(this, msh, coef, gs, params, user, &
                           material_properties)
 
     ! Setup backend dependent Ax routines
-    call ax_helm_factory(this%ax)
+    call ax_helm_factory(this%ax, full_formulation=.false.)
```
timfelle, line 152:
```suggestion
    call ax_helm_factory(this%ax, full_formulation = .false.)
```

timfelle (changes requested):
In general looks good.

A general notes:
- Update the copyright now that we are touching these files.

flinter_rc.yml
timfelle: I have found a way to make flint happy about ,1 alone and not other things

njansson: > I have found a way to make flint happy about ,1 alone and not other things

Spill the beans!

timfelle: > > I have found a way to make flint happy about ,1 alone and not other things
> 
> Spill the beans!

Do the white-space manually `(\w)` expands to `([a-zA-Z0-1_])` so we can just tell it to match against `([a-zA-Z02-9_])` instead 😅

njansson: But that would still give poor scoring for `(i,1,1,1)` due to `i` right?

timfelle: No, i is before the `,` not after. it would degrade scores of `(i,1,1,e)` based on the `e`.

dauptaia: OMG, I created monsters...😅

timfelle: > OMG, I created monsters...😅

Yep 🤣 

Even better!!! `(,)(\w){2,}`
Allows `,` trailed by a single non word character. But disallows multi character words!!

njansson: > No, i is before the `,` not after. it would degrade scores of `(i,1,1,e)` based on the `e`.
> 
> Also any number starting with 1 such as 10 and 11 😅

Should be fine (^-^)v, since we only use 1, for indexing, unless we have a loop variable

njansson: > > OMG, I created monsters...😅
> 
> Yep 🤣
> 
> Even better!!! `(,)(\w){2,}` Allows `,` trailed by a single non word character. But disallows multi character words!!

🍻

src/math/ax_helm_full.f90
timofeymukha: OK, let's go for the placeholder options, undoing stuff seems complicated. :-)

src/fluid/bcknd/cpu/pnpn_res_cpu.f90
njansson: Should this comment be removed?

njansson (approved):
Let's get it in and such that we can iron out any issues together with the merge of the stress branch

src/fluid/bcknd/cpu/pnpn_res_cpu.f90
njansson: Ok, let's fix that later once the stress branch is in

PR #1253, opened 2024-04-24
Change strain rate to include the 1/2 factor

timofeymukha:
This changes the strain_rate operator to actually return the strain rate instead of its double (which I think I copied from Nek5000). The x2 convention is easy to forget about and now caused some issues with the implemention of models for the stress formulation branch.

On develop, the operator is not used.

### DIFF
--- a/src/math/operators.f90
+++ b/src/math/operators.f90
@@ -358,9 +358,9 @@ contains
        call device_cmult(s13_d, 0.5_rp, nelv*lxyz)
        call device_cmult(s23_d, 0.5_rp, nelv*lxyz)
     else
-       call cmult(s12_d, 0.5_rp, nelv*lxyz)
-       call cmult(s13_d, 0.5_rp, nelv*lxyz)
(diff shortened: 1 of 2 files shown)


timofeymukha:
@Shiyu-Sandy-Du Please double check that the math is right now :-).

timofeymukha:
> I think it is used in the lift/drag, albeit as an input argument.

You are right, somehow my search missed that. I will take a look

Shiyu-Sandy-Du:
The math is correct :D, and we need to change some stuffs in the feature/stress_formulation branch?

timofeymukha:
> I think it is used in the lift/drag, albeit as an input argument.

I think it boils down to the pointwise viscous drag computation, which needs the extra factor of 2 now. I just added that.

timofeymukha:
> The math is correct :D, and we need to change some stuffs in the feature/stress_formulation branch?

Yeah, we merge this eventually to the stress branch, and make adjustments there.

njansson (approved):

PR #1254, opened 2024-04-25
README with JSON-Fortran

MartinKarp:
To make life easier, removed some more advanced options from the readme and added how to compile json-fortran. 

Easier than looking into https://github.com/ExtremeFLOW/neko/blob/develop/.github/workflows/develop.yml for how to compile as I am doing now.

Not perfect, but I think this would be a significant improvement, in particular it's rather easy to copy-paste and get something up and running.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,11 @@ To build the project you will need: A Fortran compiler supporting the Fortran-08
 git clone --depth 1 https://github.com/ExtremeFLOW/json-fortran/
 cd json-fortran
 mkdir build && cd build
-cmake -DCMAKE_INSTALL_PREFIX=path/to/json-fortran_install -DUSE_GNU_INSTALL_CONVENTION=ON ..
+export JSON_INSTALL=/path/to/json-fortran_install
+cmake -DCMAKE_INSTALL_PREFIX=${JSON_INSTALL} -DUSE_GNU_INSTALL_CONVENTION=ON ..
 make -j4 && make install && cd ../../
-export PKG_CONFIG_PATH=${PKG_CONFIG_PATH}:path/to/json-fortran_install/lib/pkgconfig
-export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/path/to/json-fortran_install/lib/
+export PKG_CONFIG_PATH=${PKG_CONFIG_PATH}:${JSON_INSTALL}/lib/pkgconfig
+export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${JSON_INSTALL}/lib/
 ```
 
 A basic CPU version of Neko can then be installed according to the following
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ git clone https://github.com/ExtremeFLOW/neko
 Documentation for Neko is available at https://neko.cfd and most things related to the code, cases, and different features are described there. The documentation is always improving, in large part due to our active users and if something is missing or hard to understand, don't be afraid to create an issue or create a PR. It is a great way to help us improve and also to start getting involved in the project.
 
 ## Building the project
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,17 +10,34 @@ Neko is a portable framework for high-order spectral element flow simulations. W
 git clone https://github.com/ExtremeFLOW/neko
 ```
 
+## Documentation
+Documentation for Neko is available at https://neko.cfd and most things related to the code, cases, and different features are described there. The documentation is always improving, in large part due to our active users and if something is missing or hard to understand, don't be afraid to create an issue or create a PR. It is a great way to help us improve and also to start getting involved in the project.
```
timofeymukha, line 14: Maybe "discusson" instead of issues, since this is what we would like, right? Could make it a clickable link as well.

timofeymukha:
very nice!

README.md
MartinKarp: Yeah, thats better, also maybe I should write Pull requests instead of PRs since its for those new to the code.

README.md
```diff
@@ -10,17 +10,34 @@ Neko is a portable framework for high-order spectral element flow simulations. W
...
+
+```bash
+export JSON_INSTALL=/path/to/json-fortran_install # Where you want to install json-fortran
+```
+```bash
+git clone --depth 1 https://github.com/ExtremeFLOW/json-fortran/
+cd json-fortran
+mkdir build && cd build
+cmake -DCMAKE_INSTALL_PREFIX=${JSON_INSTALL} -DUSE_GNU_INSTALL_CONVENTION=ON ..
+make -j4 && make install && cd ../../
```
timfelle, line 27:
```suggestion
cmake -S json-fortran -B json-fortran/build -DCMAKE_INSTALL_PREFIX=${JSON_INSTALL} -DUSE_GNU_INSTALL_CONVENTION=ON ..
cmake --parallel --build json-fortran/build
cmake --install json-fortran/build
```
We might aswell suggest people to use modern syntax for cmake instead of the
legacy one.

timfelle (changes requested):

README.md
MartinKarp: Thats a good idea, ill make a few fixes when I get to the office.

README.md
```diff
@@ -10,17 +10,34 @@ Neko is a portable framework for high-order spectral element flow simulations. W
...
+
 ## Building the project
-To build the project you will need: A Fortran compiler supporting the Fortran-08 standard, a working MPI installation, JSON-Fortran, and BLAS/lapack. Optional dependencies are gslib and ParMETIS. We use automake to build the project. These instructions should work in general, but as the project is quickly developing, things might change.
+To build the project you will need: A Fortran compiler supporting the Fortran-08 standard, a working MPI installation, JSON-Fortran, and BLAS/lapack. Optional dependencies are gslib and ParMETIS. We use automake to build the project. These instructions should work in general, but as the project is quickly developing, things might change. While we assume MPI and BLAS are installed, if JSON-Fortran is not already available it can be cloned, installed, and the correct paths set with the following commands (Skip this step if you already have an installation of JSON-Fortran).
```
njansson, line 17:
```suggestion
To build the project you will need: A Fortran compiler supporting the Fortran-08 standard, a working MPI installation, JSON-Fortran, and BLAS/lapack. Optional dependencies are gslib and ParMETIS. We use autotools to build the project. These instructions should work in general, but as the project is quickly developing, things might change. While we assume MPI and BLAS are installed, if JSON-Fortran is not already available it can be cloned, installed, and the correct paths set with the following commands (Skip this step if you already have an installation of JSON-Fortran).
```

njansson (commented):
Looks good! We should not forget to backport this to 0.8 before tagging a release.

One thing to consider is how to deal with autotools and regen.sh . These are only needed if the source has been cloned from the repo.

MartinKarp:
Updated a little bit

MartinKarp:
@timfelle The new lines with more modern cmake command do not work for me it seems, when I do the conventional `make && make install` it works.

MartinKarp:
Seems it something do to with --parallel, now it works

timfelle:
I think i might have swapped them arround, parallel should apparently be after the other one.

njansson (approved):

timfelle (approved):

PR #1255, opened 2024-04-25
Various bugs related to BCs

MartinKarp:
Several BCs in Neko are currently not working as intended. This PR solves several of these issues.

# Fixes
- Correct application of dirichlet BCs when only an element vertex is touching the dirichlet boundary.
- Correct treatment of BCs in which the normal is involved (the normal of a corner vertex is ill defined)
- Correct initialization/finalization of the dong BC for GPUs - We really need a regression test for GPUs.


vbaconnet:
Yes this is what we had talked about! I can see if I have a small mesh to test this on.

src/fluid/fluid_scheme.f90
```diff
@@ -879,6 +879,14 @@ subroutine fluid_scheme_bc_apply_vel(this, t, tstep)
...
     call bc_list_apply_vector(this%bclst_vel,&
          this%u%x, this%v%x, this%w%x, this%dm_Xh%size(), t, tstep)
+    call this%gs_Xh%op(this%u,GS_OP_MIN)
+    call this%gs_Xh%op(this%v,GS_OP_MIN)
+    call this%gs_Xh%op(this%w,GS_OP_MIN)
+    call bc_list_apply_vector(this%bclst_vel,&
+         this%u%x, this%v%x, this%w%x, this%dm_Xh%size(), t, tstep)
+    call this%gs_Xh%op(this%u,GS_OP_MAX)
+    call this%gs_Xh%op(this%v,GS_OP_MAX)
+    call this%gs_Xh%op(this%w,GS_OP_MAX)
```
njansson, line 889: Would it be possible to only add a subset of the bclst? to reduce the work, and similarly for the gs, are these additional ops always needed, if not, can we possibly detect when they are, and have a if block around them?

src/bc/bc.f90
```diff
@@ -447,9 +490,9 @@ subroutine bc_finalize(this)
        call device_map(this%facet, this%facet_d, n)
 
        call device_memcpy(this%msk, this%msk_d, n, &
-                          HOST_TO_DEVICE, sync=.false.)
+                          HOST_TO_DEVICE, sync=.true.)
```
njansson, line 493:
```suggestion
                          HOST_TO_DEVICE, sync=.false.)
```

This is perfectly safe to have `.false.` since it is guaranteed to be synced by the call afterwards.

MartinKarp:
I have updated this PR. Please have a look when you have time!

src/fluid/fluid_scheme.f90
MartinKarp: I think this is the only bulletproof way of doing this, but we can give it a thought I guess. For now I would keep it this way.

src/bc/bc.f90
```diff
@@ -437,19 +446,58 @@ subroutine bc_finalize(this)
...
+       end if
+       msk_c = 0
+       do i = 1, this%dof%size()
+          if (test_field%x(i,1,1,1) .gt. 0.5) then
+             msk_c = msk_c + 1
+          end if
+       end do
+       !Allocate new mask
+       deallocate(this%msk)
+       allocate(this%msk(0:msk_c))
```
vbaconnet, line 477: so if I understand correctly in the case of two neighboring elements that have different BCs and shared points, if say you are building the mask for element 1, this will add the shared points with element 2 to the mask as well? Or perhaps I misunderstood... :)

Like if you have two facets and set the values 0 and 1 for the 'v' boundary like this:

1 -- 'v' -- 1 0 -- 'on' -- 0

after gs

1 -- 'v' -- 1 1 -- 'on' -- 0

So the point that was set to 1 for the 'on' boundary will be added to the new mask, correct?

vbaconnet: just checking I understand correctly, otherwise lgtm

MartinKarp: Yeah, exactly

MartinKarp: The different special cases here are to find these points which are on the boundary but not on the facet

njansson:
> @njansson can we merge this? We should probably try to get it in as soon as possible, if device-aware mpi doesn't work maybe we can let that be an exception or something.

I suggest we wait a bit, I'll have a go at the min/max kernels next week

MartinKarp:
How is this going?

src/fluid/fluid_scheme_incompressible.f90
```diff
@@ -502,6 +502,15 @@ subroutine fluid_scheme_bc_apply_vel(this, t, tstep, strong)
...
          this%u%x, this%v%x, this%w%x, this%dm_Xh%size(), t, tstep, strong)
+    call this%gs_Xh%op(this%u,GS_OP_MIN)
+    call this%gs_Xh%op(this%v,GS_OP_MIN)
+    call this%gs_Xh%op(this%w,GS_OP_MIN)
+
+    call this%bcs_vel%apply_vector(&
+         this%u%x, this%v%x, this%w%x, this%dm_Xh%size(), t, tstep, strong)
+    call this%gs_Xh%op(this%u,GS_OP_MAX)
+    call this%gs_Xh%op(this%v,GS_OP_MAX)
+    call this%gs_Xh%op(this%w,GS_OP_MAX)
```
timofeymukha, line 513:
```suggestion
    call this%gs_Xh%op(this%u, GS_OP_MIN)
    call this%gs_Xh%op(this%v, GS_OP_MIN)
    call this%gs_Xh%op(this%w, GS_OP_MIN)

    call this%bcs_vel%apply_vector(&
         this%u%x, this%v%x, this%w%x, this%dm_Xh%size(), t, tstep, strong)
    call this%gs_Xh%op(this%u, GS_OP_MAX)
    call this%gs_Xh%op(this%v, GS_OP_MAX)
    call this%gs_Xh%op(this%w, GS_OP_MAX)
```

timofeymukha (approved):

MartinKarp:
I have tested on a cylinder mesh with elements with only vertices on the outflow. I would suggest to merge it now.

vbaconnet (approved):
Tested it on my side and it works as well.

src/fluid/fluid_scheme_incompressible.f90
```diff
@@ -501,7 +501,16 @@ subroutine fluid_scheme_bc_apply_vel(this, t, tstep, strong)
...
+      this%u%x, this%v%x, this%w%x, this%dm_Xh%size(), t, tstep, strong)
+    call this%gs_Xh%op(this%u, GS_OP_MIN)
+    call this%gs_Xh%op(this%v, GS_OP_MIN)
+    call this%gs_Xh%op(this%w, GS_OP_MIN)
+
+    call this%bcs_vel%apply_vector(&
+      this%u%x, this%v%x, this%w%x, this%dm_Xh%size(), t, tstep, strong)
+    call this%gs_Xh%op(this%u, GS_OP_MAX)
+    call this%gs_Xh%op(this%v, GS_OP_MAX)
+    call this%gs_Xh%op(this%w, GS_OP_MAX)
```
njansson, line 513: Would it be possible to have a flag in `bc` or some other place, that could indicate if the min/max communication is necessary?

njansson (approved):
🍻

PR #1258, opened 2024-04-27
Add div and grad operators

timofeymukha:
Adds a strong gradient and a divergence operator. Both are wrappers around calls to dudxyz. Further optimizations (I guess mostly for the device) can be done in the future if necessary.

### DIFF
--- a/src/math/operators.f90
+++ b/src/math/operators.f90
@@ -47,15 +47,16 @@ module operators
   use space, only : space_t
   use coefs, only : coef_t
   use field, only : field_t
-  use math, only : glsum, cmult, add2, cadd
+  use math, only : glsum, cmult, add2, cadd, copy
   use device, only : c_ptr, device_get_ptr
-  use device_math, only : device_add2, device_cmult
+  use device_math, only : device_add2, device_cmult, device_copy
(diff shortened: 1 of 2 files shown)


njansson (approved):

src/math/operators.f90
```diff
@@ -85,6 +86,54 @@ subroutine dudxyz (du, u, dr, ds, dt, coef)
...
+    type(c_ptr) :: res_d
+
+    res_d = device_get_ptr(res)
+
+    call neko_scratch_registry%request_field(work, ind)
+
+    ! Get dux / dx
+    call dudxyz(work%x, ux, coef%drdx, coef%dsdx, coef%dtdx, coef)
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_copy(res_d, work%x_d, work%size())
```
adperezm, line 112: Cant we avoid this by using res in the fitst dudxyz?

timofeymukha: Yeah, it seems so, good catch!

adperezm (approved):
I had checked this PR already so I can proceed to just approve.

PR #1260, opened 2024-04-28
HDF5 Support

njansson:
This add support for reading/writing (parallel) HDF5 files. The new format has been enabled as an optional format for checkpoints, but the backend can read/write most of our output lists.

### DIFF
--- a/src/field/field_series.f90
+++ b/src/field/field_series.f90
@@ -49,6 +49,11 @@ module field_series
      procedure, pass(this) :: size => field_series_size
   end type field_series_t
 
+  !> field_series_ptr_t, To easily obtain a pointer to a field series
+  type, public :: field_series_ptr_t
+     type(field_series_t), pointer :: fs => null()
+  end type field_series_ptr_t
+
 contains
 
   !> Initialize a field series of length @a len for a field @a f
--- a/src/field/field_series.f90
+++ b/src/field/field_series.f90
@@ -51,7 +51,7 @@ module field_series
 
   !> field_series_ptr_t, To easily obtain a pointer to a field series
   type, public :: field_series_ptr_t
-     type(field_series_t), pointer :: fs => null()
+     type(field_series_t), pointer :: ptr => null()
   end type field_series_ptr_t
 
 contains
--- a/src/io/hdf5_file.F90
+++ b/src/io/hdf5_file.F90
@@ -1,0 +1,540 @@
+! Copyright (c) 2024, The Neko Authors
+! All rights reserved.
+!
+! Redistribution and use in source and binary forms, with or without
+! modification, are permitted provided that the following conditions
+! are met:
+!
+!   * Redistributions of source code must retain the above copyright
+!     notice, this list of conditions and the following disclaimer.
+!
+!   * Redistributions in binary form must reproduce the above
+!     copyright notice, this list of conditions and the following
+!     disclaimer in the documentation and/or other materials provided
+!     with the distribution.
+!
+!   * Neither the name of the authors nor the names of its
+!     contributors may be used to endorse or promote products derived
+!     from this software without specific prior written permission.
+!
+! THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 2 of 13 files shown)


adperezm (approved):
Nice!

timfelle:
Looks great.
One minor thing though, could it be useful to have the short dummy code first in the `#if` statement?

It might make it easier to read as one can quickly understand what the if statement actually contains.

``` .f90
#ifndef HAVE_HDF5
	! Print the error message
#else
	! implementation
#endif
```

src/field/field_series.f90
```diff
@@ -49,6 +49,11 @@ module field_series
      procedure, pass(this) :: size => field_series_size
   end type field_series_t
 
+  !> field_series_ptr_t, To easily obtain a pointer to a field series
+  type, public :: field_series_ptr_t
+     type(field_series_t), pointer :: fs => null()
+  end type field_series_ptr_t
```
timofeymukha, line 55: Can we call the wrapped variable `ptr` for consistency with other such types?

src/io/hdf5_file.F90
```diff
@@ -0,0 +1,540 @@
...
+       if (data%size() .gt. 0) then
+          allocate(fp(data%size()))
+       
+          dof => data%items(1)%ptr%dof
+          msh => data%items(1)%ptr%msh
+
+          do i = 1, data%size()
+             fp(i)%ptr => data%items(i)%ptr
+          end do
+       else
```
timofeymukha, line 422: Can we use any of the wrapper routines in the field_list here?

src/io/hdf5_file.F90
```diff
@@ -0,0 +1,540 @@
...
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+!
+!> HDF5 file format
+module hdf5_file
+  use num_types
+  use generic_file
+  use checkpoint, only : chkp_t
+  use utils
```
timofeymukha, line 38: Let's not give up on `only` for these.

timofeymukha:
Very happy to see hdf5 come in :-).

src/io/hdf5_file.F90
njansson: In principle yes if the `field_ptr_t(:)` is change to a `field_list_t`, however since we also have `field_series_ptr_t` data, the code will be less consistent between the different data types.

timofeymukha: A little bit confused. Here we type-select the `data` to be a `field_list_t`, right? So we could do like
```
dof => data%dof(1)
```
etc, no?

Anyway, this is not super important, just nice to skip the items%ptr stuff when possible.

njansson: But that would only work for `field_list_t`, not for the general case with checkpoints where we also have lagged arrays, which are not `field_list_t`

timofeymukha: Yes, I am just referring to this small snippet inside `type is (field_list_t)`

njansson: Unless we want to duplicate the reader/writer parts, we can't use `assign` here since the left side is not a `field_list_t`.

Another, probably have overlooked in during the overhaul of `field_list_t` is the pointer assignment from a return value of a function. Yes it's nice, but I'm not sure if compilers are there yet (or we are doing something wrong), since this is what have caused a lot of issues in the past (especially with Cray Fortran), e.g. why we don't use `%array()` in stack.

timofeymukha: Yeah, I guess only the `dof` and `msh` wrapper could be used here in 416-417. If one cannot return a pointer from a function that we are indeed in trouble. It would be very unfortunate if such basic stuff is not supported.

njansson: Hence, we we have a lot of comments in certain parts of gather-scatter

timofeymukha (approved):

PR #1261, opened 2024-04-29
Slight update to field_dirichlet bcs

timofeymukha:
My main goal here is to reduce the number of additional components in the scheme types necessary to add the user field_dirichlet bcs. Currently we need the bc itself, a field list for the dummy fields in the bc,  a bc_list to store the field_dirichlet bcs themselves, and an update procedure pointer to the user routine.

To trim this down, I put most of these things inside the field_dirichlet types themselves. The only extra thing still in the scalar_scheme is the bc_list. I did not quite dare to add a bc_list to a type to store.. a pointer to itself. However, for the field_dirichlet_vector I could store the bc_list, it just contains the 3 field_dirichlet_bcs that are stored as components inside field_dirichlet_vector.

For the fluid, I use the lists in velocity field_dirichlet to also store the field and bc for the pressure. I also use it to store the `update` pointer.

Also adds some extra routines to the field_list wrapper.

As a note, the way things are handled for the field_dirichlet_vector is a bit scary, because we always add the dummy fields for u,v,w,p, *even when they are not initialized* due to the fact that the bc for them is not asked for in the case file.  So, if you point to those uninitialized fields in the user routine, things will crash.  I have not addressed this here.

### DIFF
--- a/src/fluid/fluid_scheme.f90
+++ b/src/fluid/fluid_scheme.f90
@@ -109,8 +109,8 @@ module fluid_scheme
      class(bc_t), allocatable :: bc_inflow !< Dirichlet inflow for velocity
 
      ! Attributes for field dirichlet BCs
-     type(field_dirichlet_vector_t) :: user_bc_vel   !< User-computed Dirichlet velocity condition
-     type(field_dirichlet_t) :: user_bc_prs   !< User-computed Dirichlet pressure condition
+     type(field_dirichlet_vector_t) :: user_field_bc_vel   !< User-computed Dirichlet velocity condition
+     type(field_dirichlet_t) :: user_field_bc_prs   !< User-computed Dirichlet pressure condition
      type(dirichlet_t) :: bc_prs               !< Dirichlet pressure condition
      type(dong_outflow_t) :: bc_dong           !< Dong outflow condition
      type(symmetry_t) :: bc_sym                !< Symmetry plane for velocity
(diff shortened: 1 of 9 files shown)


njansson (commented):
Looks good, the only minor thing is the `user_bc_` name in `fluid_scheme`, not it kind of reads like field_dirichlet is the only user provided bc. I could be wrong, haven't followed the latest bc work too closely.

timofeymukha:
> Looks good, the only minor thing is the `user_bc_` name in `fluid_scheme`, not it kind of reads like field_dirichlet is the only user provided bc. I could be wrong, haven't followed the latest bc work too closely.

There is also the pointwise user inflow bc, which hides in `bc_inflow`. My idea was that it will be removed anyway, but it seems we want to keep those, so you have a point. I can change to `user_field_bc...` to distinguish them for now, perhaps.

src/fluid/fluid_scheme.f90
```diff
@@ -397,71 +392,73 @@ subroutine fluid_scheme_init_common(this, msh, lx, params, scheme, user, &
...
-         this%bc_field_vel%field_dirichlet_w%field_bc
-    this%field_dirichlet_fields%items(4)%ptr => &
-         this%bc_field_prs%field_bc
-
-    call bc_list_init(this%field_dirichlet_bcs, size=4)
-    call bc_list_add(this%field_dirichlet_bcs, this%bc_field_vel%field_dirichlet_u)
-    call bc_list_add(this%field_dirichlet_bcs, this%bc_field_vel%field_dirichlet_v)
-    call bc_list_add(this%field_dirichlet_bcs, this%bc_field_vel%field_dirichlet_w)
+
+    ! Note, some of these are potentially not initialized !
```
vbaconnet, line 446: Yea... This is dangerous and required quite a lot of explanations in the user file documentation, which probably will not be read. So I expect people running into runtime errors for this. Hopefully the example is enough to make people avoid the issue for now. At some point we should move away from this and only pass down whatever is initialized...

vbaconnet (approved):
I think it looks much cleaner now, thanks!! As I said at some point we should only pass down the fields that are initialized to the user_dirichlet_update, but for now I guess this is fine?

src/fluid/fluid_scheme.f90
vbaconnet: Sorry I meant to do this as a simple comment lol, you can mark as resolved whenever :)

timofeymukha: One option would be to just always init and make all 4 fields available. Then one does not have to do all the d_vel_u/d_vel_v stuff, just a "user_dirichlet". But it is a tradeoff of course.

njansson (approved):

src/fluid/fluid_scheme.f90
njansson: If possible, can we avoid init extra fields?

timofeymukha: Yeah, in the current implementation they are not init.

PR #1262, opened 2024-04-29
Remove legacy scalar user bc

timofeymukha:
Basically the title. There seems to be no sense to keep the old pointwise bc as it can be easily replaced by the new one, which also supports GPUs.


njansson (commented):
Fully agree with what has been stated, but removing the point wise apply also removes the possibility of running interactive with pyneko, and have the bc provided by e.g. a notebook

timofeymukha:
> Fully agree with what has been stated, but removing the point wise apply also removes the possibility of running interactive with pyneko, and have the bc provided by e.g. a notebook

Hmm, would it be too far-fetched to say that we could in the future handle most scalar functions with expression parsing from the case file?

njansson:
> > Fully agree with what has been stated, but removing the point wise apply also removes the possibility of running interactive with pyneko, and have the bc provided by e.g. a notebook
> 
> Hmm, would it be too far-fetched to say that we could in the future handle most scalar functions with expression parsing from the case file?

Difficult to say without having some kind of prototype in the pipeline

timofeymukha:
> > > Fully agree with what has been stated, but removing the point wise apply also removes the possibility of running interactive with pyneko, and have the bc provided by e.g. a notebook
> > 
> > 
> > Hmm, would it be too far-fetched to say that we could in the future handle most scalar functions with expression parsing from the case file?
> 
> Difficult to say without having some kind of prototype in the pipeline

Yeah, that's true. Also, a bit difficult to predict with pyneko as there are 0 users, and it is still not really released, right? I'm a bit torn here.

pschlatt1:
if it does not hurt, I would leave simple (as in known from Nek5000) ways of interacting with the field in the code. It is a research code, so people may want to do quick and dirty things at times.

PR #1264, opened 2024-04-30
Factory-generated boundary conditions with new case file interface [Preview]

timofeymukha:
Dear all, I think I've accumulated enough changes to launch this preview of the new bc handling. It is now almost fully implemented for the scalar (excluding the user field Dirichlet), but not yet for the fluid, therefore the *don't merge*. I do want the CI to run through, to see what complaints, I guess. 

# Major changes

## The `bc_t` type

* Adds an abstract interface for the constructor and the corresponding deferred routine `init`. Takes `json_file` and `coef_t`. Some derived types now use this to parse the JSON they need, e.g. the Dong condition, the Blasius, etc.
* Some descendants also implement an `init_from_componets` in order to construct without JSON.
* Add an abstract interface for finalization and the corresponding deferred routine `finalize`. Most types just call the `finalize_base` in `bc_t`, but `symmetry` and `non_normal` run their logic for determining the normal direction here.
* Added generic apply for scalar and vector, which looks at whether we are on the device and calls the corresponding routine. Before, one had to use a `bc_list` to do that. I haven't actually used this yet, so maybe these will be removed later on.
* In the `step` routines for the solvers, we need to be able to distinguish between weak and strong bcs. So `bc_t` no has an additional component `strong = .true.` to mark a bc as strong or weak.

## The `bc_list_t` type
* Moved to a separate file.
* All the subroutines like `bc_list_add` are now TBPs, which simplifies `use` statements quite a bit, and is generally the way to go.
* I mimic as much as possible the implementation of the `field_list_t`. Some differences remain though, e.g. for the `bc_list` we track both the capacity of the list and the number of allocated items. We need to think a bit more about this.
* The `apply_` routines are augmented with an optional `strong` parameter, which will lead to only applying strong or weak bcs. As a result, we can use a single list to store all the bcs and then call apply with either `strong=.true.` or `strong=.false.` when needed in the `step` routines.

## The `zero_dirichlet_t` type
This is a rebranding of what used to the `noslip_wall`. The bc just assigns both vectors and scalars to zero. What I realized is that we have a lot of places in the code where we use a zero-valued Dirichlet for the purpose of marking boundaries for the linear solver routines. Using `zero_dirichlet_t` helps with clarity there.

## The `bc_factory` subroutine
* A factory that inits, marks, and finalizes a bc.
* The marking is for a single zone, with the index taken from the JSON. (more on the new structure JSON below).
* Currently set up to work with the scalar. What I realized is that construction of user-drives bcs requires knowledge of the `scheme`. So perhaps, what will happen is that each `scheme` has its own factory.
* Does not support the `field_dirichlet_t` yet.

## The `scalar_scheme_t` type
* Uses a single `bc_list` called `bcs` to store the boundary conditions and constructs them with a factory. This results in quite a bit of cleanup here and there.
* No longer uses `bclabels`.
* Instead, the keyword `boundary_conditions` is used, which is an array of JSON objects, similar to how we do sources and simcomps. Each object has the `type` and `zone_index` keyword + other keywords needed, for that type.
* The RB examples are updated to demonstrate the new layout.
```json
            "boundary_conditions": [
            {
                "type": "dirichlet",
                "zone_index": 5,
                "value": 1.0
            },
            {
                "type": "dirichlet",
                "zone_index": 6,
                "value": 0.0
            },
            {
                "type": "user_pointwise"
            }
            ],
```

# Remaining work
* I don't like that with the current structure we have a lot of "no-op" routines in bc types, that don't implement either the scalar or vector apply. Or sometimes neither, as in the case of e.g. symmetry. One could split all types int scalar and vector, but mixed types would still fall out of this classification. Some mixed types, however, a just wrappers around several simple types as components. I have to look at this more.
* The big work is to accommodate the new bc structure in `fluid_scheme`. I don't think it will become as "clean" as scalar because there are quite a bit of auxiliary bcs that are defined and will probably remain as additional components of the type. 
* Support for field_dirichlet should be added. I will wait for the small cleanup of those I've done in PR #1261 to be merged first.


src/bc/bc.f90
```diff
@@ -155,6 +162,24 @@ subroutine bc_destructor(this)
...
+
+  abstract interface
+     !> Constructor
+     subroutine bc_constructor(this, coef, json)
+       import :: bc_t, coef_t, json_file
+       class(bc_t), intent(inout), target :: this
+       type(coef_t), intent(in) :: coef
+       type(json_file), intent(inout) ::json
+     end subroutine bc_constructor
+  end interface
```
timfelle, line 181: Maybe move this above destructor (just a very minor organization thing). But its
a bit more natural to read the constructor and then the destructor.

src/bc/bc.f90
```diff
@@ -244,6 +262,82 @@ subroutine bc_free_base(this)
 
   end subroutine bc_free_base
 
+  subroutine bc_apply_vector_generic(this, x, y, z, n, t, tstep)
```
timfelle, line 265: We are missing some documentation of this function

src/bc/bc.f90
```diff
@@ -244,6 +262,82 @@ subroutine bc_free_base(this)
...
+       else if (present(tstep)) then
+          call this%apply_vector(x, y, z, n, tstep=tstep)
+       else
+          call this%apply_vector(x, y, z, n)
+       end if
+    end if
+
+  end subroutine bc_apply_vector_generic
+
+  subroutine bc_apply_scalar_generic(this, x, n, t, tstep)
```
timfelle, line 306: Documentation here as well

src/bc/bcknd/device/cuda/zero_dirichlet.cu
timfelle, line 1: Remember to update documentation and copyright notice when chaning these codes

src/bc/bcknd/device/cuda/zero_dirichlet.cu
```diff
@@ -32,43 +32,43 @@
...
-#include "no_slip_wall_kernel.h"
+#include "zero_dirichlet_kernel.h"
 #include <device/device_config.h>
 #include <device/cuda/check.h>
 
 extern "C" {
 
-  /** 
+  /**
    * Fortran wrapper for device no-slip wall apply vector
```
timfelle, line 42: Remember to update documentation and copyright notice when chaning these codes

src/bc/bcknd/device/cuda/zero_dirichlet_kernel.h
```diff
@@ -32,17 +32,17 @@
  POSSIBILITY OF SUCH DAMAGE.
 */
 
-#ifndef __BC_NO_SLIP_WALL_KERNEL__
-#define __BC_NO_SLIP_WALL_KERNEL__
+#ifndef __BC_ZERO_DIRICHLET_KERNEL__
+#define __BC_ZERO_DIRICHLET_KERNEL__
 
 /**
  * Device kernel for scalar apply for a no-slip wall conditon
```
timfelle, line 39: Remember to update documentation and copyright notice when chaning these codes

timfelle:
Should we align the new BC's with the submodule structure we introduced recently before we merge this in?

njansson:
> Should we align the new BC's with the submodule structure we introduced recently before we merge this in?

I don't think we need submodules for bc, unless we have some redundant factories that are in a separate file.

njansson:
> > I don't think we need submodules for bc, unless we have some redundant factories that are in a separate file.
> 
> Well this PR introduces a bc factory.

submodules it is. I just feel that for proper derived types, like dirichlet_t, it's good to have it in a own module

timfelle:
> submodules it is. I just feel that for proper derived types, like dirichlet_t, it's good to have it in a own module

I am not sure i follow. Do you suggest that `direchlet_t` should not be part of the general BC list, but instead handled manually on its own?

timfelle:
> I mean like for ksp_t, base definition in one module, and factory in a submodule. But for the different flavours like cg and gmres, full modules are fine

I see, yes that makes sense. It would just be to have a factory for the abstract BC type in a submodule and then all the specializations are regular modules.

Sounds good to me.

PR #1265, opened 2024-05-02
Add Sigma and Smagorinsky model, fix the Vreman model.

timofeymukha:
This PR pulls in the Sigma and Smagorinsky models implemented on the stress formulation branch, developed by @Shiyu-Sandy-Du  and @muenschm . Also fixes a bug in the Vreman model (a **2 was missing).


muenschm (approved):

muenschm (approved):

muenschm (approved):

muenschm (approved):

muenschm (commented):
What does the standard specify here? To my probably outdated knowledge: acos -> REAL(4) and dacos -> REAL(8). In case this is still true, we should **not** change this.

Shiyu-Sandy-Du:
> What does the standard specify here? To my probably outdated knowledge: acos -> REAL(4) and dacos -> REAL(8). In case this is still true, we should **not** change this.

acos(X) returns the same data type as X. See https://gcc.gnu.org/onlinedocs/gfortran/ACOS.html

muenschm (approved):

Shiyu-Sandy-Du (approved):

timofeymukha:
Thanks for fixing the precision issues. Someone else but us 3 should look at the PR before we merge, though since you two are the authors :-). I will make a request.

src/les/bcknd/cpu/sigma_cpu.f90
```diff
@@ -0,0 +1,248 @@
...
+! LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+!
+!> Implements the CPU kernel for the `sigma_t` type.
+!> Following Nicoud et al. "Using singular values to build a
+!> subgrid-scale model for large-eddy simulations"
+!> https://doi.org/10.1063/1.3623274
```
njansson, line 36:
```suggestion
!> Following Nicoud et al. "Using singular values to build a
!! subgrid-scale model for large-eddy simulations"
!! https://doi.org/10.1063/1.3623274
```

njansson (approved):

src/les/bcknd/cpu/sigma_cpu.f90
```diff
@@ -0,0 +1,248 @@
...
+                            sigG33*sigG12*sigG12)
+
+                 ! G is symmetric semi-definite positive matrix:
+                 ! the invariants have to be larger-equal zero
+                 ! which is obtained via forcing
+                 Invariant1=max(Invariant1, 0.0_rp)
+                 Invariant2=max(Invariant2, 0.0_rp)
+                 Invariant3=max(Invariant3, 0.0_rp)
+
+!                ! compute the following angles from the invariants
```
timfelle, line 182:
```suggestion
                 ! compute the following angles from the invariants
```

src/les/bcknd/cpu/sigma_cpu.f90
timfelle, line 1: Please add spaces around the assignment operators.

src/les/smagorinsky.f90
```diff
@@ -0,0 +1,127 @@
...
+  use json_module, only : json_file
+  use utils, only : neko_error
+  use neko_config, only : NEKO_BCKND_DEVICE
+  use smagorinsky_cpu, only : smagorinsky_compute_cpu
+  use coefs, only : coef_t
+  implicit none
+  private
+
+  !> Implements the smagorinsky LES model.
+  !! @note Reference DOI: Smagorinsky (1963)
```
timfelle, line 50: Did you mean to include the DOI and forgot?

timfelle:
Well I dont know much about the method but it seem valid enough as is.

Two things though,
- Do you have an example that can be submitted along with the PR.
- Could you please update the documentation with some words on this method and how it can be used?

Shiyu-Sandy-Du:
> Well I dont know much about the method but it seem valid enough as is.
> 
> Two things though,
> 
> * Do you have an example that can be submitted along with the PR.
> * Could you please update the documentation with some words on this method and how it can be used?

For the first question: It could be tested with any example as long as it solves the Navier-Stokes. Now the tested one is tgv. Should we update the json file for tgv example? (I don't think we should change anything in the examples at this stage because the sgs model needs stress formulation to test, which is not included in this branch and we performed tests on that branch. As a result, sgs stress will not be coupled in this branch.)

timofeymukha:
Yes, an example will have to wait until we actually have stress formulation. With the documentation, it is a bit the same -- since we cannot actually use the models right now, we cannot add instructions. But I pinky-promise that we will add it later :-).

timfelle (approved):

PR #1266, opened 2024-05-02
Warn for ranks with 0 elements

timofeymukha:


src/mesh/mesh.f90
```diff
@@ -188,6 +188,10 @@ subroutine mesh_init_nelv(this, gdim, nelv)
 
     call this%free()
 
+    if (this%nelv < 1) then
+       call neko_warning("One of the MPI ranks has 0 mesh elements!")
```
MartinKarp, line 192: Could we perhaps also print the number of the rank?

MartinKarp (changes requested):

src/mesh/mesh.f90
```diff
@@ -211,6 +215,9 @@ subroutine mesh_init_dist(this, gdim, dist)
     call this%free()
 
     this%nelv = dist%num_local()
+    if (this%nelv < 1) then
+       call neko_warning("One of the MPI ranks has 0 mesh elements!")
+    end if
```
njansson, line 220:
```suggestion
    if (this%nelv < 1) then
       call neko_warning("Rank " // pe_rank // " ranks has zero mesh elements!")
    end if
```

@MartinKarp How about something like this?

timofeymukha: I didnt know you could concat numbers into a string like that!

adperezm:
Quick question, In which case could this happen? I only see this happening if nelv < mpiranks, or?


Am I missing something?

timofeymukha:
> Quick question, In which case could this happen? I only see this happening if global_nelv < mpiranks, or?
> 
> Am I missing something?

Yes, like if you run the RB example on >25 ranks. So, in principle, only you attempt to run a small problem on a huge number of ranks.

src/mesh/mesh.f90
timofeymukha: @njansson OK, it seems one actiually cannot :-)

MartinKarp (approved):

timofeymukha:
Wait, some rogue files got in.. let me fix that :D

timofeymukha:
There we go. I think VS auto-commits all changes when I push some button..

njansson (approved):

PR #1268, opened 2024-05-06
Dynamic Smagorinsky model

Shiyu-Sandy-Du:
two things are done:

1. Dynamic Smagorinsky model with only temporal smoothening and elementwise test filter.
2. In the old code, w is acquired as u, now it's fixed.


src/les/bcknd/cpu/dynamic_smagorinsky_cpu.f90
```diff
@@ -0,0 +1,316 @@
...
+       if (.not. (den%x(i,1,1,1) .eq. 0)) then
+          c_dyn%x(i,1,1,1) = 0.5_rp * (num%x(i,1,1,1)/den%x(i,1,1,1))
+       end if
+       c_dyn%x(i,1,1,1) = max(c_dyn%x(i,1,1,1),0.0_rp)
+       nut%x(i,1,1,1) = c_dyn%x(i,1,1,1) * delta%x(i,1,1,1)**2 * s_abs%x(i,1,1,1)
+    end do
+
+    call coef%gs_h%op(nut%x, nut%dof%size(), GS_OP_ADD)
+    call col2(nut%x, coef%mult, nut%dof%size())
+
```
timofeymukha, line 128: Let's not do that for now. I know I added it in Vreman, but is actually unclear if this is the right way to match things at the element boundaries. E.g. if one should match the actually flux, and not just nut. We will have to look into that stuff at the future.

src/les/bcknd/cpu/dynamic_smagorinsky_cpu.f90
```diff
@@ -0,0 +1,316 @@
...
+    type(field_t), intent(in) :: delta
+    integer, intent(in) :: n
+    integer, intent(inout) :: nelv
+
+    real(kind=rp), dimension(n) :: fs11, fs22, fs33, fs12, fs13, fs23, fs_abs
+    real(kind=rp) :: delta_ratio2 !test- to grid- filter ratio, squared
+    integer :: i
+    real(kind=rp) :: delta2
+    
+    delta_ratio2 = ((test_filter%nx-1)/(test_filter%nt-1))**2
```
timofeymukha, line 213: Could you maybe leave a comment here. How are the two deltas approximated here?

src/les/bcknd/cpu/dynamic_smagorinsky_cpu.f90
```diff
@@ -0,0 +1,316 @@
...
+                     mij(2)%x(i,1,1,1)*mij(2)%x(i,1,1,1) + &
+                     mij(3)%x(i,1,1,1)*mij(3)%x(i,1,1,1) + &
+                     2.0_rp*(mij(4)%x(i,1,1,1)*mij(4)%x(i,1,1,1) + &
+                     mij(5)%x(i,1,1,1)*mij(5)%x(i,1,1,1) + &
+                     mij(6)%x(i,1,1,1)*mij(6)%x(i,1,1,1))
+    end do
+    
+    ! running average over time
+    call add3s2(num%x, num%x, num_curr, alpha, 1.0_rp-alpha, n)
+    call add3s2(den%x, den%x, den_curr, alpha, 1.0_rp-alpha, n)
```
timofeymukha, line 311: Interesting. Maybe some of these details of the computational procedure can be put into the documentation of the dynamic_smagorinsky type?

On a more general note, none of us added anything to the docs for all the SGS models, this will be our debt :P.

timofeymukha:
@Shiyu-Sandy-Du Thanks for the update! Any comments on the comments above not marked as resolved?

Shiyu-Sandy-Du:
> @Shiyu-Sandy-Du Thanks for the update! Any comments on the comments above not marked as resolved?

I have left all comments there, and they should be marked as "pending" now...

src/les/bcknd/cpu/dynamic_smagorinsky_cpu.f90
```diff
@@ -0,0 +1,316 @@
...
+    do i=1, u%dof%size()
+       if (.not. (den%x(i,1,1,1) .eq. 0)) then
+          c_dyn%x(i,1,1,1) = 0.5_rp * (num%x(i,1,1,1)/den%x(i,1,1,1))
+       end if
+       c_dyn%x(i,1,1,1) = max(c_dyn%x(i,1,1,1),0.0_rp)
+       nut%x(i,1,1,1) = c_dyn%x(i,1,1,1) * delta%x(i,1,1,1)**2 * s_abs%x(i,1,1,1)
+    end do
+
+    call coef%gs_h%op(nut%x, nut%dof%size(), GS_OP_ADD)
+    call col2(nut%x, coef%mult, nut%dof%size())
```
timofeymukha, line 129:
```suggestion
```

PR #1271, opened 2024-05-07
Remove check for more arguments

MartinKarp:
Might be controversial, but enabling the input for more arguments than the case file might be good, especially if one wants to vary a lot of different parameters (especially between ranks) and make ensemble runs without changing the case file. 

I don't think it hurts to relax this check.


adperezm:
You really like command line arguments :P.

I think it is fair, and I guess that taking care of those arguments is responsibility of the user, but I still think it would be good to report them back just to be on the safe side.

If you do not agree, then ignoring this comment is fine. I do not really mind either way :D

njansson:
How about instead adding proper command line arguments, as suggested by @adperezm? Something along the line of `-v` for detailed version information (compiler flags etc), `-c  jsonfile.case` to run a case. There are of course several options, but would need some discussion on good naming for the switches.

MartinKarp:
Reviving this, would it be ok to allow more command line arguments? Im not sure I see how it hurts for those that don't need them, but it opens up of rumor customization for people who experiment with certain things (like me).

timfelle (approved):

Mr-2c:
does it make sense to have a "pro" mode for the code (maybe a command line argument) that disables checks and allows for more command line arguments and environmental variables? just so someone new and somewhat clueless does not mess things up unintentionally

timfelle:
> does it make sense to have a "pro" mode for the code (maybe a command line argument) that disables checks and allows for more command line arguments and environmental variables? just so someone new and somewhat clueless does not mess things up unintentionally

Well that is usually done with the case files. We have a lot of flexibility with default values for most arguments. Is there anything in particular you want to control that you cannot?

Mr-2c:
> > does it make sense to have a "pro" mode for the code (maybe a command line argument) that disables checks and allows for more command line arguments and environmental variables? just so someone new and somewhat clueless does not mess things up unintentionally
> 
> Well that is usually done with the case files. We have a lot of flexibility with default values for most arguments. Is there anything in particular you want to control that you cannot?

I mostly meant it as a compromise to what Martin was suggesting, to have a pro mode which disables the checks and maybe even a flag of --ignore-case-file or something and works with environmental variables or additional command line arguments, etc., just so a clueless user does not mess things up unintentionally.

timofeymukha:
I think the idea is that clueless users should not use the user file, and so makeneko becomes irrelevant. Once you have your own .f90 you can do all sorts of crazy things with no guards, so you are essentially already in pro mode. So I would not do any additional splits.

timfelle:
> I think the idea is that clueless users should not use the user file, and so makeneko becomes irrelevant. Once you have your own .f90 you can do all sorts of crazy things with no guards, so you are essentially already in pro mode. So I would not do any additional splits.

This change is not to `makeneko` it was to `neko` itself. Meaning now you can give multiple commandline arguments which you can handle if you need. Currently we don't use any. But it might be a good idea to move the current environment variables we read to be switches for the commandline.

timofeymukha:
Oh, I got confused then, sorry!

MartinKarp:
Yeah, my idea for this was primarily to allow for this type of of configuration. As Tim writes, I think adding the environment variables as command line arguments would be good in the future. The removal of the check for more arguments I think will not impact a new user very much. If one accidentally adds a bunch they will not do anything unless one has tinkered with the code at the moment.

PR #1272, opened 2024-05-07
Move away from re2/rea internal mesh zones

vbaconnet:
This PR gets rid of the internal BCs from the re2/rea formats. The idea is that the "classic" labels w,v,o,on,sym,shl (and its variations) are transformed into labeled zones with indices 1...6. 

For example, if the mesh contains boundaries with `W`, `v` and `on`, we will now have to specify `"boundary_types": ["w", "v","","","on"]`.

Note that if one has several boundaries set internally as one of those re2 labels, they will all be marked in the same labeled zone. For example, if I have a channel with a lower and upper wall they will both be marked as labeled zone 1, and one will not be able to modify them separately (say, if I want to top wall to become an open boundary instead).  

TODO before undrafting:
- [x] Do the same logic with rea files

NOTE: this has not been tested with meshes that have both internal labeled zones AND internal re2 labels like v,w,o etc.


vbaconnet:
After some time, this PR is ready for review :) I have only tested with `re2` and `rea` files that only have internal labels like `v`, `w`, etc. Not a combination of labeled zones and labels (as mentioned in the PR description). 

But I think the logic should work!

src/io/rea_file.f90
```diff
@@ -268,6 +283,7 @@ subroutine rea_file_read(this, data)
        read(9,*)
        read(9,*)
        if (.not. read_bcs) then ! Mark zones in the mesh
+          call neko_log%message("Reading boundary conditions")
```
timfelle, line 286:
```suggestion
          call neko_log%message("Reading boundary conditions", neko_log_debug)
```
Maybe assign the debug level messaging?

vbaconnet: sure, no strong feelings about it

timfelle: It just seemed more like a message one would need when debugging more than when doing a production run of the code.

timofeymukha:
Victor, do you still map to zones 1...6 based on type, or is it just appending stuff as we discussed in the meeting?

timofeymukha:
Appears to look good I think!

>> Note that if one has several boundaries set internally as one of those re2 labels, they will all be marked in the same labeled zone.

This is not a big problem I think. But generally, I think all the `case` logic based on the bc type is somewhat redundant now. You anyway do the same thing, besides for doing the addition to the zone number based on the type. It seems one could just throw away the `case` and always simply add a new zone. But this can be done later if needed.

Do we have a case with mixed internal/labled zones for Nek? To check if it works :-).

timofeymukha (approved):

MartinKarp:
I think this looks good! As for the output, when reading a re2/rea file if I understand correctly, at the first occurrence of a hardcoded BC it is printed to stdout which label it is assigned to?

vbaconnet:
> at the first occurrence of a hardcoded BC it is printed to stdout which label it is assigned to?

Yes exactly, it will assign the labeled zone based on order of appearance in the re2/rea file

timfelle (approved):

PR #1278, opened 2024-05-14
Refactor/advection

timfelle:
- Simplified the factory.
- Reduced usage to only whats used.
- Split advection into multiple files.


timofeymukha:
Very nice. What I am wondering, now that we are restructuring a bit, is whether we should have this in fluid and the different advection types in bcknd, moving forward. It is used in scalar as well, I suppose. I think we could promote it to its own directory in src. Something for you guys to consider.

timofeymukha (approved):

src/Makefile.am
```diff
@@ -197,6 +197,8 @@ neko_fortran_SOURCES = \
 	fluid/flow_ic.f90\
 	fluid/advection.f90\
 	fluid/advection_fctry.f90\
+    fluid/bcknd/advection/adv_dealias.f90\
+    fluid/bcknd/advection/adv_no_dealias.f90\
 	fluid/bcknd/cpu/pnpn_res_cpu.f90\
```
njansson, line 202:
```suggestion
	fluid/advection_fctry.f90\
        fluid/bcknd/advection/adv_dealias.f90\
        fluid/bcknd/advection/adv_no_dealias.f90\
	fluid/bcknd/cpu/pnpn_res_cpu.f90\
```

njansson (approved):

PR #1280, opened 2024-05-15
Fix/rocm_version

timfelle:
Attempt to fix the github workflows. HIP seem to be not found


timofeymukha:
Or how does this work, if one changes the CI yaml file does it run the updated yaml file when doing CI for the PR...? :D

timfelle:
> Or how does this work, if one changes the CI yaml file does it run the updated yaml file when doing CI for the PR...? :D



> Or how does this work, if one changes the CI yaml file does it run the updated yaml file when doing CI for the PR...? :D

Yes, the CI is updated to the stuff defined here. However, it seem i did not catch the actual issue..

njansson (approved):

PR #1295, opened 2024-05-30
"Rework CI for linting and use new linting in main CI"

timfelle:
This pull request includes the following changes:

- Rework of the CI for linting, including separating linting into its own file, performing checks for the repository and all changed files individually, and adding the ability to set a minimum score when calling the action.

- Using the new linting in the main CI workflow.

- Clean up of the workflow.

These changes improve the linting process and make it more clear what can be done to improve it.

As an additional bonus, we now cancel old workflows for a PR when a new commit is submitted. which should help reduce resource usage and speedup return for the valid runs.

PR #1298, opened 2024-05-31
Complete rework of the CI pipelines

timfelle:
By introducing actions specific for external dependencies and separated workflow files, maintenance should be simpler. Additionally we can better utilize caches of external dependencies to save resources.

Depends on: #1295
Requires the following repository level changes:

- Branch requirements should be assigned to the completion job for both main and develop.
- Repository variables should be set:
-- `FLINT_GLOBAL_MINIMUM`: score for repository level flint.
-- `FLINT_CHANGED_MINIMUM`: score requirement for modified files.
-- `JSON_FORTRAN_VERSION`: The JSON-Fortran version we should test against.
-- `PFUNIT_VERSION`: The version of pFunit we should test against.


timofeymukha:
Nice, the develop.yml was indeed a bit of a monster.

njansson (approved):

njansson:
Great stuff!

timofeymukha (approved):

PR #1301, opened 2024-06-07
Update error handling of actions

timfelle:
This pull request updates the error handling of actions to improve the overall reliability and stability of the code. The changes include better error messages and improved error handling logic.

Please do not manually merge this PR. Update the following repository level changes on github:

- Branch requirements should be assigned to the completion step for both main and develop.
	- The completion job should be the only requirement set `develop / Develop PR Ready` since that job checks all the others. This allow us to extend the CI pipeline later, without modifying the branch protection rules.

EDIT: The previously mentioned repository variables does not support PR's submitted from forks. Those should be removed again.

Depends on #1303

### DIFF
--- a/.github/actions/setup_json-fortran/action.yml
+++ b/.github/actions/setup_json-fortran/action.yml
@@ -44,6 +44,36 @@ outputs:
 runs:
   using: composite
   steps:
+    - name: Check the environment
+      env:
+        INSTALL_DIR: ${{ inputs.install-dir }}
+        WORKING_DIR: ${{ inputs.working-dir }}
+        OS: ${{ inputs.os }}
+        COMPILER: ${{ inputs.compiler }}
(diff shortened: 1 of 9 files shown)


timfelle:
I have added the don't merge label so we can make sure it actually works before you merge it in.

timfelle:
@njansson Alright, we are down to just the macos issues. I don't see what would cause the tests to fail. Do we run with a different version of pfunit?

I have removed the repository variables again. They do not propagate to pull requests opened from a fork, which make them kinda useless in this case. Instead i managed to use a top-level environment block in the workflow which should also make it simple to update if needed.

timfelle:
MacOS issues are now fixed as per #1303

njansson (approved):

timofeymukha (approved):

PR #1302, opened 2024-06-11
Fix CI

timfelle:
It seems a line was accedentally not migrated to the new actions for Json-Fortran.


timfelle:
Cannot complete from external fork..

PR #1304, opened 2024-06-11
The OIFS Scheme

Kamyarmvn:
The following changes relates to the implementation of the OIFS scheme:
1. Adding the OIFS advection derived type in fluid.
2. Adding the makeoifs rhs_maker in common. 
3. adding the higher order time interpolator for the convecting field.
4. changes in fluid_scheme and fluid_pnpn.
5. changes in scalar_scheme and scalar_pnpn.
6. adding the Runge_kutta operator, conv_fst_3d, and set_convect_new in math.
7. changes in simulation and case files. 
Note: There is also an implementation of conv_fst_3d available that calculates the local gradients similar to conv1 by using nested loops. However, the current implementation showed better performance so I uploaded this version.

### DIFF
--- a/examples/tgv/results/bdry0.nek5000
+++ b/examples/tgv/results/bdry0.nek5000
@@ -1,0 +1,4 @@
+filetemplate:         bdry%01d.f%05d
+firsttimestep:     0
+numtimesteps:     1
+type: binary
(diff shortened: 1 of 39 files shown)


njansson (changes requested):
Great stuff! Much needed functionality. 

However, the PR contains several binary files, would it be possible to reissue the PR without these files? (such that we don't end up with them in the history)

Kamyarmvn:
> Great stuff! Much needed functionality.
> 
> However, the PR contains several binary files, would it be possible to reissue the PR without these files? (such that we don't end up with them in the history)

Thank you for your feedback. I removed the binary files and will open a new request again.

PR #1308, opened 2024-06-12
Initial wall model implementation

timofeymukha:
Cherry-pick from stress_formulation.

Implements the standard json-based factory pattern for wall models.
- A BAC for wall models.
- A factory routine.
- Two wall models: Spalding's law and a rough log-law.

The wall model sampling is index-based, waiting for better times with more robust probes. The same index is used for the entire wall.


src/wall_models/spalding.f90
```diff
@@ -0,0 +1,232 @@
...
+      this%tau_z(i) = -utau**2 * wi / magu
+    end do
+
+  end subroutine spalding_compute
+
+  !> Newton solver for the algebraic equation defined by the law.
+  !! @param u The velocity value.
+  !! @param y The wall-normal distance.
+  !! @param guess Initial guess.
+  function solve(this, u,  y, guess) result(utau)
```
MartinKarp, line 186: Perhaps this should be lifted out for other uses in the future, but for now I guess it is fine to have it here for this specific case

src/wall_models/wall_model.f90
```diff
@@ -0,0 +1,318 @@
...
+    call this%n_z%free()
+  end subroutine wall_model_free_base
+
+  subroutine wall_model_find_points(this)
+    class(wall_model_t), intent(inout) :: this
+    integer :: n_nodes, fid, idx(4), i, linear
+    real(kind=rp) :: normal(3), p(3), x, y, z, xw, yw, zw, dot
+    real(kind=rp) :: hmin, hmax
+
+    n_nodes = this%msk(1)
```
MartinKarp, line 226: I remember we had some discussion about this, but do we want this one to also to start at 0? Or did we want to store the size outside of the mask.

src/wall_models/wall_model.f90
```diff
@@ -0,0 +1,318 @@
...
+
+  !> Base abstract type for wall-stress models for wall-modelled LES.
+  type, abstract, public :: wall_model_t
+     !> SEM coefficients.
+     type(coef_t), pointer :: coef => null()
+     !> Map of degrees of freedom.
+     type(dofmap_t), pointer :: dof => null()
+     !> The boundary condition mask. First element holds the array size!
+     integer, pointer :: msk(:) => null()
+     !> The boundary condition facet ids. First element holds the array size!
```
MartinKarp, line 60: First element in the array might be good to add. I immediately thought of the spectral elements.

src/wall_models/spalding.f90
```diff
@@ -0,0 +1,232 @@
...
+  use neko_config, only : NEKO_BCKND_DEVICE
+  use wall_model, only : wall_model_t
+  use field_registry, only : neko_field_registry
+  use json_utils, only : json_get_or_default
+  use logger, only : neko_log, NEKO_LOG_DEBUG
+
+  implicit none
+  private
+
+  !> Wall model based on Spalding's law of the wall
```
MartinKarp, line 50: While this one is famous, might be good with a ref.

src/wall_models/rough_log_law.f90
```diff
@@ -0,0 +1,180 @@
...
+  use dofmap, only : dofmap_t
+  use coefs, only : coef_t
+  use neko_config, only : NEKO_BCKND_DEVICE
+  use wall_model, only : wall_model_t
+  use field_registry, only : neko_field_registry
+  use json_utils, only : json_get_or_default, json_get
+  implicit none
+  private
+
+  !> Wall model based on the log-law for a rough wall.
```
MartinKarp, line 48: A ref might be good here too, but I am not an expert.

src/wall_models/wall_model.f90
```diff
@@ -0,0 +1,318 @@
...
+                         this%ind_e(i))
+
+
+       ! vector from the sampling point to the wall
+       p(1) = x - xw
+       p(2) = y - yw
+       p(3) = z - zw
+
+       ! project on the normal direction to get h
+       this%h%x(i) = p(1)*normal(1) + p(2)*normal(2) + p(3)*normal(3)
```
MartinKarp, line 294: Should we add something about how much the direction of `p` differs from the normal direction? Just thinking if someone has significantly deformed elements.

MartinKarp (commented):
Just some thoughts, looks really good I think.

src/wall_models/spalding.f90
timofeymukha: I am generally hoping that we will in the future manage to avoid newton solvers altogether for these models. We are working with a colleague to find some explicit expressions :-). If we fail, I think we just ditch Spalding for a better and explicit alternative, recently proposed by Sagaut's group.

src/wall_models/wall_model.f90
timofeymukha: In general, I would really rather store it outside. I didn't find a way for it to start at 0, which would be better here. Do you have an idea for how to do that?

src/wall_models/rough_log_law.f90
timofeymukha: There is probably no definitive reference here, I think we can just put in the formula.

src/wall_models/wall_model.f90
MartinKarp: I think something like `this%msk(0:mask%(0)) => msk` might work

src/wall_models/wall_model.f90
MartinKarp: Yes, if it differs a lot maybe a warning or so

src/wall_models/wall_model.f90
timofeymukha: I will give it a shot!

timofeymukha: Seems to have worked!

src/wall_models/wall_model.f90
```diff
@@ -0,0 +1,317 @@
...
+
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+      call device_memcpy(this%h%x, this%h%x_d, n_nodes, HOST_TO_DEVICE,&
+                         sync=.false.)
+      call device_memcpy(this%n_x%x, this%n_x%x_d, n_nodes, HOST_TO_DEVICE, &
+                         sync=.false.)
+      call device_memcpy(this%n_y%x, this%n_y%x_d, n_nodes, HOST_TO_DEVICE, &
+                         sync=.false.)
+      call device_memcpy(this%n_z%x, this%n_z%x_d, n_nodes, HOST_TO_DEVICE, &
+                         sync=.false.)
```
njansson, line 313:
```suggestion
                         sync=.true.)
```

Last gpu command, change to true as a failsafe

src/wall_models/wall_model_fctry.f90
```diff
@@ -0,0 +1,95 @@
+
```
njansson, line 1:
```suggestion
```

njansson (approved):

src/common/utils.f90
```diff
@@ -198,4 +199,19 @@ subroutine neko_warning(warning_msg)
     write(*,*) '*** WARNING: ', warning_msg,' ***'
   end subroutine neko_warning
 
+  !> Concatenate an array of strings into one string with array items
+  !! separated by spaces.
+  !! @param array The array of strings.
+  function concat_string_array(array) result(result)
```
timfelle, line 205: Maybe we could have an optional seperator input variable?

src/wall_models/wall_model.f90
```diff
@@ -0,0 +1,317 @@
...
+    call this%h%free()
+    call this%n_x%free()
+    call this%n_y%free()
+    call this%n_z%free()
+  end subroutine wall_model_free_base
+
+  subroutine wall_model_find_points(this)
+    class(wall_model_t), intent(inout) :: this
+    integer :: n_nodes, fid, idx(4), i, linear
+    real(kind=rp) :: normal(3), p(3), x, y, z, xw, yw, zw, dot
```
timfelle, line 223: `dot` seem to be unused?
```suggestion
    real(kind=rp) :: normal(3), p(3), x, y, z, xw, yw, zw
```

src/wall_models/wall_model_fctry.f90
```diff
@@ -0,0 +1,94 @@
...
+    call json_get(json, "model", type_name)
+
+    if (trim(type_name) .eq. "spalding") then
+       allocate(spalding_t::object)
+    else if (trim(type_name) .eq. "rough_log_law") then
+       allocate(rough_log_law_t::object)
+    else
+       call neko_error("Unknown wall model type: " // trim(type_name) // &
+          trim(type_name) // ".  Known types are: "  // type_string)
```
timfelle, line 85: For just a few types this is great. The seperator i sugested earlier
might help to format the known types by linebreaks or similar later on.

src/.depends
```diff
@@ -283,5 +283,9 @@ les/elementwise_filter.o : les/elementwise_filter.f90 math/tensor.o les/bcknd/cp
 les/bcknd/cpu/elementwise_filter_cpu.o : les/bcknd/cpu/elementwise_filter_cpu.f90 math/mxm_wrapper.o math/tensor.o math/matrix.o sem/speclib.o math/math.o config/num_types.o 
 les/sigma.o : les/sigma.f90 sem/coef.o les/bcknd/cpu/sigma_cpu.o config/neko_config.o common/utils.o common/json_utils.o sem/dofmap.o les/les_model.o field/field.o config/num_types.o 
 les/bcknd/cpu/sigma_cpu.o : les/bcknd/cpu/sigma_cpu.f90 sem/coef.o math/operators.o field/field.o field/field_registry.o field/scratch_registry.o math/math.o field/field_list.o config/num_types.o 
+wall_models/rough_log_law.o : wall_models/rough_log_law.f90 common/json_utils.o field/field_registry.o wall_models/wall_model.o config/neko_config.o sem/coef.o sem/dofmap.o config/num_types.o field/field.o 
+wall_models/spalding.o : wall_models/spalding.f90 common/log.o common/json_utils.o field/field_registry.o wall_models/wall_model.o config/neko_config.o sem/coef.o sem/dofmap.o config/num_types.o field/field.o
```
timfelle, line 287: Runing makedepend adds a few extra dependencies.


```suggestion
wall_models/rough_log_law.o : wall_models/rough_log_law.f90 common/utils.o common/json_utils.o field/field_registry.o wall_models/wall_model.o config/neko_config.o sem/coef.o sem/dofmap.o config/num_types.o field/field.o 
wall_models/spalding.o : wall_models/spalding.f90 common/utils.o common/log.o common/json_utils.o field/field_registry.o wall_models/wall_model.o config/neko_config.o sem/coef.o sem/dofmap.o config/num_types.o field/field.o
```

```suggestion
 ```

timfelle (approved):
Seems good to me. But a few random and "food for thought" comments.

src/wall_models/wall_model.f90
timofeymukha: @MartinKarp I added a comparison of distanced, instead of directly looking at the angle. I guess it is a relevant proxy in this context. Take a look :-). Once we have probes, we will sample directly along the normal and this will stop being an issue.

src/wall_models/wall_model_fctry.f90
timofeymukha: This is a good suggestion. But now I wonder, what is the symbol for new lines. Adding \n to the message just writes it verbatim :P.

timofeymukha: Found NEW_LINE('A') but not sure if compiler-specific. @njansson ?

timofeymukha: CI is happy with it in #1204 and Cray has it. Should be fine I suppose.

timofeymukha: @timfelle Updated to have a sperator, which is optionally also prepended. Now outputting a nice list of types with items starting with `- `

PR #1310, opened 2024-06-13
The OIFS Scheme

Kamyarmvn:
I fixed the bug related to the initialization of the scalar_pnpn and applied minor changes to the code. 

I also noticed that when running the Rayleigh Benard example, regardless of the scheme, there is an error in free case.


timofeymukha:
I will close and reopen the old pr

PR #1314, opened 2024-06-19
Update the hemi example to explciit bc setting

timofeymukha:
- Adds boundary_types
- Add the source re2
- Add a run.sh, which launches rea2nbin and then neko..


njansson (commented):
Not sure it is the best way to remove the nmsh file. Especially since the script is not general enough such that it can run in all environments.

I would suggest that we leave the nmsh file, but also include the re2 and the script as an example on how the mesh is generated.

timofeymukha:
> Not sure it is the best way to remove the nmsh file. Especially since the script is not general enough such that it can run in all environments.
> 
> I would suggest that we leave the nmsh file, but also include the re2 and the script as an example on how the mesh is generated.

Yes, sounds good. I restored the nmsh

njansson (approved):
A suggestion would be to mention the raw re2 format in the readme, and refer users to the script for instructions on how to generate the neko mesh

examples/hemi/hemi.case
```diff
@@ -40,6 +40,7 @@
             "absolute_tolerance": 1e-7,
             "max_iterations": 800
         },
+        "boundary_labels": ["sym", "w", "o", "v"],
```
timfelle, line 43:
```suggestion
        "boundary_labels": ["sym", "w", "v", "o"],
```
It would seem you swapped the flow direction.

timfelle (changes requested):

examples/hemi/hemi.case
vbaconnet:
```suggestion
        "boundary_types": ["sym", "w", "o", "v"],
```

This is what I get:
image

It looks okay to me, if we follow the indexing as output by `rea2nbin` no?

timfelle: Well if you run the simulation on develop and on the PR the flow is reversed on
the PR. The labels might be right but the simulation is not.

timfelle: But it seems the issue is fixed by your suggestion to change `boundary_labels`
to `boundary_types`

timofeymukha: Ah yes, jeez. My bad, I just put the wrong keyword it seems.

timfelle (approved):

PR #1317, opened 2024-06-24
Feature/simcomp-precompute

timfelle:
Introduction of interface to run preprocessing through simulation components.


njansson (approved):

src/simulation_components/simulation_component.f90
```diff
@@ -62,15 +64,21 @@ module simulation_component
...
      !> The common constructor using a JSON dictionary.
      procedure(simulation_component_init), pass(this), deferred :: init
      !> Destructor.
      procedure(simulation_component_free), pass(this), deferred :: free
+     !> The preprocessing function to be executed during the run.
+     procedure, pass(this) :: preprocess_
      !> The main function to be executed during the run.
-     procedure(simulation_component_compute), pass(this), deferred :: compute_
+     procedure, pass(this) :: compute_
   end type simulation_component_t
```
timofeymukha, line 82: Why use dummies instead of `deferred` routines? So that one can have only one of them implemented?

timfelle: Exactly.
We should probably add a nooverride on the others maybe?

timofeymukha: I see. So, is the vision, that each simcomp either does something in preprocess, or in compute, or in both? Or should the user be given some control in the case when it is not both? E.g. the vorticity, will only be able to run after the fluid in this design, or should we be able to control for it to run in the preprocess step via the user file?

Now that I think of it, if fluid and scalar become simcomps then this latter point can be just controlled by the simcomp execution order, and only simcomps that do something both in pre and compute benefit from having 2 procedures. So probably no need for extra user control.

timfelle: Well I'll let it be up to the user to define things in either preprocess or compute as they see fit, especially with PR #1185 coming along.
In general i see no point in limiting the user if they wish to do something in one or both of them. The `preprocessing_control` have also been added to allow direct control of each stage individually.

Well the question is not whether there should be a general way to swap or copy the `compute` and `preprocessing` for all simulation componments, I am quite sure that would not be useful. But yes one could think about adding a switch to the `vorticity_t` and allow the user to choose to export vorticity before and after the simulation update if you find that to be useful, I don't think I see why it would be? Maybe the field exporter could use that, if one wish to examine some values before and after an update?

timofeymukha: Yes, for the user one things are clear I think. I think I also agree that the swapping is probably not very useful for most cases already now and will be less so in the future.

timofeymukha (approved):

Issue #706, opened 2022-11-07
Generic math interface


njansson:
Add a generic math interface for all kinds of backend.

njansson:
The generic interface prevented inline of the small leaf routines in `math.f90`, thus gave a quite large performance hit.
Remove from v0.5.0 (https://github.com/ExtremeFLOW/neko/projects/21).

timofeymukha:
@njansson As a small comment, we mostly used `cpu` and not `host` as far as naming goes so far, like in `bcknd` folders.

timofeymukha:
@njansson I wonder if we can also use assumed shape arrays in `math`, so that one does not have to specify the size in stuff like `rzero`. I guess the size is always the whole array. I just had a very nasty bug, where I passed a value, which was to high, and the array was a part of a type. So it went over to other attributes in memory and zeroed them out :P. Was pretty damn hard to find.

timofeymukha:
@njansson How about a poor man's generic interface, which is just
```fortran
subroutine rzero(x, n)

if (NEKO_BCKND_DEVICE .eq. 1) then
   res_d = device_get_ptr(res)
   call device_rzero(res_d, n)
else
   call cpu_rzero(res, n)
end
end
```

This should not affect performance I suppose since this is exactly what we do all over the code? It would still clean up a lot of if-statements!

njansson:
> @njansson How about a poor man's generic interface, which is just
> 
> ```fortran
> subroutine rzero(x, n)
> 
> if (NEKO_BCKND_DEVICE .eq. 1) then
>    x_d = device_get_ptr(x)
>    call device_rzero(x_d, n)
> else
>    call cpu_rzero(x, n)
> end
> end
> ```
> 
> This should not affect performance I suppose since this is exactly what we do all over the code? It would still clean up a lot of if-statements!

Since `device_**` calls are quite expensive due to its overhead vs compute time, the additional pointer lookup adds a bit too much I would say. Furthermore, for CPU backends it the if statement will introduce the same issue with inlining as the generic interfaces :( This can of course be solved by writing out the loops where we would like inlining (which also allow for fusing). Then we add the generic interface without too much performance loss across the backends, and also add openmp for cpus.

njansson:
Maybe a bit clearer,  I would suggest we take the approach of `feature/omp` for the main part of our current `math` usage. Add the generic interface, which would then only be used in "less" performance critical part of the code. This would also be a great improvement for the user when writing user files

PR #1319, opened 2024-06-26
Feature/math_field

timfelle:
Introduction of a `field_math` library to capture the cpu and device operations.

This is related to the old issue #706.

I did attempt the generic interface which should automatically select the cpu, field or device operation based on input types. However, due to our implicit flattening of arrays this did not work, without introducting cpu operations for 1, 2, 3 and 4 dimensions.

Additionally I used the new library to hide some of the `if (NEKO_BCKND_DEVICE .eq. 1)` statements and cleanup the code for the brinkman sourceterm and the mean field.


njansson (commented):
Nice, but one concern is how this would affect inlining, that was what killed #709

timfelle:
> Nice, but one concern is how this would affect inlining, that was what killed #709

Well since this is just to avoid us manually writting something like the example bellow, I would not expect any differences in terms of inlining. But sure, there might be some compilers that only look 1 function deep to determine inlining. I am not exactly sure how they determine that.

```fortran
	call device_add2(.....)
else
	call add2(.....)
end if
```

njansson:
> > Nice, but one concern is how this would affect inlining, that was what killed #709
> 
> Well since this is just to avoid us manually writting something like the example bellow, I would not expect any differences in terms of inlining. But sure, there might be some compilers that only look 1 function deep to determine inlining. I am not exactly sure how they determine that.
> 
> ```fortran
> if (NEKO_BCKND_DEVICE .eq. 1) then
> 	call device_add2(.....)
> else
> 	call add2(.....)
> end if
> ```

In the past the issue was more inline at the call site, eg we often have a loop calling multiple `add2`, since math is rather pure, these would be inlined and the loops (from both `add2`) fused. Would be good to check before we merge this at least

timfelle:
> In the past the issue was more inline at the call site, eg we often have a loop calling multiple `add2`, since math is rather pure, these would be inlined and the loops (from both `add2`) fused. Would be good to check before we merge this at least

Ahh sure thing. Yes i would not expect kernel fusing to function as well with this.
I still think it can be worth merging, but we should probably write a note that it's only recommended for single operations and not for building large complexes that are performance critical.

timofeymukha:
Yes, this would certainly be a most welcome addition if the performance issues can be solved. I think there are quite a bit single-statement math calls in the code, so if fusion is the only problem, this would still be quite useful.

njansson:
For user experience this is of course great. I would propose that we have something like this or even a generic idea like #709 to also cover none field data, and for performance we do as in the OpenMP branch, where we avoid calling math routines in performance critical parts (more or less entire core), and let the math stuff be more user facing.

timfelle:
> For user experience this is of course great. I would propose that we have something like this or even a generic idea like #709 to also cover none field data, and for performance we do as in the OpenMP branch, where we avoid calling math routines in performance critical parts (more or less entire core), and let the math stuff be more user facing.
> 
> Edit: apparently already mentioned here: [#706 (comment)](https://github.com/ExtremeFLOW/neko/issues/706#issuecomment-2081631615)
> 
> This would allow us to offer OpenMP parallelisation on CPUs for those who want (e.g. A64FX or Grace-Grace) users

Essentially one should be able to do a generic interface which wraps up `math`, `device_math` and `field_math` modules. The only reason that is not part of this PR is the implicit casts from 4D to 1D arrays in the current `math` module. I am aware this is needed for performance reasons, but the generic interface will fail to compute as it cannot determine which interface matches.

njansson (approved):

timofeymukha (approved):

PR #1320, opened 2024-06-26
OS Version updates

timfelle:
Update the CI to determine the full OS version currently used. This improves caching of dependencies, as an update to the minor version of the runner os might require a recompilation which was not caught previously.


timofeymukha (approved):

timofeymukha:
I have to admit that I pretty much approve these devops stuff blindly :D. But I like how neat things are getting, thanks for all this work Tim.

njansson (approved):
🍻

PR #1321, opened 2024-06-26
Update auto config scripts to fix obsolete warnings

timfelle:
Please correct me if we are required to support version `2.69` for some HPC cluster purposes.
Otherwise this update clears the obsolete warnings when configuring.


timfelle:
Well that answered that question 😅

PR #1322, opened 2024-06-26
Alternative suggestion to the makeneko naming

timfelle:
Instead of changing the name of the executable, we can print a message to the log if we are running any user extended functions.

Please note, this PR is to the original PR not to the develop branch as I would not start overriding the work of Måns directly.


MaansAndersson (approved):
Sounds like a good idea, feel free to update this PR :)

PR #1331, opened 2024-07-01
SGS model updates

timofeymukha:
Final cherry-pick from stress_formulation.

- Introduces 2 ways to compute the LES filter width. Currently, just an if statement. We should probably upgrade to have a factory and types later.
- Averages the strain-rate  values at the element boundaries before using them to compute eddy viscosity.
- Some fixes to filtering that @Shiyu-Sandy-Du  knows about :P.


njansson (approved):

src/les/les_model.f90
```diff
@@ -147,55 +151,85 @@ subroutine les_model_compute_delta(this)
...
+              - this%coef%dof%y(1,int(this%coef%Xh%ly/2)+1,1,e))**2 &
+             + (this%coef%dof%z(1,int(this%coef%Xh%ly/2),1,e) &
+              - this%coef%dof%z(1,int(this%coef%Xh%ly/2)+1,1,e))**2
+
+          dk = (this%coef%dof%x(1,1,int(this%coef%Xh%lz/2),e) &
+              - this%coef%dof%x(1,1,int(this%coef%Xh%lz/2)+1,e))**2 &
+             + (this%coef%dof%y(1,1,int(this%coef%Xh%lz/2),e) &
+              - this%coef%dof%y(1,1,int(this%coef%Xh%lz/2)+1,e))**2 &
+             + (this%coef%dof%z(1,1,int(this%coef%Xh%lz/2),e) &
+              - this%coef%dof%z(1,1,int(this%coef%Xh%lz/2)+1,e))**2
```
timfelle, line 179: Why not make 3 integers up front and use those, instead of inlining 20 integer conversion calls?

timfelle (approved):

src/les/les_model.f90
timfelle: Not into 1 no. But into 3

```fortran
!> Half the dimensions
integer :: lx_half, ly_half, lz_half

lx_half = int(this%coef%Xh%lx / 2)
ly_half = int(this%coef%Xh%ly / 2)
lz_half = int(this%coef%Xh%lz / 2)
```

And then use them later on. This way we reduce the number of `int()` calls from 20 to 3. And it should make it easier to read.

timfelle: You probably don't even need the `int()` call then since you divide 2 integers with each other and store it as an integer.

Shiyu-Sandy-Du: Oh yes you are right.

timofeymukha:
The linter strikes again :D. I gotta say ,I am loving this

njansson:
> The linter strikes again :D. I gotta say ,I am loving this

I'm happy to have admin permission, and can bypass all rules  :D

timofeymukha:
@Shiyu-Sandy-Du what needs to be done is to go through the files in the les directory like, for example, this 
`flint lint -r flinter_rc.yml  src/les/les_model.f90`
and address the complaints.

njansson:
> @Shiyu-Sandy-Du what needs to be done is to go through the files in the les directory like, for example, this `flint lint -r flinter_rc.yml src/les/les_model.f90` and address the complaints.

kitten atrocities! But for a good cause (^-^)v

Shiyu-Sandy-Du:
Finally!

PR #1332, opened 2024-07-02
Pass error codes to `error stop` statements

timfelle:


njansson (approved):

timfelle:
@njansson Do we only support Fortran 2003 and not Fortran 2008?

njansson:
> @njansson Do we only support Fortran 2003 and not Fortran 2008?

We require at least Fortran 2008

timfelle:
> > @njansson Do we only support Fortran 2003 and not Fortran 2008?
> 
> We require at least Fortran 2008

Interesting, our CI gets compiled to 2003, as support for the feature i tried to implement here is fortran 2008 and the error message indicate support only ranging for the 2003 features.

https://stackoverflow.com/a/17446997

njansson:
> > > @njansson Do we only support Fortran 2003 and not Fortran 2008?
> > 
> > 
> > We require at least Fortran 2008
> 
> Interesting, our CI gets compiled to 2003, as support for the feature i tried to implement here is fortran 2008 and the error message indicate support only ranging for the 2003 features.
> 
> https://stackoverflow.com/a/17446997

I think the issue for the CI is the non constant error message

timfelle:
> I think the issue for the CI is the non constant error message

Riiight, we do not always call the function with a constant error message.

timofeymukha (approved):

timfelle:
Interesting it would seem my local compiler succeeds since because it compiles to Fortran 2018 and not Fortran 2008. This use is not allowed in Fortran 2008.

PR #1334, opened 2024-07-03
Revise stats

MartinKarp:
Moving statistics to a simcomp.

Removes the old way of doing statistics and instead introduces the fluid_stats simcomp which is intended to compute the raw statistics necessary to compute the RMS, means, TKE and Reynolds stress budgets similar to what's in the KTH framework.

Also adds the option to compute statistics and average them into a 1D csv file which greatly reducing the memory required for say channel flow. 2d is in progress, but not in this PR.

2D is only started in this PR, but perhaps I should add that too before we merge this. average_field_in_space is currently not finished.


src/fluid/fluid_stats.f90
```diff
@@ -128,7 +128,7 @@ module fluid_stats
      type(field_t) :: dwdz
 
      type(coef_t), pointer :: coef
-     integer :: n_stats = 40
+     integer :: n_stats = 44
```
njansson, line 131:
```suggestion
     integer, parameter :: n_stats = 44
```

njansson (commented):
Nice! Shows how clean solvers etc will be with optional components are moved to simcomps

src/simulation_components/fluid_stats_simcomp.f90
```diff
@@ -0,0 +1,149 @@
...
+    character(len=:), allocatable :: hom_dir
+    real(kind=rp) :: start_time
+
+    call this%init_base(json, case)
+    call json_get_or_default(json, 'avg_direction', &
+         hom_dir, 'none')
+    call json_get_or_default(json, 'start_time', &
+         start_time, 0.0_rp)
+   
+    call fluid_stats_simcomp_init_from_attributes(this, case%fluid, start_time, hom_dir)
```
timofeymukha, line 93: Just a small comment here, do you really need the fluid, or is it enough to know the field names? I am just thinking ahead a bit, if fluid becomes a simcomp itself, fetching it here might be somewhat difficult.

timfelle (commented):
Will this retire PR#1186 as they both introduce a statistics based simcomp?

MartinKarp:
No, i dont think so, this is kind of prepackaged statistics while adalbertos pr is for extending it with quantities that one as a user may need that are not covered in this one. 

I think expanding on his is a natural next step

src/simulation_components/fluid_stats_simcomp.f90
MartinKarp: yeah that should make quite a lot of sense I think

timfelle:
> No, i dont think so, this is kind of prepackaged statistics while adalbertos pr is for extending it with quantities that one as a user may need that are not covered in this one.
> 
> I think expanding on his is a natural next step

Would it not make sense to merge the two functionalities and do the setup from json on which statistics to compute?
Not that the two Pr´s should be merged together though.

MartinKarp:
Im not sure, this one actually generates a lot of fields that are computed additionally, the other one I think is more to sample fields that are available in the simulation.

timfelle:
> Im not sure, this one actually generates a lot of fields that are computed additionally, the other one I think is more to sample fields that are available in the simulation. Which differs a bit in the intended use I guess.
> 
> I am envisioning something like fluid_stats - for prebaked, common statistics user_stats - for sampling and computing statistics of fields in the simulation user_fields - for parsing expressions and add fields of derived quantities that can be added to the field registry, currently I guess this is easiest done in usercheck as wecannot parse say "u*v" or so and add it to the field registry
> 
> Perhaps the nicest though would be to kind of merge user_stats and user_fields in some sense, or as you say simply have one big stats simcomp with all kind of variations

Sure, We would need to ensure we don't allocate a lot of fields that is not needed for all users. So there would be some thinking to do if we want to have some components off and others on.

But sure having some well defined interface for users to inject their own statistics would probably be nice. But since we have the registry then all fields the user could wish to access is always there so I don't think we should have two different interfaces for the user to do statistics.

MartinKarp:
Updated the PR and tried to fix the linting (but failed). I am moving on to user statistics as I need it for CTR. This should probably be merged sooner rather than later as I intend to let all postprocessing adapt this structure instead for the statistics.

It is a rather breaking change at the same time as it is an important change on the way to make neko a bit more usable when it comes to carrying out simulations.

adperezm:
Good stuff @MartinKarp. If I may add a suggestion/request.

Is it possible to add a metadata file that contains the name of the statistics files and the time where the statistics were accumulated for (start, end)? Currently we see only the end time of the sampling period in the fld file. This would be good to make the process of collecting statistics over multiple runs less cumbersome.

This goes in line with something @Shiyu-Sandy-Du said a long time ago but I do not know if it is done in some other PR.

This can be of course done by someone else in another PR, but as I said, if stats is being refactored, I think that is a good addition. Be free to ignore the suggestion.

src/simulation_components/fluid_stats_simcomp.f90
```diff
@@ -0,0 +1,183 @@
...
+    character(len=:), allocatable :: hom_dir
+    real(kind=rp) :: start_time
+    type(field_t), pointer :: u, v, w, p
+    type(coef_t), pointer :: coef
+
+    call this%init_base(json, case)
+    call json_get_or_default(json, 'avg_direction', &
+         hom_dir, 'none')
+    call json_get_or_default(json, 'start_time', &
+         start_time, 0.0_rp)
```
vbaconnet, line 97: This means that we get rid of the `enable` keyword in the json file right? Would be nice to have it as a general attribute of `simulation_component_t` to enable/disable simcomps from the case file maybe?

vbaconnet: but perhaps this is not in the scope of this PR!

timofeymukha: Yes, that would be good. It would be easy to add I suppose

PR #1342, opened 2024-07-03
fix the amplitude of the 3D TS waves to match the Schlatter 2005 case

pschlatt1:
also added a ipynb notebook to plot the TS waves.

This case could eventually be aded to the CI to verify the code accuracy once we have an established baseline from Simson.


Shiyu-Sandy-Du:
The original setting follows p40 in Schlatter 2005. And how does it come to 5e-4?

pschlatt1:
The velocity magnitude (in centre-line velocity) in Schlatter 2005 is 0.1%, i.e. 0.001. SInce we have two oblique waves, they have individually 0.0005 as peak amplitude, which is 5e-4 for each wave.

Shiyu-Sandy-Du:
Ah yes. I overlooked we have two waves (and maybe the 1e-4 in the previous code was a typo)

Shiyu-Sandy-Du (approved):

njansson (approved):

PR #1345, opened 2024-07-04
Fix global requirement for changed files

timfelle:
This pull request fixes the issue where the global requirement was used for changed files instead of the correct requirement. This ensures that the correct requirement is used for changed files.


timofeymukha (approved):

timofeymukha:
Ah, so we were no lenient on the new stuff, or? I had a feeling that accepting that Ax PR went too easy :D

timfelle (commented):
> Ah, so we were too lenient on the new stuff, or? I had a feeling that accepting that Ax PR went too easy :D 

Yep, the requirement was only 8.55. Lets see, its currently at 10, meaning perfectly in line with our requirements. But there are some rules being enforced by flint on the number of input variables, which will be an issue for some files.

.github/workflows/main.yml
```diff
@@ -7,8 +7,8 @@ on:
   workflow_dispatch:
 
 env:
-  FLINT_GLOBAL_MINIMUM: 8.0
-  FLINT_CHANGED_FILES_MINIMUM: 8.0
+  FLINT_GLOBAL_MINIMUM: 9.0
+  FLINT_CHANGED_FILES_MINIMUM: 10.0
```
MartinKarp, line 11: Looks good I think, is 10 the maximum score?

MartinKarp (approved):
Looks good, I don't have that much intuition, but as long as 10 is relatively easy to get when writing code all cool

njansson (approved):

.github/workflows/main.yml
timfelle: Yes, 10 means, you have to adhere to all rules we have defined.
If need be, we can look into the rules if we find something odd.

njansson (approved):

PR #1348, opened 2024-07-05
Debug: Update TS wave test

Shiyu-Sandy-Du:
Two point to update with the TS wave test examples:

1. Update the outdated case file to keep accordance with the current time step controller setting. (committed already)
2. Change to a 16^3 elements mesh for a matching result with Schlatter 2005. (under discussion now)

### DIFF
--- a/examples/TS_channel/TS_channel.case
+++ b/examples/TS_channel/TS_channel.case
@@ -10,7 +10,8 @@
     "checkpoint_value": 100,
     "end_time": 200,
     "timestep": 0.02,
-    "constant_cfl": 0.4,
+    "variable_timestep": true,
+    "target_cfl": 0.4,
     "numerics": {
         "time_order": 3,
         "polynomial_order": 7,


Shiyu-Sandy-Du:
Point 2 is not needed for running and the test simulation runs well now. So maybe we can just merge the first debug point into develop

Shiyu-Sandy-Du:
@timofeymukha I guess you don't have time step varied simply because the case file in example is outdated (I should have noticed it earlier) :)

pschlatt1 (approved):
looks good to me, as long as the results is matching.

Shiyu-Sandy-Du:
> looks good to me, as long as the results is matching.

The transition process does not match exactly because of coarse resolution dampen disturbances. But as an examples, it works smoothly on a local workstation and the transition happens as expected :) A recommended mesh size is noted in README.md

njansson (approved):

PR #1352, opened 2024-07-09
updated the initial condtion for channel flow example

pschlatt1:


MartinKarp (approved):
Looks good to me, is polynomial order 5 enough? I guess it makes things go quite a lot faster so that's nice.

examples/turb_channel/turb_channel.case
```diff
@@ -41,7 +43,7 @@
         },
         "boundary_types": ["", "", "w", "w"],
         "output_control": "simulationtime",
-        "output_value": 20
+        "output_value": 1.0
```
timofeymukha, line 46: Do we really want 200 outputs?

MartinKarp: I think that would be reasonable, at least in the beginning, otherwise it takes quite a while before getting something and one can follow as the channel becomes turbulent.

njansson (approved):

PR #1351, opened 2024-07-07
Add variable properties support to fluid_scheme

timofeymukha:
Adds a `mu_field` to `fluid_scheme` and routine to update it based on a turbulence viscosity field.

The idea of putting it in the base type is the whole of eventually fusing full and simplified stress formulations into one solver.

There are also a bunch of changes to make flint happier.

PR #1354, opened 2024-07-09
Fix missing set_coef call for user inflow condition

vbaconnet:
Put the call to `bc_inflow%set_coef` back in fluid scheme that was causing `usr_inflow::usr_inflow_validate` to fail when running with a `"user"` inflow condition since the `usr_inflow_t%c` pointer was never being associated.

### DIFF
--- a/flinter_rc.yml
+++ b/flinter_rc.yml
@@ -222,11 +222,11 @@ regexp-rules:
 structure-rules:
   file-line-length: 80
   file-line-number: 2000
-  max-statements-in-context: 300 # Subroutine of function
+  max-statements-in-context: 800 # Subroutine of function
   max-declared-locals: 35
   min-varlen: 1
   max-varlen: 30
-  max-arguments: 10
+  max-arguments: 20
   min-arglen: 1
   max-arglen: 70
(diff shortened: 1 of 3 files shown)


njansson (approved):

timofeymukha:
Guys, let's wait for the varprops pr to merge in, so that we don't have to fix the linter for the fluid scheme in 2 PRs.

timofeymukha:
@vbaconnet The `bc_t` type now contains the `coef_t`, you see it being passed in `call this%bc_inflow%init_base(this%c_Xh)`.

So, the correct thing to do here is to kill the additional coef pointer in the user inflow type, and directly use the one in parent type. The fluid_scheme_t code I think can remain as is.

vbaconnet:
Ok so I guess don't merge until #1351 is merged @timofeymukha ?

timofeymukha:
> Ok so I guess don't merge until #1351 is merged @timofeymukha ?

I guess if you only have to modify the usr_inflow_t then we can merge independently.

timofeymukha (approved):

PR #1358, opened 2024-07-10
Temp fix to be able to run on LUMI

MartinKarp:
Dont know exactly what goes wrong, but something is iffy with hipGetDeviceCount. Do you have any idea what can be the issue @timfelle ?

To be clear, if one does not comment these lines Neko segfaults directly on LUMI.

### DIFF
--- a/src/device/hip_intf.F90
+++ b/src/device/hip_intf.F90
@@ -272,13 +272,13 @@ contains
   subroutine hip_init
     integer(c_int) :: num_devices
 
-    if (hipGetDeviceCount(num_devices) .ne. hipSuccess) then
-       call neko_error('Failed to query device count')
-    end if
+    !if (hipGetDeviceCount(num_devices) .ne. hipSuccess) then
+    !   call neko_error('Failed to query device count')
+    !end if
 
-    if (num_devices .ne. 1) then
-        call neko_error('Only one device is supported per MPI node')
-    end if
+    !if (num_devices .ne. 1) then
+    !    call neko_error('Only one device is supported per MPI node')
+    !end if
 
     if (hipDeviceGetStreamPriorityRange(STRM_LOW_PRIO, STRM_HIGH_PRIO) &
          .ne. hipSuccess) then


njansson:
I think the interface definition is wrong. Hipgetdevicecount takes a pointer, thus 'num_devices' shouldn't be declared with value in the interface

MartinKarp:
The check for number of devices also makes it a bit inconvenient to use the contrib scripts

njansson:
> The check for number of devices also makes it a bit inconvenient to use the contrib scripts

How about we make it possible to disable the check, e.g. when using tools from contrib? I still think it is important to have the check when we run the solvers, if not only to avoid answering repeated questions about mappings etc.


Any other thoughts @timfelle @timofeymukha ?

MartinKarp:
I also think this is good in principle, it was mostly that the use of neko changed rather significantly. I would also not be opposed to have a big warning on runtime instead as well.

njansson:
> I also think this is good in principle, it was mostly that the use of neko changed rather significantly. I would also not be opposed to have a big warning on runtime instead as well.

How about an error if more than one device is used per rank, and a warning if no devices are detected (if one now runs contrib stuff on the login node...)

MartinKarp:
I think one should be able to run contrib regardless of the number of devices.

timofeymukha:
Instead of doing things in hipInit and similar, cam we put this check in a separate function? Then, we can e.g. call it in neko's main, but not in (some of) cobtrib's mains.

njansson:
Ok, but how about todo as @timofeymukha suggested then, we move the check later down in init, e.g. after the check for a case https://github.com/ExtremeFLOW/neko/blob/f54685c88ad29e9b63c515ebcb834a520f1cc0eb/src/neko.f90#L142,

then contrib will work, and we still cover the case of support questions in a good way

MartinKarp:
Super fix Tim, if this all looks good with people I think we can merge it.

njansson (approved):

timofeymukha (approved):

PR #1359, opened 2024-07-13
update TS test

Shiyu-Sandy-Du:


njansson (approved):

Shiyu-Sandy-Du:
Two changes are made:

1. fix the bug in the 3D TS wave initial condition from the csv file
2. change the domain to be symmetric about 0 in spanwise direction in order to match the result from Schlatter 2005

examples/TS_channel/TS_channel.f90
```diff
@@ -61,8 +62,8 @@ subroutine user_ic(u, v, w, p, params)
     real(kind=rp) :: ur_3D, ui_3D, vr_3D, vi_3D, wr_3D, wi_3D
 
     ! data reading
-    open(unit=10, file='TSwave_cheb_2D.csv', status='old', &
-                  action='read', iostat=ios)
+    open(unit = 10, file = 'TSwave_cheb_2D.csv', status = 'old', &
+                  action = 'read', iostat = ios)
```
vbaconnet, line 66: Just out of curiosity, is there a reason why you chose not to use the `csv_file` reader? ;)

Shiyu-Sandy-Du: oh actually there is no specific reason ... Just picked one works and keep it as it is. Is it better to use `csv_file` reader here?

pschlatt1: to be honest, here I would even prefer to not even have csv but directly write out the binary files from python and read it in again in neko. I know, Fortran binary files are in principle not standard, but de facto they are.

vbaconnet: > Is it better to use `csv_file` reader here?

In the end the csv reader does the exact same you did so there is no "better" way to do it, personally I feel that it just declutters the code a bit...

Shiyu-Sandy-Du: > to be honest, here I would even prefer to not even have csv but directly write out the binary files from python and read it in again in neko. I know, Fortran binary files are in principle not standard, but de facto they are.

Is it because binary file have smaller size?

Shiyu-Sandy-Du: > > Is it better to use `csv_file` reader here?
> 
> In the end the csv reader does the exact same you did so there is no "better" way to do it, personally I feel that it just declutters the code a bit...

okay.. I will have a look later if we decide to use csv file in the end

njansson: I actually had a discussion with @timofeymukha  about this, how about dumping it to hdf5 ?

pschlatt1: no, the size is not a problem for these files. But you do not loose any accuracy. That is for TS waves quite important. So if anything, I would change it to binary (there is a fortran writer in python that writes nice Fortran records directly).

pschlatt1: No hdf, just fortran binary, like in the good old days (record separatora etc, works perfectly)

Shiyu-Sandy-Du: So it's .bin file? (It's what GPT told me)

pschlatt1: No idea what a bin file is. Just fortran binaries. Maybe chatgpt is too modern for that

njansson: But isn't record separated files compiler specific? Don't remember if it is standardized or not.

pschlatt1: De facto it.is standard

adperezm: Is this a performance critical part?. In my opinion writing things in binary makes everything more difficult.

You say Fortran is standard but I remember post processing simson for my case needed that I get some matlab reader from you and then I adapt that to python.

This is also a small file I guess, since every rank is reading it. I would suggest to just keep it simple. 

If you still want binary, and to show hdf5, then I would suggest that ALL of those are shown. Following other examples where there are "commented options" on how to do things. Then the user can have flexibility and learn how to read the data from an input file and adapt it to their own needs.

timofeymukha: That could be a good idea. We could put the same data in different formats in some "input" directory. I think showcasing hdf5 would be nice.

Shiyu-Sandy-Du: So I guess for now, a fortran standard file and a .hdf5 file could be provided for users to choose and we can show the use of the corresponding reader in this example?

Shiyu-Sandy-Du: Ok, after having a talk with @pschlatt1, a binary file is fairly enough for this case. However, an example showing the use of different formats and corresponding readers is possible.

njansson (approved):

pschlatt1 (approved):

PR #1365, opened 2024-07-16
Fix/hip_device_count

timfelle:
Should fix the device count interface for HIP.

We now have a sepparate function `hip_device_count` instead of the direct hip API binding.

Additionally we update to the latest develop branch.

I do not have access to a HIP based device so could you please see if this now works @MartinKarp?

### DIFF
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -10,7 +10,6 @@ on:
       - reopened
       - ready_for_review
   workflow_dispatch:
-  push:
 
 env:
   FLINT_GLOBAL_MINIMUM: 8.55
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -10,6 +10,7 @@ on:
       - reopened
       - ready_for_review
   workflow_dispatch:
+  push:
 
 env:
   FLINT_GLOBAL_MINIMUM: 8.55
--- a/src/device/hip_intf.F90
+++ b/src/device/hip_intf.F90
@@ -235,7 +235,7 @@ contains
 
   subroutine hip_init
     if (hip_device_count() .ne. 1) then
-       call neko_error('Only one device is supported per MPI node')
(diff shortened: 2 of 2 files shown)


njansson:
Sorry haven't had time to try it on lumi yet, however I have some doubt about the api binding. afaik it's not compatible with what the api provides, https://rocm.docs.amd.com/projects/HIP/en/docs-6.0.0/doxygen/html/group___device.html#ga8555d5c76d88c50ddbf54ae70b568394. The return values should be the error flag, and the number of devices is given by the argument (which shouldn't be passed as `value`, which was the issue in the previous implementation)

timfelle:
> Sorry haven't had time to try it on lumi yet, however I have some doubt about the api binding. afaik it's not compatible with what the api provides, [rocm.docs.amd.com/projects/HIP/en/docs-6.0.0/doxygen/html/group___device.html](https://rocm.docs.amd.com/projects/HIP/en/docs-6.0.0/doxygen/html/group___device.html#ga8555d5c76d88c50ddbf54ae70b568394). The return values should be the error flag, and the number of devices is given by the argument (which shouldn't be passed as `value`, which was the issue in the previous implementation)

I see... I accidentally ended up on the python documentation for hip instead of the c++ one...

njansson:
How about we add a `device_count` in `device.f90` such that we can query it from elsewhere in the code ?

timfelle:
> How about we add a `device_count` in `device.f90` such that we can query it from elsewhere in the code ?

That was the ultimate idea, yes. However I wanted to fix the lumi runs first and then make a dedicated PR where the other device types are also tied together.

MartinKarp (changes requested):
Tiptop! Sorry for being a bit late on the review. I tested it on LUMI on GPU and the TGV case runs just fine. However, I managed to get spurious crashes for the turb_channel. When I reran however I did not crash. I don't know if this is a problem, but ikely not related to this. Only thing is changing to mpi rank I think rather than node.

Issue #1283, opened 2024-05-20
Enable initial condition from field file


vbaconnet:
With #1279 in preparation I thought it would be nice to be able to use a field file as an initial condition. A suggested plan for this would be (for the general case of a 3D simulation):

1. Initialize from a 3D field file that has the exact same mesh
2. Initialize from a 2D field file that has the same 2D slice mesh, that is then extruded on the 3D mesh
3. Initialize from a 3D and/or 2D field file from a different mesh (interpolation)

As far as I understand point 1. should be a simple `copy` if the field file was output from neko? From what I understand, the rest would be more or less the same as in `chkp_t` for the interpolation, and the 2D extrusion could somehow make use of the `map_1d`?

timofeymukha:
Hi! So, 1 is essentially "restart from a .fld instead of the checkpoint". I think this is generally useful. 

Point 3 we already have, or at least would if the probes worked will, right? But not easily through the case file.  Perhaps I am a bit confused...

 Point 2 I think is the most tricky one by far. Considering we don't solve for 2D in Neko, what would be the scenario for using it?

njansson:
Great, the reading part for point 1 already in place, what is missing is the infrastructure duct tape to glue it together from the case file.

point 2 is for sure the tricky part ;)

njansson:
I guess one idea would be to have an "initial condition": "filename" field in the case file, and the filename can then be a nek5000 field, or even better HDF5

vbaconnet:
> Point 3 we already have, or at least would if the probes worked will, right? But not easily through the case file. Perhaps I am a bit confused...

Yes exactly, as Niclas said we have all the tools for it, it just needs some putting together.

> Point 2 I think is the most tricky one by far. Considering we don't solve for 2D in Neko, what would be the scenario for using it?

Well it would be just running the "pseudo 2D" case in Neko with only 1 element in the spanwise direction, and reusing that result on the same case but extruded in the spanwise direction. So then not really 2D but it has that idea of extruding?

vbaconnet:
I guess depending on the case one may run the same spanwise length but with a different number of elements... So then it would just be interpolation.

timofeymukha:
Yeah, if you are going to do pseudo-2d then you might as well match the width I thought

ronithstanly:
Adding the capability to give Nek5000 fields as IC to Neko would be very helpful to move several of my cases to Neko:)

njansson:
[...]

PR #1366, opened 2024-07-16
Initial condition from fld or chkp

vbaconnet:
Adds some functionalities outlined in #1283 for using fields as an initial condition.

Basically allows to initialize the simulation from `fld` or `chkp` files with the keyword `case.<fluid/scalar>.initial_condition.type = field`.

- ~~For chkp the already existing interpolations (mesh to mesh and from one polynomial order to another) are usable.~~ 
- For fld we basically do the exact same interpolation workflow as chkp, but we wrap the findpts part in `fld_data_file` instead of having it in the reader. ~~Chkp should perhaps be modified to do that as well if we decide to go that way~~.

# Tests

Tested with:

- [x] AMD GPUs
- [x] Nvidia GPUs
- [x] CPUs 

Functionalities
- [x] ~~Read from chkp with or without interpolation~~
- [x] Read from fld without interpolation
- [x] Read from fld with interpolation (different polynomial order)
- [x] Read from fld with interpolation (mesh to mesh)
- [x] ~~Read from fld with 2D extrusion~~ (I think this is already quite a lot so probably we can skip this for now?)

# Usage

The name of the fld file name is given in `file_name`. The interpolation between two different polynomial orders is done automatically without having to specify anything (same as for chkp).

In order to use the mesh to mesh interpolation, we use the `interpolate` keyword and can also provide the file that contains the mesh coordinates if necessary:

```json
"initial_condition": {
    "type": "field",
    "file_name": "field0.f00003",
    "interpolate": true,
    "mesh_file_name": "field0.f00000"
}
```

# Other changes

I added some logging for the initial condition so it becomes cleaner when using the `fld` reader that outputs some messages, that gives something like:

```text
   ---Fluid initial condition----  
   Type: field
   File name: results/field0.fld
   Reading meta file for fld series
   Name: field0
   Start counter    :       0
   Number of samples:      17
   Reading fld file results/field0.f00000
   Reading fld file results/field0.f00016
  
   ---Scalar initial condition---  
   Type: field
   File name: results/field0.fld
   Sample index: 00005
   Reading meta file for fld series
   Name: field0
   Start counter    :       0
   Number of samples:      17
   Reading fld file results/field0.f00005
```

But it will also output all the values provided if one uses the other initial condition types like `blasius`, `uniform`, `point_zone` etc.

```text
   ---Fluid initial condition----  
   Type: blasius
   delta       :   0.001000
   Approximation: sin
   Value: [  1.000000,  0.000000,  0.000000]
```


src/fluid/flow_ic.f90
```diff
@@ -256,4 +350,115 @@ subroutine set_flow_ic_point_zone(u, v, w, base_value, zone_name, zone_value)
...
+    type(field_t), intent(inout) :: w
+    type(field_t), intent(inout) :: p
+    character(len=*), intent(in) :: file_name
+    integer, intent(in) :: sample_idx
+
+    type(fld_file_data_t) :: fld_data
+    type(file_t) :: f
+
+    ! Should this init be init(u%msh%nelv, u%msh%offset_el)?
+    call fld_data%init
```
vbaconnet, line 375: Should this init be `init(u%msh%nelv, u%msh%offset_el)`?

src/fluid/flow_ic.f90
```diff
@@ -256,4 +350,119 @@ subroutine set_flow_ic_point_zone(u, v, w, base_value, zone_name, zone_value)
...
+         (fld_data%lz .ne. u%dof%Xh%lz)) then
+       call neko_error("The fld file must match the current mesh")
+    end if
+
+    ! Note: we do not copy on the GPU since `set_flow_ic_common` does the
+    ! copy for us, except for the pressure
+    call copy(u%x, fld_data%u%x, u%dof%size())
+    call copy(v%x, fld_data%v%x, v%dof%size())
+    call copy(w%x, fld_data%w%x, w%dof%size())
+    call copy(p%x, fld_data%p%x, p%dof%size())
```
vbaconnet, line 414: Do we need to copy the pressure too? Is that what one would want from this feature?

src/scalar/scalar_ic.f90
```diff
@@ -187,4 +272,105 @@ subroutine set_scalar_ic_point_zone(s, base_value, zone_name, zone_value)
...
+    type(chkp_t) :: chkp_data
+    type(file_t) :: f, meshf
+
+    u => neko_field_registry%get_field("u")
+    v => neko_field_registry%get_field("v")
+    w => neko_field_registry%get_field("w")
+    p => neko_field_registry%get_field("p")
+
+    ! Could we also init with chkp_data%init(s,s,s,s) for simplicity?
+    call chkp_data%init(u,v,w,p)
```
vbaconnet, line 356: Could we also init with `chkp_data%init(s,s,s,s)`? Or is it dangerous?

adperezm:
Great work!

A comment before reading the code:
I think it is very good also that one can specify directly the file name (avoiding .nek5000) files. In that sense, it could also be good that one gives that option for the file with the mesh, rather than just the index.

I think I see your reasoning, however, since you would restart in a folder that already has an up to date .nek5000 and all that. But it is good to have options.

src/common/global_interpolation.F90
```diff
@@ -85,7 +85,7 @@ module global_interpolation
      !> Error code for each point, needed for gslib
      integer, allocatable :: error_code(:)
      !> Tolerance for distance squared between original and interpolated point
-     real(kind=rp) :: tol = 5e-13
+     real(kind=rp) :: tol = 5d-13
```
adperezm, line 88: I do not know how relevant, but one could run into trouble if trying to restart from an fld file that is in single precision. I suppose that we do not need to change the defaults here, but rather give a warning to the user when they try to do that.

vbaconnet:
> it could also be good that one gives that option for the file with the mesh, rather than just the index.

Actually yes it would make sense to be able to do it via a file too. Also I am now wondering if it even makes sense to have the option to give the index?

src/sem/dofmap.f90
adperezm, line 1: Nice, I think this sort of init is very good to have for post processing scripts.

src/io/fld_file_data.f90
```diff
@@ -171,4 +182,97 @@ subroutine fld_file_data_free(this)
...
+
+    ! The initialization is done based on the variables created from
+    ! fld data
+    call global_interp%init(fld_dof, tol = tolerance, Xh = fld_Xh, &
+         msh = fld_empty_mesh)
+
+    ! Note that to_dof%size() should be equal to
+    ! to_msh%nelv * to_Xh%lxyz
+    if (to_dof%size() .ne. to_msh%nelv*to_Xh%lxyz) &
+         call neko_error("something weird happening")
```
adperezm, line 268: Excellent name :P.

I can imagine a user frustration if they get that error message haha. Maybe something more descriptive.

adperezm (commented):
I did not check the ic files since I am not familiar with them, but everything is very nice! just some minor comments.

src/common/global_interpolation.F90
vbaconnet: You mean the change from `e` to `d` or the fact that the tolerance could be too low for single precision?

I was running into single precision issues on LUMI, and it made me wonder if perhaps it would be good that we always output the first field file with the mesh in double precision, and the rest is up to the user?

adperezm: I meant that the tolerance might be too low. In your .nmsh you have double, but then the fld will have single, so there will be a lot of rubbish after some digits.

I think that since the initialization from fld is a very specific thing (I would hope and assume that people use the chkp which are always in double) the files can be left as they are. But someone else can comment on you second point, it does not matter for me.

One thing about writing the first file in double and then the rest in single is that I do not know how visit and the sort would behave.

src/io/fld_file_data.f90
adperezm: Haha yeah, I thought so :)

src/common/global_interpolation.F90
vbaconnet: > I think that since the initialization from fld is a very specific thing the files can be left as they are ( (I would assume that the default use would be the chkp which are always in double). But someone else can comment on you second point, it does not matter for me.

I think this is definitely worth mentioning in the documentation, and adding a warning if the fld file is in single precision.

> One thing about writing the first file in double and then the rest in single is that I do not know how visit and the sort would behave.

Oh that is true, I have no idea either.

src/scalar/scalar_ic.f90
Shiyu-Sandy-Du: I don't think it would be dangerous anyway since in `chkp_t%init()`, all arguments are declared to be `intent(in)`

src/io/fld_file_data.f90
```diff
@@ -171,4 +182,97 @@ subroutine fld_file_data_free(this)
...
+    allocate(z_coords(to_Xh%lx,  to_Xh%ly, to_Xh%lz, to_msh%nelv))
+
+    !> To ensure that each point is within an element
+    !! Remedies issue with points on the boundary
+    !! Technically gives each point a slightly different value
+    !! but still within the specified tolerance
+    do e = 1, to_msh%nelv
+       center_x = 0d0
+       center_y = 0d0
+       center_z = 0d0
```
Shiyu-Sandy-Du, line 241: Will these double precision assignment pass the PR test for single precision compiler?

src/scalar/scalar_ic.f90
vbaconnet: Yes, that is true. But then `chkp_t` has pointers to u,v,w,p in `checkpoint::chkp_init` and the reader seems to write over them in `chkp_file::chkp_file_read`:

```.f90
    call this%read_field(fh, byte_offset, u%x, nel)
! ...
    call this%read_field(fh, byte_offset, v%x, nel)
! ...
    call this%read_field(fh, byte_offset, w%x, nel)
! ...
    call this%read_field(fh, byte_offset, p%x, nel)
! ...
    if (read_scalar) then
       call this%read_field(fh, byte_offset, s%x, nel)
    end if
```

So I wonder if this means that a call to `chkp_file%read(chkp_data)` will overwrite `u,v,w,p` and `s`, potentially messing things up.

adperezm: I think this is a good place where maybe @MartinKarp knows from the top of his head.

The other alternative is to of course track at which point the data is given to the u,v,w,p fields that are part of the fluid scheme.

A quick look (which I guess you have seen) tells that the u%x that appears in those read_file() calls are part of the field_data type, so I guess they do not directly affect the fields in the fluid scheme. But maybe they do downstream..

vbaconnet: Actually now that I am testing just initializing the scalar from a checkpoint I see that it's not reading the scalar properly anymore, but `u` instead. Even with `init(s,s,s,s)`, kind of weird

src/io/fld_file_data.f90
vbaconnet: The same lines were merged previously in `chkp_file` with the mesh to mesh interpolation so I'm expecting it to work

src/scalar/scalar_ic.f90
vbaconnet: Ok so the issue is that we don't read the `ulag, vlag, wlag` fields even though they are in the file, so when it is time to read `scalar` we instead read `ulag`.

vbaconnet: I pushed the first option so you can see what it looks like

timofeymukha: >Ok so the issue is that we don't read the ulag, vlag, wlag fields even though they are in the file, so when it is time to read scalar we instead read ulag. 

Sigh, this is exactly the type of crap HDF5 will fix.

adperezm: Well, having data in a file and not accessing it can happen with everything. That's more on us

vbaconnet: > Is there something fundamentally different in what you are doing here?

Yes, I do not initialize any lag arrays in the `chkp_t` object before reading the chkp file, because I do not have access to those from `flow_ic` or `scalar_ic`.

vbaconnet: For a normal restart you add the lag arrays in [fluid_pnpn](https://github.com/ExtremeFLOW/neko/blob/411e4c2a2f1fa8b121d00c4f94be29dd378b0b00/src/fluid/fluid_pnpn.f90#L315):
```.f90
call this%chkp%add_lag(this%ulag, this%vlag, this%wlag)
```

but those are in `fluid`, which is not visible from `scalar_ic`

MartinKarp (changes requested):
Good work! The only thing I would like to raise is whether we should support set ic from chkp. For the fld file I think it is great, but why would one not restart from the chkp? 

Is there some interpolation we do when setting it as ic we don't do when doing the restart?

I would personally skip support for chkp, but now you have done a lot of work on it so.

Sorry for the late comment, but Im on vacation until the 11th.

src/io/fld_file_data.f90
MartinKarp: these are scalars so they are correctly casted by the compiler.

src/scalar/scalar_ic.f90
MartinKarp: Is the ic from chkp working now? Otherwise, as I wrote in my review I would perhaps skip it. I am also not entirely sure when one wants to set the ic from a checkpoint.

src/common/global_interpolation.F90
```diff
@@ -98,29 +98,45 @@ module global_interpolation
...
 contains
-  !> Initialize user defined variables.
+
+  !> Initialize the global interpolation object on a dofmap.
   !! @param dof Dofmap on which the interpolation is to be carried out.
   !! @param tol Tolerance for Newton iterations.
-  subroutine global_interpolation_init(this, dof, tol)
+  !! @param Xh Space on which to interpolate.
+  !! @param msh Mesh on which to interp
+  subroutine global_interpolation_init(this, dof, tol, Xh, msh)
```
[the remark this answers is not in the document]
vbaconnet: It was in case we passed down a "stripped-down" `dofmap` that only had x,y,z coordinates. But as you said it makes more sense to just remove dofmap from the init directly,

vbaconnet:
> why would one not restart from the chkp?

I guess it was just to have that option. In case the restart somehow does not work for any reason one can still use the fields for a first order restart. Thinking about some simulation component that does not restart properly for example. But in practice as you say this is probably quite rare.

src/scalar/scalar_ic.f90
vbaconnet: Yes it's working. Just have to hardcode the size of lag arrays so...

vbaconnet:
I think this is pretty much done :)

doc/pages/user-guide/case-file.md
```diff
@@ -502,6 +538,36 @@ file documentation.
...
+   ~~~~~~~~~~~~~~~
+   The output `#std 4 ...` indicates single precision, 
+   whereas `#std 8 ...` indicates double precision.
+   Neko write single precision `fld` files by default. To write your 
+   files in double precision, set `case.output_precision` to
+   `"double"`. 
+  
+   @attention Neko does not detect wether interpolation is needed or not. 
+   Interpolation will always be performed if `"interpolate"` is set 
+   to `true` even if the field file matches with the current simulation.
```
timofeymukha, line 570: Maybe we can just point to the doc for fluid here.

src/common/global_interpolation.F90
```diff
@@ -98,43 +106,73 @@ module global_interpolation
...
     type(dofmap_t), target :: dof
     real(kind=rp), optional :: tol
-    integer :: lx, ly, lz, nelv, max_pts_per_iter
 
-    this%dof => dof
-    this%Xh => dof%Xh
-    this%mesh => dof%msh
-    if(present(tol)) this%tol = tol
+    call this%init_xyz(dof%x(:,1,1,1), dof%y(:,1,1,1), dof%z(:,1,1,1), &
+         dof%msh%gdim, dof%msh%nelv, dof%Xh, tol = tol)
```
timofeymukha, line 132: Does this actually send the whole x y z arrays and not just a small slice?

timofeymukha (commented):
Very nice! 

I am a bit conflicted on whether we need the logic for figuring out where stuff is instead of just forcing the user to point to concrete filenames. But I guess now that you added it, let's keep it :-).

src/common/global_interpolation.F90
vbaconnet: It sends the whole array! it's a trick I got from Martin :) or rather from what was done in the fld_file read/write routines! https://github.com/ExtremeFLOW/neko/blob/2969cdbc1fe039511bc7f4e6c003070e0a5b9cb2/src/io/fld_file.f90#L185-L189

vbaconnet:
> I am a bit conflicted on whether we need the logic for figuring out where stuff is instead of just forcing the user to point to concrete filenames. But I guess now that you added it, let's keep it :-).

I mean I don't mind removing it if it makes the usage less confusing :) Whatever is best for the user, also at this point it is easier to remove things than add more functionality, especially since the logic in `flow_ic` and `scalar_ic` is not the easiest to follow (from my experience of coming back to it after the summer vacations).

src/common/global_interpolation.F90
timofeymukha: Black magic. Add a comment maybe, for sure this will surprise people.

vbaconnet: hahaha sure

timofeymukha: Yeah just a note is fine!

timofeymukha (approved):

src/fluid/flow_ic.f90
```diff
@@ -69,28 +77,133 @@ subroutine set_flow_ic_int(u, v, w, p, coef, gs, type, params)
...
+                     read_str)
        call json_get(params, 'case.fluid.blasius.freestream_velocity', uinf)
-       call set_flow_ic_blasius(u, v, w, delta, uinf, blasius_approximation)
+
+       write (log_buf, '(A,F10.6)') "delta       : ", delta
+       call neko_log%message(log_buf)
+       call neko_log%message("Approximation: " // trim(read_str))
+       write (log_buf, '(A,"[",2(F10.6,","),F10.6,"]")') "Value: ", uinf(1), &
+            uinf(2), uinf(3)
+       call neko_log%message(log_buf)
```
timfelle, line 115: Maybe this should be inside the set_flow_ic_blasius? It's great with some info
on the initial condition though.

src/fluid/flow_ic.f90
```diff
@@ -69,28 +77,133 @@ subroutine set_flow_ic_int(u, v, w, p, coef, gs, type, params)
...
+& file format must be e.g. 'mean0.f00001'")
+
+             write (log_buf, '(A,A)') "Coordinates file : ", &
+                  trim(mesh_fname)
+             call neko_log%message(log_buf)
+          end if
+
+       else ! if interpolate
+          call neko_log%message("Interpolation: no")
+       end if
```
timfelle, line 202: Same here, I think it would make sense if checks for validity is done inside the
`set_flow_ic_fld` and not in the factory.

src/simulation_components/probes.F90
```diff
@@ -245,7 +243,7 @@ subroutine read_point(this, json)
 
     ! Ensure only rank 0 reads the coordinates.
     if (pe_rank .ne. 0) return
-    call json_get(json, 'coordinates', rp_list_reader)
+    call json%get('coordinates', rp_list_reader, found)
 
     ! Check if the coordinates were found and were valid
     if (.not. found) then
```
timfelle, line 249: Maybe just remove the if statement below. `json_get` does the check for us.

src/case.f90
```diff
@@ -287,6 +287,10 @@ subroutine case_init_common(C)
     !
     call json_get(C%params, 'case.fluid.initial_condition.type',&
                   string_val)
+
+    call neko_log%section("Fluid initial condition ")
+    call neko_log%message("Type: " // trim(string_val))
```
timfelle, line 292: Maybe let each type message who they are themselves.

timfelle (approved):
All in all, looks good. My comments are more suggestions and "food for thought"

src/case.f90
vbaconnet: Do you mean something like:
```.f90
if (trim(type) .eq. 'uniform') then
   call neko_log%message("Type: uniform")
else if (trim(type) .eq. 'blasius') then
   call neko_log%message("Type: blasius")
else if (trim(type) .eq. 'point_zone') then
   call neko_log%message("Type: point zone")
else if (trim(type) .eq. 'field') then
   call neko_log%message("Type: field")
else
   call neko_error('Invalid initial condition')
end if
```

or in their respective `set_flow_ic_*`? 

The reason I did it this way initially was to avoid calling it in both `set_flow_ic_int` and `set_flow_ic_usr`.

timfelle: I would say it should be done at least in the set_flow_ic_* and not in `case.f90`.

Lets assume we build a case with two different fluids. Now we need to move the print statement. If we place it inside the ic factory we don't have to worry about it.

MartinKarp (approved):

PR #1375, opened 2024-07-18
Feature/Device count

timfelle:
Adding a unified interface to get the number of available backend devices.

In PR #1358 it was mentioned this could be used to move the check out of initialization. Where should that be done then?


njansson:
Isn't the indentation for new line a bit off, now we have a several fixes which is not what eMacs Fortran mode (which we follow) gives ?

timofeymukha:
Is there a comprehensive ruleset of eMacs Fortran mode written somewhere when it comes to indentation after &?

timofeymukha:
@timfelle regarding the check location, maybe somewhere in the beginning of neko_init ?

timfelle:
I have looked into the line continuation. Judging from https://emacsdocs.org/docs/emacs/Fortran-Indent
It seem the continuation lines must be indented by 5 additional space compared to the current context.

Is that what we should follow @njansson? Because then i will update the documentation to reflect that.

njansson:
> I have looked into the line continuation. Judging from https://emacsdocs.org/docs/emacs/Fortran-Indent It seem the continuation lines must be indented by 5 additional space compared to the current context.
> 
> Is that what we should follow @njansson? Because then i will update the documentation to reflect that.

I would say yes. It also makes it consistent with most of the continuation lines in Neko then, which followed their style.

timfelle:
Done, @njansson and @timofeymukha 
However, Depends on #1391 to be merged

njansson (approved):

njansson:
@timfelle Should we merge this one?

timfelle:
> @timfelle Should we merge this one?

Should be good yes

PR #1376, opened 2024-07-18
Add generic vector/matrix operations

vbaconnet:
Overloading some operators for the `vector_t` and `matrix_t` classes, and add some additional math routines like `cadd2`, `device_cadd2` (a(i) = b(i) + c) and `device_add3` (a(i) = b(i) + c(i)).

This is to facilitate the usage of `vector_t` and `matrix_t` especially with GPUs. This should not break anything since it's just adding to what we have, only thing I changed is the intent on some arguments in `sub3` and `add3`, @njansson will this be a problem?

Tested on cpus, nvidia, and amd gpus. Not tested with OpenCL.

Contrary to the assignment operator where if we do `v = w` and `v%x` is already allocated, we free `v` and re-initialize it to have the same size as `w`, all the other operations assume that if one does `v = a + b` and `v` is already allocated then `v` should have the same size as `b` and `a`. So there is no implicit reallocation except for the assignment operator.


vbaconnet:
Not sure I understand what is wrong with the checks, if someone could take a look that would be much appreciated :)

src/math/vector.f90
```diff
@@ -32,27 +32,59 @@
...
 
   type, public ::  vector_t
-     real(kind=rp), allocatable :: x(:) !< Vector entries.
-     type(c_ptr) :: x_d = C_NULL_PTR    !< Device pointer.
-     integer :: n  = 0                  !< Size of vector.
+     !< Vector entries.
+     real(kind=rp), allocatable :: x(:)
+     !< Device pointer.
+     type(c_ptr) :: x_d = C_NULL_PTR
+     !< Size of vector.
```
timofeymukha, line 51: Should be !>, right?

src/math/vector.f90
```diff
@@ -126,7 +158,7 @@ subroutine vector_assign_vector(v, w)
     if (NEKO_BCKND_DEVICE .eq. 1) then
        call device_copy(v%x_d, w%x_d, v%n)
     else
-       v%x = w%x
+       call copy(v%x, w%x, v%n)
     end if
```
timofeymukha, line 162: I think we decided that in these cases we should actually use the intrinsic = etc, so the cpu math.f90 is somewhat deprecated. But we can still use it. I guess this is a good opportunity to reaffirm the decision, @njansson

tests/vector/vector_parallel.pf
```diff
@@ -118,5 +118,186 @@ contains
...
+
+    @assertEqual(u%n, v%n)
+    @assertTrue(allocated(u%x))
+
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_memcpy(u%x, u%x_d, n, DEVICE_TO_HOST, sync=.true.)
+    end if
+
+    do i = 1, n
+       @assertEqual(7.0_rp, u%x(i), tolerance=NEKO_EPS)
```
timofeymukha, line 154: We add 1 to 4, so the result should be 5. Check out the other fails, I think the CI may actually correct here. 
[test-suite.log](https://github.com/user-attachments/files/16325384/test-suite.log)

src/math/vector.f90
vbaconnet: :) yep

src/math/matrix.f90
```diff
@@ -161,6 +185,196 @@ subroutine matrix_assign_scalar(m, s)
...
+  !> Matrix-matrix addition \f$ v = m + b \f$.
+  function matrix_add_matrix(m, b) result(v)
+    class(matrix_t), intent(in) :: m, b
+    type(matrix_t) :: v
+
+    if (m%nrows .ne. b%nrows .or. m%ncols .ne. b%ncols) &
+         call neko_error("Matrices must be the same size!")
+
+    if (.not. allocated(v%x)) then
+
```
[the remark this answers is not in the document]
vbaconnet: Oh yes I guess you're right, I got confused with the logic. Of course there is no way `v` can be previously allocated :)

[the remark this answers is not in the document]
vbaconnet: I guess it's the same everywhere then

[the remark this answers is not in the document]
vbaconnet: Is there a way to somehow let the operator know that we're doing an in-place operation? Like if i do `v = v + b` will it always create (v+b) and copy that into v?

tests/vector/vector_parallel.pf
vbaconnet: Wow that's weird, my `make check` locally did not pick up all of those errors. Maybe a stupid question but how can I access the logs produced by the github checks?

timofeymukha: You cant get the logs for everything but for the check step we produce an "artifact", when you click on checks it will be below that fancy diagram connecting all the steps of the CI.

src/math/matrix.f90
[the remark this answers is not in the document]
timofeymukha: I would assume that for a derived type it first has to run the routine you override + with and the then the routine you override = with. Which would mean that we always copy, and perhaps limit the usability of these operators quite a lot. But I am not sure. Let's ask @njansson

tests/vector/vector_parallel.pf
vbaconnet: Nice, thanks! It seems my pFUnit is kind of messed up, in for example `vector_parallel.F90`, I have lines like:
```.f90
call assertEqual(42.0_rp, v%x(i), tolerance=NEKO_EPS, &
 & location=SourceLocation( &
 & 'vector_parallel.pf', &
 & 116) )

if (anyException()) return
```
My interpretation of this is that if there is an error, the program will return exit status 0 and `test-driver` will not pick it up as an error when it checks for the exit status `estatus=$?` 

```bash
# Test script is run here. We create the file first, then append to it,
# to ameliorate tests themselves also writing to the log file. Our tests
# don't, but others can (automake bug#35762).
: >"$log_file"
"$@" >>"$log_file" 2>&1
estatus=$?
```
Because if I change the `return` to `stop 99` it does pick it up. I guess that is probably what is happening for the github checks that pass?

njansson: If you haven't patched pfunit (look at our workflows), it will unfortunately not exit with an error code with gfortran. This was something they changed recently since people complained about getting stack traces upon errors

src/math/matrix.f90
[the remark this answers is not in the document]
njansson: unfortunately, afaik that is how bad it will be (T_T)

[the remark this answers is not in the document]
timofeymukha: Sad times. But then I guess it is better to give up on operator overloading and transform the implemented functions to subroutines, where v is one of the dummy arguments. So it will be like a small subset of field_math but for vectors and matrices. It is really a shame, though...
Or we can leave the overloading, but warn that it should only be used for small `n`.

src/math/vector.f90
njansson: I agree, getting rid of copy would be nice, and use the intrinsics

src/math/matrix.f90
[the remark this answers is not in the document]
vbaconnet: Indeed having a copy every time is quite annoying. My idea was to do something similar to field math, how about having type bound procedure for in-place operations? something like:

```.f90
  !> to be called like this%add(v)
  subroutine vector_add(this, v)
    class(vector_t), intent(inout) :: this
    type(vector_t), intent(in) :: v

    if (v%n .ne. this%n) call neko_error("Vectors must be the same size!")

       call device_add2(this%x_d, v%x_d, v%n)
    else
       call add2(this%x, v%x, v%n)
    end if

  end subroutine vector_add
```

But perhaps that is a bit ugly. Reflecting a bit, would it be interesting to replicate some kind of `field_math` thingy with vector and matrix? Since it's basically what this PR is doing?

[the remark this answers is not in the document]
njansson: But with a different type, thus closer to the idea of a generic math interface. As discussed elsewhere,  this is of course great for end users (and developers), as long as we ensure that we don't use these wrappers in performance critical code (due to lack of inlining capabilities, bad when e.g. we have a bunch of `add2` calls)

njansson:
I would say we should merge this, and I can take upon me to fix the generic interface (+ ensure correct inlining in key kernels)

njansson (approved):

timfelle (approved):

PR #1377, opened 2024-07-18
Update indentation of logs

timfelle:
Ensure the neko header is not indented.

Reduce the log default indentation to 0 instead of the entire log being indented by 1 space.

### DIFF
--- a/src/common/log.f90
+++ b/src/common/log.f90
@@ -115,7 +115,7 @@ contains
 
     if (pe_rank .eq. 0) then
        do i = 1, this%indent_
-          write(*,'(A)', advance='no') ' '
+          write(*,'(A)', advance = 'no') ' '
        end do
     end if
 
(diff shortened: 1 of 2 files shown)


timofeymukha:
Makes sense, I guess. Was there some reason to start with a space?

timofeymukha (approved):

timfelle:
> Makes sense, I guess. Was there some reason to start with a space?

The header had an additionaly space added for some wiered reason that i could not figure out. `write (*,*)` prepends a space `write (*,'(A)')` did not.
I expect the extra space in the logger was to make it consistent with the header.

njansson:
The space is added by Fortran, so a reason to keep it would be that user added write statement wouldn't be too far off

timfelle:
> The space is added by Fortran, so a reason to keep it would be that user added write statement wouldn't be too far off  

Lets close it then

PR #1378, opened 2024-07-18
Reduce probe output in the log file

vbaconnet:
Fixes a maximal number of probes to write to the log file to not flood the output if we use a large number of probes with the variable `NEKO_MAX_LOG_PROBES`.

### DIFF
--- a/src/simulation_components/probes.F90
+++ b/src/simulation_components/probes.F90
@@ -572,7 +572,7 @@ contains
   subroutine probes_show(this)
     class(probes_t), intent(in) :: this
     character(len=LOG_SIZE) :: log_buf ! For logging status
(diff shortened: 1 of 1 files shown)


timofeymukha:
I would honestly put the whole output under the DEBUG log level.

timfelle:
> I would honestly put the whole output under the DEBUG log level.

Thats a good point. I guess if one is interested in reading through the log for the probe, then one is probably debugging something.

vbaconnet:
Ok! I agree that it makes more sense, then I just removed the max limit on the printing and only output with the debug level

timofeymukha (approved):

adperezm (approved):

PR #1381, opened 2024-07-19
Output the commands used with genmeshbox in a separate log file

vbaconnet:
Added a log that shows which parameters were used with genmeshbox, to have a trace.

### DIFF
--- a/contrib/genmeshbox/genmeshbox.f90
+++ b/contrib/genmeshbox/genmeshbox.f90
@@ -97,13 +97,13 @@ write(*,*) 'This examples generates a cube (box.nmsh) with side length 1 and &
         ! if the original genmeshbox.log does not exist, we create new
         ! files with genmeshbox.log.1, .2, .3, etc
         if (i .lt. 10) then
-           write(log_fname,'(A,I1)') "genmeshbox.log.", i
+           write(log_fname,'(A,I1,A)') "genmeshbox_", i, ".log"
         else if (i .lt. 100) then
-           write(log_fname,'(A,I2)') "genmeshbox.log.", i
+           write(log_fname,'(A,I2,A)') "genmeshbox_", i, ".log"
         else if (i .lt. 1000) then
-           write(log_fname,'(A,I3)') "genmeshbox.log.", i
+           write(log_fname,'(A,I3,A)') "genmeshbox_", i, ".log"
         else
-           write(log_fname,'(A,I4)') "genmeshbox.log.", i
+           write(log_fname,'(A,I4,A)') "genmeshbox_", i, ".log"
         end if
 
      end do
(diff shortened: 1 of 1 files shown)


timofeymukha:
Naive question: would it be easier / better to print things into stdout by default, and then if the user wants it logged the standard linux output redirection can be used?

timfelle:
> Naive question: would it be easier / better to print things into stdout by default, and then if the user wants it logged the standard linux output redirection can be used?

The benefit of this approach is that one could separate run time messages from the parameters more easily.

Just a little fever dream:
A cool feature based on this could be to dump it as a json, which could then be reinserted into genmeshbox. This could allow us to recreate the exact mesh and the json could then follow the dataset so you dont have to remember how you generated it.

Once that is done, you could even put the specification into the casefile and have the neko driver call genmeshbox if the mesh specifications are there, but the mesh is not.

vbaconnet:
> Just a little fever dream: A cool feature based on this could be to dump it as a json, which could then be reinserted into genmeshbox. This could allow us to recreate the exact mesh and the json could then follow the dataset so you dont have to remember how you generated it.
>
> Once that is done, you could even put the specification into the casefile and have the neko driver call genmeshbox if the mesh specifications are there, but the mesh is not.


That does sound really nice :) Unfortunately I am not so well-versed with the json fortran stuff and perhaps this will be a  bit more work than I can do for now since this was supposed to be a quick fix... But I agree it would be super nice. 

I would suggest we add the json stuff as todo and either merge or we close this PR :)

njansson (approved):

contrib/genmeshbox/genmeshbox.f90
```diff
@@ -67,6 +72,43 @@ program genmeshbox
...
+        ! files with genmeshbox.log.1, .2, .3, etc
+        if (i .lt. 10) then
+           write(log_fname,'(A,I1)') "genmeshbox.log.", i
+        else if (i .lt. 100) then
+           write(log_fname,'(A,I2)') "genmeshbox.log.", i
+        else if (i .lt. 1000) then
+           write(log_fname,'(A,I3)') "genmeshbox.log.", i
+        else
+           write(log_fname,'(A,I4)') "genmeshbox.log.", i
+        end if
```
timfelle, line 107: Could we maybe format it as `genmeshbox_N.log` instead to avoid modifying the file extension.

vbaconnet: sure!

MartinKarp (commented):
Ok to merge this one before before #1407 @adperezm?

adperezm:
> Ok to merge this one before before #1407 @adperezm?

Yeah, that sounds good

MartinKarp (approved):

PR #1385, opened 2024-07-22
Add full stress tensor and LES capability to fluid

timofeymukha:
This PR adds the capability to perform LES to fluid_pnpn scheme, making use of the full stress formulation and the coupled CG solver.

- The bulk of the PR is changes to the residual interface for the constant material properties case. It is changed to have mu and rho as fields to match the interface currently used for the full stress residuals.
- The residual types for both full and simplified stress formulations now derive from the same base type.
- A rho_field is added to fluid_scheme to allow for variable density. Currently, it is not used in any way, but I guess this is more future-proof.
- The les_simcomp is simplified, since the update of material properties in scalar and fluid are handled by the schemes. To that end, `update_material_properties` is used in fluid_pnpn. The routine was added in a previous PR.


timofeymukha:
Omg, flint approved directly :D.

src/simulation_components/les_simcomp.f90
```diff
@@ -69,12 +69,23 @@ subroutine les_simcomp_init_from_json(this, json, case)
...
     character(len=:), allocatable :: name
+    character(len=:), allocatable :: nut_field
+    character(len=20) :: fields(2)
 
     call this%free()
 
-    call json_get(json, "model", name)
+    ! Add fields keyword to the json so that the field_writer picks it up.
+    ! Will also add fields to the registry if missing.
+    call json_get(json, "nut_field", nut_field)
```
Shiyu-Sandy-Du, line 79: why we want a name for nut_field from json?

timofeymukha: We need it here in order to give the name to the field_writer simcomp, which takes care of saving it to disk. Generally, it is used to configure the name of the field that will store the turbulent viscosity. We could perhaps default to `nut`.

Shiyu-Sandy-Du: Yes I think read it from json, and if not found then default it to `nut` would be a good idea. :)

src/fluid/bcknd/cpu/pnpn_res_cpu.f90
```diff
@@ -58,22 +60,24 @@ subroutine pnpn_prs_res_cpu_compute(p, p_res, u, v, w, u_e, v_e, w_e, f_x, &
 
     n = c_Xh%dof%size()
 
-    do i = 1, n
-       c_Xh%h1(i,1,1,1) = 1.0_rp / rho
-       c_Xh%h2(i,1,1,1) = 0.0_rp
-    end do
+    call invers2(c_Xh%h1, rho%x, n)
+    call rzero(c_Xh%h2, n)
```
njansson, line 64: Along the lines of math routines and kernels, here I would suggest we keep the loop such that we can use openmp on cpus

src/fluid/bcknd/cpu/pnpn_res_cpu.f90
```diff
@@ -136,19 +141,17 @@ subroutine pnpn_vel_res_cpu_compute(Ax, u, v, w, u_res, v_res, w_res, &
...
     type(field_t), pointer :: ta1, ta2, ta3
     integer, intent(in) :: n
     integer :: i
 
-    do i = 1, n
-       c_Xh%h1(i,1,1,1) = mu
-       c_Xh%h2(i,1,1,1) = rho * (bd / dt)
-    end do
+    call copy(c_Xh%h1, mu%x, n)
+    call cmult2(c_Xh%h2, rho%x, bd / dt, n)
```
njansson, line 154: Same comment as above, would keep the loops as is to make both compiler and openmp happy

src/fluid/bcknd/sx/pnpn_res_sx.f90
```diff
@@ -58,25 +60,26 @@ subroutine pnpn_prs_res_sx_compute(p, p_res, u, v, w, u_e, v_e, w_e, f_x, &
 
     n = c_Xh%dof%size()
 
-    do i = 1, n
-       c_Xh%h1(i,1,1,1) = 1.0_rp / rho
-       c_Xh%h2(i,1,1,1) = 0.0_rp
-    end do
+    call invers2(c_Xh%h1, rho%x, n)
+    call rzero(c_Xh%h2, n)
```
njansson, line 64: Same as above, keep loops such that they can be fused and vectorised and openmpified

src/fluid/bcknd/sx/pnpn_res_sx.f90
```diff
@@ -137,19 +141,17 @@ subroutine pnpn_vel_res_sx_compute(Ax, u, v, w, u_res, v_res, w_res, &
...
     integer :: temp_indices(3)
     type(field_t), pointer :: ta1, ta2, ta3
     integer :: i
 
-    do i = 1, n
-       c_Xh%h1(i,1,1,1) = mu
-       c_Xh%h2(i,1,1,1) = rho * (bd / dt)
-    end do
+    call copy(c_Xh%h1, mu%x, n)
+    call cmult2(c_Xh%h2, rho%x, bd / dt, n)
```
njansson, line 154: Same as above, keep loops such that they can be fused and vectorised and openmpified

njansson (commented):
Nice!

My only comments would be regarding the multiple Ax and the mu, rho fields.

Ax is a minor thing, but somehow it feels a bit redundant for none-stress runs to have two (since both will be the same, and are written to support both pressure and velocity) 

Regarding the mu and rho fields, a bit picky, but having these as fields for none-stress runs would require a much more data movement, thus wouldn't it be better to have to pnpn solvers, one with mu/rho scalar and one with mu/rho as a field (to support stress)

src/fluid/bcknd/cpu/pnpn_res_cpu.f90
timofeymukha: We decided to kill these loops in the PR for the residual routines for the full stress, so I just followed suite here. I can revert of course, but we should figure out what to do with CPU kernels. Again, e.g. `c_Xh%h2 = 0.0_rp` is also a possibility.

timofeymukha:
> Regarding the mu and rho fields, a bit picky, but having these as fields for none-stress runs would require a much more data movement, thus wouldn't it be better to have to pnpn solvers, one with mu/rho scalar and one with mu/rho as a field (to support stress)

I guess the trajectory of the previous PRs was to have one solver. I think it would be better because there will otherwise be quite a bit of overhead in adding new things to the fluid solvers. You would pretty much always do the same changes in two types. But where will we have extra movement? Once we init the fields, doesn't everything get passed by reference?

njansson:
> > Regarding the mu and rho fields, a bit picky, but having these as fields for none-stress runs would require a much more data movement, thus wouldn't it be better to have to pnpn solvers, one with mu/rho scalar and one with mu/rho as a field (to support stress)
> 
> 
> 
> I guess the trajectory of the previous PRs was to have one solver. I think it would be better because there will otherwise be quite a bit of overhead in adding new things to the fluid solvers. You would pretty much always do the same changes in two types. But where will we have extra movement? Once we init the fields, doesn't everything get passed by reference?

There will be movement when loading from memory, before it was one scalar now there will be ndofs loads of the same value inside gpu kernels.

Maybe we could keep the kernels for none stress with scalars and only pass the first entry of the arrays

timofeymukha:
> Maybe we could keep the kernels for none stress with scalars and only pass the first entry of the arrays

Yeah, this we can certainly do.

timofeymukha:
@njansson @timfelle If you can look at this one in the next 2 days, it would be great -- then we can try to finish before I leave for vacation.

src/fluid/bcknd/cpu/pnpn_res_cpu.f90
```diff
@@ -59,21 +61,25 @@ subroutine pnpn_prs_res_cpu_compute(p, p_res, u, v, w, u_e, v_e, w_e, f_x, &
     n = c_Xh%dof%size()
 
     do i = 1, n
-       c_Xh%h1(i,1,1,1) = 1.0_rp / rho
-       c_Xh%h2(i,1,1,1) = 0.0_rp
+      c_Xh%h1(i,1,1,1) = 1 / rho%x(i,1,1,1)
```
timfelle, line 64:
```suggestion
      c_Xh%h1(i,1,1,1) = 1.0_rp / rho%x(i,1,1,1)
```

src/fluid/bcknd/sx/pnpn_res_sx.f90
```diff
@@ -59,24 +61,27 @@ subroutine pnpn_prs_res_sx_compute(p, p_res, u, v, w, u_e, v_e, w_e, f_x, &
     n = c_Xh%dof%size()
 
     do i = 1, n
-       c_Xh%h1(i,1,1,1) = 1.0_rp / rho
-       c_Xh%h2(i,1,1,1) = 0.0_rp
+      c_Xh%h1(i,1,1,1) = 1 / rho%x(i,1,1,1)
```
timfelle, line 64:
```suggestion
      c_Xh%h1(i,1,1,1) = 1.0_rp / rho%x(i,1,1,1)
```

timofeymukha:
Thanks @timfelle 

@Shiyu-Sandy-Du maybe you can also take another look?

njansson (commented):
Looks good, my only concerns are the additional memory loads for `rho` and `mu` (even on cpus). It's fine if we waste some memory in pnpn, but for the kernels these loads are not for free.

I've added some suggestions on how we can avoid it for cpu and sx backends (following what's already in place for devices)

Shiyu-Sandy-Du:
Nice job! I don't have many further suggestion but move the test filter setup in `compute` into `init` is good for performance. Apart from that, I think the rest is good!

Shiyu-Sandy-Du:
Two questions regarding accelerators for stress formulation:

1. Why don't we have `prs_stress_res_part2` for cuda?
2. We don't have hip backend for `prs_stress_res`, do we?

I might ignore some parts of the code, and will be grateful if someone remind me where they are :)

src/fluid/bcknd/cpu/pnpn_res_cpu.f90
```diff
@@ -57,23 +59,27 @@ subroutine pnpn_prs_res_cpu_compute(p, p_res, u, v, w, u_e, v_e, w_e, f_x, &
...
     do concurrent (i = 1:n)
-       ta1%x(i,1,1,1) = f_x%x(i,1,1,1) / rho &
-            - ((wa1%x(i,1,1,1) * (mu / rho)) * c_Xh%B(i,1,1,1))
-       ta2%x(i,1,1,1) = f_y%x(i,1,1,1) / rho &
-            - ((wa2%x(i,1,1,1) * (mu / rho)) * c_Xh%B(i,1,1,1))
-       ta3%x(i,1,1,1) = f_z%x(i,1,1,1) / rho &
-            - ((wa3%x(i,1,1,1) * (mu / rho)) * c_Xh%B(i,1,1,1))
+       ta1%x(i,1,1,1) = f_x%x(i,1,1,1) / rho_val &
+             - ((wa1%x(i,1,1,1) * (mu_val / rho_val)) * c_Xh%B(i,1,1,1))
+        ta2%x(i,1,1,1) = f_y%x(i,1,1,1) / rho &
```
njansson, line 79:
```suggestion
        ta2%x(i,1,1,1) = f_y%x(i,1,1,1) / rho_val &
```

njansson (approved):

PR #1387, opened 2024-07-22
Add logical operators and combine point zones

vbaconnet:
A little addition to the point zones, with a new `combine` point zone, that allows the combination of several point zones. I also add another attribute to the `point_zone_t` which allows to select the "inverse" of the point zone:

```json
"point_zones": 
[
    {
        "name": "mysphere",
        "geometry": "sphere",
        "center": [0.0, 0.0, 0.0],
        "radius": 0.01
    },

    {
        "name": "combined_box_and_sphere",
        "geometry": "combine",
        "operator": "OR",
        "subsets":
        [
            {
                "name": "mysphere"
            },
            {
                "name": "mybox",
                "geometry": "box",
                "x_bounds": [-1.0, 1.0],
                "y_bounds": [-1.0, 1.0],
                "z_bounds": [-1.0, 1.0]
            },

        ],
    },
    {
        "name": "inverted_combination",
        "geometry": "combine",
        "operator": "OR",
        "invert": true,
        "subsets":
        [
            {"name": "combined_box_and_sphere"},
        ],
    },

]
```

The combination of point zones can be done with a logical operation `OR`, `AND` or `XOR`. The default is "OR".


src/mesh/point_zones/combine_point_zone.f90
```diff
@@ -0,0 +1,215 @@
+! Copyright (c) 2019-2021, The Neko Authors
```
njansson, line 1:
```suggestion
! Copyright (c) 2024, The Neko Authors
```
If this is a new file?

njansson (approved):

src/mesh/point_zones/combine_point_zone.f90
```diff
@@ -0,0 +1,215 @@
...
+          allocate(sphere_point_zone_t::this%internal_zones(i)%pz)
+       else if (trim(type_name) .eq. "cylinder") then
+          allocate(cylinder_point_zone_t::this%internal_zones(i)%pz)
+       else
+          type_string =  concat_string_array(KNOWN_TYPES, NEW_LINE('A') // "-  ", &
+               .true.)
+          call neko_error("Unknown point zone type: " &
+               // trim(type_name) // ".  Known types are: " &
+               // type_string)
+       end if
```
timofeymukha, line 130: Question, is there a reason not to use the point zone factory here? Is there actually a limitation on not using a combined zone as a part of a combined zone? It would be cool if one could build a big monster by nesting combines :-).

src/mesh/point_zones/combine_point_zone.f90
```diff
@@ -0,0 +1,215 @@
...
+
+       ! Initialize the sub-point zone but do not map anything, here we
+       ! initialize the stack with just size 1.
+       call this%internal_zones(i)%pz%init(source_subdict, 1)
+
+    end do
+
+    ! Chcek that we got the proper operator
+    call json_get_or_default(json, "operator", this%operator, "OR")
+    select case (trim(this%operator))
```
timofeymukha, line 140: I would not default to anything and just use json_get. It's not like OR makes more sense as a default than say AND.

timofeymukha (commented):
Great stuff!

src/mesh/point_zones/combine_point_zone.f90
vbaconnet: Well, the initial reason was that I didn't manage to do it in a way that does not cause circular imports :( If the factory uses `combine` and `combine uses `factory`, is that not a problem?

Also, the factory needs a `dof` that is not part of `point_zone::init_base` which is also a bit annoying, but perhaps we could have another factory that looks like 

```.f90
point_zone_factory(object

File: src/mesh/point_zone_fctry.f90 [hunk 4]
```diff
@@ -75,6 +76,8 @@ subroutine point_zone_factory(object, json, dof)
        allocate(sphere_point_zone_t::object)
     else if (trim(type_name) .eq. "cylinder") then
        allocate(cylinder_point_zone_t::object)
+    else if (trim(type_name) .eq. "combine") then
+       allocate(combine_point_zone_t::object)
     else
        type_string =  concat_string_array(KNOWN_TYPES, NEW_LINE('A') // "-  ", &
```
timfelle, line 82:
```suggestion
       type_string =  concat_string_array(KNOWN_TYPES, new_line('A') // "-  ", &
```

I guess we dont need the old fortran of all caps

src/mesh/point_zones/combine_point_zone.f90
vbaconnet: Well, the initial reason was that I didn't manage to do it in a way that does not cause circular imports :( If the factory uses `combine` and `combine uses `factory`, is that not a problem?

timofeymukha: Yeah, I have not thought of that :(.

timfelle: Each point zone have a name. Why not let the combined pointzone just refer to the names of existing zones. Yes you need to create them in some order, but we can put expectations on the user to do it in a logical fashion.
That way the combined zone can combine with other combined zones without the circular dependency.

timofeymukha: Sounds like a good idea!

vbaconnet: You mean that the combined point zone doesn't build anything but just uses the registry? Doesn't that create a circular dependency between the factory and the registry? Factory imports combine, combine imports registry, and registry imports factory?

vbaconnet: perhaps I didn't understand what you meant :)))

timfelle: Yes, if you run the combination in the factory. But why would it? It could just as easy be a subroutine of the registry directly instead. So after in the initializer, after all the primitive zones have been constructed, you can run the combination constructor.

src/mesh/point_zones/combine_point_zone.f90
```diff
@@ -0,0 +1,215 @@
...
+     !> Number of zones to construct.
+     integer :: n_zones = 0
+     !> Operator with which to combine the point zones (AND, OR, XOR)
+     character(len=:), allocatable :: operator
+   contains
+     !> Constructor from json object file.
+     procedure, pass(this) :: init => combine_point_zone_init_from_json
+     !> Destructor.
+     procedure, pass(this) :: free => combine_point_zone_free
+     !> Defines the criterion of selection of a GLL point in the combine point zone.
```
timfelle, line 66:
```suggestion
     !> Defines the criterion of selection of a GLL point in the combine point 
     !! zone.
```

src/mesh/point_zones/combine_point_zone.f90
```diff
@@ -0,0 +1,215 @@
...
+
+       if (trim(type_name) .eq. "box") then
+          allocate(box_point_zone_t::this%internal_zones(i)%pz)
+       else if (trim(type_name) .eq. "sphere") then
+          allocate(sphere_point_zone_t::this%internal_zones(i)%pz)
+       else if (trim(type_name) .eq. "cylinder") then
+          allocate(cylinder_point_zone_t::this%internal_zones(i)%pz)
+       else
+          type_string =  concat_string_array(KNOWN_TYPES, NEW_LINE('A') // "-  ", &
+               .true.)
```
timfelle, line 126:
```suggestion
          type_string = concat_string_array(KNOWN_TYPES, &
               NEW_LINE('A') // "-  ", .true.)
```

src/mesh/point_zones/combine_point_zone.f90
```diff
@@ -0,0 +1,215 @@
...
+  !> Defines the criterion of selection of a GLL point in the combined point
+  !! zone.
+  !! @param x x-coordinate of the GLL point.
+  !! @param y y-coordinate of the GLL point.
+  !! @param z z-coordinate of the GLL point.
+  !! @param j 1st nonlinear index of the GLL point.
+  !! @param k 2nd nonlinear index of the GLL point.
+  !! @param l 3rd nonlinear index of the GLL point.
+  !! @param e element index of the GLL point.
+  pure function combine_point_zone_criterion(this, x, y, z, j, k, l, e) result(is_inside)
```
timfelle, line 179:
```suggestion
  pure function combine_point_zone_criterion(this, x, y, z, j, k, l, e) &
       result(is_inside)
```

src/mesh/point_zones/combine_point_zone.f90
vbaconnet: I think I managed to somehow put your idea to life :) But i do not use the registry directly in combine, I give `combine` an extra array of point zone pointers, such that `combine%init` reads the names and allocates the array of pointers. 

The registry then takes care of assigning the pointers in `combine` in the point zone pointer array with the subroutine `build_combine_point_zone`

vbaconnet: Ok I made it now so that one can do both, construct a new point zone one and/or refer to already existing point zones. And the logic is that point zones created internally are not actually allocated but it is only their `criterion` which is used

src/mesh/point_zones/combine_point_zone.f90
```diff
@@ -0,0 +1,263 @@
...
+  use logger, only: neko_log
+  use point_zone_fctry, only: point_zone_factory
+  implicit none
+  private
+
+  ! List of all possible types that one can combine
+  character(len=20) :: KNOWN_TYPES(3) = [character(len=20) :: &
+       "box", &
+       "sphere", &
+       "cylinder"]
```
timfelle, line 52: Seem to be unused

src/mesh/point_zones/combine_point_zone.f90
```diff
@@ -0,0 +1,263 @@
...
+  !! @param json Json object file.
+  !! @param size Size with which to initialize the stack
+  subroutine combine_point_zone_init_from_json(this, json, size)
+    class(combine_point_zone_t), intent(inout) :: this
+    type(json_file), intent(inout) :: json
+    integer, intent(in) :: size
+
+    ! Json low-level manipulator.
+    type(json_core) :: core
+    ! Pointer to the source_terms JSON object and the individual sources.
```
timfelle, line 93: Seems like there is some artifacts from where you copied the framework

src/mesh/point_zones/cylinder_point_zone.f90
timfelle, line 1: It seems like you forgot to actually use the inversion flag in the criterion function.

timfelle: Ah i see, you put it in the point_zone.f90.
Maybe we should have a more generic base initializer so stuff like the name and
inversion flag cannot be forgotten by someone adding a new zone type?

vbaconnet: How about adding the inversion flag to the base initialiser? 

```.f90
  subroutine point_zone_init_base(this, size, name, inverse)
    class(point_zone_t), intent(inout) :: this
    integer, intent(in), optional :: size
    logical, intent(in) :: inverse

    call point_zone_free_base(this)

    if (present(size)) then
       call this%scratch%init(size)
    else
       call this%scratch%init()
    end if

    this%name = trim(name)
    this%inverse = inverse

  end subroutine point_zone_init_base

```

timfelle: Sure thing, that's probably a good way to do it, without modifying the interface too much.

timfelle (approved):

PR #1388, opened 2024-07-23
performance fix in gmres

pohm01:


njansson (approved):
Great stuff! 🍻 🐈

Shiyu-Sandy-Du (approved):

njansson:
@timfelle Seems like we have another hickup with flinter preprocessed files, it will 💣 on `#endif` asking it to be expanded with a space like `#end if`

timfelle:
> @timfelle Seems like we have another hickup with flinter preprocessed files, it will 💣 on `#endif` asking it to be expanded with a space like `#end if`

On it, tim mentioned it yesterday as well

njansson:
> > @timfelle Seems like we have another hickup with flinter preprocessed files, it will 💣 on `#endif` asking it to be expanded with a space like `#end if`
> 
> On it, tim mentioned it yesterday as well. I'll bundle it with another cornercase he found that i haven't sorted just yet

🍺 🍻 🐈‍⬛

PR #1389, opened 2024-07-23
Missing stack init in overlap computation

njansson:


src/mesh/aabb_tree.f90
```diff
@@ -609,7 +630,8 @@ subroutine aabb_tree_query_overlaps(this, object, object_index, overlaps)
 
        if (this%nodes(node_index)%aabb%overlaps(object_box)) then
           if (this%nodes(node_index)%is_leaf()) then
-             if (.not. this%nodes(node_index)%object_index == object_index) then
+             if (.not. &
```
timfelle, line 633:
```suggestion
             if (this%nodes(node_index)%object_index .ne. object_index) then
```

njansson: Somethings up with gh, had to fix this with a manual push

timfelle: Yea dont know what went wrong with that

timfelle (approved):

timofeymukha (approved):

PR #1394, opened 2024-07-25
Fix bug in probes and wrong documentation

vbaconnet:
As the title says, there was a bug causing probes to not work when using other the new inputs like `line`, `circle`, etc.


njansson (approved):
🎉

Shiyu-Sandy-Du:
Nice job! Since `this%seq_io` is turned on in `probes_init_from_json` when calling another subroutine, it's not that obvious to see it in `probes_init_from_attributes`.

Shiyu-Sandy-Du (approved):

timfelle (approved):

PR #1404, opened 2024-08-01
Fix invalid loop bounds for triangle elements

njansson:


src/mesh/tri.f90
```diff
@@ -157,7 +157,7 @@ function tri_diameter(this) result(res)
     p2 => this%p(2)
     p3 => this%p(3)
 
-    do i = 1, NEKO_TRI_GDIM
+    do i = 1, NEKO_TRI_NPTS
        d1 = d1 + (p2%x(i) - p1%x(i))**2
        d2 = d2 + (p3%x(i) - p2%x(i))**2
        d3 = d3 + (p1%x(i) - p3%x(i))**2
```
timfelle, line 163: Are you sure you want to change this?
Arrent we looping through the entries of x and not the list of points?

njansson: 100% sure, it gave me a flat earth pancake instead of a sphere

timfelle: I agree that it was wrong to use the dimensionality of the triangle (2D) and not
the space it is embedded in (3D). But we should make a new variable or simply
write 3, since we dont use the number of points

njansson: fair enough, it's correct for the diameter

src/mesh/tri.f90
```diff
@@ -157,7 +157,7 @@ function tri_diameter(this) result(res)
     p2 => this%p(2)
     p3 => this%p(3)
 
-    do i = 1, NEKO_TRI_GDIM
+    do i = 1, NEKO_TRI_NPTS
```
njansson, line 160:
```suggestion
    do i = 1, NEKO_TRI_GDIM
```

timfelle: I think you are right to loop across all 3 spatial dimensions. otherwise you compute the diameter of the triangle projected into the x,y-plane, which is not what you want.

src/mesh/tri.f90
```diff
@@ -179,7 +179,7 @@ function tri_centroid(this) result(res)
     p3 => this%p(3)
     res%x = 0d0
 
-    do i = 1, this%gdim()
+    do i = 1, NEKO_TRI_NPTS
```
timfelle, line 182:
```suggestion
    do i = 1, size(p1%x)
```

You are not looping through the points of the triangle, but the spatial coordinates.

njansson: but that is the same as `NEKO_TRI_NPTS`, or maybe it's getting too late here?

timfelle: I know the value stored are the same, but it makes it difficult to read for the user since we indicate that we are looping through something we are not.
If someone were to use this as the base for a quad, it would now segfault because it tries to loop through 4 spatial dimensions.

timfelle: So maybe we could introduce `NEKO_TRI_SDIM` to differentiate between the geometric and spatial dimension?

njansson: Sure, sounds good

src/mesh/tri.f90
timfelle, line 1: @njansson I have added a paramter for the space the triangle is embedded in,
essentially always 3D and used that across the code. Please resolve this
conversation if you agree

timfelle (approved):

timfelle (approved):

timfelle:
@timofeymukha and @MartinKarp, This Pr seem to have been forgotten, could we get it in?

MartinKarp (approved):
Looks good. A heads up again that subroutines such as getting the diameter and centroid only work on non-deformed elements. This is I guess just a general note if someone tries to use these element routines after having combined the element with a function space of higher-order than 1.

PR #1407, opened 2024-08-01
Added posibility to provide a point distribution

adperezm:
As the title says.

Added the possibility to add the distribution of points. Check if you agree with this. I think I replicate many operations to somehow be consistent, but the code can be written better/(more efficiently)


MartinKarp (approved):
This is really nice! I think the duplication is fine for this, it is easy and short enough to understand anyway.

contrib/genmeshbox/genmeshbox.f90
```diff
@@ -66,24 +79,86 @@ program genmeshbox
...
+  end if
+
+  if (trim(dist_y_fname) .eq. 'uniform') then
+    el_len_y(:) = (y1 - y0)/nely
+  else
+    dist_y_file = file_t(trim(dist_y_fname))
+    call dist_y%init(nely+1)
+    call dist_y_file%read(dist_y)
+    do i = 1, nely
+      el_len_y(i) = dist_y%x(i+1) - dist_y%x(i)
```
timofeymukha, line 118: A slight concern here, can we get a huge loss in precision due to subtracting very similar numbers (e.g. in the boundary layer)?

adperezm: I think an obvious change to address this is to just give as an input the spacing between points instead of the locations as is now.

I have not checked how other codes do/deal with this

MartinKarp: I think that would be unlikely as it is the element lengths, not the distance between gll points that are computed.

timofeymukha: Yeah, Martin is probably right.

timofeymukha (approved):

contrib/genmeshbox/genmeshbox.f90
```diff
@@ -6,9 +6,9 @@ program genmeshbox
   implicit none
 
   character(len=NEKO_FNAME_LEN) :: inputchar, fname='box.nmsh'
```
timfelle, line 8:
```suggestion
  character(len=NEKO_FNAME_LEN) :: inputchar, fname = 'box.nmsh'
```

adperezm: Noooo, please I will fix it! :)

PR #1419, opened 2024-08-09
Add enforcement of `only` specifications.

timfelle:
According to our guidelines we must use `only` when using modules.
https://neko.cfd/docs/develop/d0/d47/dev_patterns.html#autotoc_md16

This PR will enforce this.

Closes: #1418

### DIFF
--- a/doc/pages/developer-guide/dev_patterns.md
+++ b/doc/pages/developer-guide/dev_patterns.md
@@ -89,9 +89,10 @@ base as of now, but we should at least make all new code as clean as possible.
 
 ## B. Scope
 
-1. Always use `only` when `using` something from another module.  The `neko`
-   module is an exception and imports everything. 
-   
+1. Always use `only` when `using` something from another module. The `neko`
+   module is an exception and imports everything. Additionally, current compiler
+   limitations make it impossible to use `only` with `comm` and `device`.
+
    This latter is done so that the user `.f90` files need only `use` the `neko`
    module to get access to everything.
 
--- a/flinter_rc.yml
+++ b/flinter_rc.yml
@@ -136,6 +136,12 @@ regexp-rules:
     replacement: null
     active: true
 
+  scope-not-specified:
+    message: Module usage should be explicit
+    regexp: ((use +)(?!\bneko\b)\w+\b)(?!, only)
+    replacement: null
+    active: true
+
   # -------------------------------------------------------------------------- #
   # Formatting and code style rules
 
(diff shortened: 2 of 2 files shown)


njansson:
We need to exclude enforcing `only` for `comm` and `device` due to various observed compiler issues

timfelle (commented):
> We need to exclude enforcing `only` for `comm` and `device` due to various observed compiler issues  

done

timfelle (commented):
> We need to exclude enforcing `only` for `comm` and `device` due to various observed compiler issues  

done

njansson:
> > We need to exclude enforcing `only` for `comm` and `device` due to various observed compiler issues
> 
> done

Great!

Sorry but I forgot one thing, external modules is probably also a good idea to exclude, like `mpi_f08` and `hdf5`, if not also due to the issues we saw before with `comm` and `device`

timfelle (commented):
> Great!

> Sorry but I forgot one thing, external modules is probably also a good idea to exclude, like `mpi_f08` and `hdf5`, if not also due to the issues we saw before with `comm` and `device`  

What are the issues? It makes no sense that it would cause issues? Is it bugs in the compilers, or are some of our code relying on some implicit inclusions somewhere?

njansson:
> > Great!
> 
> > Sorry but I forgot one thing, external modules is probably also a good idea to exclude, like `mpi_f08` and `hdf5`, if not also due to the issues we saw before with `comm` and `device`
> 
> What are the issues? It makes no sense that it would cause issues? Is it bugs in the compilers, or are some of our code relying on some implicit inclusions somewhere?

For example: https://github.com/ExtremeFLOW/neko/pull/1179

This is not related to implicit inclusion, this is something worse

timfelle:
> > > Great!
> > 
> > 
> > > Sorry but I forgot one thing, external modules is probably also a good idea to exclude, like `mpi_f08` and `hdf5`, if not also due to the issues we saw before with `comm` and `device`
> > 
> > 
> > What are the issues? It makes no sense that it would cause issues? Is it bugs in the compilers, or are some of our code relying on some implicit inclusions somewhere?
> 
> For example: #1179
> 
> This is not related to implicit inclusion, this is something worse. In the above issue, the compiler couldn't figure out the correct MPI constants, so reductions simply became zeros...this was a very difficult bug to track down

I added an issue to track our progress, feel free to add more information in there if you have any.
https://github.com/ExtremeFLOW/neko/issues/1420

njansson (approved):

timofeymukha (approved):

PR #1431, opened 2024-08-16
Fixed dangling pointers

njansson:
A couple of dangling pointers found by NAG Fortran

### DIFF
--- a/src/scalar/scalar_user_source_term.f90
+++ b/src/scalar/scalar_user_source_term.f90
@@ -66,11 +66,11 @@ module scalar_user_source_term
      !> Device pointer for `s`.
      type(c_ptr) :: s_d = C_NULL_PTR
      !> Compute the source term for a single point
-     procedure(scalar_source_compute_pointwise), nopass, pointer :: compute_pw_ &
-       => null()
+     procedure(scalar_source_compute_pointwise), nopass, pointer :: &
+          compute_pw_ => null()
      !> Compute the source term for the entire boundary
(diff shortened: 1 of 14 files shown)


timofeymukha (approved):

njansson:
@timfelle 
Format strings as in:
https://github.com/ExtremeFLOW/neko/blob/4e03353765effd37387450e0923f5b7ee4145c3b/src/simulation_components/probes.F90#L613 

is another challenge for the linter, e.g as in the example above, where one might not want spaces after `,` in the output

src/scalar/scalar_user_source_term.f90
```diff
@@ -163,7 +164,8 @@ subroutine scalar_user_source_term_init_from_components(this, fields, coef, &
     if (trim(source_term_type) .eq. 'user_pointwise' .and. &
               present(eval_pointwise)) then
        if (NEKO_BCKND_DEVICE .eq. 1) then
-          call neko_error('Pointwise source terms not supported on accelerators')
+          call neko_error('Pointwise source terms not &
+               supported on accelerators')
```
njansson, line 168:
```suggestion
&supported on accelerators')
```

timfelle:
> @timfelle 
> Format strings as in:
> https://github.com/ExtremeFLOW/neko/blob/4e03353765effd37387450e0923f5b7ee4145c3b/src/simulation_components/probes.F90#L613 
> 
> is another challenge for the linter, e.g as in the example above, where one might not want spaces after `,` in the output 

Yes, i have it in the "known issues". I think the mixing of string delimiters make it fail at the moment.

PR #1434, opened 2024-08-21
Move imp of opr to submodules

njansson:
WIP of moving operators to submodules.

In general seems to work fine, but `make depend` can't automatically figure out dependencies


njansson:
@timfelle @timofeymukha 
Looks promising, but `make depend` is as mentioned an issue (one has to manually add the main module as a dependency)

timfelle (commented):
@njansson, Interesting, which version of makedepf90 do you have installed? Mine does it just fine.

njansson:
> @njansson, Interesting, which version of makedepf90 do you have installed? Mine does it just fine.

I've v2.8.8

timfelle:
> > @njansson, Interesting, which version of makedepf90 do you have installed? Mine does it just fine.
> 
> I've v2.8.8

Mine is 2.9.0 and seem to work fine for it.

njansson:
> > > @njansson, Interesting, which version of makedepf90 do you have installed? Mine does it just fine.
> > 
> > 
> > I've v2.8.8
> 
> Mine is 2.9.0 and seem to work fine for it. Im qurious about the CI thing we set up then

Hmm, I can only find v2.8.9 out in the wild. Where did you find v2.9.0?

timfelle:
> > > > @njansson, Interesting, which version of makedepf90 do you have installed? Mine does it just fine.
> > > 
> > > 
> > > I've v2.8.8
> > 
> > 
> > Mine is 2.9.0 and seem to work fine for it. Im qurious about the CI thing we set up then
> 
> Hmm, I can only find v2.8.9 out in the wild. Where did you find v2.9.0?

Well I just installed it through `apt-get`.
But when I attempt to update it claims my makedepf90 version is 3.0.0-1. 
But calling `makedepf90 --version` claims 2.9.0

src/math/bcknd/cpu/opgrad.f90
```diff
@@ -31,25 +31,176 @@
...
+              drdx(1,1,1,e_start), dsdx(1,1,1,e_start), dtdx(1,1,1,e_start), &
+              drdy(1,1,1,e_start), dsdy(1,1,1,e_start), dtdy(1,1,1,e_start), &
+              drdz(1,1,1,e_start), dsdz(1,1,1,e_start), dtdz(1,1,1,e_start), &
+              Xh%w3, e_len)
+      case (17)
+         call cpu_opgrad_lx17(ux, uy, uz, u, &
+              Xh%dx, Xh%dy, Xh%dz, &
+              drdx(1,1,1,e_start), dsdx(1,1,1,e_start), dtdx(1,1,1,e_start), &
+              drdy(1,1,1,e_start), dsdy(1,1,1,e_start), dtdy(1,1,1,e_start), &
+              drdz(1,1,1,e_start), dsdz(1,1,1,e_start), dtdz(1,1,1,e_start), &
```
njansson, line 71: I think we should accept constructs like these when they are arguments to functions, since we already supports it in normal expressions (@timfelle @timofeymukha )

njansson:
Not sure what it is, byt the linter is still unhappy, despite us allowing `x(j,k,l)` expressions. Maybe it's due to `kk` two letter loop variables?

timfelle:
> Not sure what it is, byt the linter is still unhappy, despite us allowing `x(j,k,l)` expressions. Maybe it's due to `kk` two letter loop variables?

Yes, what we set up was that we allow single character variables or double digit numbers to be used without the space.

src/math/bcknd/cpu/opgrad.f90
timfelle: For the linting you mean?

njansson: Yes, but I guess it's the same as with `kk`, it can't handled `e_start`

njansson:
Now the issue is a big ICE with fj, which is kind of a blocker. Not sure I'm in favour of a merge unless I can find a workaround and add a patch (similar to what is needed for Nvidia) to the repo

src/math/bcknd/cpu/opgrad.f90
timfelle: We would have to change the rules for spaces after the `,` to change it yes.

timfelle:
> Now the issue is a big ICE with fj, which is kind of a blocker. Not sure I'm in favour of a merge unless I can find a workaround and add a patch (similar to what is needed for Nvidia) to the repo

Are we certain it's not just some implicitly imported stuff which is now broken? I see a lot of that when i start adding only statements.

src/math/bcknd/cpu/opgrad.f90
njansson: Ok, then I would change the names

timfelle: Well it would have to be a single letter name to avoid the space which i think would deduct quite a bit of readability.

njansson:
> > Now the issue is a big ICE with fj, which is kind of a blocker. Not sure I'm in favour of a merge unless I can find a workaround and add a patch (similar to what is needed for Nvidia) to the repo
> 
> Are we certain it's not just some implicitly imported stuff which is now broken? I see a lot of that when i start adding only statements.

It's certainly related `only` but rather the opposite, it will not go belly up 🐋  if `only` is removed (all these are related to `operators`

timfelle (commented):
Should we clean the imports as well?

njansson:
Good news #1442 seems to fix the ICE with Fujitsu, so I'm fine removing the don't merge label here

Issue #1405, opened 2024-08-01
Known linter issues


timfelle:
- [ ] Double spaces should be allowed inside strings.
- [ ] Comment spacing rule triggered in strings ended with line breaks.

MartinKarp:
It might just be me, maybe I installed the wrong version, but the options of flint that I have locally are: `config, dump, score, and tree`.

So when I run `flint lint` I get an error.

My flinter is located in:
```
$ whereis flint
flint: /home/martin/anaconda3/bin/flint
```
and my version should be
```
$ conda list flint
# packages in environment at /home/martin/anaconda3:
#
# Name                    Version                   Build  Channel
flinter                   0.4.0                    pypi_0    pypi
```

timfelle:
Yes, if you look at the CI you will see that we actually use 2 different versions of flint.

The latest Flint crashes with some of the complex recursive structures in the `stack` and `htable`. Therefore we use version 0.4.0 to `score` the entire repository.

We use the latest Flint to `lint` all the changed files, since the output is a lot more readable than `0.4.0`.

If you upgrade your flint then the `lint` command will work. Otherwise I seem to recall that you need to use `dump` for `0.4.0`, since score only works for directories I think.

MartinKarp:
Thanks! How should we do with the docs you think? Right now its 0.4.0, but lint would be more helpful for contributors.

PR #1437, opened 2024-08-25
Fix invalid module dependencies

njansson:


timfelle (commented):
Linter is broken for allocation statements with "::". It always triggers it seems. Issue #1405 have been updated to track the issue

timofeymukha (approved):

timfelle:
I guess `json_file` is no longer implicitly made available when asking for `use neko`. Do we want to fix it by explicitly adding `use json_module, only: json_file` to `neko.f90` or do we wish that users ask for the json library manually since it is an external dependency?

njansson:
> I guess `json_file` is no longer implicitly made available when asking for `use neko`. Do we want to fix it by explicitly adding `use json_module, only: json_file` to `neko.f90` or do we wish that users ask for the json library manually since it is an external dependency?

I think the best would be in neko, such that we minimize the amount of boilerplate code for users of the user interface

timfelle (approved):
Well I guess you might want to manually merge this to avoid all the missing only statements in `neko.f90`

PR #1438, opened 2024-08-26
Fix for 'sym' and 'on' bc

MartinKarp:
Perhaps not optimal, maybe one should change the the init to something like mark components? 

Can take that in another PR though so we have this working. 

I have tried to check if there are more similar bugs, but couldn't find any.

Let's continue to be on the lookout for similar things!

Moves the urgent bugfix out of #1255 for on and symmetry, but the geometric exceptions with edges on the boundary is still there.


timofeymukha:
I didn't quite understand, why can one not init base here? One should probably at least point this%coeff to coef manually then, right?

Regarding renaming init, can we wait with that a bit? I need to sync my bc pr with the latest develop once I am back and probably start issuing some PRs with cherrypicks from it. I want to think a bit about how to deal with these composite bcs..

adperezm:
On this. I think we have talked before, but why do we oppose to call the case file <casefile>.json?

Then you get stuff like this directly if a comma is missing:
image

Just wondering. A colleague apparently spent hours trying to debug this sort of error. This reformulation of the message would help for sure though.

MartinKarp:
Yes lets wait. The reason is here

https://github.com/ExtremeFLOW/neko/blob/93cb557022456295d8963d2dfdece2bda69c0228/src/bc/bc.f90#L201C1-L220C34

timofeymukha:
But dont we need to run this code form the constructor somewhere for the object to be valid?

````
    this%dof => coef%dof
    this%coef => coef
    this%Xh => this%dof%Xh
    this%msh => this%dof%msh
````

**Comment by MartinKarp** [comment]
It is done before we add the facets

**Comment by timofeymukha** [comment]
> It is done before we add the facets into marked_facets.
> 
> Procedure right now is
> 
> ```
> init_base
> mark facets with some operation
> finalize
> init(sym,on)
> ```
> 
> This is not very nice of course, but it is what we have currently. So we cannot do init_base after add facets. Here it is for 'sym'.
> 
> https://github.com/ExtremeFLOW/neko/blob/a213e423c3fd962fdeabfc3572a11f91c538c850/src/fluid/fluid_scheme.f90#L417

Ok, thanks for the clarification. I am hoping this will be solved by having deferred init eventually. I will have to keep in mind of this.

**Review by timofeymukha — APPROVED** [review]

**Comment by njansson** [comment]
> > It is done before we add the facets into marked_facets.
> 
> > 
> 
> > Procedure right now is
> 
> > 
> 
> > ```
> 
> > init_base
> 
> > mark facets with some operation
> 
> > finalize
> 
> > init(sym,on)
> 
> > ```
> 
> > 
> 
> > This is not very nice of course, but it is what we have currently. So we cannot do init_base after add facets. Here it is for 'sym'.
> 
> > 
> 
> > https://github.com/ExtremeFLOW/neko/blob/a213e423c3fd962fdeabfc3572a11f91c538c850/src/fluid/fluid_scheme.f90#L417
> 
> 
> 
> Ok, thanks for the clarification. I am hoping this will be solved by having deferred init eventually. I will have to keep in mind of this.

But free should be fine to call or?

**Comment by MartinKarp** [comment]
The issue is that free_base makes this%marked_facet empty. So one will never mark the different components of the bc when calling the init that's specific for sym and on

File: src/bc/non_normal.f90 [hunk 1]
```diff
@@ -62,9 +62,6 @@ subroutine non_normal_init(this, coef)
     type(tuple_i4_t) :: bc_facet
     integer :: facet, el
 
-    call this%free()
-
-    call this%init_base(coef)
     call this%bc_x%init_base(this%coef)
```
njansson, line 65: If we can't call free due to free_base, we should at least do some housekeeping to ensure that at least our object is clean 
```suggestion
    call this%bc_x%free()
    call this%bc_y%free()
    call this%bc_z%free()
    
    call this%bc_x%init_base(this%coef)
```

src/bc/symmetry.f90
```diff
@@ -70,9 +70,6 @@ subroutine symmetry_init(this, coef)
     type(tuple_i4_t) :: bc_facet
     integer :: facet, el
 
-    call this%free()
-
-    call this%init_base(coef)
     call this%bc_x%init_base(this%coef)
```
njansson, line 73:
```suggestion
    call this%bc_x%free()
    call this%bc_y%free()
    call this%bc_z%free()
    
    call this%bc_x%init_base(this%coef)
```

timfelle:
> The issue is that free_base makes this%marked_facet empty. So one will never mark the different components of the bc when calling the init that's specific for sym and on

Should marking the facets not be a part of the initialization? Modifying parts of an object before initialization is usualy not a good approach

timofeymukha:
> > The issue is that free_base makes this%marked_facet empty. So one will never mark the different components of the bc when calling the init that's specific for sym and on
> 
> Should marking the facets not be a part of the initialization? Modifying parts of an object before initialization is usualy not a good approach

In principle, yes, but the issue is that we have to allow for marking the facets to be a step by step process. E.g. we start by marking one boundary, then add some more facets etc. This is the reason currently we have an init, then marking of the facets, and then a finalize, which finishes up the construction of the object. In #1264 I make use of the fact that usually the facets are just one zone, so a bc factory inits, marks, and finalizes based on the type of bc and the zone index.

src/bc/symmetry.f90
```diff
@@ -70,9 +70,10 @@ subroutine symmetry_init(this, coef)
     type(tuple_i4_t) :: bc_facet
     integer :: facet, el
 
-    call this%free()
-
-    call this%init_base(coef)
+    call this%bc_x%free()
+    call this%bc_y%free()
+    call this%bc_z%free()
+
```
timfelle, line 76:
```suggestion

```

src/bc/non_normal.f90
```diff
@@ -62,9 +62,10 @@ subroutine non_normal_init(this, coef)
     type(tuple_i4_t) :: bc_facet
     integer :: facet, el
 
-    call this%free()
-
-    call this%init_base(coef)
+    call this%bc_x%free()
+    call this%bc_y%free()
+    call this%bc_z%free()
+
```
timfelle, line 68:
```suggestion

```

timfelle (changes requested):

MartinKarp:
Perhaps this should also be added to the latest release?

njansson:
> Perhaps this should also be added to the latest release?

Sounds good, let me bake a bug fix release

njansson (approved):

PR #1440, opened 2024-08-26
Add more intelligible error for when simulation dies instantly due to issue in the json file

MartinKarp:
Helps troubleshooting when the simulation dies due to `ERROR: Parameter case.mesh_file missing from the case file` and gives hint that the json file is incorrect.

### DIFF
--- a/src/case.f90
+++ b/src/case.f90
@@ -158,8 +158,8 @@ contains
     !
     call json_get_or_default(C%params, 'case.mesh_file', string_val, 'no mesh')
     if (trim(string_val) .eq. 'no mesh') then
-       call neko_error('No mesh found, often caused by that the json ' //&
-                       '(.case) file is incorrectly formatted somewhere.')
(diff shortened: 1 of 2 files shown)


src/case.f90
```diff
@@ -156,7 +156,11 @@ subroutine case_init_common(C)
     !
     ! Load mesh
     !
-    call json_get(C%params, 'case.mesh_file', string_val)
+    call json_get_or_default(C%params, 'case.mesh_file', string_val, 'no mesh')
+    if (trim(string_val) .eq. 'no mesh') then
+       call neko_error('No mesh found, often caused by that the json ' //&
```
timofeymukha, line 161: I suggest a slight reformulation: "The mesh_file keyword could not be found in the .case file. Often caused by incorrectly formatted json."

timofeymukha:
We may have to merge in dictator mode to dodge the :: error.

timfelle (approved):

PR #1441, opened 2024-08-26
Add timer for generating mesh connectivity

MartinKarp:
Adds time for generating mesh connectivity in the log. 

Helps when one has a big case and it takes a long time and one does not know if the simulation has frozen.

If we think this is a good idea perhaps we should add something for gs as well as that also sometimes takes some time. I was thinking one could perhaps generate the buffers from a previous gs_t to drive down the initialization time.


MartinKarp:
Is there any easy way the same output as I get here locally? While I can score the directory, I am not clear how I get the same error messages as when my PR fails, I get less detail and only a score per subroutine.

timofeymukha:
> Is there any easy way the same output as I get here locally? While I can score the directory, I am not clear how I get the same error messages as when my PR fails, I get less detail and only a score per subroutine.

Does using the same command as executed by the CI not work?

timofeymukha (approved):

njansson (approved):

PR #1442, opened 2024-08-27
Move factories to submodules

njansson:
Move factories to submodule under respective base type. Avoids circular dependencies, reduces use statements and compilation deps


timfelle (commented):
Looks really good. It seem like a great way of doing this.

njansson:
Clean build with fj, so I would suggest we proceed with this once CI is happy

adperezm:
May I ask, does this one affect OIF? One should try to get that in, no? It seems this one here touches a lot of files.

njansson:
> May I ask, does this one affect OIF? One should try to get that in, no? It seems this one here touches a lot of files.

Good point. 

These submodules will make that cleanup less painful

PR #1445, opened 2024-08-28
Make brinkman fields private

timfelle:


njansson (approved):

timfelle:
@njansson Seem like the MacOS machines are acting up. The linker complain about not having symbols for x86_64, are we sometimes compiling on arm chips with the MacOS?

njansson:
> @njansson Seem like the MacOS machines are acting up. The linker complain about not having symbols for x86_64, are we sometimes compiling on arm chips with the MacOS?

I suspect this is something related to the home-brew packages. I do daily builds on macOS, both arm and x64

timfelle:
> I suspect this is something related to the home-brew packages. I do daily builds on macOS, both arm and x64

Right, it was just since it's realated to jsonfortran which we compile manually.

njansson:
> > I suspect this is something related to the home-brew packages. I do daily builds on macOS, both arm and x64
> 
> Right, it was just since it's realated to jsonfortran which we compile manually.

Let's flush the caches and see if it fixes the issue with a fresh build

MartinKarp:
I got the same before and reran it and then those checked passed.

MartinKarp (approved):

Issue #1423, opened 2024-08-13
Renew the linter CI.


timfelle:
Martin suggested a change to the linting CI that could reduce the burden put on each PR while still converging to improved code.

- Lint new files according to the rulesets.
- Check for degradation of the repository wide score.
- If the PR reduce the repository score, then run linter for the changed files.

Additionally, we should let the linter run in parallel with the other tests to facilitate improvements to correctness of the code before the style.
This will additionally make it easier to judge if the linting errors can be allowed and the merge of a PR can be forced, while still having the code tested with all other tests.

timofeymukha:
We have this implemented, right @timfelle ?

PR #1448, opened 2024-08-28
Update linting process

timfelle:
This PR includes changes to improve the linting process and adds a dummy change to observe its effect.
We now only stop merger of a PR if it actively degrades the quality of the repository.
Otherwise we just report what can be improved.

Closes issue #1423


timfelle:
@njansson maybe its nice to squash this. Lots of kinda empty commits

timofeymukha:
So essentially the higher the repo score will be, the higher the requirements for incoming code, right :P?

.github/workflows/check_lint.yml
```diff
@@ -144,7 +170,7 @@ jobs:
                   grep -oP '(?<=\>\|)[^\|\<]+(?=\|\<)')
             printf ": $score\n"
 
-            if (($(echo "$score < $changed_min" | bc -l))); then
+            if (($(echo "$score < 10" | bc -l))); then
```
MartinKarp, line 173: Is this for all new files or only if the overall score does not deteriorate (regardless if new files were added or not)?

MartinKarp (commented):
This looks great! Just a question regarding how we handle adding new files.

timfelle:
> So essentially the higher the repo score will be, the higher the requirements for incoming code, right :P?

Sort of yes. We only check if the entire repository was degraded, we don't check for every file.
That could also be a way to do it. Make sure that any individual file changes are better than the average.

.github/workflows/check_lint.yml
timfelle: We only exit with an error for the CI if the score of the entire repository goes down.

But we always print a list of possible improvements, which is the 10 there.

timfelle: New files are not handled differently than old files here.

MartinKarp: Top! Sounds great to me.

MartinKarp (approved):

timofeymukha:
I guess there is some fairness in not demanding more than what we have now. But I feel like this is temporary in nature, since eventually we would push the score up, and so be in the same dilemma as now. I think in the end of the day, we need to decided what rules are really important to us and lint those hard. People seem to mostly be upset about spaces, for example. I would rather sacrifice some space rules, but be sure that all code is 80col and uses `only`. That said, I am not against merging this as.

timfelle:
> I guess there is some fairness in not demanding more than what we have now. But I feel like this is temporary in nature, since eventually we would push the score up, and so be in the same dilemma as now. I think in the end of the day, we need to decided what rules are really important to us and lint those hard. People seem to mostly be upset about spaces, for example. I would rather sacrifice some space rules, but be sure that all code is 80col and uses `only`. That said, I am not against merging this.

Yea over lunch i also considered making two different versions of the linter rules. Exactly as you suggest with one being a hard requirement always, and the other to be more lax. But I am still of the mind that if we have a rule, it must be followed by all. I am not against removing rules that we feel do not make sense, but I am against not enforcing the rules we decide to have.

njansson:
> @njansson maybe its nice to squash this. Lots of kinda empty commits

Sure, I'll put a squash commit

PR #1452, opened 2024-08-28
Make write of fld_file possible when not all ranks have an element 

MartinKarp:
Usecase for this is when some rank has 0 elements (which can be the case when one distributes a 2d mesh between ranks for example).

Perhaps we want to create a different communicator in this case, but this worked for 2d stats.

### DIFF
--- a/src/io/fld_file.f90
+++ b/src/io/fld_file.f90
@@ -405,96 +405,98 @@ contains
             (int(lxyz, i8) * &
             int(FLD_DATA_SIZE, i8))
     end do
+    
+    if (gdim .eq. 3) then
 
-
-    !> Include metadata with bounding boxes (Just copying from nek5000)
-    if (write_mesh) then
-       !The offset is:
-       ! mpioff + element_off * 2(min max value)
-       !           * 4(single precision) * gdim(dimensions)
-       byte_offset = int(mpi_offset, i8) + &
-            int(offset_el, i8) * &
-            int(2, i8) * &
-            int(MPI_REAL_SIZE, i8) * &
-            int(gdim, i8)
-       call fld_file_write_metadata_vector(this, fh, byte_offset, &
-            x%ptr, y%ptr, z%ptr, gdim, lxyz, nelv)
-       mpi_offset = int(mpi_offset, i8) + &
-            int(glb_nelv, i8) * &
-            int(2, i8) * &
-            int(MPI_REAL_SIZE, i8) * &
-            int(gdim, i8)
-    end if
-
-    if (write_velocity) then
-       byte_offset = int(mpi_offset, i8) + &
-            int(offset_el, i8) * &
-            int(2, i8) * &
-            int(MPI_REAL_SIZE, i8) * &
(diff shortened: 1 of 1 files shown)


njansson (approved):

src/io/fld_file.f90
```diff
@@ -115,6 +115,14 @@ subroutine fld_file_write(this, data, t)
 
     select type (data)
       type is (fld_file_data_t)
+       nelv = data%nelv
+       lx = data%lx
+       ly = data%ly
+       lz = data%lz
```
adperezm, line 121: Lz is equal to 1 in data when the mesh is 2d, right?

adperezm:
When the file is 2d, one does need to write the meta data. I guess visit/paraview will just ignore that, but it will make it a bit lighter if one does not include it.

Additionally, I do not know if you have noticed, but when you write a 2d file, one should avoid writing in the z velocity, as it will not be seen in visit (I do not know if the same happens in paraview). I do not know if a safety check should be added in the file directly or if this is something people should just know

pschlatt1:
maybe just to add to this (for my understanding): When you write 2D: Do you mean 2D2C (2D mesh with 2 componenents), or could there also be support for 2D3D (2D mesh with 3 components)? Nek5000 for instance had 2D2C, and I guess the visit reader too, but one could always trick it with a scalar if need be.

adperezm:
@pschlatt1 I guess you did not ask me, but based on just the file that is here, I would say this is 2D2C as you call it.

I did a quick test in the morning and in summary, I would just write everything as scalars when one does this :), that's what I ultimately decided to do in the python 2D averaging.

======================================================================================

If you want more details:

To write: 2D3D as you call it, can be achieved by keeping lz=1 but making gdim remain being 3. This will keep the offsets correct and write 3 fields in the _vector_write function. (One must ensure that the buffers are the correct size, so they must also be allocated with gdim = 3).

To read: The reader relies on lz to determine if the data is 3d (lz>1 => gdim=3). So if one leaves the reader unmodified, then the data read will be all scrambled, given that the vel components are written by element. One can avoid the problems if one makes a simple modification to the reader with an optional input that says "read 3 velocity fields", in which case one just sets gdim to be 3 as well.

But based on this: 2D3D can be done easily to contain data that we can post process, but to visualize, then one would also need to modify the visit reader. Perhaps by adding an extra parameter to the header, which I am not sure it can be done.

adperezm:
@pschlatt1 I have answered this message twice haha. I guess you can see those in the zulip log if you want a more detailed answer. 

So currently it is 2D2C. 2D3C can certainly be done, but I think we will just get more trouble from it, specially from the readers.

src/io/fld_file.f90
MartinKarp: yeah

MartinKarp:
Yeah, ill remove the thing for 2d to plot the metadata. I think we will always do 2d2c, and then if one wants to save the z field we will have to add a scalar as you say philipp.

adperezm (approved):

PR #1453, opened 2024-08-29
Not recomputing integration weights in cdtp

gloopydoop:
The integration weights are recomputed as w3 = B / jac in cdtp, but we can pass the integration weights directly.

The current implementation is incorrect (I'm pretty confident...) when working in a different space, ie, calling cdtp on the dealiased GL mesh for instance.

This PR only contains CPU and Cuda, but I'm happy to include the other backends if we agree to the change.


njansson:
Always good to save some redundant flops 🍻

njansson:
I need to rethink how this can be done efficiently on SX with the long vector lengths

njansson:
@gloopydoop If we can't use `w3` for certain backends, wouldn't it only be a matter of extra flops (unless for creative use with different function spaces?)

gloopydoop:
> @gloopydoop If we can't use `w3` for certain backends, wouldn't it only be a matter of extra flops (unless for creative use with different function spaces?)

To the best of my knowledge, it shouldn't make a difference to how neko operates currently. 
To work in different function spaces I would guess one would need to include B and jac in the `coef_init_empty` and map them in a similar fashion to drdx etc in the `init_dealias` or something to that effect. Passing `w3` seemed like a cleaner option.

njansson:
> > @gloopydoop If we can't use `w3` for certain backends, wouldn't it only be a matter of extra flops (unless for creative use with different function spaces?)
> 
> To the best of my knowledge, it shouldn't make a difference to how neko operates currently. To work in different function spaces I would guess one would need to include B and jac in the `coef_init_empty` and map them in a similar fashion to drdx etc in the `init_dealias` or something to that effect. Passing `w3` seemed like a cleaner option.

Good, then I suggest we keep the old formulation for SX (it's a pita to get it to work efficiently with the smaller `w3` array), and just merge this in

njansson (approved):

timfelle (approved):

njansson:
@gloopydoop should we merge then?

PR #1464, opened 2024-09-02
Add linting action and create post merge job for develop

timfelle:
This pull request adds a linting action to execute the Flint linter on the specified directory. It also creates a post merge job for the develop branch that runs compilation checks and creates badges based on the results. Additionally, the badges for the linting score and compiler support are used to show the status of the code.

At the moment, the post merge job on develop should trigger whenever a PR is merged into develop succesfully.
Alternatively we can let it run as part of the nightly documentation update, this could be part of a nightly release job.

### DIFF
--- a/.github/actions/flint/action.yml
+++ b/.github/actions/flint/action.yml
@@ -1,0 +1,52 @@
+name: Execution of Flint
+description: |
+  This action executes the Flint linter on the specified directory.
+
+inputs:
+  directory:
+    description: "The directory to lint."
+    required: true
+    default: "."
+
+  depth:
+    description: "The depth to lint."
+    required: false
+    default: "10"
+
+  rc-file:
+    description: "The configuration file to use."
+    required: false
+    default: "flinter_rc.yml"
+
+outputs:
+  score:
+    description: "The score of the linting."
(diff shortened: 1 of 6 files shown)


timofeymukha:
Cool, badges are nice :-)

njansson (approved):

timofeymukha:
Aren't things guaranteed to compile right after a merge, if CI greenlights it?

timfelle:
Btw, the linting job for this one cannot complete, since i created a new action. The workflow will run the action from develop, but develop does not have the action yet.

timfelle:
> Aren't things guaranteed to compile right after a merge, if CI greenlights it?

Well we recently broke NVfortran, but otherwise yes

timofeymukha:
I see. So it is more to just show the status of what we got rather than to check for unexpected failures.

timofeymukha (approved):

timfelle:
It would be nice if we could avoid the full retest though 🤔

timofeymukha:
I would say in the long run it is better to do it nightly, then it doesn't really matter if it takes a bit of time.

PR #1465, opened 2024-09-02
Move badge creation to checks

timfelle:
Move badge creation to the PR check.
This is also to test if we can commit to the branch without creating an infinite loop


timfelle:
Well Directly pushing to the PR does not work either.

timfelle:
Alright, New approach for the compiler support.
The badges are standard status badges from the workflow runs, and the workflows now execute on pushes to develop.
This mean the status will be reported for develop as soon as a PR is pulled in.
I was not able to figure out how to do the linting badge yet, so I removed it again.

timofeymukha (approved):

njansson (approved):

PR #1468, opened 2024-09-08
Add Coriolis force source term

timofeymukha:
Adds the Coriolis force on the CPU under source terms. 

I provide multiple ways of defining the strength, as described in the documentation. Two of them are to be friendly to our ABL users.


pschlatt1:
should one add the centrifugal term as well?

timofeymukha:
> should one add the centrifugal term as well?

That's a good question. I had mostly ABLs in mind here, where it is not needed. But perhaps one should add it and have the possibility to disable either the Coriolis or the centrifugal.

doc/pages/user-guide/case-file.md
```diff
@@ -262,11 +262,29 @@ The following types are currently implemented.
    - `g`: The gravity vector.
    - `beta`: The thermal expansion coefficient, defaults to the inverse of
       `ref_value`.
-3. `user_pointwise`, the values are set inside the compiled user file, using the
+3. `coriolis`, a source term introducing a Coriolis force, define as \f$ -2
```
timfelle, line 265:
```suggestion
3. `coriolis`, a source term introducing a Coriolis force, defined as \f$ -2
```

doc/pages/user-guide/case-file.md
```diff
@@ -262,11 +262,29 @@ The following types are currently implemented.
...
+   \f$ u \f$ is the velocity vector, and \f$ U_g \f$ is the geostrophic wind.
+   Several ways of setting \f$ \Omega \f$ are provided via the following
+   keywords.
+
+   - `rotation_vector`: Array with 3 values. Directly assigns \f$ \Omega \f$ to
+     the provided vector.
+   - `omega` and `phi`: Both scalars. The latitude `phi` should be provided in
+     degrees. Sets \f$ \Omega = [0, \omega \cos \phi, \omega \sin \phi ] \f$.
+     Common notation when modelling the atmosphere. This assumes that the \f$ z
+     \f$ axis is normal to the ground.
```
timfelle, line 276:
```suggestion
     \f$ axis is normal to the ground.
```

doc/pages/user-guide/simcomps.md
```diff
@@ -212,7 +212,7 @@ fields to disk via the usual common keywords.
 ### weak_grad {#simcomp_weak_grad}
 Computes the weak gradient of a field. The weak gradient is value of the
 gradeint multiplied by the local value of the mass matrix. This is how a
```
timfelle, line 214:
```suggestion
gradient multiplied by the local value of the mass matrix. This is how a
```

src/source_terms/bcknd/cpu/coriolis_source_term_cpu.f90
```diff
@@ -0,0 +1,79 @@
...
+    u => fields%get_by_index(1)
+    v => fields%get_by_index(2)
+    w => fields%get_by_index(3)
+
+    do concurrent (i = 1:n)
+       ui = u%x(i,1,1,1) - u_geo(1)
+       vi = v%x(i,1,1,1) - u_geo(2)
+       wi = w%x(i,1,1,1) - u_geo(3)
+
+       u%x(i,1,1,1) = u%x(i,1,1,1) + 2.0_rp * (vi * omega(3) - wi * omega(2))
```
adperezm, line 72: In the docs it says -2 omega cross u.

In mechanics it also appears with the negative sign when the term is on the right hand side. Is there a reason here it is positive? Sorry if this is a very stupid question.

src/source_terms/coriolis_source_term.f90
```diff
@@ -0,0 +1,174 @@
...
+
+    call this%free()
+    call this%init_base(fields, coef, start_time, end_time)
+
+    if (fields%size() .ne. 3) then
+       call neko_error("Number of fields for the Coriolis force must be 3.")
+    end if
+
+    this%omega = omega
+    this%u_geo = u_geo
```
timfelle, line 148:
```suggestion
    this%u_geo = u_geo
```

pschlatt1:
> > should one add the centrifugal term as well?
> 
> That's a good question. I had mostly ABLs in mind here, where it is not needed. But perhaps one should add it and have the possibility to disable either the Coriolis or the centrifugal.

I guess it could be helpful. Having the centrifugal force may affect the pressure convergence, and of course give a different pressure field, which may be relevant at times.

pschlatt1:
Perhaps a more fundamental question: Does one really want to have these different forcings in the code? That for instance relates to the Brinkman forcing and the Coriolis force (but pehhaps also Boussinesq etc.). I would in fact recommend to not have it, and rather have examples for cases that may implement such forces. The main reason is to keep the main code as simple and clean as possible. Counterexample is for instance Simson where everything was implemented in the main code - simply because there was no other way. This makes the code very large, and potentially half of the features do not work any longer anyway.

src/source_terms/bcknd/cpu/coriolis_source_term_cpu.f90
timofeymukha: You are correct. I just accidentally implemented u x Omega instead, and compensated by changing sign. I should rewrite it, it will cause confusion.

adperezm: Haha I see, I had "verified" the cross product to see if that is what you had done, but I realize now that I did that wrong :).

timofeymukha:
> Perhaps a more fundamental question: Does one really want to have these different forcings in the code? That for instance relates to the Brinkman forcing and the Coriolis force (but pehhaps also Boussinesq etc.). I would in fact recommend to not have it, and rather have examples for cases that may implement such forces. The main reason is to keep the main code as simple and clean as possible. Counterexample is for instance Simson where everything was implemented in the main code - simply because there was no other way. This makes the code very large, and potentially half of the features do not work any longer anyway.

It can be discussed, of course. My vision was always that the asymptote is that most users don't ever have to use the user file. Some stuff will also be very difficult to write examples for because of the GPU. This one is an example, you can't just use device_math here, so one would need to write the kernels.

adperezm:
I think having these terms is good. 

A user can always of course do the same by putting in a custom forcing term with the same effect if they want.

I was skeptical, for example, when the buoyancy term was added. However after using it for a little bit I realized it was very nice to have :)

If you are concerned about modularity and lightweight, maybe in the future one could have a lightweight neko that only compiles the core functionality and not the simcomps or something like that? :)

adperezm (approved):

pschlatt1:
I would be skeptical having all these different "case-specific" forcings and potentially other features to be in the main version. Having examples (or perhaps a yet to be defined extensions/plugins) would in my opinion be more appropriate. The main reason is to make the code more streamlined, but most importantly to make sure that whatever is in the code actually work. Essentially, it means that you need to make sure that all new featuers need to work with all old features, or there needs to be some logic making sure that combinations that don't work (or are not tested) cannot be chosen. Or even worse, there may be situations where you have features that are simply not compatible (cannot think of an example just now).

Again, my point here is motivated by Simson, where people implemented various things, which then simply were not compatible (e.g. there is wall-roughness implemented, very similar to the Brinkman forcing, but it did not work e.g. for scalars, not for the perturbation formulation etc.).

timofeymukha:
> I would be skeptical having all these different "case-specific" forcings and potentially other features to be in the main version. Having examples (or perhaps a yet to be defined extensions/plugins) would in my opinion be more appropriate. The main reason is to make the code more streamlined, but most importantly to make sure that whatever is in the code actually work. Essentially, it means that you need to make sure that all new featuers need to work with all old features, or there needs to be some logic making sure that combinations that don't work (or are not tested) cannot be chosen. Or even worse, there may be situations where you have features that are simply not compatible (cannot think of an example just now).
> 
> Again, my point here is motivated by Simson, where people implemented various things, which then simply were not compatible (e.g. there is wall-roughness implemented, very similar to the Brinkman forcing, but it did not work e.g. for scalars, not for the perturbation formulation etc.).

I think being restrictive to what is in main is generally a good idea, but for that approach to work we need to make sure that we can easily load-in 3rd party code into neko. For all types (not only simcomps), and also for GPUs (not just device_math). This is not easy to do, in fact, I am not sure if it is actually possible. I have given a lot of thought on how to make factories externally extendible and haven't found a reliable way :(. For example, implementing a forcing would not be possible now, except for the user file. But I agree that we should be careful about compatibility!

timofeymukha:
I think it is somewhat a question of scope. You can use the user file for simple things on the CPU and trivial things on the GPU via device_math. But having anything substantial there is very cumbersome. I guess this is the reason why KTH_FRAMEWORK appeared for Nek5000, right? Nobody wants the tripping line forcing dumped into their user file, etc. Gluing together code from multiple user files is also quite error-prone, since they may e.g. use the same variable names but for different things.
I am not sure one can expect people to be ready to jump in and write code also. There are a lot of applied researchers in CFD, which don't really know how to write code, and certainly won't attempt GPU code! I don't know if it is to dream too big to aim for attracting this cohort as Neko users, but I guess that was the ambition inside my head. And even for myself, it is very, very convenient that so much stuff in e.g. OpenFOAM can be done via the case file.

However, one must also be realistic, and maybe avoid targeting a theoretical user, which does not yet exist and perhaps never will? Almost all current users are ex Nek5000 hardcore turbulence guys, who are used to things working in certain ways. It may make more sense to pivot the code towards precisely that. It would be a buzz killer for me personally, but should not stop us from making the right decisions. And in that case we are kind of over-engineering already, so we should decide!

timofeymukha:
To take the example of OpenFOAM again. It is really nice to have all that stuff built in. And it is easy to extend the code as long as you are just deriving a new child from some abstract class and don't want to do anything too crazy. But going in really deep and modifying stuff close to basics is very difficult, because it is like fighting against "the system" :-). So there is certainly a tradeoff. And you have people who hate OpenFOAM because of the latter difficulty, but also on the other end of the spectrum, those who hate it because there is no GUI :D.

doc/pages/user-guide/case-file.md
```diff
@@ -262,11 +262,29 @@ The following types are currently implemented.
...
+   - `f`: Scalar, referred to as the Coriolis parameter, \f$ f = 2 \omega \sin
+     \phi \f$. Sets \f$ \Omega = [0, 0, 0.5f ] \f$. This further assumes that the
```
linneahuusko, line 278: I would suggest explicitly stating both assumptions under this point, since otherwise it is not completely clear what "further" refers to.

linneahuusko (approved):

timofeymukha:
merge?

MartinKarp:
I think so

njansson (approved):

PR #1481, opened 2024-09-11
Introduction of source term handler

timfelle:
This PR introduces a common ancestor for the fluid and scalar source term handling.

This reduces code duplication without modifying the interface to either. While opening the opportunity to create source term handlers for other objects. An adjoint source term handler comes to mind.

Depends on: #1480


timfelle:
@njansson, @timofeymukha and @MartinKarp I think this should be done as is. I guess you three would have a direct interest in this source term handler as well.

MartinKarp (approved):
I think this looks good, what comes to mind is kind of how deeply we have integrated json, I guess related a bit to the discussion from Tuesday about using Neko as a library.

timfelle:
> I think this looks good, what comes to mind is kind of how deeply we have integrated json, I guess related a bit to the discussion from Tuesday about using Neko as a library.

Yes, that is exactly why the `add_source_term` subroutine exists. This allow the user to append programmatically constructed source terms to the handler. 
One improvement could be to have the json path be an optional input, making it clearer that the handler can be used without the json?

timofeymukha:
The deep JSON integration I think should ideally be balanced by having a constructor that does not use it, but directly inits from all the stuff that one would ordinarily read from the JSON. I tried to implement this idea with some of the `init_from_components` constructors, but I have to admit that sometimes I failed to keep this consistent across all the code I pushed. I think with the "use as library" case becoming potentially more prominent, we might want to try harder on this front.

MartinKarp:
> > I think this looks good, what comes to mind is kind of how deeply we have integrated json, I guess related a bit to the discussion from Tuesday about using Neko as a library.
> 
> Yes, that is exactly why the `add_source_term` subroutine exists. This allow the user to append programmatically constructed source terms to the handler. One improvement could be to have the json path be an optional input, making it clearer that the handler can be used without the json?

But json is still required for the init or? I guess if it is still required there it doesn't matter too much if it's in the `add` subroutine as well.

timfelle:
> > > I think this looks good, what comes to mind is kind of how deeply we have integrated json, I guess related a bit to the discussion from Tuesday about using Neko as a library.
> > 
> > 
> > Yes, that is exactly why the `add_source_term` subroutine exists. This allow the user to append programmatically constructed source terms to the handler. One improvement could be to have the json path be an optional input, making it clearer that the handler can be used without the json?
> 
> But json is still required for the init or? I guess if it is still required there it doesn't matter too much if it's in the `add` subroutine as well.

At the moment it is, I suggest making it optional.
The `add` subroutine takes a fully initialized `source_term_t` which can be done however you want, and inserts it into the handlers list of source_terms. So no json is needed to add a source term, assuming the source term itself can be constructed without.

timfelle:
> The deep JSON integration I think should ideally be balanced by having a constructor that does not use it, but directly inits from all the stuff that one would ordinarily read from the JSON. I tried to implement this idea with some of the `init_from_components` constructors, but I have to admit that sometimes I failed to keep this consistent across all the code I pushed. I think with the "use as library" case becoming potentially more prominent, we might want to try harder on this front.

I think making the json stuff in the handler optional would be sufficient here. But yes, individual source terms should be possible to initialize with no json.

timfelle:
I have updated the handler:

- Json is now not parsed in the initializer of the handler base class. 
- Source terms can now be read from a json_file and name of the key to look at.

Additionally, the system now also allow user defined source terms to adhere to a start and end time.

timofeymukha (approved):

timfelle:
Fluid and scalar source term, no longer knows JSON. It is parsed directly through `source_term%add` in the schemes after initialization.

njansson (approved):

PR #1484, opened 2024-09-12
Add overview of the most important types to the docs

timofeymukha:
If you think I missed an important one, please advise. Also if you think the desciption is bad.

### DIFF
--- a/doc/pages/developer-guide/important_types.md
+++ b/doc/pages/developer-guide/important_types.md
@@ -36,7 +36,7 @@ as subroutines.
 - [case_t](#case::case_t): Holds some data shared by both the fluid and the
   scalar solver, which it also stores as components.
 - [fluid_scheme_t](#fluid_scheme::fluid_scheme_t): Navier-Stokes solvers.
-- [sclar_scheme_t](#scalar_scheme::scalar_scheme_t): Scalar
+- [scalar_scheme_t](#scalar_scheme::scalar_scheme_t): Scalar
   advection-diffusion-reaction solvers.
 - [bc_t](#bc::bc_t): Boundary conditions.
 - [time_scheme_t](#time_scheme::time_scheme_t): Time integration schemes.
--- a/doc/pages/developer-guide/important_types.md
+++ b/doc/pages/developer-guide/important_types.md
@@ -24,17 +24,20 @@ explore the descendants.
   essentially everything else that lives on the mesh. 
 
 ## Basic math routines
-Here, we list file names rather than types, since the basic math is implemented
+Here, we also list file names rather than types, since the basic math is implemented
 as subroutines.
 
 - `math.f90`: Basic math operations on raw arrays.
 - `field_math.f90`: Basic math operations on [field_t](#field::field_t).
 - `operators.f90`: Various explicit operators, including derivatives, etc.
+- [vector_t](#vector::vector_t) and [matrix_t](#matrix::matrix_t): 1D and 2D arrays
+  with support for computation on [accelerators](#accelerators).
 
 ## Governing equation solvers and related types
 
-- [case_t](#case::case_t): Holds some data shared by both the fluid and the
-  scalar solver, which it also stores as components.
+- [case_t](#case::case_t): An abstraction for the simulation case. Stores the
+  fluid and scalar solver as components, handles IO via
+  [sampler_t](#sampler::sampler_t).
 - [fluid_scheme_t](#fluid_scheme::fluid_scheme_t): Navier-Stokes solvers.
 - [scalar_scheme_t](#scalar_scheme::scalar_scheme_t): Scalar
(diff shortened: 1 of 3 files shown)


doc/pages/developer-guide/important_types.md
```diff
@@ -0,0 +1,66 @@
...
+- [gs_t](#gather_scatter::gs_t): Gather-scatter kernels used to make the
+  solution continuous, i.e. perform direct stiffness summation. 
+- [field_t](#field::field_t): The main type for storing the unknowns, and
+  essentially everything else that lives on the mesh. 
+
+## Basic math routines
+Here, we list file names rather than types, since the basic math is implemented
+as subroutines.
+
+- `math.f90`: Basic math operations on raw arrays.
```
timfelle, line 30:
```suggestion
- `math.f90`: Basic math operations on raw arrays.
- `device_math.f90`: Basic math operations on device arrays.
```

doc/pages/developer-guide/important_types.md
```diff
@@ -0,0 +1,66 @@
...
+- `field_math.f90`: Basic math operations on [field_t](#field::field_t).
+- `operators.f90`: Various explicit operators, including derivatives, etc.
+
+## Governing equation solvers and related types
+
+- [case_t](#case::case_t): Holds some data shared by both the fluid and the
+  scalar solver, which it also stores as components.
```
timfelle, line 37: I guess it actually holds, well everything.

doc/pages/developer-guide/important_types.md
njansson: Should we also mention `matrix_t` and `vector_t` here?

doc/pages/developer-guide/important_types.md
njansson: I would avoid `data` here, yes it holds the solvers etc, but not really the actual data. How about saying that it defines a given case to be solved?

doc/pages/developer-guide/important_types.md
timfelle: Sure, maybe in a new list with some text between `Useful types for mathematical operations:` or something like that? Still under the `Basic Math` header ofcourse.

doc/pages/developer-guide/important_types.md
timofeymukha: Yeah, fair enough, I was mainly referring to objects like the sampler etc, data is a bad word. But it doesn't hold the simcomps, so I would not say everything :-).

timfelle: Well, should it not contain the simcomp?

njansson: hear hear!

timfelle: An example here is when we are working with the adjoint solver, we want to support simulation components for the adjoint, which are different than the forward simulation. Atleast i think that might be useful 😅

timfelle: I would even say the field registry should be a member of the case, so one could have multiple sets of simulations, without having to construct a unique naming across all simulations, but just within a given simulation.

timofeymukha: I updated the text and added the vector and matrix types.

Regarding the case_t, things are not easy because of the prohibition on 2-way deps in Fortran. If `case_t` holds important shared stuff, like the mesh, etc. then it needs to be easily accessed. This automatically means that it cannot hold too much stuff inside. For example, if we put the registry there, then every type that uses the registry would need to `use case`, and none of those types can be components of the case.

I think using singletons has made a lot of things very easy, but I agree that they currently don't work with having more than one case. A solution could be to add another dimension to them, so that you have an array of registires, etc.

timfelle: In any case, this is a discussion, not a comment for the PR itself 😅

doc/pages/developer-guide/important_types.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Basic math routines
+Here, we also list file names rather than types, since the basic math is implemented
+as subroutines.
+
+- `math.f90`: Basic math operations on raw arrays.
+- `device_math.f90`: Basic math operations on device arrays.
+- `field_math.f90`: Basic math operations on [field_t](#field::field_t).
+- `operators.f90`: Various explicit operators, including derivatives, etc.
+- [vector_t](#vector::vector_t) and [matrix_t](#matrix::matrix_t): 1D and 2D arrays
```
MartinKarp, line 34: I think we should really encourage the use of vector, but maybe one does not need to do it explicitly.

MartinKarp (approved):
Good start I think, one can add more as we go along I guess.

njansson (approved):

njansson (approved):

PR #1489, opened 2024-09-18
Add logging levels and improve error handling

timfelle:
This pull request includes several commits that add logging levels and improve error handling in the codebase. 

The commits introduce new function for logging, `log_header`, and update `neko_error` to output in stderr. Additionally, a new environment variable `NEKO_LOG_FILE` is introduced to specify the log file name, with `stdout` used as the default if not set. The pull request also includes documentation updates for the new variable.

Additionally the json utilities have been extended to handle extraction of entire sub structures.

### DIFF
--- a/src/common/log.f90
+++ b/src/common/log.f90
@@ -34,7 +34,8 @@
 module logger
   use comm, only : pe_rank
   use num_types, only : rp
-  use, intrinsic :: iso_fortran_env, only: stdout => output_unit
+  use, intrinsic :: iso_fortran_env, only: stdout => output_unit, &
+       stderr => error_unit
   implicit none
   private
 
@@ -208,7 +209,7 @@ contains
 
     if (pe_rank .eq. 0) then
        call this%indent()
-       write(this%unit_, '(A,A,A)') '*** ERROR: ', trim(msg),'  ***'
+       write(stderr, '(A,A,A)') '*** ERROR: ', trim(msg),'  ***'
     end if
 
   end subroutine log_error
@@ -336,7 +337,7 @@ contains
        end do
 
        call neko_log%indent()
-       write(*, '(A)') trim(msg(1:len))
+       write(neko_log%unit_, '(A)') trim(msg(1:len))
     end if
 
   end subroutine log_message_c
@@ -358,7 +359,7 @@ contains
        end do
 
        call neko_log%indent()
-       write(*, '(A,A,A)') '*** ERROR: ',trim(msg(1:len)),'  ***'
+       write(stderr, '(A,A,A)') '*** ERROR: ',trim(msg(1:len)),'  ***'
     end if
 
   end subroutine log_error_c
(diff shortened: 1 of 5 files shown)


src/common/log.f90
```diff
@@ -149,19 +176,39 @@ subroutine log_message(this, msg, lvl)
...
+
   !> Write an error message to a log
   subroutine log_error(this, msg)
     class(log_t), intent(in) :: this
     character(len=*), intent(in) :: msg
 
     if (pe_rank .eq. 0) then
        call this%indent()
-       write(*, '(A,A,A)') '*** ERROR: ', trim(msg),'  ***'
+       write(this%unit_, '(A,A,A)') '*** ERROR: ', trim(msg),'  ***'
```
MartinKarp, line 211: This is an error but does not kill the simulation then?

MartinKarp (commented):
Looks good I think. If I understand correctly everything is the same, only we have expanded the functionality, but the defaults are as before?

src/common/log.f90
```diff
@@ -34,6 +34,7 @@
 module logger
   use comm, only : pe_rank
   use num_types, only : rp
+  use, intrinsic :: iso_fortran_env, only: stdout => output_unit
```
njansson, line 37: Should we also use `error_unit` to be consistent

timofeymukha:
Is there any difference between using an env variable and just redirecting to a file?

timfelle:
> Is there any difference between using an env variable and just redirecting to a file?

Not all print statements are done through the logger. This way you can pipe official logs to a file and add your own prints to the terminal during debugging for example.

src/common/log.f90
timfelle: Sure.

src/common/log.f90
timfelle: Nope, this just prints to the log. I don't know that anyone is using it though.

njansson: Not used, it has been like this for a while. But let's keep it if we ever unify stuff

src/common/log.f90
```diff
@@ -232,15 +305,12 @@ subroutine log_status(this, t, T_end)
...
-    call this%message('----------------------------------------------------------------', &
-                      NEKO_LOG_QUIET)
-    write(log_buf, '(A,E15.7,A,F6.2,A)') &
-    't = ', t, '                                  [ ',t_prog,'% ]'
-
+    call this%message(repeat('-', 64), NEKO_LOG_QUIET)
     call this%message(log_buf, NEKO_LOG_QUIET)
-    call this%message('----------------------------------------------------------------', &
-                      NEKO_LOG_QUIET)
+    call this%message(repeat('-', 64), NEKO_LOG_QUIET)
```
njansson, line 312: This changes everything. `repeat` quality of life++ 🍻

njansson (approved):

src/common/log.f90
timfelle: Oh yes, it is a lot better being able to use repeat. I guess we could avoid the loops in the indentation as well then 🤔

MartinKarp (approved):

Issue #1487, opened 2024-09-16
CTR porting adventure


MartinKarp:
Porting of features implemented for the CTR summer program.

- [x] Lift/drag/torque calculation
- [x] map_2d_t and map_1d_t updates
- [x] 1D/2D output of statistics
- [x] mean_field simcomp "user_stats"
- [x] fluid_stats in simcomp
- [x] fld_file_datta updates
- [x] fld_file fixes
- [x] simcomp restart

MartinKarp:
#1486 fixes first point.

MartinKarp:
#1493 second point.

MartinKarp:
#1494 covers points 3 and 5.

MartinKarp:
#1519 covers the final point, point 2

PR #1493, opened 2024-09-19
Add map_2d

MartinKarp:
Adds the map_2d_t type and enables averaging fields into 2D as a post-processing step. Continues #1487 

Cleans up averaging in space contrib script a bit. Functionality is primarily tailored for the CPU, so don't expect any major performance from the averaging routines.

Next steps will be dependent on this functionality.


njansson (approved):
Should we have a (C) header in the files?

contrib/calc_lift_from_field/calc_lift_from_field.f90
```diff
@@ -1,4 +1,4 @@
-!> Program to calculate the force and acting on a single boundary zone as well as
+	!> Program to calculate the force and acting on a single boundary zone as well as
```
timfelle, line 1: Accidental tab it seems
```suggestion
!> Program to calculate the force and acting on a single boundary zone as well as
```

src/io/fld_file_data.f90
```diff
@@ -70,6 +72,71 @@ function fld_file_data_size(this) result(i)
...
+
+  end subroutine fld_file_init_same
+
+  !> Genereate same fields as in another fld_file
+  subroutine fld_file_init_n_fields(this, n_fields, n)
+    class(fld_file_data_t), target, intent(inout) :: this
+    integer, intent(in) :: n, n_fields
+    integer :: i, j
+    
+
```
timfelle, line 111: Whitespace
```suggestion


```

src/io/fld_file_data.f90
```diff
@@ -157,7 +224,9 @@ subroutine fld_file_data_free(this)
        do i = 1, this%n_scalars
           call this%s(i)%free()
        end do
+       deallocate(this%s)
     end if
+
```
timfelle, line 229: Whitespace
```suggestion

```

src/sem/map_1d.f90
timfelle, line 1: There are quite a few lines that are rather long here.

src/sem/map_2d.f90
timfelle, line 1: Also quite a lot of long lines here

timfelle (approved):
Generally seem pretty good. There are a few formatting things to consider as well as old print statements you probably didn't mean to include

MartinKarp:
Thanks I will fix this!

PR #1494, opened 2024-09-19
Fluid statistics as simcomp and 1D/2D stats

MartinKarp:
Makes fluid_stats a simcomp and makes it possible to average in 2 or 1 homogeneous directions and significantly compress the output stats files.

The syntax for the statistics simcomp is the following:
```json
{ 
       "type": "fluid_stats",
       "compute_control": "tsteps",
       "compute_value": 4,
       "output_value": 20,
       "start_time":50.0,
       "avg_direction":"xz",
       "set_of_stats":"basic"
}
```

Builds on #1493 and replaces #1334.

Slightly different functionality than previously as one now outputs a "empty" stats file as soon as sampling begins. Makes it easier to keep track of when one starts to sample the statistics as one now has a file output at that time. Adds the functionality to instead sample statistics based on simulation time.

Also updated the documentation a bit, especially the information on the order of the statistics.


MartinKarp:
I might have pulled in force_torque here as well, but I guess that will be merged before so I hope that's ok.

ronithstanly (commented):
Looks good. Just found a couple of typos in the comments.

doc/pages/user-guide/statistics-guide.md
```diff
@@ -29,84 +29,88 @@ where \f$ u_0 \f$ is the fields value at \f$ T_0 \f$ and \f$ N \f$ is the number
...
+      "set_of_stats":"basic"
+    }1
+  ]
 ~~~~~~~~~~~~~~~
-When the output is written one obtains two .fld files called mean_field and stats. 
+The argument "avg\_direction" is optional and if ignored we output 3d fields. The statistics are saved in a fld file according to the following in 2D and 3D. Observe that in 2D the mean velocity in the homogenous direction is not stored in Z-velocity, but is stored in a last scalar field. The correct velocity components are stored in X-velocity and Y-velocity respectively. All other fields are kept the same. This is due to a limitation of the fld file format.
 
-## List of fields in output files 
-In `mean_field` the following averages are stored. The stored in variable column is which field one finds the computed statistic if one opens the file in paraview or visit.
+For 1D statistics a CSV file is outputted. The first column is the time at which the statistics are collected, the second column the spatial coordinate, and the rest of the data is stored in the order below. In this case all statistics are kept in the same order as in 3D.
```
ronithstanly, line 62: It is good to add somewhere in the documentation what the first entry in the csv (in case of 2 homogenous directions for averaging) and the first fld file (in case of only 1 homogenous direction for averaging) corresponds to and that it differs from the real statistics which will be included in the later entries

MartinKarp: But does it differ? It does output the average from the specified start_time to the current time. Current time and output time is the same the weight will simply be 0.

ronithstanly: Why I mentioned this is because when I loaded the first fld stat file that came out of averaging, I forgot that to see the real average fields I should start looking from the second file instead of looking at the very first file that is outputted.

MartinKarp: Right, let's add something

doc/pages/user-guide/statistics-guide.md
```diff
@@ -29,84 +27,91 @@ where \f$ u_0 \f$ is the fields value at \f$ T_0 \f$ and \f$ N \f$ is the number
 In the statistics in Neko, various averages of the the different velocity components, derivatives and pressure are computed. In total, 44 "raw statistics" are computed that are required to compute the Reynolds stress budgets, mean fields, and the different terms in the turbulent kinetic energy equation.
 
 ## Using statistics
-Statistics are enable in the the case file as the following:
+Statistics are enable in the the case file as a simcomp with the added argument avg_direction, set_of_stats, and start_time:
```
timfelle, line 30:
```suggestion
Statistics are enable in the the case file as a simcomp with the added argument `avg_direction`, `set_of_stats`, and `start_time`:
```

Shiyu-Sandy-Du (changes requested):

src/fluid/fluid_stats.f90
```diff
@@ -48,8 +48,8 @@ module fluid_stats
   use neko_config, only : NEKO_BCKND_DEVICE
   use utils, only : neko_warning
   implicit none
-  private
-
+  private  
+  
   type, public, extends(stats_quant_t) :: fluid_stats_t
      type(field_t) :: stats_u !< Not reasonable to allocate 20 something fields
```
timofeymukha, line 54: These work arrays can probably be taken from the registry.

src/fluid/fluid_stats.f90
```diff
@@ -645,9 +693,9 @@ subroutine fluid_stats_make_strong_grad(this)
 
   end subroutine fluid_stats_make_strong_grad
 
-  subroutine fluid_stats_post_process(this, mean, reynolds, pressure_flatness, &
-       pressure_skewness, skewness_tensor, &
-       mean_vel_grad, dissipation_tensor)
+  subroutine fluid_stats_post_process(this, mean, reynolds, pressure_flatness,&
+                                      pressure_skewness, skewness_tensor, &
+                                      mean_vel_grad, dissipation_tensor)
     class(fluid_stats_t) :: this
```
timofeymukha, line 699: Is this routine going to be deprecated to Python?

timofeymukha (commented):
I've started going through the files. Very happy to see some base types getting slimmer. I issues a doc and style pr, mostly for fluid_stats, which is what I went through so far.

Question, can we directly use grad instead of opgrad and kill the make_strong_grad routine?

src/fluid/fluid_stats.f90
timofeymukha: This is regarding the stats_ fields. I realize you don't touch this in the PR though. So can be skipped.

src/simulation_components/fluid_stats_simcomp.f90
```diff
@@ -0,0 +1,205 @@
...
+       call MPI_Barrier(NEKO_COMM, ierr)
+
+       sample_start_time = MPI_WTIME()
+
+       call this%stats%update(deltaT)
+       call MPI_Barrier(NEKO_COMM, ierr)
+       this%time = t
+
+       sample_time = MPI_WTIME() - sample_start_time
+
```
timofeymukha, line 192: How does this timing fit with the profiling added by @njansson ? Can that be used instead? I feel like outputting this at every timestep is perhaps superflous.

src/simulation_components/fluid_stats_simcomp.f90
```diff
@@ -0,0 +1,205 @@
...
+  !! @param hom_dir directions to average in
+  !! @param stat_set Set of statistics to compute (basic/full)
+  subroutine fluid_stats_simcomp_init_from_attributes(this, u, v, w, p, coef, &
+                                                      start_time, hom_dir, &
+                                                      stat_set)
+    class(fluid_stats_simcomp_t), intent(inout) :: this
+    character(len=*), intent(in) :: hom_dir
+    character(len=*), intent(in) :: stat_set
+    real(kind=rp), intent(in) :: start_time
+    type(field_t), intent(inout) :: u, v, w, p !>Should really be intent in I think
```
timofeymukha, line 127: But it cannot be?

examples/turb_channel/turb_channel.case
```diff
@@ -43,26 +43,29 @@
...
+      "center": [0,0,0],
+      "scale": 0.01899
+    },    
+    { 
+      "type": "fluid_stats",
+      "output_control": "simulationtime",
+      "output_value": 20,
+      "start_time": 60.0,
+      "compute_control": "tsteps",
+      "compute_value": 50,
```
timofeymukha, line 65: So we can sample based on simulation time as well, like @Mr-2c wanted?

src/simulation_components/fluid_stats_simcomp.f90
njansson: I think it's good to have both, the one can always extract data from the log if something extra is needed.

So here I would suggest to add an additional profiler region such that it can also end up in both NVTX traces as well as the runtime stats

src/simulation_components/fluid_stats_simcomp.f90
```diff
@@ -0,0 +1,205 @@
...
+ 
+    call neko_log%end_section()
+  
+  end subroutine fluid_stats_simcomp_init_from_attributes
+
+  !> Destructor.
+  subroutine fluid_stats_simcomp_free(this)
+    class(fluid_stats_simcomp_t), intent(inout) :: this
+    call this%free_base()
+  end subroutine fluid_stats_simcomp_free
```
timofeymukha, line 168: Should we call free on the stats object?

src/io/fluid_stats_output.f90
```diff
@@ -35,63 +35,128 @@ module fluid_stats_output
...
+                  output_2d%w%x(i) = u
+               else if (this%map_2d%dir .eq. 2) then
+                  output_2d%u%x(i) = u
+                  output_2d%v%x(i) = w
+                  output_2d%w%x(i) = v
+               else if (this%map_2d%dir .eq. 3) then
+                  output_2d%u%x(i) = u
+                  output_2d%v%x(i) = v
+                  output_2d%w%x(i) = w
+               end if
```
timofeymukha, line 153: Are we sure this is a good idea? I think it will be much less error prone to just always store the 3 velocity components first and let the user figure out which ones are relevant later in Python.

timofeymukha:
OK, I am done reviewing now :-). It seems Niclas merged my style PR, there is a new small one opened for another file. I did not expect anyone to be active now, but ofc there is the time difference. So, it seems @MartinKarp will wake up to a slightly updated branch. Hopefully I did not mess anything up.

src/fluid/fluid_stats.f90
MartinKarp: Yes, I was thinking one can add warnings when pynektools becomes public.

src/simulation_components/fluid_stats_simcomp.f90
MartinKarp: Because we never modify these fields, but only use them, but I guess there are some routines that I use that have inout.

examples/turb_channel/turb_channel.case
MartinKarp: Yup

src/io/fluid_stats_output.f90
MartinKarp: Otherwise visualizing the field becomes extremely confusing.

src/simulation_components/fluid_stats_simcomp.f90
MartinKarp: Good point

src/simulation_components/fluid_stats_simcomp.f90
timofeymukha: Yeah, we have in general been a bit too sloppy with this in the code base

examples/turb_channel/turb_channel.case
timofeymukha: Great!

src/io/fluid_stats_output.f90
timofeymukha: How so? Perhaps I am missing something.

MartinKarp: It depends what people expect I guess. What is nice with this is that when one plots "x+velocity" it will be the x velocity in the new coordinate axes. But for post processing perhaps it would be best to just keep the original axis as far as possible. For averaging in z (which I guess it the most common), it will not make any difference.

timofeymukha: In the end, this is just about the index in the output, right? If I open the .fld, under case

                  output_2d%u%x(i) = u
                  output_2d%v%x(i) = w
                  output_2d%w%x(i) = v

Will the array called v in the field list in Paraview actually hold w? That, I think, we should really avoid. Or are they all just saved as scalars, like it was with kth toolbox?

I think having to figure out this logic based on the plane the data is in most situations going to be an extra hoop to jump. I mean, surely, one knows what component one wants to plot? You will still have to do it for all the other statistics anyway, e.g. you are not swapping vv for ww.

MartinKarp: Pushed a change.

timofeymukha (approved):

examples/turb_channel/turb_channel.case
```diff
@@ -43,26 +43,29 @@
...
+      "avg_direction": "xz",
```
vbaconnet, line 66: Getting a segfault when changing the `avg_direction: "z"` and `start_time: 0.0`, could someone else test it to see if it's just me? :)

```
Backtrace for this error:

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x7f66cd4218c2 in ???
#1  0x7f66cd420a55 in ???
#2  0x7f66cd25b04f in ???
	at ./signal/../sysdeps/unix/sysv/linux/x86_64/libc_sigaction.c:0
#3  0x563e5d2d2600 in __fld_file_MOD_fld_file_write
	at io/fld_file.f90:396
#4  0x563e5d1b3c90 in __file_MOD_file_write
	at io/file.f90:150
#5  0x563e5d2e4acf in __fluid_stats_output_MOD_fluid_stats_output_sample
	at io/fluid_stats_output.f90:158
#6  0x563e5d1b52b5 in __sampler_MOD_sampler_sample
	at common/sampler.f90:259
#7  0x563e5d149223 in __simulation_MOD_neko_solve
	at /scratch/baconnet/software/neko/src/simulation.f90:93
#8  0x563e5d11f78c in usrneko
	at /scratch/baconnet/software/neko/examples/turb_channel/usr_driver.f90:8
#9  0x563e5d11eaae in main
	at /scratch/baconnet/software/neko/examples/turb_channel/usr_driver.f90:2
```

vbaconnet: Fixed with commit [81a5d78](https://github.com/ExtremeFLOW/neko/pull/1494/commits/81a5d7897cbe1f14770dec42d1f0ff8fdf5b8545)

Shiyu-Sandy-Du (approved):

src/field/mean_field.f90
```diff
@@ -96,10 +107,11 @@ subroutine mean_field_reset(this)
   end subroutine mean_field_reset
 
 
-  !> Update a mean field
+  !> Update the mean field with a new sample.
+  !! @param k Time since last sample.
   subroutine mean_field_update(this, k)
     class(mean_field_t), intent(inout) :: this
-    real(kind=rp), intent(in) :: k !< Time since last sample
+    real(kind=rp), intent(in) :: k
```
timfelle, line 114:
```suggestion
    real(kind=rp), intent(in) :: k
```

src/fluid/fluid_stats.f90
```diff
@@ -48,23 +48,26 @@ module fluid_stats
   use neko_config, only : NEKO_BCKND_DEVICE
   use utils, only : neko_warning
   implicit none
-  private
-
+  private  
+
```
timfelle, line 52:
```suggestion
  private

```

timfelle (approved):

vbaconnet (approved):
Tested it on a few cases and it is super nice!

PR #1496, opened 2024-09-20
Add SEM types to case

timofeymukha:
This is a bit of a demonstrator, somewhat in line with the discussion of the sem_t type. 

What I do here is add all the basic SEM types to the case_t, kind of solidifying the case as a collection of simcomps on a given sem_t (to be introduced). 

For now I still continue using coef_t as a proxy for sem_t, so the interfaces to fluid and scalar become simpler, as I remove everything that can be reached through the coef_t. Also, I demonstrate how the dependency on fluid_scheme drops out from the simcomps. Although instead they now depend on case, it seems we can inverse the relationship if we pass coef_t (sem_t) to the constructor of the simcomps instead of the case, and the case can own the simcomps, like @timfelle and @njansson wanted.


timfelle (commented):
Would it not make sense that a `sem_t` contain a mesh and all spectral element information. The case can own one and then a simcomp and a fluid scheme could be initialized based on that `sem_t` instead of case itself.

timofeymukha:
> Would it not make sense that a `sem_t` contain a mesh and all spectral element information. The case can own one and then a simcomp and a fluid scheme could be initialized based on that `sem_t` instead of case itself.

Yes, that is exactly the intention!

timofeymukha:
Now added sem_t, sits as a component in case_t. Also pass it to the scalar and the simcomps as a demo. 

At this point we should decide definitively if we go forward. If we do, there is a lot of interfaces that can be changed to just taking sem_t instead of some combo of its components. The combos are often not very logical, as I mentioned before. And more recently, we just tend to pass coef for everything.

MartinKarp:
How about if we have components with say different polynomial orders or similar?

timofeymukha:
> How about if we have components with say different polynomial orders or similar?

Yeah, the assumption here is that you always have some "main" space, and others would be derivatives of them, like say in pn-pn2.

So if one really wants to solve on a space entirely independent of the one in the case_t then it should be in a different case_t, i.e. one should have two cases. It kind of reflects how we set things up now, where the lx is essentially set for the entire case in the json.

timofeymukha:
@njansson Thoughts?

njansson:
> @njansson Thoughts?

I'll give it a look next week

MartinKarp:
Thinking about it, I have been stumbling across more and more things nowadays which depend on only the geometry and not the connectivity. As such I think perhaps assembling everything into a SEM_t requires us to think about that a bit. 

Currently we have:

- mesh_t contains information about element connectivity, global numbering of vertices and essentially represents the information in a nmsh file
- dofmap_t contains the coordinates of the gll points and their global numbering
- coef_t contains the information regarding the jacobian of the spectral elements as well as the multiplicity of the points
- gs_t, uses the global numbering in the dofmap to do the direct stiffness summation and such.

As I see it the information required would be more similar to something like:

- mesh_t similar to now, but without generating the connectivity
- connectivity_t connectivity data of the vertices of a number of hexes (can be initialized from the mesh, but also a series of elements with numbered vertices)
- geometry_t (this could perhaps be the sem_t?) the actual geometry + whatever function space one uses. Computes the jacobians, everything needed to compute gradients, normals, and whatnot. Two initialize functions, one based on mesh_t and one on xyz coordinates.
- point_numbering_t, computes the global numbering of all the gll points, does not depend on the geometry_t.
- gs_t, uses the point_numbering_t for the direct stiffness summation, also stores the the multiplicity of the collocation points and such.

timofeymukha:
Interesting suggestion! I have a question, how can the point_numbering_t not depend on the geometry_t (which holds the dofs, I guess). The underlying space should coincide, right?

MartinKarp:
Given the number of points in a hex it does not really needthe geometry, only the global numbering of the vertices of the hex.

timofeymukha:
Hopelessly out of date, so I will close. And I guess we decided for now to allow for different spaces in different components / solvers, although for now we are not ready for that :-).

PR #1497, opened 2024-09-20
Purge material_properties_t

timofeymukha:
The `material_properties_t` type served us well for parsing dimensional and non-dimensional input for material properties of the scalar and fluid.

Unfortunately, it relies on a particular structure of the case file to find the input, and that we have exactly one fluid and one scalar. As we move away from this structure, we have to decouple the computing of the material properties as well.

What this PR does is essentially breaking up the fluid and scalar parsers inside `material_properties_t` and puts them inside the `scalar_` and `fluid_` scheme types. The `material_properties_t` is entirely removed, so it is no longer in the `case_t`.
The scalar also takes `rho` as part of its constructor now. We may need some more flexibility in the future, for e.g. when `rho` is not available from the fluid, but for now this is fine.

There is some awkwardness remaining with the user routine, since it is also hardcoded for a single fluid + scalar pair. Again, for now this works, but we will need to take some more steps elsewhere to make it more flexibile.


njansson (approved):

timfelle (approved):

timofeymukha:
@njansson merge?

njansson:
> @njansson merge?

Sure but shouldn't the automerge take care of that once checka are passed?

PR #1498, opened 2024-09-22
fix/parallelizing loop using DO CONCURRENT construct

Kamyarmvn:
Optimizing loops with DO CONCURRENT for parallel execution for time_interpolator_scalar, rhs_maker_oifs_cpu, and scalar_rhs_maker_oifs_cpu


njansson (approved):

timfelle (approved):

Kamyarmvn:
should I change something in the code to avoid the error?

timfelle:
No, The Nvidia ones are expected as they are slow to implement submodules.
The Hip one is an error during fetching HIP from the official sources, which happens sometimes. I'll poke it to see if they fixed their server.

PR #1500, opened 2024-09-23
Expanded examples + lambda2 fix on cpu + quality of life changes in force_torque

MartinKarp:
Expanded the turb_channel example and added a cylinder case to give a better idea of how to run a case in Neko and how to handle the results. Is based on #1494 and incorporates #1511 to assure proper behavior for single precision runs.

Also fixes some things in the force_torque calculation for the linter and using a scratch registry.

Also fixes the lambda2 calculation in single precision on the cpu.

These fixes are mostly related because they make the examples nicer to run.


njansson:
Great stuff! I guess we should wait to merge this until #1494 ?

MartinKarp:
Yes, exactly

examples/turb_channel/turb_channel.case
```diff
@@ -46,24 +46,42 @@
...
+      "scale": 0.01899,
+      "_scale_comment": "1 over area of wall (scale=1/(4pi*4/3pi))."
     },    
     { 
       "type": "fluid_stats",
       "output_control": "simulationtime",
-      "output_value": 20,
-      "start_time": 60.0,
+      "output_value": 5,
+      "_output_value_comment":"For example purposes we output relatively often, usually we do larger batches.",
```
timofeymukha, line 78: As a side comment, json_fortran supports having comment strings in JSON. I think we might want to allow that for the purpose of documentation like this.

timofeymukha (approved):

examples/turb_channel/turb_channel.case
timfelle: I agree. It seems odd to have an entry in order to comment.

To those who don't know, the comment syntax is:
```jsonc
// Some comment about what we do
"value": 10.0, // Comment about the variable
```

MartinKarp: Is this general for json or only for json_fortran?

timfelle: Yes it works out of the box with json fortran. I did not even consider that it was not standard since i have always used it.

MartinKarp: Okay, I will change to // then?

timofeymukha: Sounds good to me!

njansson (approved):

PR #1501, opened 2024-09-24
Add intersection detector for meshes

njansson:
Intersection detector for meshes, based on the aabb search tree.

Can be used to return a list of elements that overlaps with a given point 

Usage:

```f90
type(intersect_detector_t) :: intersect
type(stack_i4_t) :: overlaps

call intersect%init(mesh, aabb_padding)
call intersect%overlap(point, overlaps)
do while(.not. overlaps%is_empty())
     Process overlaped elements
end do
```

### DIFF
--- a/src/mesh/aabb_tree.f90
+++ b/src/mesh/aabb_tree.f90
@@ -337,12 +337,13 @@ contains
   end subroutine aabb_tree_init
 
   !> @brief Builds the tree.
-  subroutine aabb_tree_build_tree(this, objects)
+  subroutine aabb_tree_build_tree(this, objects, padding)
     use utils, only: neko_error
     implicit none
 
     class(aabb_tree_t), intent(inout) :: this
     class(*), dimension(:), intent(in) :: objects
+    real(kind=dp), optional, intent(in) :: padding
 
     integer :: i_obj, i_node, i
     logical :: done
@@ -352,6 +353,8 @@ contains
     type(aabb_t), dimension(:), allocatable :: box_list
     integer, dimension(:), allocatable :: sorted_indices
 
+    real(kind=dp) :: aabb_padding
+
     call this%init(size(objects) * 2)
 
     ! ------------------------------------------------------------------------ !
@@ -360,8 +363,14 @@ contains
 
     allocate(box_list(size(objects)))
 
+    if (present(padding)) then
+       aabb_padding = padding
+    else
+       aabb_padding = 0.0_dp
+    end if
+       
     do i_obj = 1, size(objects)
-       box_list(i_obj) = get_aabb(objects(i_obj))
+       box_list(i_obj) = get_aabb(objects(i_obj), aabb_padding)
     end do
     sorted_indices = sort(box_list)
 
@@ -608,14 +617,14 @@ contains
     class(aabb_tree_t), intent(in) :: this
     class(*), intent(in) :: object
     integer, intent(in) :: object_index
-    integer, intent(out) :: overlaps(:)
+    type(stack_i4_t), intent(inout) :: overlaps
 
     type(stack_i4_t) :: simple_stack
     type(aabb_t) :: object_box
 
     integer :: root_index, left_index, right_index
 
-    integer :: node_index
+    integer :: node_index, tmp_index
 
     object_box = get_aabb(object)
     root_index = this%get_root_index()
@@ -631,7 +640,8 @@ contains
        if (this%nodes(node_index)%aabb%overlaps(object_box)) then
           if (this%nodes(node_index)%is_leaf()) then
              if (this%nodes(node_index)%object_index .ne. object_index) then
-                overlaps = [this%nodes(node_index)%object_index, overlaps]
+                tmp_index = this%nodes(node_index)%object_index
+                call overlaps%push(tmp_index)
              end if
           else
              left_index = this%get_left_index(node_index)
--- a/src/mesh/intersection_detector.f90
+++ b/src/mesh/intersection_detector.f90
@@ -1,0 +1,117 @@
+! Copyright (c) 2024, The Neko Authors
+! All rights reserved.
+!
+! Redistribution and use in source and binary forms, with or without
+! modification, are permitted provided that the following conditions
+! are met:
+!
+!   * Redistributions of source code must retain the above copyright
+!     notice, this list of conditions and the following disclaimer.
+!
+!   * Redistributions in binary form must reproduce the above
+!     copyright notice, this list of conditions and the following
+!     disclaimer in the documentation and/or other materials provided
+!     with the distribution.
+!
+!   * Neither the name of the authors nor the names of its
+!     contributors may be used to endorse or promote products derived
+!     from this software without specific prior written permission.
+!
+! THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+! "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+! LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+! FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+! COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+! INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+! BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+! LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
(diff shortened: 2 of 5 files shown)


src/mesh/aabb_tree.f90
```diff
@@ -608,14 +617,14 @@ subroutine aabb_tree_query_overlaps(this, object, object_index, overlaps)
     class(aabb_tree_t), intent(in) :: this
     class(*), intent(in) :: object
     integer, intent(in) :: object_index
-    integer, intent(out) :: overlaps(:)
+    type(stack_i4_t), intent(inout) :: overlaps
```
timfelle, line 620: How about overloading instead? I could see people wanting to keep the list around if the same computations are done at every iteration, but the mesh is stationary.

njansson: How do you mean? Not sure if I follow

src/mesh/intersection_detector.f90
```diff
@@ -0,0 +1,117 @@
...
+  end type intersect_detector_t
+
+contains
+
+  !> Initialise an intersector detector for a given mesh @a msh
+  subroutine intersect_detector_init(this, msh, padding)
+    class(intersect_detector_t), intent(inout) :: this
+    type(mesh_t), target, intent(in) :: msh
+    real(kind=dp), intent(in), optional :: padding
+    type(hex_t), allocatable :: elements(:)
```
timfelle, line 62: I don't think you have to manually build the element list. Can you not pass msh%elements

njansson: No it didn't work, were some issues (later, probably in aabb) with the type of `elements` since it's a special type inside our mesh

njansson: As mentioned in the todo, hopefully we can remove this once the new mesh structure is in place

src/mesh/aabb_tree.f90
timfelle: You have removed the option to get a list of integers with the overlapping elements, and replaced it with a stack. Assume you wish to run the same computations after each timestep. Now you have to either save a deep copy of your stack, so you can run through it every time, or you have to call this to build a new stack every time.

With the integer list, you could run through the same list as many times as you need to.

So my suggestion would be to have two versions of the function, one returning a list of integers, one returning a stack. And then let the tree have a generic pointer to the two functions.

src/mesh/intersection_detector.f90
timfelle: You already added the `class is (element_t)`, could you not do the same for `class is (mesh_element_t)` and get the desired result?

src/mesh/aabb_tree.f90
njansson: Sure make sense. 

I recall that I had some issues with the prev. overlaps implementation and the growing of the list

src/mesh/intersection_detector.f90
timfelle: But sure, with a rework of the mesh it might not be worth doing.

src/mesh/aabb_tree.f90
timfelle: Sure that might be, possible one have to do some move_alloc stuff to make it better and not having to rebuild the array at every added overlap.

src/mesh/intersection_detector.f90
timfelle: In general we should probably have a build function that just takes a mesh at some point 😅

MartinKarp (approved):
I will use this for my interpolation and perhaps make some modifications later, looks good for me and good to have in develop.

src/mesh/intersection_detector.f90
```diff
@@ -0,0 +1,117 @@
...
+  private
+
+  type, public :: intersect_detector_t
+     type(mesh_t), private, pointer :: msh => null()
+     type(aabb_tree_t), private :: search_tree
+   contains
+     procedure, pass(this) :: init => intersect_detector_init
+     procedure, pass(this) :: free => intersect_detector_free
+     procedure, pass(this) :: overlap => intersect_detector_overlap
+  end type intersect_detector_t
```
timofeymukha, line 53: Please add docstrings here for the type, components and procedures.

For this lower-level algorithmic stuff, we really need to give a bit of detail on when / for what to use the type.

MartinKarp:
Does the overlap take the mesh but not the dofmap as input? With deformations, the actual GLL points might be important, but I will see.

Also, this will be a rather performance critical part of the code, have we made any checks with say thousands/millions of points?

njansson:
> Does the overlap take the mesh but not the dofmap as input? With deformations, the actual GLL points might be important, but I will see.
> 

The bounding box + padding should take care of that. Thus if a mesh has massive deformations the padding has to be increased 

> 
> 
> Also, this will be a rather performance critical part of the code, have we made any checks with say thousands/millions of points?

Yes, it's sufficiently fast, per point : O(n log n) for n number of elements.

I tried it with large models without any issues

MartinKarp:
After playing a bit with local and global searches, why does one use this and not the aabb_tree directly?

timfelle (commented):
> After playing a bit with local and global searches, why does one use this and not the aabb_tree directly?  

I guess this locks the mesh and the associated AABB tree together. Which could allow you to keep a collected object for accessing elements that overlap with a given point. I guess it makes it simple to update the underlying query structure if we need to later, assuming we find a better solution than the AABB trees.

In general i would not expect the AABB tree to be efficient on GPU, which might merrit a different approach when running on GPU.

timfelle (commented):
I have updated a bunch of the code.

We now have both an array and a stack version of the overlaps, if one wish to save the list of overlapping elements for later, the array would be simpler to use without having to recompute.
The destructor for the AABB tree is implemented.
I also added some docstrings which @timofeymukha asked for.

njansson:
> > After playing a bit with local and global searches, why does one use this and not the aabb_tree directly?
> 
> I guess this locks the mesh and the associated AABB tree together. Which could allow you to keep a collected object for accessing elements that overlap with a given point. I guess it makes it simple to update the underlying query structure if we need to later, assuming we find a better solution than the AABB trees.
> 
Exactly that was the intent, and also to create a simpler interface than having to deal with the aabb boxes everywhere if needed.

timfelle:
> > > After playing a bit with local and global searches, why does one use this and not the aabb_tree directly?
> > 
> > 
> > I guess this locks the mesh and the associated AABB tree together. Which could allow you to keep a collected object for accessing elements that overlap with a given point. I guess it makes it simple to update the underlying query structure if we need to later, assuming we find a better solution than the AABB trees.
> 
> Exactly that was the intent, and also to create a simpler interface than having to deal with the aabb boxes everywhere if needed.

Well even with just the AABB tree you don't need to do anything with the boxes themselves as a user.

MartinKarp:
I see, the main thing for me then is that since I need to use the actual coordinates of the elements to get all candidates, in addition to the mesh not always being available when one does interpolation. 

Would it be ok if I change it to use x,y,z instead of the mesh then?

timfelle:
> I see, the main thing for me then is that since I need to use the actual coordinates of the elements to get all candidates, in addition to the mesh not always being available when one does interpolation.

That's exactly why this is useful. You only need to have the mesh available upon initialization, after that the reference is stored in the object and you can do your look-ups whenever you wish.

 
> Would it be ok if I change it to use x,y,z instead of the mesh then?

I'm, not sure i follow? Do you wish to create an intersection detector for an array of points instead of for a mesh?

MartinKarp:
The issues with using mesh is the following, to avoid confusion I will refer to the actual spectral element mesh (consisting of the coordinates in the computational grid) as the spectral element grid:

1. Often one only has access to the grid points and not the connectivity needed to generate the mesh. Say if one reads an fld  file and wants to get the value of some coordinate in that grid. 
2. The mesh does not give any information of the deformations of the elements, but only the vertices (if even). It is as such not a reliable way of finding whether a point is in the spectral element grid or not. Adding padding is not a reliable way of solving that for all mesh geometries.
3.  The mesh is quite costly to build. This is a bit related to 1., but generating the connectivity often takes large amounts of time for large cases. 

Overall the issue is that one needs more geometrical information than what exists in the mesh and less of the connectivity, so it is both a bit overkill and too little at the same time.

njansson:
> The issues with using mesh is the following, to avoid confusion I will refer to the actual spectral element mesh (consisting of the coordinates in the computational grid) as the spectral element grid:
> 
> 1. Often one only has access to the grid points and not the connectivity needed to generate the mesh. Say if one reads an fld  file and wants to get the value of some coordinate in that grid.
> 2. The mesh does not give sufficient information of the deformations of the elements, but only the vertices (although most is somehow accessible in the deformed facets). It is as such not a reliable way of finding whether a point is in the spectral element grid or not. Adding padding is not a reliable way of solving that for all mesh geometries.

If the padding is based on the deformed mesh (dof coordinates) then it's reliable. The reason for using a mesh was to cut down the number of boxes in the tree. If one starts to doing it by the dof coordinates the cost will increase significantly (even if it is logarithmic).

Also, even if the tree is based on deformed coordinates, padding will still be essential to minimise the number of false positives. 

> 3. The mesh is quite costly to build. This is a bit related to 1., but generating the connectivity often takes large amounts of time for large cases.
> 
> Overall the issue is that one needs more geometrical information than what exists in the mesh and less of the connectivity, so it is both a bit overkill and too little at the same time.

MartinKarp:
But it would be the same number of boxes right? Just that we need to use the coordinates in the grid to calculate the bounding box.

PR #1503, opened 2024-09-24
Refactor device math and improve scoping

timfelle:
This pull request includes several commits that refactor the device math code and improve scoping. The commits move the device math code to the backend folder, move interfaces to device-specific code, and reduce the scope of modules. Additionally, the commits fix the processing of macros and add missing flags. These changes aim to improve the organization and efficiency of the codebase.


src/math/bcknd/device/device_math.F90
```diff
@@ -31,1092 +31,51 @@
...
+
+  use opencl_math, only: opencl_copy, opencl_rzero, opencl_cmult, &
+       opencl_cmult2, opencl_cadd, opencl_cadd2, opencl_cfill, opencl_add2, &
+       opencl_add4, opencl_add2s1, opencl_add2s2, opencl_addsqr2s2, &
+       opencl_add3, opencl_add3s2, opencl_invcol1, opencl_invcol2, &
+       opencl_col2, opencl_col3, opencl_subcol3, opencl_sub2, opencl_sub3, &
+       opencl_addcol3, opencl_addcol4, opencl_vdot3, opencl_glsc3, &
+       opencl_glsc3_many, opencl_add2s2_many, opencl_glsc2, opencl_glsum, &
+       opencl_masked_copy, opencl_cfill_mask
+
```
njansson, line 67: How about using submodules here as in the e.g. opr_cpu?

timfelle: Well that might be nice, but would not allow us to move the definition of the interface, so we would have the same long list of interface definitions as before. I wanted to make the file shorter and more readable.

timfelle: And on that note, I'm not sure you can do that since what we define are interface bindings to C. Not sure that works with submodules

njansson: Indeed, that's a good point

timfelle: This is a particular example of where the enforcement of `only:` seem a bit silly 😅 We manualy import everything from a large library of things instead of just using the library.

njansson: Yes, for these cases I think we should be more reasonable, but 100% of the file in an `only` statement doesn't really make sense.

src/math/bcknd/device/device_math.F90
```diff
@@ -31,1092 +31,51 @@
...
+       cuda_masked_copy, cuda_cfill_mask, cuda_masked_red_copy, cuda_vcross
+
+  use opencl_math, only: opencl_copy, opencl_rzero, opencl_cmult, &
+       opencl_cmult2, opencl_cadd, opencl_cadd2, opencl_cfill, opencl_add2, &
+       opencl_add4, opencl_add2s1, opencl_add2s2, opencl_addsqr2s2, &
+       opencl_add3, opencl_add3s2, opencl_invcol1, opencl_invcol2, &
+       opencl_col2, opencl_col3, opencl_subcol3, opencl_sub2, opencl_sub3, &
+       opencl_addcol3, opencl_addcol4, opencl_vdot3, opencl_glsc3, &
+       opencl_glsc3_many, opencl_add2s2_many, opencl_glsc2, opencl_glsum, &
+       opencl_masked_copy, opencl_cfill_mask
```
timfelle, line 66:
```suggestion
  use hip_math
  use cuda_math
  use opencl_math
```
As suggested

timofeymukha (approved):

njansson (approved):

PR #1505, opened 2024-09-24
Reverse dofmap to older version

MartinKarp:
Somewhere among these changes dofmap brakes the use of midpoint deformations (and perhaps other things as well).

Perhaps source of issue on Dardel I noted earlier as well. 

Quick check to see the issue is to run the basic cyl_bl example.

### DIFF
--- a/src/sem/dofmap.f90
+++ b/src/sem/dofmap.f90
@@ -252,73 +252,49 @@ contains
           global_id = msh%get_global(edge)
           edge_id = edge_offset + int((global_id - 1), i8) * num_dofs_edges(1)
           !Reverse order of tranversal if edge is reversed
-          if(int(edge%x(1), i8) .ne. this%dof(1,1,1,i)) then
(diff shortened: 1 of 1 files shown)


njansson:
I guess one mistake is edge_id which isn't incremented, so a simple fix could be to increase edge is with the iteration count after each loop nest (the ones split into two to avoid the if statement inside)

njansson:
> I guess one mistake is edge_id which isn't incremented, so a simple fix could be to increase edge is with the iteration count after each loop nest (the ones split into two to avoid the if statement inside)

Nope that's is not the issue, edge is is always recomputed before each nest

njansson:
Seems to be some invalid ranges for the merged nests, eg missing -1 etc.

njansson:
> Seems to be some invalid ranges for the merged nests, eg missing -1 etc.

yes that seems to have been the issue, a fix has been pushed to #1506.

PR #1511, opened 2024-09-26
Introduce extra precision - various bug fixes for sp and added test

MartinKarp:
Fixes reproducibility issues with single precision runs for the CPU as global reductions are carried in higher precision.

On the GPU reductions are less sensitive to the precision due to the structure off the reduction which avoids round off.

If this looks good I was thinking we could add it on all backends as well. Currently I set this to double precision, but I want to introduce xp as it might not be hardcoded to double in the future.

This got expanded now to quite a few places which calculate important constants. In particular, places including division and intricinsi operations got some consideration.


timofeymukha:
What exactly is the reproducibility issue?

Also, should it always be extra precision? Would it not be interesting to also be able to have communication in lower precision? E.g. communicate in sp when the computation is in dp? There is a paper with SSDC on this, and it seems to work pretty well for a rather applied case. So, something like `MPI_COMM_PRECISION`.

MartinKarp:
This is primarily introduced to make global reductions more reliable. An easy test to see why this is necessary is to run the code in single precision on a different number of MPI ranks, if one has around 4000 elements (array length of 4000*512=2 million) one gets that the value of global reductions can differ on order of 1-10%. 

For the GPU it is less significant as one first computes the reduction into batches and then adds those batches together, so the problem is more amplified  when running on a few number of CPU cores.

For gather scatter maybe lower is fine, but that would not be the same as this use case as it involves adding together relatively few numbers, not an entire reduction.

MartinKarp:
I thought a bit about different names, but thought extra was correct as the intended use of this one would be for operations that are extra sensitive to precision, rather than where one can use lower precision.

timofeymukha (approved):

njansson (commented):
Needs some more checks if we configure for quad, since then xp will be less accurate than rp

MartinKarp:
From what I gather qp does not compile at the moment in develop. In addition, I am not sure we have fixed the issue with MPI for quad precision. 

The intent of this PR was not to fix support for quad precision, but to start the journey toward making sp (or rather some mixed precision setting) more reliable and introduce a precision for areas where lower precision formats might need more precision.

timofeymukha:
I think some extra stuff got into the PR by accident.

reg_test/cylinder/ref1_dp.log
timfelle, line 1: Probably not intended

MartinKarp: I guess the tests are not in this PR, but I included them just to illustrate that the develop branch fails tests, but this one does not.

If we could merge #1524 I would be quite happy.

src/common/checkpoint.f90
```diff
@@ -76,6 +76,7 @@ module checkpoint
 
      real(kind=dp) :: t         !< Restart time (valid after load)
      type(mesh_t) :: previous_mesh
+     type(space_t) :: previous_Xh
```
timofeymukha, line 79: Is this related? It just looks very lonely :-).

src/simulation.f90
```diff
@@ -259,13 +259,14 @@ subroutine simulation_restart(C, t)
...
-    call C%fluid%chkp%previous_mesh%free()
     do i = 1, size(C%dtlag)
        call C%ext_bdf%set_coeffs(C%dtlag)
     end do
 
     call C%fluid%restart(C%dtlag, C%tlag)
-    if (allocated(C%scalar)) call C%scalar%restart( C%dtlag, C%tlag)
+    call C%fluid%chkp%previous_mesh%free()
+    if (allocated(C%scalar)) &
+    call C%scalar%restart( C%dtlag, C%tlag)
```
timofeymukha, line 269:
```suggestion
       call C%scalar%restart( C%dtlag, C%tlag)
```

src/fluid/fluid_pnpn.f90
```diff
@@ -427,19 +427,22 @@ subroutine fluid_pnpn_restart(this, dtlag, tlag)
...
+       call this%gs_Xh%op(this%v, GS_OP_ADD)
+       call this%gs_Xh%op(this%w, GS_OP_ADD)
+       call this%gs_Xh%op(this%p, GS_OP_ADD)
+
+       do i = 1, this%ulag%size()
+          call this%gs_Xh%op(this%ulag%lf(i), GS_OP_ADD)
+          call this%gs_Xh%op(this%vlag%lf(i), GS_OP_ADD)
+          call this%gs_Xh%op(this%wlag%lf(i), GS_OP_ADD)
+       end do
+    end if
```
timofeymukha, line 445: The stuff in this file also seems to pertain to something else, or?

MartinKarp: Otherwise this does not fix the test for restart

timofeymukha: Aah, I see

timfelle (commented):
Is it your intention to introduce the regression tests with this PR or is that only because you intended for the regression test PR to be pulled first?

src/fluid/fluid_pnpn.f90
```diff
@@ -427,19 +427,22 @@ subroutine fluid_pnpn_restart(this, dtlag, tlag)
...
-       call this%gs_Xh%op(this%ulag%lf(i), GS_OP_ADD)
-       call this%gs_Xh%op(this%vlag%lf(i), GS_OP_ADD)
-       call this%gs_Xh%op(this%wlag%lf(i), GS_OP_ADD)
-    end do
-
+    ! Make sure that continuity is maintained (important for interpolation)
+    ! Do not do this for lagged rhs
+    ! (derivatives are not necessairly coninous across elements)
+       
+    if (allocated(this%chkp%previous_mesh%elements) .or. this%chkp%previous_Xh%lx .ne. this%Xh%lx) then
```
timfelle, line 434:
```suggestion
    if (allocated(this%chkp%previous_mesh%elements) &
         .or. this%chkp%previous_Xh%lx .ne. this%Xh%lx) then
```

configure.ac
```diff
@@ -14,9 +14,10 @@ AC_ARG_ENABLE(contrib,
 
 AC_ARG_ENABLE(real,
 	      AS_HELP_STRING([--enable-real=Xp  Specify working precision of REAL types:
-                    sp -- REAL(kind=4)
-	            dp -- REAL(kind=8) (default)
-	            qp -- REAL(kind=16)]),
+                    ssp -- REAL(kind=4) xp -- sp 
+                    sp -- REAL(kind=4) xp -- dp  
+	            dp -- REAL(kind=8) xp -- dp (default)
+	            qp -- REAL(kind=16) xp -- qp]),
```
timfelle, line 20: Inconsistent whitespace.
```suggestion
                    dp -- REAL(kind=8) xp -- dp (default)
                    qp -- REAL(kind=16) xp -- qp]),
```

configure.ac
```diff
@@ -179,21 +180,42 @@ AX_ADIOS2
 AX_GSLIB
 
 # Select Working precision of REAL
-if test "x$enable_real" = xsp; then
+if test "x$enable_real" = xssp; then
    AC_SUBST(NEKO_REAL_TYPE, "sp")
    AC_SUBST(NEKO_C_REAL_TYPE, "c_float")
    AC_SUBST(NEKO_MPI_REAL_TYPE, "MPI_REAL")
    AC_SUBST(NEKO_DEV_REAL_TYPE, "float")
```
timfelle, line 187: Trailing whitespace
```suggestion
   AC_SUBST(NEKO_DEV_REAL_TYPE, "float")
```

configure.ac
```diff
@@ -179,21 +180,42 @@ AX_ADIOS2
...
-if test "x$enable_real" = xsp; then
+if test "x$enable_real" = xssp; then
    AC_SUBST(NEKO_REAL_TYPE, "sp")
    AC_SUBST(NEKO_C_REAL_TYPE, "c_float")
    AC_SUBST(NEKO_MPI_REAL_TYPE, "MPI_REAL")
    AC_SUBST(NEKO_DEV_REAL_TYPE, "float")   				
+   AC_SUBST(NEKO_XREAL_TYPE, "sp")
+   AC_SUBST(NEKO_C_XREAL_TYPE, "c_float")
+   AC_SUBST(NEKO_MPI_XREAL_TYPE, "MPI_REAL")
+   AC_SUBST(NEKO_DEV_XREAL_TYPE, "float")
```
timfelle, line 191: Trailing whitespace
```suggestion
   AC_SUBST(NEKO_DEV_XREAL_TYPE, "float")
```

configure.ac
```diff
@@ -179,21 +180,42 @@ AX_ADIOS2
...
    AC_SUBST(NEKO_DEV_REAL_TYPE, "float")   				
+   AC_SUBST(NEKO_XREAL_TYPE, "sp")
+   AC_SUBST(NEKO_C_XREAL_TYPE, "c_float")
+   AC_SUBST(NEKO_MPI_XREAL_TYPE, "MPI_REAL")
+   AC_SUBST(NEKO_DEV_XREAL_TYPE, "float")   				
+elif test "x$enable_real" = xsp; then
+   AC_SUBST(NEKO_REAL_TYPE, "sp")
+   AC_SUBST(NEKO_C_REAL_TYPE, "c_float")
+   AC_SUBST(NEKO_MPI_REAL_TYPE, "MPI_REAL")
+   AC_SUBST(NEKO_DEV_REAL_TYPE, "float")
```
timfelle, line 196: Trailing whitespace
```suggestion
   AC_SUBST(NEKO_DEV_REAL_TYPE, "float")
```

configure.ac
```diff
@@ -179,21 +180,42 @@ AX_ADIOS2
...
+   AC_SUBST(NEKO_DEV_REAL_TYPE, "float")   				
+   AC_SUBST(NEKO_XREAL_TYPE, "dp")
+   AC_SUBST(NEKO_C_XREAL_TYPE, "c_double")
+   AC_SUBST(NEKO_MPI_XREAL_TYPE, "MPI_DOUBLE_PRECISION")
+   AC_SUBST(NEKO_DEV_XREAL_TYPE, "double") 
 elif test "x$enable_real" = xdp; then
    AC_SUBST(NEKO_REAL_TYPE, "dp")
    AC_SUBST(NEKO_C_REAL_TYPE, "c_double")
    AC_SUBST(NEKO_MPI_REAL_TYPE, "MPI_DOUBLE_PRECISION")
    AC_SUBST(NEKO_DEV_REAL_TYPE, "double")
```
timfelle, line 205: Trailing whitespace
```suggestion
   AC_SUBST(NEKO_DEV_REAL_TYPE, "double")
```

configure.ac
```diff
@@ -179,21 +180,42 @@ AX_ADIOS2
...
    AC_SUBST(NEKO_DEV_REAL_TYPE, "double")   				
+   AC_SUBST(NEKO_XREAL_TYPE, "dp")
+   AC_SUBST(NEKO_C_XREAL_TYPE, "c_double")
+   AC_SUBST(NEKO_MPI_XREAL_TYPE, "MPI_DOUBLE_PRECISION")
+   AC_SUBST(NEKO_DEV_XREAL_TYPE, "double")   				
 elif test "x$enable_real" = xqp; then
    AC_SUBST(NEKO_REAL_TYPE, "qp")
    AC_SUBST(NEKO_C_REAL_TYPE, "c_long_double")
    AC_SUBST(NEKO_MPI_REAL_TYPE, "MPI_REAL16")
    AC_SUBST(NEKO_DEV_REAL_TYPE, "long double")
```
timfelle, line 214: Trailing whitespace
```suggestion
   AC_SUBST(NEKO_DEV_REAL_TYPE, "long double")
```

configure.ac
```diff
@@ -179,21 +180,42 @@ AX_ADIOS2
...
+   AC_SUBST(NEKO_DEV_XREAL_TYPE, "double")   				
 elif test "x$enable_real" = xqp; then
    AC_SUBST(NEKO_REAL_TYPE, "qp")
    AC_SUBST(NEKO_C_REAL_TYPE, "c_long_double")
    AC_SUBST(NEKO_MPI_REAL_TYPE, "MPI_REAL16")
    AC_SUBST(NEKO_DEV_REAL_TYPE, "long double")   				
+   AC_SUBST(NEKO_XREAL_TYPE, "qp")
+   AC_SUBST(NEKO_C_XREAL_TYPE, "c_long_double")
+   AC_SUBST(NEKO_MPI_XREAL_TYPE, "MPI_REAL16")
+   AC_SUBST(NEKO_DEV_XREAL_TYPE, "long double")
```
timfelle, line 218: Trailing whitespace
```suggestion
   AC_SUBST(NEKO_DEV_XREAL_TYPE, "long double")
```

configure.ac
```diff
@@ -14,9 +14,10 @@ AC_ARG_ENABLE(contrib,
 
 AC_ARG_ENABLE(real,
 	      AS_HELP_STRING([--enable-real=Xp  Specify working precision of REAL types:
-                    sp -- REAL(kind=4)
-	            dp -- REAL(kind=8) (default)
-	            qp -- REAL(kind=16)]),
+                    ssp -- REAL(kind=4) xp -- sp 
+                    sp -- REAL(kind=4) xp -- dp
```
timfelle, line 18: Trailing whitespace
```suggestion
                    sp -- REAL(kind=4) xp -- dp
```

MartinKarp:
> Is it your intention to introduce the regression tests with this PR or is that only because you intended for the regression test PR to be pulled first?

My original intent was to merge the regression tests first, but as one should not add tests that do not pass I not thought why not merge them with this PR.

vbaconnet (approved):

PR #1519, opened 2024-09-27
User statistics

MartinKarp:
User defined statistics. 

Picks a field from the field_registry and computes the temporal average. Going to look through for documentation and such, but most things are in place.

The ordering of the output fields is kind of all over the place. Maybe best if everything would be output into the scalars?

Finishes the CTR porting adventure, interpolation next.


src/io/mean_field_output.f90
```diff
@@ -0,0 +1,192 @@
...
+            !do i = 1, this%map_2d%n_2d
+            !   u = output_2d%v%x(i)
+            !   v = output_2d%w%x(i)
+            !   w = output_2d%p%x(i)
+            !   p = output_2d%u%x(i)
+            !   output_2d%p%x(i) = p
+            !   output_2d%u%x(i) = u
+            !   output_2d%v%x(i) = v
+            !   output_2d%w%x(i) = w
+            !end do
```
timfelle, line 178: Maybe a bit of cleanup is in order?
```suggestion
```

timofeymukha:
Don't forget the documentation :-).

src/simulation_components/user_stats.f90
```diff
@@ -0,0 +1,186 @@
...
+       end do
+       deallocate(this%mean_fields)
+    end if
+    if (allocated(this%field_names)) then
+       deallocate(this%field_names)
+    end if
+
+  end subroutine user_stats_free
+
+  !> Compute the vorticity field.
```
timofeymukha, line 167:
```suggestion
  !> Update the statistics.
```

timfelle:
> Everything almost works here, except averaging exactly one field over one dimension. This dies with
> 
> ```
> #3  0x104396763 in __fld_file_MOD_fld_file_write
>         at io/fld_file.f90:393
> #4  0x104396763 in __fld_file_MOD_fld_file_write
>         at io/fld_file.f90:393
> #5  0x1043af9cb in __file_MOD_file_write
>         at io/file.f90:167
> #6  0x1043b675f in __mean_field_output_MOD_mean_field_output_sample
>         at io/mean_field_output.f90:168
> #7  0x1043b876f in __output_controller_MOD_output_controller_execute
>         at io/output_controller.f90:283
>         
>         
> ```
> 
> So in this code block
> 
> ```fortran
>     if (write_velocity) then
>        byte_offset = mpi_offset + int(offset_el, i8) * &
>             (int(gdim * (lxyz), i8) * int(FLD_DATA_SIZE, i8))
>        call fld_file_write_vector_field(this, fh, byte_offset, &
>             u%ptr, v%ptr, w%ptr, n, gdim, lxyz, nelv)
> 
>        mpi_offset = mpi_offset + int(glb_nelv, i8) * &
>             (int(gdim * (lxyz), i8) * &
>             int(FLD_DATA_SIZE, i8))
> ```
> 
> I don't know very well how this stuff works. When you init an fld_file_data_t (the object written here) with less or equal than 3 fields, it inits the velocity compoents, so e.g. for one field it only inits `u`.
> 
> ```fortran
> subroutine fld_file_data_init_n_fields(this, n_fields, n)
>     class(fld_file_data_t), target, intent(inout) :: this
>     integer, intent(in) :: n, n_fields
>     integer :: i, j
> 
> 
>     if(n_fields .gt. 0) then
>        call this%u%init(n)
>     end if
>     if(n_fields .gt. 1) then
>        call this%v%init(n)
>     end if
>     if(n_fields .gt. 2) then
>        call this%w%init(n)
>     end if
> ```
> 
> So I thought that maybe the bug is that the writer tries to write all u,v,w, but only u is inited. However, everything works with 2 fields to average, which should also be broken if my theory was correct. @njansson @timfelle any ideas?

So if the gdim value is set to 2, then the z is just pointed to y and the `fld_file_write_metadata_vector` function does not export z. Try to place if gdim .ge. 2 around the y part. It might be related to that bit

timofeymukha:
> > So I thought that maybe the bug is that the writer tries to write all u,v,w, but only u is inited. However, everything works with 2 fields to average, which should also be broken if my theory was correct. @njansson @timfelle any ideas?
> 
> So if the gdim value is set to 2, then the z is just pointed to y and the `fld_file_write_metadata_vector` function does not export z. Try to place if gdim .ge. 2 around the y part. It might be related to that bit
> 
> ```fortran
>   subroutine fld_file_write_metadata_vector(this, fh, byte_offset, x, y, z, &
>        gdim, lxyz, nelv)
>     class(fld_file_t), intent(inout) :: this
>     type(MPI_File), intent(inout) :: fh
>     integer, intent(in) :: gdim, lxyz, nelv
>     real(kind=rp), intent(in) :: x(lxyz, nelv), y(lxyz, nelv), z(lxyz, nelv)
>     integer (kind=MPI_OFFSET_KIND), intent(in) :: byte_offset
>     integer :: el, j, ierr, nout
>     type(MPI_Status) :: status
>     real(kind=sp) :: buffer(2*gdim*nelv)
> 
>     j = 1
>     do el = 1, nelv
>        buffer(j+0) = real(vlmin(x(1, el), lxyz), sp)
>        buffer(j+1) = real(vlmax(x(1, el), lxyz), sp)
>        buffer(j+2) = real(vlmin(y(1, el), lxyz), sp)
>        buffer(j+3) = real(vlmax(y(1, el), lxyz), sp)
>        j = j + 4
>        if (gdim .eq. 3) then
>           buffer(j+0) = real(vlmin(z(1, el), lxyz), sp)
>           buffer(j+1) = real(vlmax(z(1, el), lxyz), sp)
>           j = j + 2
>        end if
>     end do
> 
>     ! write out data
>     nout = 2*gdim*nelv
> 
>     call MPI_File_write_at_all(fh, byte_offset, buffer, nout, &
>          MPI_REAL, status, ierr)
> 
>   end subroutine fld_file_write_metadata_vector
> ```

Adding these extra ifs resolves this
```fortran
       if (data%u%size() .gt. 0) then
          u%ptr => data%u%x
          ! In case only u is actually allocated, point the other comps to u
          ! so that we don't die on trying to write them
          if (data%v%size() .le. 0) v%ptr => data%u%x
          if (data%w%size() .le. 0) w%ptr => data%u%x
          write_velocity = .true.
       end if
```
I added for w for good measure. It is just a coincidence that it works for 2 fields because of the dimensionality thing that you mentioned, Tim. This stuff feels kinda hacky, not gonna lie :). I think the better way is to allocate the field data in a different way: allocate velocity only of you have at least 3 fields. But I guess for now, this will work.

timofeymukha:
Ready for review. @timfelle do you know off the top of your head why genmeshbox is not in the PATH for CI?

adperezm:
> It is just a coincidence that it works for 2 fields because of the dimensionality thing

write_vector expects a vector field, so in 3d it needs 3 and in 2d 2. there is a write_scalar routine that can be used per field. I would argue that if you are averaging only 1 field, then the correct thing would be to write it as a scalar.

timofeymukha:
> > It is just a coincidence that it works for 2 fields because of the dimensionality thing
> 
> write_vector expects a vector field, so in 3d it needs 3 and in 2d 2. there is a write_scalar routine that can be used per field. I would argue that if you are averaging only 1 field, then the correct thing would be to write it as a scalar.

Yeah, I agree. The reason the vector write is called now is because when you init a fld_file_data requesting n fields to be initialized, it starts from the velocity, then pressure, then temperature and then scalars. So if you request 1 field, it just inits u. However, in the fld writer, it checks if u is allocated and if it is, it assumes that v and w are as well (which is actually reasonable, I think). So ultimately we need to change how the fld_file_data inits I think. But I am a bit scared to do a quick change there in this PR, because I am not sure what other existing stuff may rely on status quo.

adperezm:
> > > It is just a coincidence that it works for 2 fields because of the dimensionality thing
> > 
> > 
> > write_vector expects a vector field, so in 3d it needs 3 and in 2d 2. there is a write_scalar routine that can be used per field. I would argue that if you are averaging only 1 field, then the correct thing would be to write it as a scalar.
> 
> Yeah, I agree. The reason the vector write is called now is because when you init a fld_file_data requesting n fields to be initialized, it starts from the velocity, then pressure, then temperature and then scalars. So if you request 1 field, it just inits u. However, in the fld writer, it checks if u is allocated and if it is, it assumes that v and w are as well (which is actually reasonable, I think). So ultimately we need to change how the fld_file_data inits I think. But I am a bit scared to do a quick change there in this PR, because I am not sure what other existing stuff may rely on status quo.

Oh that's interesting. I do not remember the fld_data itself, but I remember a file where these checks were made and then the order of checks were different. For 1 field it assigned to pressure, then for 2 it assigned pressure and temperature, then for 3 it assigned u,v,w and so on. 

Maybe things have changed. But yes, I agree. It seems the problem is in that assign logic. We can wait to see if someone knows the dependencies of that.

timofeymukha:
Unclear what is happening with the sp macOS, according the log, the file is written.

```
   --------Postprocessing--------

    --------Writer output---------
    File name     : user_stats.csv
    Output number :     1
    Writing at time:   500.000000 Output time (s):  0.000007

   ---------Step summary---------
   Total time for step      500 (s):   0.2180000E-03
   Total elapsed time (s):             0.1271020E+00

  --------Writer output---------
  File name     : user_stats.csv
  Output number :     2
  Writing at time:   500.000000 Output time (s):  0.000009
 Normal end.
```

timfelle (approved):
Looks good 👍

PR #1520, opened 2024-09-27
Harmonize dofmap constructor and destructor, bump style

timofeymukha:
Converts the dofmap constructor to a %init procedure, just like all other types have. Same for destructor. Bumps code style score.

Question, why is the dofmap destructor not called in some places. There is a comment in schwarz.f90 on line 155. I was generally surprised that I didn't have to refactor a single destructor call, excluding the one from the init.

### DIFF
--- a/bench/gs/driver.f90
+++ b/bench/gs/driver.f90
@@ -38,7 +38,7 @@ program gsbench
 
   call Xh%init(GLL, lx, lx, lx)
 
-  dm = dofmap_t(msh, Xh)
+  call dm%init(msh, Xh)
   call gs_h%init(dm)
 
   n = Xh%lx * Xh%ly * Xh%lz * msh%nelv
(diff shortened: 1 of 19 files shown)


njansson (approved):

MartinKarp (approved):
The comment is from me, in general I have found that for types with a constructor like 

```fortran
blob = blob_t(args)
```

often segfaults and leads to weird memory problems (especially if one tries to free it as well). Recently I found this because I think it is the way several of the older output_t types (mean_flow__output_t for example) are initialized which lead to weird errors until I changed to %init like is done here.

timofeymukha:
> The comment is from me, in general I have found that for types with a constructor like
> 
> ```fortran
> blob = blob_t(args)
> ```
> 
> often segfaults and leads to weird memory problems (especially if one tries to free it as well). Recently I found this because I think it is the way several of the older output_t types (mean_flow__output_t for example) are initialized which lead to weird errors until I changed to %init like is done here.

OK, it's good to know that we ended up settling for the good way to init then :-). I guess we need to be on the look for unfreed dofmap stuff.

PR #1521, opened 2024-09-30
Check for existence of csv file in `count_lines`

vbaconnet:
Fixes a bug in `csv_file_count_lines` that caused segfault without a proper error message if the csv file did not exist.

Noticed it when using probes with an input csv that did not exist. Since we call `csv_file_count_lines` before `csv_file_read` the usual safeguards were bypassed.

Perhaps could be good to cherry-pick this into `release` as well?

### DIFF
--- a/src/io/csv_file.f90
+++ b/src/io/csv_file.f90
@@ -310,9 +310,9 @@ contains
 
     call this%check_exists()
 
-    open(file = trim(f%fname), status = 'old', newunit = file_unit, &
+    open(file = trim(this%fname), status = 'old', newunit = file_unit, &
          iostat = ierr)
-    if (ierr .ne. 0) call neko_error("Error while opening " // trim(f%fname))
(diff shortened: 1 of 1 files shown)


src/io/csv_file.f90
```diff
@@ -308,8 +308,11 @@ function csv_file_count_lines(this) result(n)
     integer :: n
     integer :: ierr, file_unit
 
-    open(file = trim(this%fname), status = 'old', newunit = file_unit)
-    rewind(file_unit)
```
timofeymukha, line 312: What does `rewind` do? It's not there in the new code.

vbaconnet: Oops that's my bad. It's supposed to position the file to its initial starting point to not skip anything. And it's supposed to be there... a bad copy/paste :)

timofeymukha (approved):

njansson (approved):

PR #1525, opened 2024-10-01
Ignore warnings in GNU

timofeymukha:
Adds -w to GNU checks to avoid unloadable logs with 80k lines.

For intel, I try to disable the ifort deprecation remark, which also adds some noise to the log. I thinkt the `-warn errors` already kill the warning for intel and the logs are indeed short.

### DIFF
--- a/.github/workflows/check_intel.yml
+++ b/.github/workflows/check_intel.yml
@@ -81,4 +81,4 @@ jobs:
         run: |
           ./regen.sh
           ./configure FC=${FC} CC=${CC} MPIFC"=mpiifort -fc=${FC}" --enable-real=${RP}
-          make FCFLAGS="-diag-disable:10448 -O2 -stand f08 -warn errors `pkg-config --cflags json-fortran`" -j$(nproc)
+          make FCFLAGS="-O2 -stand f08 -warn errors `pkg-config --cflags json-fortran`" -j$(nproc)


MartinKarp (approved):

njansson (commented):
My concern is that disabling this would also disable the strict standard conforming checks

timfelle:
> My concern is that disabling this would also disable the strict standard conforming checks

It should only disable the checks which report warnings, not the ones that report errors. So if we standard conforming warnings then yes they would be turned off. But whu can tell if they are being reported or not among the more than 70.000 lines of output.

The "true" solution would be to fix the warnings. There must be a way to do so without ruining the performance. I have seen some hints at `associate` might be able to improve matters, but i haven't investigated it. My best bet would be to move away from the N-D definitions inside the operators and use linear indices instead.

njansson:
> > My concern is that disabling this would also disable the strict standard conforming checks
> 
> It should only disable the checks which report warnings, not the ones that report errors. So if we standard conforming warnings then yes they would be turned off. But who can tell if they are being reported or not among the more than 70.000 lines of output.

For intel we have `-warn error` so any standard warnings will result in an error

> 
> The "true" solution would be to fix the warnings. There must be a way to do so without ruining the performance. I have seen some hints at `associate` might be able to improve matters, but i haven't investigated it. 

I've investigated associated,  and is truly a mixed bag experience. Most compilers will see the associated value as a pointer,  and thus refuse to vectorise it.

> My best bet would be to move away from the N-D definitions inside the operators and use linear indices instead.

Not sure, some of them well then be rather tricky to implement I would say

timfelle:
> > > My concern is that disabling this would also disable the strict standard conforming checks
> > It should only disable the checks which report warnings, not the ones that report errors. So if we standard conforming warnings then yes they would be turned off. But who can tell if they are being reported or not among the more than 70.000 lines of output.
> For intel we have `-warn error` so any standard warnings will result in an error
Oh but intel is not the issue here, the warning suppression is for gfortran alone.

> > The "true" solution would be to fix the warnings. There must be a way to do so without ruining the performance. I have seen some hints at `associate` might be able to improve matters, but i haven't investigated it.
> I've investigated associated, and is truly a mixed bag experience. Most compilers will see the associated value as a pointer, and thus refuse to vectorise it.
> > My best bet would be to move away from the N-D definitions inside the operators and use linear indices instead.
> Not sure, some of them well then be rather tricky to implement I would say
Sure, I can see that they would be.

timfelle (approved):

PR #1526, opened 2024-10-01
Integration tests with pytest

timofeymukha:
I have now developed this into something that is mergable. For now, parallel to reframe and the simple regression test. However, this deprecates the smoke tests in reframe, so I commented it out of the CI, at least for now.

So, this adds a pytest-based integration test framework, which can be run on any machine (I think), provided a launcher scripts to start neko for the tested environment. Please read the README.md for details.

I add two tests

- Smoke-screen example executions for tgv, hemi and rayleigh_bendard. We can easily add any other example.
- Compilation of all the example user files with makeneko.

There is also a dummy test that just demonstrates how to add a test.

We will pass over the regression test with the cylinder we have to this frameworks, once we have the updated log format.


timofeymukha:
#1525 got pulled in here by accident.

MartinKarp (commented):
Sure interesting, can pytest run MPI as well?

timofeymukha:
> Sure interesting, can pytest run MPI as well?

Yeah, good point. So, I just launch a subprocess in the shell to run neko, so one should just add mpirun to that command, which is passed to the shell. Should be no problem. Currently, I am trying to jsut launch neko on the CI :-)).

timofeymukha:
So pytest is essentially just a wrapper here, to make things pretty and organized. If you really feel like it, you can even launch your awk stuff as a subprocess from the test as well and then parse the output :-).

timofeymukha:
OK, I am happy now. So, all the neko runs should put the logs under "logs". If the CI fails, this folder is put up as an artifact, available for download, just like with the unit tests.

We can tweak all of this, but @MartinKarp maybe you can try transferring your test into this system. Adding a test is very easy, each test is just 1 python function starting with test_. A single .py file defines a test suite, which sits in its own directory.

njansson (commented):
A modified reframe suite that uses a staged build of Neko would remove a lot of redundancy in these tests and make them more maintainable

examples/hemi/hemi.case
njansson, line 1: Is this intended to be changed?

timofeymukha: No :-)

njansson:
> tgv on MacOS 15 dies with
> 
> `Program received signal SIGBUS: Access to an undefined portion of a memory object.`
> 
> Not sure what the issue is.

Tried with MacOS 15, gcc 14.2.0 and NAG 7.2 but I couldn't reproduce the issue :(

timofeymukha:
Any ideas for a remedy are welcome!

timofeymukha:
I changed tgv to cylinder, and MacOS is happy now. I guess it is still 1 extra smoke test compared to what we have now.

Very much ready for review, especially with the new log style merged in!

njansson:
Nice!

Since there's a lot of overlap with our reframe tests I would suggest that we remove the reframe part of the integration tests, and move + extended these for some nightly (long) tests instead. 

One reason for keeping it as reframe is that any center could then easily test neko on their machine in a somewhat portable and "standardises" HPC way

njansson:
No don't remove. We will need it very soon

njansson:
> > No don't remove. We will need it very soon
> 
> Anything blocking a merge then?

How about adding the integration tests to nvhpc, and Nvidia as well?

timofeymukha:
@njansson I restored the cylinder test. 

For nvidia, I finally managed to make things run, but the tests don't pass. One just segfaults.

```
 ---Fluid initial condition----
  Type : uniform
  Value: [1.000000E+00, 0.000000E+00, 0.000000E+00]
[pkrvmxyh4eaekms:35795:0:35795] Caught signal 11 (Segmentation fault: Sent by the kernel at address (nil))

  -----Adding write output------
  File name        : /tmp/pytest-of-runner/pytest-0/test_demo0/field.fld
  Write control    : simulationtime
  Output precision : single
  Writes per time unit (Freq.):  5.000000E-01
  Time between writes:  2.000000E+00
[pkrvmxyh4eaekms:35794:0:35794] Caught signal 11 (Segmentation fault: Sent by the kernel at address (nil))
BFD: /lib/x86_64-linux-gnu/libc.so.6: unknown type [0x13] section `.relr.dyn'
BFD: /lib/x86_64-linux-gnu/libc.so.6: unknown type [0x13] section `.relr.dyn'
BFD: /lib/x86_64-linux-gnu/libc.so.6: unknown type [0x13] section `.relr.dyn'
BFD: /lib/x86_64-linux-gnu/libc.so.6: unknown type [0x13] section `.relr.dyn'
BFD: /lib/x86_64-linux-gnu/libc.so.6: unknown type [0x13] section `.relr.dyn'
BFD: /lib/x86_64-linux-gnu/libc.so.6: unknown type [0x13] section `.relr.dyn'
BFD: /lib/x86_64-linux-gnu/libc.so.6: unknown type [0x13] section `.relr.dyn'
BFD: /lib/x86_64-linux-gnu/libc.so.6: unknown type [0x13] section `.relr.dyn'
BFD: /lib/x86_64-linux-gnu/libc.so.6: unknown type [0x13] section `.relr.dyn'
BFD: /lib/x86_64-linux-gnu/libc.so.6: unknown type [0x13] section `.relr.dyn'
==== backtrace (tid:  35795) ====
 0 0x0000000000045330 __sigaction()  ???:0
 1 0x00000000002ddf72 pgf90_dealloc_mbr03_i8()  ???:0
 2 0x0000000000394f5a pgf90_dealloc_poly_mbr03a_i8()  ???:0
 3 0x000000000069f07f output_output_init_()  /home/runner/work/neko/neko/src/io/output.f90:0
 4 0x0000000000574cde chkp_output_chkp_output_init_()  /home/runner/work/neko/neko/src/io/chkp_output.f90:79
 5 0x0000000000480351 case_case_init_common_()  /home/runner/work/neko/neko/src/case.f90:445
 6 0x000000000047caf6 case_case_init_from_file_()  /home/runner/work/neko/neko/src/case.f90:120
 7 0x0000000000413236 neko_neko_init_()  /home/runner/work/neko/neko/src/neko.f90:212
 8 0x0000000000406e97 MAIN_()  /home/runner/work/neko/neko/src/driver.f90:6
 9 0x0000000000406e31 main()  ???:0
10 0x000000000002a1ca __libc_init_first()  ???:0
11 0x000000000002a28b __libc_start_main()  ???:0
```
The funny thing is that this runs the cylinder case file, and the smoke cylinder example does run...

The other one is the RB smoke example, which dies with
```
 ------------Scalar------------
  Type       : Modular (Pn/Pn)
  Name       : s
  Ksp scalar : (cg, jacobi)
   `-abs tol : 1.000000E-06
  lambda     : 1.000000E+00
  cp         : 1.000000E+00
  LES        : F
 *** ERROR WHEN SELECTING TYPE ***
 Type dirichlet does not exist for scalar_pnpn boundary conditions
 Valid types are:
     dirichlet                
     user_pointwise           
     user                                     
     neumann                  
```
No idea why.

If someone has a setup with nvfortran, I would say it should be looked into. But for now maybe we can comment out the integration test stage.

timofeymukha:
My bad, the cylinder smoke test also fails, so that is consistent at least.

njansson:
> I think we should keep with GNU for now and merge this in. There are many things to build on here, merging the cylinder test in the new framework for starters.
> 
> There is something with paths for Intel, and for nvidia I guess we just have to debug with a local nvfortran installation.

is the cylinder tests now portred over? Run, checkpoint, restart and compare?

timofeymukha:
> > I think we should keep with GNU for now and merge this in. There are many things to build on here, merging the cylinder test in the new framework for starters.
> > There is something with paths for Intel, and for nvidia I guess we just have to debug with a local nvfortran installation.
> 
> is the cylinder tests now portred over? Run, checkpoint, restart and compare?

Not ported over, but I simply added the old test to the CI, so it still exists.  At the next stage, we will port it into the pytest framework.

timofeymukha:
Alright, commented out intel and nvidia for now, so that checks are passed.

timfelle (approved):

PR #1528, opened 2024-10-02
Fix scalar scheme message for material properties

timfelle:
The commit "scalar scheme message surpassed `LOG_SIZE`" fixes an issue where the specific heat capacity message resulted in a `Fortran runtime error: End of record` due to exceeding the log size. The commit updates the message to ensure it is displayed correctly.

### DIFF
--- a/src/scalar/scalar_scheme.f90
+++ b/src/scalar/scalar_scheme.f90
@@ -712,8 +712,9 @@ contains
           write(log_buf, '(A)') 'Non-dimensional scalar material properties &
           & input.'
           call neko_log%message(log_buf, lvl = NEKO_LOG_VERBOSE)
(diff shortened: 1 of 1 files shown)


njansson (approved):

njansson:
ps. should we also bump the buffer size?

timfelle (commented):
> ps. should we also bump the buffer size? 

Well do we want longer lines in the log than 80 chars? That is kinda what the `LOG_SIZE` controls.

njansson:
> > ps. should we also bump the buffer size?
> 
> Well do we want longer lines in the log than 80 chars? That is kinda what the `LOG_SIZE` controls.

Thanks for the reminder, that is a good point ;)

timofeymukha (approved):

PR #1530, opened 2024-10-02
Appending a source term to a case with no source terms

gloopydoop:
I was having trouble appending source terms to a case with no source terms read from the JSON.

n_sources = size(this%source_terms) would return 1, then it would append the additional source term in position 2.
Then it would segfault trying to execute the unallocated source_terms(1).

### DIFF
--- a/src/source_terms/source_term_handler.f90
+++ b/src/source_terms/source_term_handler.f90
@@ -226,9 +226,9 @@ contains
     integer :: n_sources, i
 
     if(allocated(this%source_terms)) then
-    	n_sources = size(this%source_terms)
+       n_sources = size(this%source_terms)
     else
-    	n_sources = 0
+       n_sources = 0
     endif
 
     call move_alloc(this%source_terms, temp)
--- a/src/source_terms/source_term_handler.f90
+++ b/src/source_terms/source_term_handler.f90
@@ -225,7 +225,12 @@ contains
 
     integer :: n_sources, i
 
-    n_sources = size(this%source_terms)
+	 if(allocated(this%source_terms)) then
+    	n_sources = size(this%source_terms)
+    else
+    	n_sources = 0
+    endif
+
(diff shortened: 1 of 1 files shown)


njansson (approved):
Great catch! Only some standard violations to fix before ci is happy (the Intel log will reveal the issue)

src/source_terms/source_term_handler.f90
```diff
@@ -225,7 +225,12 @@ subroutine source_term_handler_add_source_term(this, source_term)
 
     integer :: n_sources, i
 
-    n_sources = size(this%source_terms)
+	 if(allocated(this%source_terms)) then
+    	n_sources = size(this%source_terms)
+    else
```
njansson:
```suggestion
    if(allocated(this%source_terms)) then
    	n_sources = size(this%source_terms)
    else
```

njansson: Fix standard conformance

njansson (approved):

src/source_terms/source_term_handler.f90
```diff
@@ -225,7 +225,12 @@ subroutine source_term_handler_add_source_term(this, source_term)
 
     integer :: n_sources, i
 
-    n_sources = size(this%source_terms)
+    if(allocated(this%source_terms)) then
+    	n_sources = size(this%source_terms)
```
timfelle, line 229:
```suggestion
       n_sources = size(this%source_terms)
```

src/source_terms/source_term_handler.f90
```diff
@@ -225,7 +225,12 @@ subroutine source_term_handler_add_source_term(this, source_term)
...
+    else
+    	n_sources = 0
```
timfelle, line 231:
```suggestion
       n_sources = 0
```

timfelle (approved):

PR #1539, opened 2024-10-04
Rename s to sampler

timofeymukha:
Renames case%s to case%sampler. 

s is just too short to convey any meaning. I sometimes confused with scalar also, because that is the typical name of the scalar field.

@MartinKarp I know you don't like the type being called sampler in the first place (me as well I think), but we can maybe have this small improvement for now.

### DIFF
--- a/src/case.f90
+++ b/src/case.f90
@@ -69,7 +69,7 @@ module case
      real(kind=rp) :: dt
      real(kind=rp) :: end_time
      character(len=:), allocatable :: output_directory
-     type(sampler_t) :: s
+     type(sampler_t) :: sampler
      type(fluid_output_t) :: f_out
      type(chkp_output_t) :: f_chkp
      type(user_t) :: usr
@@ -383,7 +383,7 @@ contains
     !
     ! Setup sampler
     !
-    call this%s%init(this%end_time)
+    call this%sampler%init(this%end_time)
     if (scalar) then
        this%f_out = fluid_output_t(precision, this%fluid, this%scalar, &
             path = trim(this%output_directory))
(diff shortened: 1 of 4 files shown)


MartinKarp:
Could we go with writer or similar? 

I think that is more or less what the class will be renamed to as well. Big improvement regardless if you go with sampler or writer over s.

 Sampler also confuses with sampling the field for say statistics sometimes.

timofeymukha:
> Could we go with writer or similar?
> 
> I think that is more or less what the class will be renamed to as well. Big improvement regardless if you go with sampler or writer over s.
> 
> Sampler also confuses with sampling the field for say statistics sometimes.

I would like to avoid the variable name and the type name to be in disharmony. Kind of intended this as a quick fix / improvement. But yeah, I think generally writer would be a better name than sampler, even if I somehwat understand the logic behind the latter as well.

MartinKarp (approved):
Okay, lets go with this then!

timofeymukha:
Haha, @MartinKarp I have changed mind somewhat right when you approved :P. Anyway, I added a bit of refactoring to the sampler, and standardized the output_t pointer.

So, now that we are here, maybe let's use this PR after all to determine the type name we like, and I will implement it. I think `writer_t` is pretty straightforward, or `output_executor` similar to `simcomp_executor`.

njansson (approved):

njansson:
Writer I think is a bit wrong since the module doesn't write anything, the only thing it does is to sample a flow field, so some kind of a exec or controller name is more suitable

MartinKarp:
I guess output_controller/executor is good. I was more thinking that a writer (write_t) would write an output(output_t) to file (file_t).

Overall I think the type itself is quite confusing though.

timofeymukha:
Refactored to output_controller. The word sampling is now avoided. Added param docstrings

njansson:
> Refactored to output_controller. The word sampling is now avoided. Added param docstrings

I would suggest to add field to the name since it does only deals with fields nothing else

MartinKarp (approved):
Looks good! Thank you for you for starting to look at this Tim.

timofeymukha:
> > Refactored to output_controller. The word sampling is now avoided. Added param docstrings
> 
> I would suggest to add field to the name since it does only deals with fields nothing else

Not necessarily, right? In principle, I can put any output in it, even if the ones we have now all deal with fields. For example, the probes could in principle be organized by creating an csv output type, and add that to the controller.

PR #1541, opened 2024-10-10
Add option to create aabb tree directly from aabb and fix sorting bug

MartinKarp:
Maybe it was possible to use the aabb directly previously, but I used it to create my global search tree of the whole domain for the global interpolation.

The sorting seems to have fixed a bug with some kind of race condition I had.

### DIFF
--- a/src/mesh/aabb_tree.f90
+++ b/src/mesh/aabb_tree.f90
@@ -439,8 +439,9 @@ contains
        minidx = -1
        do imin = 1, size(array)
           if (.not. visited(imin) .and. minidx .eq. -1) minidx = imin
-
-          if (visited(imin) .and. array(imin) .lt. array(minidx)) minidx = imin
+          if (minidx .gt. -1) then
+             if (visited(imin) .and. array(imin) .lt. array(minidx)) minidx = imin
+          end if
        end do
 
        indices(i) = minidx
@@ -641,6 +642,7 @@ contains
           end if
        end if
     end do
+    call simple_stack%free()
   end subroutine aabb_tree_query_overlaps
 
   ! -------------------------------------------------------------------------- !
--- a/src/mesh/search_tree/aabb.f90
+++ b/src/mesh/search_tree/aabb.f90
@@ -188,7 +188,7 @@ contains
     select type(object)
 
       type is (aabb_t)
-       box%init(object%get_min(), object%get_max())
+       call box%init(object%get_min(), object%get_max())
       type is (tri_t)
        box = get_aabb_element(object)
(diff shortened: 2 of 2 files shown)


src/mesh/search_tree/aabb.f90
```diff
@@ -187,6 +187,8 @@ function get_aabb(object, padding) result(box)
 
     select type(object)
 
+      type is (aabb_t)
+       call box%init(object%get_min(), object%get_max())
```
timfelle, line 191: So you just want a copy of the existing `aabb` box? I guess padding is added so it's not always just a copy, is that what you were missing?

src/mesh/aabb_tree.f90
```diff
@@ -439,8 +439,9 @@ function sort(array) result(indices)
        minidx = -1
        do imin = 1, size(array)
           if (.not. visited(imin) .and. minidx .eq. -1) minidx = imin
-
-          if (visited(imin) .and. array(imin) .lt. array(minidx)) minidx = imin
+          if (minidx .gt. -1) then
+             if (visited(imin) .and. array(imin) .lt. array(minidx)) minidx = imin
+          end if
```
timfelle, line 444: I guess an index of 0 would also cause issues since fortran is 1 indexed.
Also lets just have one if statement instead of nested ones.
```suggestion
          
          if (visited(imin) .and. minidx .gt. 0 &
               .and. array(imin) .lt. array(minidx)) then
             minidx = imin
          end if
```

timfelle (approved):
> So you just want a copy of the existing `aabb` box? I guess padding is added so it's not always just a copy, is that what you were missing? 

I see. I read through your interpolation PR, it makes a lot more sense now.

timfelle (commented):
How about adding the modifications for point_t and element_t to this PR as well. That would be helpful for the immersed boundary thing Niclas was working on I think.

src/mesh/aabb_tree.f90
timfelle: Oh then forget my suggestion

MartinKarp:
Did I change anything for element?

MartinKarp:
I guess in pricinple I would jus like to work with arrays for 2x3 points and as little extra things as possible, and for that I wanted to use the aabb directly.

timfelle:
> Did I change anything for element?

In your other PR you have updated the `get_aabb` function to also handle `point_t` and `element_t` I just figured it would be nice to have all those changes in the same PR.

timfelle (approved):

njansson (approved):

PR #1544, opened 2024-10-14
Move mass matrix multiply to source handler

timofeymukha:
Moves the multiplication by the mass matrix from the step loops to the source term handler.


MartinKarp:
This is essentially permitting source terms which directly act on the weak form of the equations. This would of course enable us to do most everything on the res in the source term framework, but I am not sure if it makes sense from a physical standpoint? These source terms would not be possible to write like NS(u) = f, which is kind of how I think of source terms.

I guess I would somehow prefer to keep it as a stabilization or add some kind of 1/B factor at gradient_jump_perform.

timofeymukha:
PR updated to just move the B multiply to the source handler. We decided to keep source terms as something that is multiplied by B. Special ones should just account for that by having 1 / B in front. The idea is that we have only one such exception and it is unclear whether there will be more.

pschlatt1:
This is much better to do it this way. Source terms should be physical in my view, where you always have the mass matrix included.  Numerical "tricks" may be included using inv B. So I think this is better.

timfelle (approved):

MartinKarp (approved):

PR #1551, opened 2024-10-23
Resolve some outstanding issue with brinkman

timfelle:
Brinkman computation was not using the devices.

- Signed distance fields still compute on host, but is shared to device properly.
- Filters have been implemented for devices.
- A few math operations added in order to work with filters.
- Additionally the examples were cleaned slightly.


njansson (commented):
Looks good! One question, the massive changes in the device dependency, what is the cause of that, formatting?

timfelle:
> Looks good! One question, the massive changes in the device dependency, what is the cause of that, formatting?

My editor had identified the file as a YAML file and auto formatted the "` : `" to "`: `", removing the leading space, which is not the same format as the `.depend` file so i have now reverted it.

njansson (approved):

PR #1552, opened 2024-10-23
A mesh checker utility

timofeymukha:
Adds a mesh checker utility, currently in rudimentary and serial form, but already quite useful to detect what kind of zones one has in a mesh.

The idea is to add more diagnostics further on, and output associated fields as an option.


MartinKarp (approved):
Looks good! I guess things to add further on would be things like the extent of the mesh and such.

timofeymukha:
Yeah, I think we can all add useful functionality as we go along..

njansson (approved):

PR #1555, opened 2024-10-25
Spectral error indicator as simcomp

vbaconnet:
Moves the spectral error indicator into a simcomp. 

The behaviour is the same as lambda2: the simcomp uses a field_writer to append the 3 spectral error fields as scalar fields in the field file output. 
- The default output frequency is that of the field files. 
- The default compute frequency is such that the `compute_control` matches the field file `output_control` by default, instead of doing it at every timestep (which is the default for all simcomps). To do this I manually add the parameters to the json in the init_from_json to bypass the get_or_default in the simcomp init_base:

```f90
    call case%params%get("case.fluid.output_control", fluid_output_control, &
         found)
    call case%params%get("case.fluid.output_value", fluid_output_value, &
         found)

    call json%get("compute_control", str, found)
    if (.not. found) then
       call json%add("compute_control", fluid_output_control)
       ! Use "update" and not "add" just in case compute_value is present in the json file.
       call json%update("compute_value", fluid_output_value, found)
    end if
```

- all of the above now also applies to the lambda2, derivative and vorticity simcomps, such that we only compute when we need to output.
- small fix for the derivative simcomp that did not compute the proper derivatives.

Tested locally with CPUs and on LUMI with GPUs.


njansson (approved):

src/simulation_components/lambda2.f90
```diff
@@ -87,11 +87,34 @@ subroutine lambda2_init_from_json(this, json, case)
...
 
+    call case%params%get("case.fluid.output_control", fluid_output_control, &
+         found)
+    call case%params%get("case.fluid.output_value", fluid_output_value, &
+         found)
+
+    ! See if the user has set a compute control, otherwise
+    ! set it to the fluid output control.
+    ! NOTE: We do not use get_or_default here, as we need to manually add
+    ! entries to the json to bypass the get_or_default in the init_base later.
```
timofeymukha, line 107: Maybe I am wrong, but doesn't the fact that the _or_default routines have a call to `add` inside solve this? I actually don't think I've every checked that that works properly :-).

doc/pages/user-guide/simcomps.md
```diff
@@ -32,13 +32,23 @@ in Neko. The list will be updated as new simcomps are added.
...
 
 ## Controling execution and file output
-Each simulation component is, by default, executed once per time step to perform
-associated computations and output. However, this can be modified by using the
+Most simulation components are, by default, executed once per time step to perform
+associated computations and output. The exceptions are:
+- \ref simcomp_lambda2, 
+- \ref simcomp_speri,
+- \ref simcomp_derivative,
+- \ref simcomp_vorticity,
```
timofeymukha, line 43: For the derivative and vorticity, I am not sure the new default makes (more) sense. I would assume a relatively common scenario for vorticity would be to average it (e.g. in addition to a "basic" stats package), and the same for a derivative of some user-defined field. For that, the default to compute each time-step makes more sense, although it is still superfluous if the stats are not sampled every timestep. In TGV, we use the vorticity in user check to compute some measure of the flow, so also uncoupled to write time.

There are for sure cases where your default makes more sense. But I wonder if it is worth introducing an exception to a shared default.

Note that the *output* control can be easily matched with the fluid by setting it to "global" in the json for the simcomp. I think maybe we need a similar shortcut for matching the compute with the output of the simcomp, like ``compute_control : "on_output"` or something like that.

src/simulation_components/spectral_error.f90
```diff
@@ -30,8 +30,8 @@
 ! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 ! POSSIBILITY OF SUCH DAMAGE.
 !
-!> Implements type spectral_error_indicator_t.
-module spectral_error_indicator
+!> Implements type spectral_error_t.
+module spectral_error
```
timofeymukha, line 34: Very nice work! Two small comments:

1. I think it is generally better to split the simcomp type and the background implementation, like it is for lambda, the les simcomp etc. For stuff like vorticity, it makes little sense, but for more substantial functionality it may be good to have it as just a type, without the bells and whistles of a simcomp always attached. I am not forcing you to do this here, though :-). But for the future.
2. Can we think a bit of how the simcomp can be tested? I.e. can one easily construct a field for which the error would be known, or something like that? Again, not for this PR, but as a general theme of raising our error-proof levels.

src/simulation_components/lambda2.f90
vbaconnet: Oh I didn't check either, I thought that the get_or_default would just set the value of the variable but not modify the json file... Now I am curious

doc/pages/user-guide/simcomps.md
vbaconnet: > For the derivative and vorticity, I am not sure the new default makes (more) sense. I would assume a relatively common scenario for vorticity would be to average it (e.g. in addition to a "basic" stats package), and the same for a derivative of some user-defined field. For that, the default to compute each time-step makes more sense, although it is still superfluous if the stats are not sampled every timestep. In TGV, we use the vorticity in user check to compute some measure of the flow, so also uncoupled to write time.

The idea was to make it such that if the user wants to compute the vorticity/derivatives more frequently they should explicitly set it that way. But to be honest I don't have strong feeling about it so it's whatever makes most sense to people :)

> Note that the _output_ control can be easily matched with the fluid by setting it to "global" in the json for the simcomp. I think maybe we need a similar shortcut for matching the compute with the output of the simcomp, like ``compute_control : "on_output"` or something like that.

Yea, I thought that making that change would be a bit too controversial so I didn't dare to do it...

src/simulation_components/lambda2.f90
vbaconnet: Oh you are right! It does modify the json file object directly, good to know!

doc/pages/user-guide/simcomps.md
vbaconnet: I like the idea of a `compute_control: "on_output"`. I think that encapsulates the idea quite well. Ideally that would be the default behaviour for lambda2 and spectral_error. 

As I said it makes more sense to me that the computation should not be done more than necessary by default, unless you want it computed more often for other purposes. But maybe I am missing the bigger picture here..

timofeymukha: What I am now wondering is whether we can have a parameter that can be of different type, like both 1 and "on_output".  When we do the json read, we pass the variable of the associated type to be filled with the parameter's value. I woder what happens if we send the wrong type :P. I.e. ask for an integer, but the parameter is actually a string.

MartinKarp (approved):

src/simulation_components/vorticity.f90
```diff
@@ -101,6 +103,16 @@ subroutine vorticity_init_from_json(this, json, case)
...
+    call json_get(case%params, "case.fluid.output_control", &
+         fluid_output_control)
+    call json_get(case%params, "case.fluid.output_value", &
+         fluid_output_value)
+
+    ! See if the user has set a compute control, otherwise
+    ! set it to the fluid output control.
+    call json_get_or_default(json, "compute_control", str, fluid_output_control)
+    call json_get_or_default(json, "compute_value", val, fluid_output_value)
+
```
timfelle, line 115:
```suggestion
```
This is already handled by `init_base` so you don't have to.

vbaconnet: The reason I did this is because `compute_control` does not support the `global` keyword. This avoids the need for the user to have to manually match the compute control with the fluid output, so with this code I set the parameters that I want inside the json object so that the init_base sets things properly. Of course it's not so pretty but it works fine for just getting the spectral error in

timfelle: If you do not set a `compute_control` value for the simulation component, then it follows the fluid. But sure, we should probably add a `global` keyword to ensure a user can explicitly match the fluid instead of it just be done implicitly if you do not specify anything.

timfelle: The reason i flagged this is that since you used `json_get`, it is now mandatory to assign a value to `case.fluid.output_control` and `case.fluid.output_value` which cause the reframe test to fail, since we rely on default values for one of them.

vbaconnet: Isn't the default to compute once per timestep and output by default follows what is in compute? https://github.com/ExtremeFLOW/neko/blob/62405f3ee0d6d883266908ff06e80253cb1b911b/src/simulation_components/simulation_component.f90#L144-L161

timfelle: What did you want to achieve with the code you wrote then?

timfelle: I'll add `fluid_out` as a control type for the simcomps, in my open simcomp PR.
That should avoid you having to do something like this for a single simcomp, I guess many of them would kinda be nice to only compute on output.

vbaconnet: > Alright so what you want is to compute the simcomp, on the output of fluid?
Yes exactly

timofeymukha: I still like matching one's own output better, but since both you and Tim think this is better, it probably is :-).

We may have to rethink this stuff if fluid becomes a simcomp. But since this transition is not really an urgent priority (I guess), we should have some good settings in the code now as well.

timfelle: > Based on your "fluid_output" suggestion, how about this then?
> 
> The default behavior will be to compute on fluid output and if they want the simcomp computed more often they should specify it explicitly.

I am not really convinced it would be a good idea to have different default behavior for the same setting, in different simcomps. 🤔 That would probably just cause confusion.
Also as @timofeymukha mentioned, if we go with this suggestion, i think it would be good to match the default of the compute and output.

vbaconnet: Yea matching the compute and output makes more sense now to me as well... Then perhaps we should keep the scope of this PR to actually only add the spectral error as a simcomp.

doc/pages/user-guide/simcomps.md
timfelle: @vbaconnet and @timofeymukha, is this conversation resolved?

vbaconnet: I would say yes since in the end the PR doesn't change affect the time controllers anymore

PR #1557, opened 2024-10-28
Fix legendre_poly

MartinKarp:
Quick check is to compare with PNLEG which also computes the Legendre polynomials in a point

### DIFF
--- a/src/sem/speclib.f90
+++ b/src/sem/speclib.f90
@@ -985,7 +985,8 @@ contains
     L(1) = x
 
     do j=1, N-1
-       L(j+1) = ( (2.0_xp*j+1.0_xp) * x * L(j) - real(j,xp) * L(j-1) ) / (j+1.0_xp) 
+       L(j+1) = ( (2.0_xp * real(j, xp) + 1.0_xp) * x * L(j) &
+            - real(j, xp) * L(j-1) ) / (real(j, xp) + 1.0_xp) 
     end do
   end subroutine legendre_poly
 
--- a/src/sem/speclib.f90
+++ b/src/sem/speclib.f90
@@ -976,15 +976,16 @@ contains
   
   subroutine legendre_poly(L, x, N)
     ! Evaluate Legendre polynomials of degrees 0-N at point x
-    real(kind=rp), intent(inout):: L(1:N+1)
+    real(kind=rp), intent(inout):: L(0:N)
     real(kind=rp) :: x
     integer :: N, j
 
-    L(1) = 1.0_rp
-    L(2) = x
+    L(0) = 1.0_xp
+    if (N .eq. 0) return
+    L(1) = x
 
-    do j=3, N+1
-       L(j) = ( (2*j-1) * x * L(j-1) - (j-1) * L(j-2) ) / j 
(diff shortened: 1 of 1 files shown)


src/sem/speclib.f90
```diff
@@ -976,15 +976,16 @@ end function PNLEG
   
   subroutine legendre_poly(L, x, N)
     ! Evaluate Legendre polynomials of degrees 0-N at point x
-    real(kind=rp), intent(inout):: L(1:N+1)
+    real(kind=rp), intent(inout):: L(0:N)
```
timfelle, line 979: Didn't we earlier discuss that we wanted to move away from non standard array indexing?

MartinKarp: This is purely to comply with the usual definition of legendre polynomials. I could change, but then it is quite difificult to translate to wikipedia I found.

MartinKarp: Or whatever other resource poeple use

timfelle: Well it will be more difficult for anyone working with the code to remember that this particular array is different from all the other ones, and then your errors will be segmentation faults at random spots in your code. 
Translating from a reference to an implementation is done now that you wrote the code, anyone using this will just need to know, "this returns me an array with the legendre polynimials."

src/sem/speclib.f90
```diff
@@ -976,15 +976,16 @@ end function PNLEG
...
-    L(1) = 1.0_rp
-    L(2) = x
+    L(0) = 1.0_xp
+    if (N .eq. 0) return
+    L(1) = x
 
-    do j=3, N+1
-       L(j) = ( (2*j-1) * x * L(j-1) - (j-1) * L(j-2) ) / j 
+    do j=1, N-1
+       L(j+1) = ( (2.0_xp*j+1.0_xp) * x * L(j) - real(j,xp) * L(j-1) ) / (j+1.0_xp)
```
timfelle, line 988: Add missing type cast
```suggestion
     L(j+1) = ( (2.0_xp * real(j, xp) + 1.0_xp) * x * L(j) &
          - real(j, xp) * L(j-1) ) / (j + 1.0_xp) 
```

src/sem/speclib.f90
timofeymukha: I keep forgetting, if I pass an ordinary array to this, without specifying any special indexing, what happens? Will be returned 0-indexed, or will L(1) be equivalent to L(0) inside the routine? I think it is the latter.

timofeymukha: Yeah, that is what I thought. So it is essentially a local view into the array with changed indexing. It is perhaps OK then. The only real issue is when uncommon indexing is used in arrays stored as type components.

src/sem/speclib.f90
```diff
@@ -976,15 +976,17 @@ end function PNLEG
   
   subroutine legendre_poly(L, x, N)
```
timfelle, line 977: I guess we are missing a doxygen string describing what this does.

adperezm (approved):

njansson (approved):

PR #1558, opened 2024-10-28
Fix volflow type

timofeymukha:
The volflow type crashes when a coupled solver for velocity is used. This is because the decoupling of the velocity and pressure Ax in the fluid scheme did not propagate there, and the ksp solver was not called via the coupled call.

This fixes this. But I have to admit that I don't fully understand what the type does all together, and why one needs ksp solves for enforcing the volflow. It would be good if someone can add documentation. 

So please check if this is correct :-).

### DIFF
--- a/src/fluid/fluid_pnpn.f90
+++ b/src/fluid/fluid_pnpn.f90
@@ -787,9 +787,9 @@ contains
 
       if (this%forced_flow_rate) then
          call this%vol_flow%adjust( u, v, w, p, u_res, v_res, w_res, p_res, &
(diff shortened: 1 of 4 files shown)


timofeymukha:
@muenschm I think this is why your periodic hill LES crashes, and hemi doesn't for the same setup.

timofeymukha:
Would be good to have in 0.9!

njansson (approved):

njansson:
Good catch, perfect timing before I drafted a `release/0.9` branch 🍻

MartinKarp (approved):
Looks good to me.

PR #1559, opened 2024-10-28
Improvements to simulation components

timfelle:
- We can now add a fully initialized simcomp to the executor.
- We call finalize internally if needed instead of outside.
- Fix member accesibility.
- Add `fluid_output` as a compute control type.

### DIFF
--- a/src/simulation_components/simcomp_executor.f90
+++ b/src/simulation_components/simcomp_executor.f90
@@ -61,21 +61,21 @@ module simcomp_executor
      logical, private :: finalized = .false.
    contains
      !> Constructor.
-     procedure, public, pass(this) :: init => simcomp_executor_init
+     procedure, pass(this) :: init => simcomp_executor_init
      !> Destructor.
-     procedure, public, pass(this) :: free => simcomp_executor_free
+     procedure, pass(this) :: free => simcomp_executor_free
      !> Appending a new simcomp to the executor.
-     procedure, public, pass(this) :: add_user_simcomp => simcomp_executor_add
+     procedure, pass(this) :: add_user_simcomp => simcomp_executor_add
      !> Execute preprocess_ for all simcomps.
-     procedure, public, pass(this) :: preprocess => simcomp_executor_preprocess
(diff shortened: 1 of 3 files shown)


src/simulation_components/simcomp_executor.f90
```diff
@@ -50,25 +50,28 @@ module simcomp_executor
   !! The execution order is based on the order property of each simcomp.
   !! By default, the order is by the order of apparence in the case file.
   type, public :: simcomp_executor_t
+     private
      !> The simcomps.
```
njansson, line 54: What will be private here? Only simcomps or all the data?

timfelle: All components of the simcomp_executor_t that is not explicitly marked public.

Meaning it is not possible to tamper with the list of simcomps and so on from the outside.

MartinKarp: I havent looked carefully in to it, but does this make it so one can not for example look through the simcomps outside?Just thinking if one might need the information from other simcomps sometimes.

timfelle: You would need to use something like the field registry to communicate things out of the simcomps, or retain a pointer to the actual simcomp you want to work with.

This prevents you from accessing `neko_simcomps%simcomps` directly. What we could do if you somehow need it, is to provide a `get` function which could return a pointer to a specific simulation component.

timfelle: @MartinKarp and @njansson Is this clear and reasonable, or should we keep the public access to the members of the executor?

MartinKarp: I think it might be good to keep the access to the members for now while we figure out how we want communication between components to work.

timfelle: I have reverted the access to public for the default. I retain private on the
internal components that we do not want others to access, size of array and
such.
I have added a get_n (maybe it should be called get_size) to return the number
of components if needed.

MartinKarp (approved):

PR #1560, opened 2024-10-28
Add axis-aligned bcs for shear stress and wall modelling

timofeymukha:
Adds sh and wm boundary types, and associated configuration for them in terms of json objects in fluid. See the new documenttaion for details on the user side.

The shear stress bc uses a nested symmetry bc to take care of the non-penetration condition. Three nested Neumann bcs add the forcing to the RHS. The wall model bc inherits from the shear_stress, and just sets the flux based on the wall model output.

The init of these nested bcs, just like the symmetry one, is quite delicate. This should be address in the bc overhaul, which I have paused to get this stuff in. At least, that is my hope :-).

The PR also adds a json utility to extract an object by name.

Fixes looping in bc.f90 in the marking by zone name: it loops over garbage atm, because the bc_labels are reallocated by the json module to match the size of the array in the JSON file.

Depends on #1558 

I verified the shear_stress bc using a laminar channel flow. The wall modelling remains unverified.


njansson (approved):

timofeymukha:
There is some nasty issue with the wall model part. Segfaults with -O2 when running in parallel. But not with -O0.

njansson:
> There is some nasty issue with the wall model part. Segfaults with -O2 when running in parallel. But not with -O0.

badness :( but given the fix to that reshaped array, I think it's good to get this in as soon as possible. The blocker now seems to be the depends file

timofeymukha:
> > There is some nasty issue with the wall model part. Segfaults with -O2 when running in parallel. But not with -O0.
> 
> badness :( but given the fix to that reshaped array, I think it's good to get this in as soon as possible. The blocker now seems to be the depends file

The issue seems to be this code

```fortran
    hmin = glmin(this%h%x, n_nodes)
    hmax = glmax(this%h%x, n_nodes)

    if (pe_rank .eq. 0) then
       write(*, "(A, F10.4, F10.4)") "   h min / max:", hmin, hmax
    end if
```
Am I not using these functions right? this%h is a vector_t.

MartinKarp:
If you change the min(tmp,a(i)) to an if else statement, what happens then?

timofeymukha:
Using `if (tmp > a(i)) tmp = a(i)` still makes things die

timofeymukha:
Oh, ideed!

timofeymukha:
Ah, it always takes a(1) even if n is 0, I guess that is the issue.

timofeymukha:
What we could do is set tmp to `huge` in that case, so it does not affect the minumum?

MartinKarp:
I think that would be excellent, and likewise for glmax I guess

src/fluid/fluid_scheme.f90
```diff
@@ -430,6 +440,35 @@ subroutine fluid_scheme_init_common(this, msh, lx, params, scheme, user, &
...
+    if (this%bc_sh%msk(0) .gt. 0) then
+       call params%get('case.fluid.shear_stress.value', real_vec, logical_val)
+       if (.not. logical_val .and. this%bc_sh%msk(0) .gt. 0) then
+          call neko_warning("No stress values provided for sh boundaries, &
+               & defaulting to 0. Use fluid.shear_stress.value to set.")
+          allocate(real_vec(3))
+          real_vec = 0.0_rp
+       else if (size(real_vec) .ne. 3) then
+          call neko_error ("The shear stress vector provided in &
+               fluid.shear_stress.value should have 3 components.")
```
timfelle, line 464:
```suggestion
               &fluid.shear_stress.value should have 3 components.")
```

timfelle (approved):

timofeymukha:
I'll do a small style bump and merge.

PR #1562, opened 2024-10-29
PDE filter for Brinkman source term

gloopydoop:
This is a PDE filter used to smooth the boundaries of a Brinkman source term, to avoid oscillations associated with sharp interfaces + spectral methods.

The theory is based on B. S. Lazarov, O. Sigmund (2010) "Filters in topology optimization based on Helmholtz-type differential equations".


src/source_terms/brinkman/PDE_filter.f90
```diff
@@ -0,0 +1,280 @@
...
+  use logger, only: neko_log, LOG_SIZE
+  use neko_config, only: NEKO_BCKND_DEVICE
+  use dofmap, only: dofmap_t
+  use jacobi, only: jacobi_t
+  use device_jacobi, only: device_jacobi_t
+  use sx_jacobi, only: sx_jacobi_t
+  use hsmg, only: hsmg_t
+  use utils, only: neko_error
+  use device_math, only: device_cfill, device_col3
+    implicit none
```
timfelle, line 68:
```suggestion
  implicit none
```

src/source_terms/brinkman/PDE_filter.f90
```diff
@@ -0,0 +1,280 @@
...
+  subroutine PDE_filter_init_from_json(this, json, coef)
+    class(PDE_filter_t), intent(inout) :: this
+    type(json_file), intent(inout) :: json
+    type(coef_t), intent(inout) :: coef
+
+    ! user parameters
+    call json_get_or_default(json, "filter.radius", this%r, 1.0_rp)
+
+    call json_get_or_default(json, "filter.tolerance", this%abstol_filt, &
+         0.0000000001_rp)
```
timfelle, line 129: Maybe use exponential notation instead?
```suggestion
         1.0e-10_rp)
```

src/source_terms/brinkman_source_term.f90
```diff
@@ -32,7 +32,7 @@
 !
 !> Implements the `brinkman_source_term_t` type.
 module brinkman_source_term
-  use num_types, only: rp, dp
+  use num_types, only: rp, dp, sp
```
timfelle, line 35:
```suggestion
  use num_types, only: rp, dp
```

You didn't add any variables of single precision?

timfelle (dismissed):

timofeymukha:
As a general comment, could we come up with more specific names? filter_t is definitely too generic, since we already have e.g. a filter for LES. PDE_filter is better, but perhaps could be concretized as well. I would also use `pde`, we generally avoid capitals unless really needed.

src/source_terms/brinkman_source_term.f90
gloopydoop: I did, but this is still a good comment to bring up.
I hard coded `fld_file_output_t` to output in single precision instead of leaving it up to the user...

timfelle:
> As a general comment, could we come up with more specific names? filter_t is definitely too generic, since we already have e.g. a filter for LES. PDE_filter is better, but perhaps could be concretized as well. I would also use `pde`, we generally avoid capitals unless really needed.

A counter proposal would be, shouldn't the LES filter derive from the generic filter type then, we will probably add more filters later, meaning the abstract base class should have some name which is easy to understand. 

on the PDE_filter name, sure, it might be better to call it a poisson_filter, since that is the PDE implemented for this particular filter.

src/source_terms/brinkman_source_term.f90
timfelle: Ah i missed that.

gloopydoop:
Both good suggestions.
I just had a little read of the `elementwise_filter_t` and I suppose with the exception of fields vs arrays, the `elementwise_field_filter_3d` fits `filter_apply`. @timofeymukha, I guess you've done a lot LES stuff, what are your thoughts?

Regarding the names, yeh I sort of expected this to be problematic. We also now have `filter.f90` and `filters.f90` :sweat_smile:

I'm happy with `helmholz_filter`, that's a good suggestion.

timfelle:
> Both good suggestions. I just had a little read of the `elementwise_filter_t` and I suppose with the exception of fields vs arrays, the `elementwise_field_filter_3d` fits `filter_apply`. @timofeymukha, I guess you've done a lot LES stuff, what are your thoughts?
> 
> Regarding the names, yeh I sort of expected this to be problematic. We also now have `filter.f90` and `filters.f90` 😅
> 
> I'm happy with `helmholz_filter`, that's a good suggestion.

Yes that is also a thing 😅 I guess we need to move the smooth_step functions somewhere else as well.

timofeymukha:
@Shiyu-Sandy-Du implemented the LES filtering. I think if you guys can figure out a good abstract interface that would fit a reasonably large class of filters, that would be great!

src/les/bcknd/cpu/dynamic_smagorinsky_cpu.f90
```diff
@@ -153,57 +153,58 @@ end subroutine dynamic_smagorinsky_compute_cpu
...
-       lij(4)%x(i,1,1,1) = fu(i) * fv(i)
-       lij(5)%x(i,1,1,1) = fu(i) * fw(i)
-       lij(6)%x(i,1,1,1) = fv(i) * fw(i)
-    end do
+    call col3(lij(1)%x, fu%x, fu%x, n)
+    call col3(lij(2)%x, fv%x, fv%x, n)
+    call col3(lij(3)%x, fw%x, fw%x, n)
+    call col3(lij(4)%x, fu%x, fv%x, n)
+    call col3(lij(5)%x, fu%x, fw%x, n)
+    call col3(lij(6)%x, fv%x, fw%x, n)
```
njansson, line 180: To allow vectorisation, and not put too much faith in compiler inlining + optimisation 

```suggestion
    do concurrent (i = 1:n)
       lij(1)%x(i,1,1,1) = fu%x(I,1,1,1) * fu%x(i,1,1,1)
       lij(2)%x(i,1,1,1) = fv%x(i,1,1,1) * fv%x(i,1,1,1)
       lij(3)%x(i,1,1,1) = fw%x(i,1,1,1) * fw%x(i,1,1,1)
       lij(4)%x(i,1,1,1) = fu%x(i,1,1,1) * fv%x(i,1,1,1)
       lij(5)%x(i,1,1,1) = fu%x(i,1,1,1) * fw%x(i,1,1,1)
       lij(6)%x(i,1,1,1) = fv%x(i,1,1,1) * fw%x(i,1,1,1)
```

src/source_terms/brinkman/PDE_filter.f90
```diff
@@ -0,0 +1,329 @@
...
+contains
+
+  !> Constructor from json.
+  subroutine PDE_filter_init_from_json(this, json, coef)
+    class(PDE_filter_t), intent(inout) :: this
+    type(json_file), intent(inout) :: json
+    type(coef_t), intent(in) :: coef
+
+    ! user parameters
+    call json_get_or_default(json, "filter.radius", this%r, 1.0_rp)
```
timfelle, line 123: Maybe don't have this as a default since it is not relative and a value of
1.0_rp will not always be useful.

```suggestion
    call json_get(json, "filter.radius")
```

njansson (approved):

timfelle (approved):

timofeymukha:
@gloopydoop Don't forget to add docs, we have a section on source terms as well as the Brinkman ;-)

PR #1563, opened 2024-10-29
allocation of xyz

gloopydoop:
The error I was getting was

`Error termination. Backtrace:                                                   
At line 474 of file common/global_interpolation.F90                             
Fortran runtime error: Attempt to DEALLOCATE unallocated 'xyz'`

I'm not sure if this is the correct solution, can someone who knows more about interpolation take a look at this?

### DIFF
--- a/src/common/global_interpolation.F90
+++ b/src/common/global_interpolation.F90
@@ -471,7 +471,9 @@ contains
     end do
 
     n_points = this%n_points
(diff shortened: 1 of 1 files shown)


njansson (approved):

MartinKarp (approved):
This is a good solution. I will revise this later to make the allocation and such better.

PR #1565, opened 2024-10-29
Fix invalid access to unallocated components of a case

njansson:
Fix potential access to unallocated memory.

Update of material properties expected cp and lambda from the scalar solver. However, C%scalar might not be allocated, causing an invalid access

### DIFF
--- a/src/simulation.f90
+++ b/src/simulation.f90
@@ -64,6 +64,7 @@ contains
     ! for variable_tsteping
     real(kind=rp) :: cfl_avrg = 0.0_rp
     type(time_step_controller_t) :: dt_controller
+    real(kind=rp) :: rho, mu, cp, lambda
 
     t = 0d0
     tstep = 0
@@ -149,6 +150,10 @@ contains
           write(log_buf, '(A,E15.7)') &
             'Total elapsed time (s):', end_time-start_time_org
           call neko_log%end_section(log_buf)
+
+          !> @todo Temporary fix until we have reworked the material properties
+          cp = C%scalar%cp
+          lambda = C%scalar%lambda
        end if
 
        call neko_log%section('Postprocessing')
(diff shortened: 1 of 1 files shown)


MartinKarp:
How do we want to do with material properties here? I would like to avoid having the if present statement in a user specified subroutine if possible.

njansson:
> How do we want to do with material properties here? I would like to avoid having the if present statement in a user specified subroutine if possible.

Then I guess we need to have a material properties type, that both fluid and scalar can refer to (or store it self), otherwise we will have potential issues with people access optional arguments.

Question is if one can find a good solution quickly, or accept this while working on something. Either way the current code is not safe, and we are just lucky that we haven't seen anything worse than one segfault

njansson:
Ps. one option would be to have a generic interface for the user file. Then a user has to make an active choice which material properties routine to use.
 
something along the line of selecting one that only has fluid params, and one with both.

MartinKarp:
Perhaps its ok.  I think TIm talked about moving the material properties from case as well. Overall I think the user file might need some love soon.

MartinKarp:
Actually, another idea

src/simulation.f90
```diff
@@ -158,11 +158,17 @@ subroutine neko_solve(C)
...
+          call C%usr%material_properties(t, tstep, C%fluid%rho,&
+                                         C%fluid%mu, &
+                                         C%scalar%cp, &
+                                         C%scalar%lambda, &
+                                         C%params)
+       else
+          call C%usr%material_properties(t, tstep, rho = C%fluid%rho,&
+                                         mu = C%fluid%mu, &
+                                         params = C%params)
+       end if
```
MartinKarp, line 171:
```suggestion
lambda = 0.0_rp
cp = 0.0_rp
 if (allocated(C%scalar)) then
          lambda = C%scalar%lambda
          cp = C%scalar%cp
  end if
  call C%usr%material_properties(t, tstep, C%fluid%rho,&
                                         C%fluid%mu, &
                                         cp, &
                                         lambda, &
                                         C%params)
       if (allocated(C%scalar)) then
          C%scalar%lambda = lambda
          C%scalar%cp = cp
       end if
```

MartinKarp:
I dunno if that came across well in the suggestion, but I guess it would avoid changing things for anyone using the userfile currently.

timofeymukha:
Yeah, I realize that killing the material properties type made things worse temporarily. The long-term issue is that if you have more than one scalar, this interface kind of stops working.

What I also realize now is that we actually have support for changing the material properties in time, right? I need to account for that in the wall modelling code somehow... But that is a different story :-)/

njansson:
Since we don't want to tag a release with obvious memory issues, I would suggest that we fix this in the `release/0.9` branch, following @MartinKarp suggestions to copy data back and forth between scalars, while we work on a more long term solution that can handle arbitrary number of scalars (as pointed out by @timofeymukha )

MartinKarp:
We could also do this fix in develop I think, to avoid the memory issue i dev also

njansson:
> We could also do this fix in develop I think, to avoid the memory issue i dev also

Sounds good, I'll rework this or accordingly

njansson:
> > We could also do this fix in develop I think, to avoid the memory issue i dev also
> 
> Sounds good, I'll rework this or accordingly

@MartinKarp @timofeymukha Done! 

I think this should work as a temporary solution for both cpu and device backends, also supporting variable properties, and without any need to modify user files

MartinKarp (approved):

njansson:
Ping @timofeymukha (^-^)

timofeymukha (approved):

PR #1577, opened 2024-11-01
Order of user-init

gloopydoop:
Hoping to get people's opinion on switching the order of the user-init to occur before the simcomps.

For me personally, this is motivated by the probes simcomp. 
If I want to probe some quantity I plan to calculate in the future, eg dudx etc, I need a way to put them in the registry prior to the probes simcomps initialization.

### DIFF
--- a/src/neko.f90
+++ b/src/neko.f90
@@ -277,6 +277,12 @@ contains
        call case_init(C, case_file)
 
        !
+       ! User-init
+       !
+       call C%usr%user_init_modules(t, C%fluid%u, C%fluid%v, C%fluid%w,&
+                                    C%fluid%p, C%fluid%c_Xh, C%params)
+
+       !
        ! Setup runtime statistics
        !
        call neko_rt_stats%init(C%params)
--- a/src/simulation.f90
+++ b/src/simulation.f90
@@ -88,12 +88,10 @@ contains
        call neko_simcomps%restart(t)
     end if
 
-    !> Execute outputs and user-init before time loop
+    !> Execute outputs before time loop
     call neko_log%section('Postprocessing')
     call C%output_controller%execute(t, tstep)
 
-    call C%usr%user_init_modules(t, C%fluid%u, C%fluid%v, C%fluid%w,&
-                                 C%fluid%p, C%fluid%c_Xh, C%params)
     call neko_log%end_section()
     call neko_log%newline()
 


njansson (commented):
My only concern is that if the user init has a routine that depends on u,v,w,p, I don't think this would work if a restart is performed?

gloopydoop:
> My only concern is that if the user init has a routine that depends on u,v,w,p, I don't think this would work if a restart is performed?

I've noticed exactly that watching all the compilations fail :sweat_smile:
Perhaps this should have been submitted as an issue rather than a PR.

Nevertheless, my problem with the probes still remains, anysuggestions?

timofeymukha:
Wasnt there another user routine, that runs really early? Precisely for initializing fields and stuff.

njansson:
> Wasnt there another user routine, that runs really early? Precisely for initializing fields and stuff.

maybe, but the checkpoint payload is loaded in simulation.f90

timfelle:
It sound like you are worried people have used the initializer as a preprocessor. I guess we should provide a preprocessor for the people who wish to do something more elaborate than initialize their variables.

timofeymukha:
OK so neko_init_modules is what I had in mind -- and the one Harry wants to move. Sorry I didn't look a the code before.

@njansson I guess the fields will be there after the case init, but not filled with values from the restart? So the probes will work, but just sample zeros or whatever. 

@timfelle A good solution would also be to use a user simcomp to compute the fields, if we could control the execution order of user simcomps. But currently they always come last, or?

njansson:
> OK so neko_init_modules is what I had in mind -- and the one Harry wants to move. Sorry I didn't look a the code before.
> 
> 
> 
> @njansson I guess the fields will be there after the case init, but not filled with values from the restart? So the probes will work, but just sample zeros or whatever. 
> 
> 


Yes probes will work, I'm more concerned if a user init routine needs the actual field during init, maybe I'm overthinking 
> 
> @timfelle A good solution would also be to use a user simcomp to compute the fields, if we could control the execution order of user simcomps. But currently they always come last, or?

timfelle:
> > OK so neko_init_modules is what I had in mind -- and the one Harry wants to move. Sorry I didn't look a the code before.
> > @njansson I guess the fields will be there after the case init, but not filled with values from the restart? So the probes will work, but just sample zeros or whatever.
> 
> Yes probes will work, I'm more concerned if a user init routine needs the actual field during init, maybe I'm overthinking
> 
> > @timfelle A good solution would also be to use a user simcomp to compute the fields, if we could control the execution order of user simcomps. But currently they always come last, or?

Maybe we should add a new `user_allocator` instead then? That might more clearly indicate that we are only supposed to allocate stuff inside it. Then the initializer can remain unchanged.

timfelle:
> @timfelle A good solution would also be to use a user simcomp to compute the fields, if we could control the execution order of user simcomps. But currently they always come last, or?

All the user simcomps are initialized at the end. What we might be able to do is allow each simcomp to store their own json settings. That way we can call all the initializers in the `finalize` step instead of in the factory. That way we can enforce strict initialization order, also with simcomps added other ways than through the initializer.

timofeymukha:
> > @timfelle A good solution would also be to use a user simcomp to compute the fields, if we could control the execution order of user simcomps. But currently they always come last, or?
> 
> All the user simcomps are initialized at the end. What we might be able to do is allow each simcomp to store their own json settings. That way we can call all the initializers in the `finalize` step instead of in the factory. That way we can enforce strict initialization order, also with simcomps added other ways than through the initializer.

I would kind of like to avoid that because I think we have JSON a bit too deep into our code, and it is mean to just be an IO thing. I would really like to keep it only in constructors / factories, and ideally have non-JSON alternatives for those as well. I think we moved in that good direction a bit now in the last months and we should continue.

timfelle:
> > > @timfelle A good solution would also be to use a user simcomp to compute the fields, if we could control the execution order of user simcomps. But currently they always come last, or?
> > 
> > 
> > All the user simcomps are initialized at the end. What we might be able to do is allow each simcomp to store their own json settings. That way we can call all the initializers in the `finalize` step instead of in the factory. That way we can enforce strict initialization order, also with simcomps added other ways than through the initializer.
> 
> I would kind of like to avoid that because I think we have JSON a bit too deep into our code, and it is mean to just be an IO thing. I would really like to keep it only in constructors / factories, and ideally have non-JSON alternatives for those as well. I think we moved in that good direction a bit now in the last months and we should continue.

Very reasonable, I think the initialization order will be tricky to solve, without resorting to locking the user to a single simcomp.

timofeymukha:
> > > OK so neko_init_modules is what I had in mind -- and the one Harry wants to move. Sorry I didn't look a the code before.
> > > @njansson I guess the fields will be there after the case init, but not filled with values from the restart? So the probes will work, but just sample zeros or whatever.
> > 
> > 
> > Yes probes will work, I'm more concerned if a user init routine needs the actual field during init, maybe I'm overthinking
> > > @timfelle A good solution would also be to use a user simcomp to compute the fields, if we could control the execution order of user simcomps. But currently they always come last, or?
> 
> Maybe we should add a new `user_allocator` instead then? That might more clearly indicate that we are only supposed to allocate stuff inside it. Then the initializer can remain unchanged.

I am pretty sure that the init_modules is meant to be an allocator, and then you populate the values in the user_ic (initial conditions). So I think the suggestion of the PR is kind of spot on. I guess the only issue is that you cannot init your custom fields based on stuff loaded from a checkpoint? I haven't look at the order there in detail.

I think we better create some flow chart of things for the documentation, and also ourselves :-).

MartinKarp:
I think the user interface needs a bit of love soon in general. I think when I added init_modules it was more or less as late as possible to be able to use as many fields and such when initializing. Overall, all functions which take u,v,w as arguments are not really scalable as one often needs way more things as a user. 

I would not change the current interface and order that much, but perhaps make everything in a larger overhaul where we go through all subroutines and interfaces.

For Harrys needs I guess initializing a derivative simcomp will be sufficient to allocate the field you want in your probes.

PR #1578, opened 2024-11-01
Add support for building shared/dynamic libraries

njansson:
Refactor build system to enable shared/dynamic libraries. This will ease with interfacing to python/Julia and to some extent linking with various external libraries

This is a **breaking change**, introducing the following
- libtool is requirement to build Neko
- Dependencies change from `.o` -> `.lo` 
- shared libraries (if requested) are named `libneko-${PACKAGE_VERSION}.so`
- old build trees must regenerate their toolchains (`./regen.sh`) 

**Note:** the default behaviour is to only build a static `libneko.a` (as before), shared libraries must be requested via `--enable-shared`


timofeymukha:
Appears to work! I updated gitignore for the .lo and some other new stuff.

timofeymukha (approved):

doc/pages/user-guide/installation.md
```diff
@@ -108,6 +108,8 @@ Features are enabled and disabled by passing either `--enable-FEATURE[=arg]` or
 | `--enable-contrib`    | Compile various tools                                                                                                                 |
 | `--enable-device-mpi` | Enable device aware MPI                                                                                                               |
 | `--enable-openmp`     | Enable OpenMP                                                                                                                         |
+| `--enable-shared`     | Build shared libraries (default: no)                                                                                                  |
+| `--enable-static`     | Build static libraries (default: yes)                                                                                                 |
```
timfelle, line 112: What happens if you assign `yes` to both?

njansson: Then both a static and a shared version of the library will be built (i.e standard libtool behaviour)

njansson: In the future we need to add logic and disable things if some of the options can't handle either of the options

timfelle: Yea, or atleast just stop with an error telling the user that certain features are now compatible.

timfelle (approved):

timfelle (approved):

PR #1580, opened 2024-11-04
Performing LES on devices

Shiyu-Sandy-Du:
In this PR,

- ax_helm_full_device is implemented for hip and cuda
- sigma and vreman SGS models are implemented for hip and cuda as well
- bug is fixed in the pnpn pressure res computation for full stress formulation on devices

Some other things to be discussed:

- [x] In pnpn_res_stress_device.F90, device math is called for multiple times. If we merge those kernels together as much as possible, how much can we benefit from it considering the solving time for the pressure equation is not short?
- [x] Implementation for 1D in ax_helm_full_device has not been implemented and thus autotune is not yet implemented as well. But do we really need it? I cannot see it in the vector version of ax_helm_device implementation. Is the 1D or kstep discussion for vector version of ax_helm happened before?


timofeymukha:
I'll leave it to the GPU experts to review, just gonna say -- very impressive work, Shiyu!

njansson:
This must be synced with develop before we can consider a merge ( due to all the dependency changes)

Shiyu-Sandy-Du:
> This must be synced with develop before we can consider a merge ( due to all the dependency changes)

Exactly, I am working on the conflicts. And wait PR #1581 to be merged for a proper git push

njansson:
Nice!

src/les/bcknd/device/hip/hip_sigma_nut.f90
njansson, line 1: Should we declare these interface in `device_sigma_nut.F90`, as with our other device implementations? One motivation would be to avoid the multiple inclusion warnings with CCE

src/les/bcknd/device/hip/hip_vreman_nut.f90
njansson, line 1: Same comment as for `hip_sigma_nut`

njansson:
To answer some of the questions,

> * [ ]  In pnpn_res_stress_device.F90, device math is called for multiple times. If we merge those kernels together as much as possible, how much can we benefit from it considering the solving time for the pressure equation is not short?

There are gains to be made here, by fusing the e.g. the cluster of col3 and sub2. Remember that each call to a device math functions comes with a launch latency, which is not negligible.

> * [ ]  Implementation for 1D in ax_helm_full_device has not been implemented and thus autotune is not yet implemented as well. But do we really need it? I cannot see it in the vector version of ax_helm_device implementation. Is the 1D or kstep discussion for vector version of ax_helm happened before?

Exactly, there's no 1d version of these, since 1d will run out of shared memory for most polynomial orders.

src/les/bcknd/device/hip/hip_sigma_nut.f90
Shiyu-Sandy-Du: Yes it should be done.

njansson (approved):

src/les/bcknd/device/vreman_device.f90
```diff
@@ -0,0 +1,136 @@
...
+
+    call dudxyz (a21%x, v%x, coef%drdx, coef%dsdx, coef%dtdx, coef)
+    call dudxyz (a22%x, v%x, coef%drdy, coef%dsdy, coef%dtdy, coef)
+    call dudxyz (a23%x, v%x, coef%drdz, coef%dsdz, coef%dtdz, coef)
+
+    call dudxyz (a31%x, w%x, coef%drdx, coef%dsdx, coef%dtdx, coef)
+    call dudxyz (a32%x, w%x, coef%drdy, coef%dsdy, coef%dtdy, coef)
+    call dudxyz (a33%x, w%x, coef%drdz, coef%dsdz, coef%dtdz, coef)
+
+    call coef%gs_h%op(a11%x, a11%dof%size(), GS_OP_ADD)
```
MartinKarp, line 116: I think you can just replace this with `gs_h%op(a11, GS_OP_ADD)` for example

src/les/bcknd/device/cuda/sigma_nut_kernel.h
```diff
@@ -0,0 +1,183 @@
...
+  for (int i = idx; i < n; i += str) {
+    T sigG11, sigG12, sigG13, sigG22, sigG23, sigG33;
+    T sigma1, sigma2, sigma3;
+    T Invariant1, Invariant2, Invariant3;
+    T alpha1, alpha2, alpha3;
+    T Dsigma;
+    T pi_3;
+    T tmp1;
+    
+    pi_3 = 4.0/3.0*atan(1.0);
```
MartinKarp, line 71: dont we have pi somewhere here?

MartinKarp (approved):
Great stuff Shiyu! I just added two comments, feel free to take them to heart or not. 

One thing I noted is the use of quite a lot of math functions such as pow, divisions and so on. I think we should just be aware this might be an issue if people run in single precision and one can perhaps use xreal in some places then. This is just a note for the future though.

src/les/bcknd/device/cuda/sigma_nut_kernel.h
Shiyu-Sandy-Du: Do we have it defined somewhere else?

Shiyu-Sandy-Du:
> Great stuff Shiyu! I just added two comments, feel free to take them to heart or not.
> 
> One thing I noted is the use of quite a lot of math functions such as pow, divisions and so on. I think we should just be aware this might be an issue if people run in single precision and one can perhaps use xreal in some places then. This is just a note for the future though.

I changed the unnecessary pow into multiplications here. And I think we should note the sin, cos and sqrt here for future maybe.

src/les/bcknd/device/cuda/sigma_nut_kernel.h
njansson: it can be fetched from the header files

njansson: Given that this is a templated variable, I would say that we leave it for later to fix.

PR #1586, opened 2024-11-06
Fix reported "DoF"s

MartinKarp:
DoFs reported are not the degrees of freedom and if for one rank. This changes to the clobal number of grid points instead.

### DIFF
--- a/src/fluid/fluid_scheme.f90
+++ b/src/fluid/fluid_scheme.f90
@@ -370,7 +370,7 @@ contains
        write(log_buf, '(A, I3)') 'Poly order : ', lx-1
     end if
     call neko_log%message(log_buf)
-    write(log_buf, '(A, I0)')    'GLLs pts.  : ', int(this%msh%glb_nelv,i8)*int(this%Xh%lxyz,i8)
+    write(log_buf, '(A, I0)')    'GLL points : ', int(this%msh%glb_nelv,i8)*int(this%Xh%lxyz,i8)
     call neko_log%message(log_buf)
     write(log_buf, '(A, I0)')    'Unique pts.: ', int(glsum(this%c_Xh%mult,this%dm_Xh%size()),i8)
     call neko_log%message(log_buf)
--- a/src/fluid/fluid_scheme.f90
+++ b/src/fluid/fluid_scheme.f90
@@ -370,9 +370,9 @@ contains
        write(log_buf, '(A, I3)') 'Poly order : ', lx-1
     end if
     call neko_log%message(log_buf)
-    write(log_buf, '(A, I0)')    'GLL points : ', int(this%msh%glb_nelv,i8)*int(this%Xh%lxyz,i8)
+    write(log_buf, '(A, I0)')    'GLL points : ', int(this%msh%glb_nelv, i8)*int(this%Xh%lxyz, i8)
     call neko_log%message(log_buf)
-    write(log_buf, '(A, I0)')    'Unique pts.: ', int(glsum(this%c_Xh%mult,this%dm_Xh%size()),i8)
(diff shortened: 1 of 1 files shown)


njansson:
Adding info about uniques are good. But not so much in favour of using terms as grid points, dofs are more aligned to what we do

MartinKarp:
I just think it is quite hard to report the number of DoFs, this would then be to multiply the unique points by the number of fields and then one would also need to take BCs into account. 

I think people in general are interested in the number of points. I think DoFs are quite difficult to report correctly.

src/fluid/fluid_scheme.f90
```diff
@@ -370,7 +370,9 @@ subroutine fluid_scheme_init_common(this, msh, lx, params, scheme, user, &
        write(log_buf, '(A, I3)') 'Poly order : ', lx-1
     end if
     call neko_log%message(log_buf)
-    write(log_buf, '(A, I0)') 'DoFs       : ', this%dm_Xh%size()
+    write(log_buf, '(A, I0)')    'GLL points : ', int(this%msh%glb_nelv,i8)*int(this%Xh%lxyz,i8)
```
timofeymukha, line 373:
```suggestion
    write(log_buf, '(A, I0)')    'GLL points : ', int(this%msh%glb_nelv, i8)*int(this%Xh%lxyz, i8)
```

src/fluid/fluid_scheme.f90
```diff
@@ -370,7 +370,9 @@ subroutine fluid_scheme_init_common(this, msh, lx, params, scheme, user, &
...
+    call neko_log%message(log_buf)
+    write(log_buf, '(A, I0)')    'Unique pts.: ', int(glsum(this%c_Xh%mult,this%dm_Xh%size()),i8)
```
timofeymukha, line 375:
```suggestion
    write(log_buf, '(A, I0)')    'Unique pts.: ', int(glsum(this%c_Xh%mult, this%dm_Xh%size()), i8)
```

timofeymukha (commented):
My initial impression is that it is better to report this under the mesh section of the log rather than fluid. But we need c_Xh to get the unique points?

timofeymukha:
I would agree that grid points (unique and not) is less ambiguous then dofs and is typically what one wants to look at, i.e. "what is my grid size?"

njansson:
> I would agree that grid points (unique and not) is less ambiguous then dofs and is typically what one wants to look at, i.e. "what is my grid size?"

Maybe, if one comes from the FV/FD world. But with FEM glasses on, grid size is very ambiguous

MartinKarp:
This reports the number of collocation points in the local representaiton ("GLL" points) and in the global representation "unique points". DoFs I think is missleading as people would think this includes the number of unique points times the number of fields we solve for and subtracting the prescribed boundary conditions.

timofeymukha:
I guess one can report all 3, as long as the definitions are clear. I am actually not sure what "dofs" means, I always thought it is essentially the grid size :-). I recall that dofmat%size returns the private component called ndofs, which is pretty much the grid size, right?  But if there is a clear definition of what that is for FEM, one can output that.

MartinKarp:
ndofs is also a misnomer as the number of gll points != ndofs. It is the number of points in the local representation. 

The glsum over mult would be more like the DoFs (excluding bcs). Degree of freedom is well defined and we should not report it if it is not the actual degrees of freedom in my opinion. 

I guess one could report dofs by calculating the number of unique points per field and then calculate how many dirichlet conditions are applied to each field and then adding this number together.

I just think we should make sure the info is true, and not make someone think it is something that it is not.

MartinKarp:
Can we merge this?

timofeymukha:
After the discussion at the meeting 2 weeks back, I am very much in favour.

timofeymukha (approved):

PR #1592, opened 2024-11-06
Fixes segfault due to trying to apply wm%symmetry%bcX 

MartinKarp:
I dunno, ran into segfaults locally and it smee ed to be relate dot the new bcs. I didnt have time to check it our properly, but perhaps Tim knows why this seems to fix the problem.

### DIFF
--- a/src/fluid/fluid_scheme.f90
+++ b/src/fluid/fluid_scheme.f90
@@ -525,6 +525,8 @@ contains
     if (this%bc_wallmodel%msk(0) .gt. 0) then
        call json_extract_object(params, 'case.fluid.wall_modelling', wm_json)
        call this%bc_wallmodel%init_wall_model_bc(wm_json, this%mu / this%rho)
+    else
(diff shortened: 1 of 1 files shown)


njansson (approved):

timofeymukha:
Yeah, the symmetry constructor is ran in the shear_stress constructor, and that is ran in turn in 
`call this%bc_wallmodel%init_wall_model_bc(wm_json, this%mu / this%rho)`

But I guess we always need to run it since otherwise this line
call bc_list_add(this%bclst_vel, this%bc_wallmodel%symmetry)
may cause an issue.

I guess an alternative, perhaps cleaner, solution is to put the addition to the bc_lists under the `if` checking the size of the bc.

The reason it is not done is because bc_list_add actually checks for the size being > 0, and doesn't add otherwise. But probably things go wrong since the constructor was not ran.

timofeymukha (approved):

PR #1598, opened 2024-11-08
Add the Guermond and Shen MMS for the Stokes equations

timofeymukha:
Relies on PR #1597 for solving the Stokes equations.

See the Readme for the descriptions of the case. I think this can be a useful demonstrator, but we can also later branch off test cases from this. For example, you can exactly compute all statistics for the case and check that `fluid_stats` work properly @MartinKarp.

Unfortunately, I must report an issue associated with the experiments with this case. Essentially, the best error is achieved with time order 1, whereas with 2 and 3 things get worse. Moreover, with 3 I have seen strong errors appearing in the corners of the domain. So, we have to investigate what is happening.

The inspiration for adding this case came from this paper
https://doi.org/10.1016/j.jcp.2018.06.062
where it is used to evaluate pressure-velocity coupling schemes and associated errors. Since the solution is very smooth in space, it is easy to make the spatial error quite negligible and focus on what is happening in time.

With a bit more work, we can add the advection terms. But I think just having a pure linear Stokes problem has its merits.


timofeymukha:
The errors in pressure are generally about 2 magnitudes higher than velocity, and are concentrated near boundaries

image

MartinKarp:
Great job, I will have a look!

Without having run it I wonder whether what we are seeing at the boundary is related to the "numerical boundary layer" where the error should be concentrated in out splitting, this should in principle be quite large when Re -> 0, but get smaller and smaller as Re -> inf. I think I wrote a bit about it in my thesis.

timofeymukha:
yeah, I think this is a good case to look at all these things in some isolation. My main concern right now is that 1st order time-stepping is the most accurate :-)

MartinKarp:
For me when I change it I get that the error decreases I think. 

Order 1:
Screenshot 2024-11-08 at 16 28 59

Order 3:

njansson (approved):

MartinKarp:
I have had a look and the behavior is related to the following lines, something that actually @gloopydoop and Martin Berggren drew to my attention earlier this year.

https://github.com/ExtremeFLOW/neko/blob/bbf0a129bdb52024745da1a3880aadce22cbe147/src/fluid/bcknd/cpu/pnpn_res_cpu.f90#L71C1-L73C64

timofeymukha:
OK, very interesting. But I guess that is a non-negligible change to the PnPn algorithm, right?

adperezm:
I have a very stupid question but this case is set up for 2D and we use the "trick" of the 1 element with periodic bc. Right?

Are we sure that one would not somehow affect?

timofeymukha:
> I have a very stupid question but this case is set up for 2D and we use the "trick" of the 1 element with periodic bc. Right?
> 
> Are we sure that one would not somehow affect?

Yes, it is the 1 element trick. I don't think it should completely destroy things, but maybe affect formal convergence rates? I thought 2D is kinda useless before, but I've changed my mind a bit now, I think it would be nice to have pure 2D. The problem is that it will always be a very low priority.

pschlatt1:
the one-element-2D should not affect the convergence rates, as there is no change (hopefully) in the third direction, so there is no solve needed, and thus all is dominated by the other two directions. I guess Kovasznay flow (or TS waves) should show the formal convergence rates.

Having 2D capability is very nice, as most implementations could be checked in an cheap way (e.g. convergence, stability, adjoints, ...), so indeed having a native capability would be excellent. But yet having the if2d flag all across the code is probably also not what you want. Isnt there a good idea how this could be done?

njansson:
> the one-element-2D should not affect the convergence rates, as there is no change (hopefully) in the third direction, so there is no solve needed, and thus all is dominated by the other two directions. I guess Kovasznay flow (or TS waves) should show the formal convergence rates.
> 
> Having 2D capability is very nice, as most implementations could be checked in an cheap way (e.g. convergence, stability, adjoints, ...), so indeed having a native capability would be excellent. But yet having the if2d flag all across the code is probably also not what you want. Isnt there a good idea how this could be done?

It could be solved by more abstractions and objects, could be quite messy

PR #1600, opened 2024-11-13
A little change on LES models

Shiyu-Sandy-Du:
In this PR, several things are modified regarding LES models:

1.  Another option of the computation of the filter size is available which is based on the average GLL spacing
2.  A little bit change on sigma model
3.  Activate nut computation when restart before time stepping
4.  Add gs operation for nut field


src/les/les_model.f90
```diff
@@ -205,7 +205,35 @@ subroutine les_model_compute_delta(this)
           dk = sqrt(dk)
           this%delta%x(:,:,:,e) = (di * dj * dk)**(1.0_rp / 3.0_rp)
        end do
+    else if (this%delta_type .eq. "elementwise_avg") then
```
muenschm, line 208: I was also already thinking about that, but isn't there an easier way to compute this: elementsize/(Polynomial order + 1)

src/les/bcknd/cpu/sigma_cpu.f90
muenschm, line 1: not sure if we should do this, it is about estimating the eigenvalues, if eps is "too" large the result may be wrong, here it is a clear statement sigma1: has to be grater than 0 and not greater than eps. However, as it appears here: the use of .gt., >, 0.0_rp and 0.0 is a bit inconsistent, maybe this causes some issues.

src/les/bcknd/device/cuda/sigma_nut_kernel.h
muenschm, line 1: same here

src/les/bcknd/device/hip/sigma_nut_kernel.h
muenschm, line 1: and here as well

src/les/bcknd/cpu/sigma_cpu.f90
muenschm: yes, but this is not zero, her it is just about some comparison. Have you made some good experience with this or why you want to change this?

src/les/les_model.f90
Shiyu-Sandy-Du: Do you mean the diameter of the element/(Polynomial order + 1)?

src/les/bcknd/cpu/sigma_cpu.f90
Shiyu-Sandy-Du: The reason here is mainly to avoid the problem of 1/0.0_rp. If it's 1/e-18 per say, I am a bit afraid the risk of blowing up.

muenschm: zero is a value which can be represented exactly, thus don't mess it up

src/les/les_model.f90
muenschm: well to be consistent, also for 3D, it has to be (Volume^1/3)/(N+1)

Shiyu-Sandy-Du:
Seems that we don't have any more opinions on it. Maybe we could merge it. @njansson

timofeymukha:
But you didn't merge in my PR to your branch, right?

timofeymukha:
Sorry that I didn't have time to look at this one.  A quick comment would be to rename elementwise_avg to elementwise_average, just to avoid remembering the shortening.

Shiyu-Sandy-Du:
> But you didn't merge in my PR to your branch, right?

I did it in another PR.

njansson (approved):

timofeymukha (approved):

PR #1604, opened 2024-11-13
Enhance/cfl deviation tol

Shiyu-Sandy-Du:
Let the user to change the tolerance of the CFL number deviation in case file


njansson (approved):

timfelle (approved):

Mr-2c:
I don't see a control for when cfl exceeds the stability cfl in the if statement of lines 115-117... is that included somewhere else? I think the behaviour should be slightly more complex. something like:
if ( (cfl.gt.cfl_max) .or. 
(abs(cfl_avrg - this%set_cfl) .ge. &
             this%dev_tol * this%set_cfl .and. &
             this%dt_last_change .ge. this%max_update_frequency) ) then ...
with cfl_max being the stability limit.

Shiyu-Sandy-Du:
Shouldn't it be bounded by the user when setting this%set_cfl and this%dev_tol?

Mr-2c:
what if the user sets the limit incorrectly, for example using cfl=0.6 and no dev_tol set, or sets them correctly but the update is not frequent enough, or sets them for third order method and switched to second order but forgets to change them... 
I think keeping the time stepping stable should supersede everything and so there should be a separate and explicit check for that, perhaps with an explicit warning that dt was changed due to stability reasons. just my opinion though

Mr-2c:
I actually agree... 
what if we add one more switch to keep dt in the stable region independent of the other parameters? 
and another question is why are not these parameters related to variable dt part of the variable dt block, but instead part of the full case block? I think that is more intuitive that if variable dt is switched off, none of them get activated

PR #1605, opened 2024-11-14
nvhpc: workaround for submodule bug

gcorbin:
Work around this bug in NVHPC:
https://forums.developer.nvidia.com/t/submodule-matching-error-of-declaration/211278 by removing target attribute from dummy args to module subroutine

### DIFF
--- a/src/krylov/krylov.f90
+++ b/src/krylov/krylov.f90
@@ -189,7 +189,7 @@ module krylov
      !! @param monitor Enable/disable monitoring, optional.
      module subroutine krylov_solver_factory(object, n, type_name, &
           max_iter, abstol, M, monitor)
-       class(ksp_t), allocatable, target, intent(inout) :: object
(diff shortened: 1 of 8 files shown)


gcorbin:
Tested with NVHPC 24.3:
- OpenMPI/4.1.6. works (tested with `examples/tgv`)
- ParaStationMPI/5.9.2: compiles but runs into this error: `0: nvf_to_cfi_type: unexpected nvf kind 281470681743361`. I am not sure whether this is related to the patches. Wild guess: it is related to conversion of some nvhpc internal descriptor to the `CFI_cdesc_t` of `ISO_Fortran_binding.h`

njansson:
Great stuff! Many thanks for finding the cause and fixing the issue

njansson (approved):

tuananhdao (approved):

PR #1608, opened 2024-11-15
added check for running in serial and only using re2 to avoid potential issues

Mr-2c:


contrib/rea2nbin/rea2nbin.f90
```diff
@@ -16,8 +16,16 @@ program rea2nbin
   end if
   
   call neko_init 
+
+  if (pe_size.gt.1) then
```
timofeymukha, line 20:
```suggestion
  if (pe_size .gt. 1) then
```

timofeymukha (approved):

contrib/rea2nbin/rea2nbin.f90
```diff
@@ -16,8 +16,16 @@ program rea2nbin
...
+
+  if (pe_size .gt. 1) then
+     call neko_error("rea2nbin can only run on 1 rank")
+  end if
   
   call get_command_argument(1, fname) 
+  call filename_suffix(fname, suffix)
+  if (suffix .ne. "re2") then
+     call neko_error("rea is no longer supported. Please convert to re2 first")
+  end if
```
njansson, line 28: Is this really the case? I guess @MartinKarp and @vbaconnet knows better since they have been working with these readers lately

vbaconnet: I mean technically the rea reader works. So perhaps just warning the user to use `re2` if the rea --> nmsh conversion doesn't work would be enough?

Mr-2c: I am not entirely sure... it is based on my experience that it generated nans when converted from rea, but worked fine when converted the rea to re2 and then re2 to nmsh. given what Victor said, it could be that maybe read from rea some faces do not close since it is ASCII while maybe reatore2 fixes such mismatches at least, but even then it would be mean that the converter does not support rea... maybe we can change the error to a warning instead, that use at own risk and possible that it generates nans, and if it did try using .re2 instead

Mr-2c: I had a very quick look at nek5000's reatore2. it does not seem to be doing anything other than reading and writing.

pschlatt1:
Why not drop rea altogether? Is there any reason?

njansson:
Let's remove rea then, or at least keep the error in the converter as a start

njansson (approved):

PR #1614, opened 2024-11-21
Cleans the Krylov factory a bit

MartinKarp:
In preparation for adding better settings for tolerances. Still allows for different inits for the different solvers (especially if one want a precon or not).


timfelle:
As far as i understand you can pass optional parameters down to other functions without explicitly needing to define them. So i think we should be able to reduce it quite a lot more.

```.f90
subroutine func(a, b)
   integer :: a
   integer :: optional b

   if preset(b)
      ! do something
   end if

   ! ... Do stuff

end subroutine func

subroutine func2(a, b)
   integer :: a
   integer :: optional b

   call func(a, b)
end subroutine func2
```

MartinKarp:
But then b will be present but without being allocated or? So the call to func in func2 would never return false on the present(b), but it would act as if it was set.

timfelle:
> But then b will be present but without being allocated or? So the call to func in func2 would never return false on the present(b), but it would act as if it was set.

The point is, `func2` does not need to know if b is set or not, all it does is delegate access down to `func`, that is my understanding of it:

If `func2(a)` is called then no `b` is passed to `func` and `if present(b)` will return false.
If `func2(a, b)` is called then `b` is passed to `func` and `if present(b)` will return true.

timfelle:
https://godbolt.org/z/svGE9GoKP

```.f90
module test
  implicit none
contains
  subroutine func(a, b)
    integer :: a
    integer, optional :: b

    if (present(b)) then
       ! do something
       write(*,*) 'Is present'
    else
       ! do something else
       write(*,*) 'Is not present'
    end if

    ! ... Do stuff

  end subroutine func

  subroutine func2(a, b)
    integer :: a
    integer, optional :: b

    write(*,*) 'Calling func'
    write(*,*) 'a = ', a
    if (present(b)) then
       write(*,*) 'b = ', b
    end if

    call func(a, b)


  end subroutine func2
end module test

program name
  use test
  implicit none

  write(*,*) 'Single argument'
  call func2(1)

  write(*,*) 'Two arguments'
  call func2(1, 2)

end program name
```

```
Program returned: 0
 Single argument
 Calling func
 a =            1
 Is not present
 Two arguments
 Calling func
 a =            1
 b =            2
 Is present
```

MartinKarp:
That could make us simplify many things I think! I will try it too

MartinKarp:
Since there is not a deferred subroutine for init I think one still needs the type select

timfelle:
> Since there is not a deferred subroutine for init I think one still needs the type select

That might very well be the case. I did not check anything but a simple set of subroutines.

MartinKarp:
I guess one could add a deferred subroutine, but I guess this is a good start for now.

timfelle (approved):

timofeymukha:
Looks very good. But maybe, just to make sure, @njansson can confirm that there was no obscure reason for the select type stuff in e.g. the destructor? Due to some compiler or something. Otherwise, we can totally merge.

njansson:
> Looks very good. But maybe, just to make sure, @njansson can confirm that there was no obscure reason for the select type stuff in e.g. the destructor? Due to some compiler or something. Otherwise, we can totally merge.

There was at least one compiler that went belly up on this. Can't remember which, but probably Cray.

timofeymukha:
> > Looks very good. But maybe, just to make sure, @njansson can confirm that there was no obscure reason for the select type stuff in e.g. the destructor? Due to some compiler or something. Otherwise, we can totally merge.
> 
> There was at least one compiler that went belly up on this. Can't remember which, but probably Cray.

OK. I am hoping that this is no longer the case now. We `free` like this all over the code. But maybe check things with Cray on some machine just in case, @MartinKarp ?

MartinKarp:
Ill check now

MartinKarp:
Seems to work on Dardel (cce 16.0.1)

PR #1616, opened 2024-11-22
Warning when performing LES

Shiyu-Sandy-Du:
Considering the users might not familiar with the code, then it's better to have a warning when having LES simcomp but not enable it in fluid.


timofeymukha:
Can't we do this directly in the les simcomp?

Shiyu-Sandy-Du:
> Can't we do this directly in the les simcomp?

We could do it!

timfelle:
Can LES be performed without the nut_field? I am just thinking whether it makes sense to allow a user to run a simulation with the simcomp without enabeling the field?

Shiyu-Sandy-Du:
> Can LES be performed without the nut_field? I am just thinking whether it makes sense to allow a user to run a simulation with the simcomp without enabeling the field?

No, LES in neko is based on eddy viscosity hypothesis now, which actually vary the local viscosity by nut_field.

timfelle:
> > Can LES be performed without the nut_field? I am just thinking whether it makes sense to allow a user to run a simulation with the simcomp without enabeling the field?
> 
> No, LES in neko is based on eddy viscosity hypothesis now, which actually vary the local viscosity by nut_field.

In  that case, why are you only printing a warning and not throwing an error. The simulation is clearly not valid according to what the user attempts to execute.

Shiyu-Sandy-Du:
> > > Can LES be performed without the nut_field? I am just thinking whether it makes sense to allow a user to run a simulation with the simcomp without enabeling the field?
> > 
> > 
> > No, LES in neko is based on eddy viscosity hypothesis now, which actually vary the local viscosity by nut_field.
> 
> In that case, why are you only printing a warning and not throwing an error. The simulation is clearly not valid according to what the user attempts to execute.

It's possible that the users want to check the nut_field of a DNS result, though it doesn't make much sense but I don't think we should block them to do it.

timfelle:
> > > > Can LES be performed without the nut_field? I am just thinking whether it makes sense to allow a user to run a simulation with the simcomp without enabeling the field?
> > > 
> > > 
> > > No, LES in neko is based on eddy viscosity hypothesis now, which actually vary the local viscosity by nut_field.
> > 
> > 
> > In that case, why are you only printing a warning and not throwing an error. The simulation is clearly not valid according to what the user attempts to execute.
> 
> It's possible that the users want to check the nut_field of a DNS result, though it doesn't make much sense but I don't think we should block them to do it.

But we are talking about enabeling the simcomp without having a nut_field right? Not enabeling the nut_field without having a the LES_simcomp.

Shiyu-Sandy-Du:
> > > In that case, why are you only printing a warning and not throwing an error. The simulation is clearly not valid according to what the user attempts to execute.
> > 
> > 
> > It's possible that the users want to check the nut_field of a DNS result, though it doesn't make much sense but I don't think we should block them to do it.
> 
> But we are talking about enabeling the simcomp without having a nut_field right? Not enabeling the nut_field without having a the LES_simcomp.

Enabling "nut_field" in fluid means using stress formulation for the fluid equation instead of only computing and/or outputing the nut_field.

timfelle:
> > > > In that case, why are you only printing a warning and not throwing an error. The simulation is clearly not valid according to what the user attempts to execute.
> > > 
> > > 
> > > It's possible that the users want to check the nut_field of a DNS result, though it doesn't make much sense but I don't think we should block them to do it.
> > 
> > 
> > But we are talking about enabeling the simcomp without having a nut_field right? Not enabeling the nut_field without having a the LES_simcomp.
> 
> Enabling "nut_field" in fluid means using stress formulation for the fluid equation instead of only computing and/or outputting the nut_field. In this way, we also allow user defined "nut_field".

All of that does not change the simcomp you are working on here. The simcomp require a variable viscosity field, if that is not there, regardless of how it was defined you are throwing a warning. My question is, why would a user want to use the simcomp without the variable field?

We have the same issue in the bousinesq source term. The source term can be enabled without the scalar field which should be driving it even existing, resulting in a lot of computations being executed, with no effect.

timfelle:
> > > > Can LES be performed without the nut_field? I am just thinking whether it makes sense to allow a user to run a simulation with the simcomp without enabeling the field?
> > > 
> > > 
> > > No, LES in neko is based on eddy viscosity hypothesis now, which actually vary the local viscosity by nut_field.
> > 
> > 
> > In that case, why are you only printing a warning and not throwing an error. The simulation is clearly not valid according to what the user attempts to execute.
> 
> It's possible that the users want to check the nut_field of a DNS result, though it doesn't make much sense but I don't think we should block them to do it.

They should do that with the field writer.

timofeymukha:
This is kinda tough. I could imagine a scientific scenario, where one would want to run an LES model in the background, without it affecting the flow. For example, you can have a DNS grid, implement some kind of filter (maybe another simcomp) and feed that the LES model as the velocity field, to just see how the model performs. There are quite a bit of studies like that out there, although they have obvious limitations. So, I would perhaps keep this at the level of warning.

I will add the variable properties to the fluid's and scalar's log.

Shiyu-Sandy-Du:
> This is kinda tough. I could imagine a scientific scenario, where one would want to run an LES model in the background, without it affecting the flow. For example, you can have a DNS grid, implement some kind of filter (maybe another simcomp) and feed that the LES model as the velocity field, to just see how the model performs. There are quite a bit of studies like that out there, although they have obvious limitations. So, I would perhaps keep this at the level of warning.
> 
> I will add the variable properties to the fluid's and scalar's log.

I think it's a good answer to @timfelle 's question.

timofeymukha:
Yeah, but the LES simcomp will create it. It is the one in charge of the field, so to speak. The fluid and scalar just use it, if you tell them to.

timfelle:
> Yeah, but the LES simcomp will create it. It is the one in charge of the field, so to speak. The fluid and scalar just use it, if you tell them to.

Alright, that was the question I wanted to have answered.

src/simulation_components/les_simcomp.f90
```diff
@@ -73,6 +74,12 @@ subroutine les_simcomp_init_from_json(this, json, case)
 
     call this%free()
 
+    ! Check for whether eddy viscosity is enabled in fluid_scheme
+    if (case%fluid%variable_material_properties .eqv. .false.) then
+        call neko_warning("Eddy viscosity is not acting &
+                          &on the equations!")
```
timfelle, line 80:
```suggestion
                          &on the equations")
```

Please do not create comments mid string.

timofeymukha:
@Shiyu-Sandy-Du I opened a PR to your branch :-).

muenschm (commented):
Well, maybe we could formulate a warning together with some more details such that a user (not a developer) knows what has to be done. "Eddy viscosity is not acting on the equations" -> And, what's next?

Mr-2c:
> > > > In that case, why are you only printing a warning and not throwing an error. The simulation is clearly not valid according to what the user attempts to execute.
> > > 
> > > 
> > > It's possible that the users want to check the nut_field of a DNS result, though it doesn't make much sense but I don't think we should block them to do it.
> > 
> > 
> > But we are talking about enabeling the simcomp without having a nut_field right? Not enabeling the nut_field without having a the LES_simcomp.
> 
> Enabling "nut_field" in fluid means using stress formulation for the fluid equation instead of only computing and/or outputing the nut_field. In this way, we also allow user defined "nut_field".

can we do stress formulation without enabling LES or is that directly set when we select a variable property? I think they should be made independent for maximum flexibility and allow for tests

njansson:
> > > > > In that case, why are you only printing a warning and not throwing an error. The simulation is clearly not valid according to what the user attempts to execute.
> > > > 
> > > > 
> > > > It's possible that the users want to check the nut_field of a DNS result, though it doesn't make much sense but I don't think we should block them to do it.
> > > 
> > > 
> > > But we are talking about enabeling the simcomp without having a nut_field right? Not enabeling the nut_field without having a the LES_simcomp.
> > 
> > 
> > Enabling "nut_field" in fluid means using stress formulation for the fluid equation instead of only computing and/or outputing the nut_field. In this way, we also allow user defined "nut_field".
> 
> can we do stress formulation without enabling LES or is that directly set when we select a variable property? I think they should be made independent for maximum flexibility and allow for tests

It should be independent now, but not easy to enable, the current way is to turn on `variable_material_properties`

timofeymukha:
> > > > > In that case, why are you only printing a warning and not throwing an error. The simulation is clearly not valid according to what the user attempts to execute.
> > > > 
> > > > 
> > > > It's possible that the users want to check the nut_field of a DNS result, though it doesn't make much sense but I don't think we should block them to do it.
> > > 
> > > 
> > > But we are talking about enabeling the simcomp without having a nut_field right? Not enabeling the nut_field without having a the LES_simcomp.
> > 
> > 
> > Enabling "nut_field" in fluid means using stress formulation for the fluid equation instead of only computing and/or outputing the nut_field. In this way, we also allow user defined "nut_field".
> 
> can we do stress formulation without enabling LES or is that directly set when we select a variable property? I think they should be made independent for maximum flexibility and allow for tests

As soon as you add `nut_field` to the fluid, stress formulation is triggered. You can create that field in the user file with 0 values. That effectively gives you stress formulation with no turbulence modelling.

Mr-2c:
> > > > > > In that case, why are you only printing a warning and not throwing an error. The simulation is clearly not valid according to what the user attempts to execute.
> > > > > 
> > > > > 
> > > > > It's possible that the users want to check the nut_field of a DNS result, though it doesn't make much sense but I don't think we should block them to do it.
> > > > 
> > > > 
> > > > But we are talking about enabeling the simcomp without having a nut_field right? Not enabeling the nut_field without having a the LES_simcomp.
> > > 
> > > 
> > > Enabling "nut_field" in fluid means using stress formulation for the fluid equation instead of only computing and/or outputing the nut_field. In this way, we also allow user defined "nut_field".
> > 
> > 
> > can we do stress formulation without enabling LES or is that directly set when we select a variable property? I think they should be made independent for maximum flexibility and allow for tests
> 
> As soon as you add `nut_field` to the fluid, stress formulation is triggered. You can create that field in the user file with 0 values. That effectively gives you stress formulation with no turbulence modelling.

two questions: stress formulation here only means we are solving three laplacians with velocity gradient at once, right? if so, perhaps we could rename it to a better more descriptive name.
this sounds more like a hack, while given what Niclas said about the improved performance when solving for all three components as once, I think we should make it directly selactable from the case file.

Shiyu-Sandy-Du:
> > > > > > > In that case, why are you only printing a warning and not throwing an error. The simulation is clearly not valid according to what the user attempts to execute.
> > > > > > 
> > > > > > 
> > > > > > It's possible that the users want to check the nut_field of a DNS result, though it doesn't make much sense but I don't think we should block them to do it.
> > > > > 
> > > > > 
> > > > > But we are talking about enabeling the simcomp without having a nut_field right? Not enabeling the nut_field without having a the LES_simcomp.
> > > > 
> > > > 
> > > > Enabling "nut_field" in fluid means using stress formulation for the fluid equation instead of only computing and/or outputing the nut_field. In this way, we also allow user defined "nut_field".
> > > 
> > > 
> > > can we do stress formulation without enabling LES or is that directly set when we select a variable property? I think they should be made independent for maximum flexibility and allow for tests
> > 
> > 
> > As soon as you add `nut_field` to the fluid, stress formulation is triggered. You can create that field in the user file with 0 values. That effectively gives you stress formulation with no turbulence modelling.
> 
> two questions: stress formulation here only means we are solving three laplacians with velocity gradient at once, right? if so, perhaps we could rename it to a better more descriptive name. this sounds more like a hack, while given what Niclas said about the improved performance when solving for all three components as once, I think we should make it directly selactable from the case file.

No, stress formulation means we have the viscosity in the first derivative in the laplacians. But if we change the "stress formulation" into "conjugate gradietn solver" in your question, it would be an unanswered one I guess.

muenschm:
> > > > > > In that case, why are you only printing a warning and not throwing an error. The simulation is clearly not valid according to what the user attempts to execute.
> > > > > 
> > > > > 
> > > > > It's possible that the users want to check the nut_field of a DNS result, though it doesn't make much sense but I don't think we should block them to do it.
> > > > 
> > > > 
> > > > But we are talking about enabeling the simcomp without having a nut_field right? Not enabeling the nut_field without having a the LES_simcomp.
> > > 
> > > 
> > > Enabling "nut_field" in fluid means using stress formulation for the fluid equation instead of only computing and/or outputing the nut_field. In this way, we also allow user defined "nut_field".
> > 
> > 
> > can we do stress formulation without enabling LES or is that directly set when we select a variable property? I think they should be made independent for maximum flexibility and allow for tests
> 
> As soon as you add `nut_field` to the fluid, stress formulation is triggered. You can create that field in the user file with 0 values. That effectively gives you stress formulation with no turbulence modelling.

you could also set C=0 or C=eps for this within the SGS model to "switch off" turbulence modeling

src/simulation_components/les_simcomp.f90
```diff
@@ -73,6 +74,14 @@ subroutine les_simcomp_init_from_json(this, json, case)
 
     call this%free()
 
+    ! Check for whether eddy viscosity is enabled in fluid_scheme
+    if (case%fluid%variable_material_properties .eqv. .false.) then
+        call neko_error("Eddy viscosity is not acting &
+                          &on the equations. &
+                          &Plese set up a nut_field option &
```
timofeymukha, line 81:
```suggestion
                          &Please set up a nut_field option &
```

timofeymukha (approved):

PR #1617, opened 2024-11-25
Rename cpldcg

MartinKarp:
Im not really working on the LES, but I found the cpldcg name to be very confusing. 

Let me know if this is reasonable

### DIFF
--- a/src/krylov/krylov_fctry.f90
+++ b/src/krylov/krylov_fctry.f90
@@ -62,7 +62,8 @@ submodule (krylov) krylov_fctry
      "gmres", &
      "cheby", &
      "bicgstab", &
-     "cpldcg"]
+     "fusedcoupledcg", &
+     "coupledcg"]
 
 contains
 
@@ -98,7 +99,7 @@ contains
        else
           allocate(cg_t::object)
        end if
-    else if (trim(type_name) .eq. 'cpldcg') then
+    else if (trim(type_name) .eq. 'coupledcg') then
        allocate(cg_cpld_t::object)
        if (NEKO_BCKND_DEVICE .eq. 1) then
           call neko_error('Coupled CG only supported for CPU')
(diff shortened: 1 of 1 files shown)


timofeymukha:
Certainly an improvement in terms of readability. I wonder if adding _ as a separtor would be even better, e.g. coupled_cg

Shiyu-Sandy-Du (approved):

MartinKarp:
This is what is most consistent with the other types (pipecg, fusedcg, etc.), but we can go with _ if that is better

muenschm (approved):

muenschm (approved):

timofeymukha:
@MartinKarp You have a point.. The underscore is the most consistent with the rest of the JSON keywords but the solvers are their own island right now. I think using _ would be the most future-proof. We can probably rename the other ones with little inconvenience to users, since I think only cg and gmres are really in active use, right?

MartinKarp:
Conventionally I would say different variants of cg/gmres use a dash - (e.g. CA-CG, Pipelined-CG etc.). I should admit I do not really care about separator vs no separator. I think the others currently are used mostly by Niclas and perhaps for others who benchmark the code. The scope of my PR was primarily to make cpld more tangible.

timfelle:
> Conventionally I would say different variants of cg/gmres use a dash - (e.g. CA-CG, Pipelined-CG etc.). I should admit I do not really care about separator vs no separator. I think the others currently are used mostly by Niclas and perhaps for others who benchmark the code. The scope of my PR was primarily to make cpld more tangible.

Given the scope of the PR, I completely agree that we should get this closed. But having the case file be as "human-readable" as possible would probably be a good idea and make it easier to convince people to use it. I like your proposal with "Variant-Solver" I think that translates and reads quite easily.

timofeymukha:
I really think we should stick to one "space" binding character for the whole case file, and not mix -, _ etc. And we already use _  everywhere.

MartinKarp:
That is also a good point. We can add an issue on it.

PR #1622, opened 2024-11-30
Feature/veldiv

Mr-2c:
Added velocity divergence as a simulation component. still need to clean up the code. have to figure out and correct if divergence/lambda2 simcomps are computed every time step and fix that if necessary


vbaconnet:
About the computing once per time step, we had this discussion with the spectral error simcomp and I don't think we landed anywhere for sure, but the idea of adding a "compute_control": "on_output" type keyword came up. I don't remember if this ever got in with your simcomp PR @timfelle ? 

But anyways yes at the moment you have to manually set the compute control to match that of the fluid output, otherwise the default is to compute once per time step.

timfelle:
> About the computing once per time step, we had this discussion with the spectral error simcomp and I don't think we landed anywhere for sure, but the idea of adding a "compute_control": "on_output" type keyword came up. I don't remember if this ever got in with your simcomp PR @timfelle ?
> 
> But anyways yes at the moment you have to manually set the compute control to match that of the fluid output, otherwise the default is to compute once per time step.

We implemented a "fluid_output" keyword for "compute_control", such that one can match with whenever the fluid gets saved. The default of "output_control" is to follow "compute_control".

Mr-2c:
> > About the computing once per time step, we had this discussion with the spectral error simcomp and I don't think we landed anywhere for sure, but the idea of adding a "compute_control": "on_output" type keyword came up. I don't remember if this ever got in with your simcomp PR @timfelle ?
> > But anyways yes at the moment you have to manually set the compute control to match that of the fluid output, otherwise the default is to compute once per time step.
> 
> We implemented a "fluid_output" keyword for "compute_control", such that one can match with whenever the fluid gets saved. The default of "output_control" is to follow "compute_control".
> 
> I think we called it "fluid_output" to avoid the ambiguity of "is it the fluid output or the simcomp output"

so is it already there? do I need to implement it in a new simcomp or is it managed by the main one (I assume the latter)?

Mr-2c:
and can we have different defaults for different sim comps? for example, the correct default behaviour for lambda2 and divergence seems to be when they a field is written, while the "correct" behaviour for nut seems to be every time step

timfelle:
> > > About the computing once per time step, we had this discussion with the spectral error simcomp and I don't think we landed anywhere for sure, but the idea of adding a "compute_control": "on_output" type keyword came up. I don't remember if this ever got in with your simcomp PR @timfelle ?
> > > But anyways yes at the moment you have to manually set the compute control to match that of the fluid output, otherwise the default is to compute once per time step.
> > 
> > 
> > We implemented a "fluid_output" keyword for "compute_control", such that one can match with whenever the fluid gets saved. The default of "output_control" is to follow "compute_control".
> > I think we called it "fluid_output" to avoid the ambiguity of "is it the fluid output or the simcomp output"
> 
> so is it already there? do I need to implement it in a new simcomp or is it managed by the main one (I assume the latter)?

It is already implemented and available. When setting up a simcomp in the case file all you do is:

```json
{
    "order": 1,
    "type": "what_ever_type_you_want",
    "compute_control": "fluid_output"
},
```

It is all handled by the parent class, so no new code is needed.

timfelle:
> and can we have different defaults for different sim comps? for example, the correct default behaviour for lambda2 and divergence seems to be when they a field is written, while the "correct" behaviour for nut seems to be every time step

No, that is not possible as the base class reads the variable on initialization.

Mr-2c:
> Thanks, Siavash. I would probably avoid pinning velocity, and just have a generic divergence simcomp, with a `field_names` keyword (or similar, we should check what we usually call that in other simcomps) keyword that provides the name of the 3 field components, to be plucked from the registry, as you do now with `u` `v` `w`. So like the `weak_grad` https://github.com/ExtremeFLOW/neko/blob/develop/src/simulation_components/weak_grad.f90

that's a good idea. do we have a way of identifying vectors? since the divergence is only defined for vectors (and tensors, but that is too much for now), ideally we want to have the user select one vector and then have the components taken from registry. (that would also make the naming more straightforward, but that not be a big problem)

timofeymukha:
@Mr-2c Do you think we can close, since we now have the `divergence` simcomp?

Mr-2c:
> @Mr-2c Do you think we can close, since we now have the `divergence` simcomp?

yes, definitely :-)

timfelle:
> > @Mr-2c Do you think we can close, since we now have the `divergence` simcomp?
> 
> yes, definitely :-)

PR #1627, opened 2024-12-03
Fix pressure when no Dirichlet BCs

MartinKarp:
Another BC bug.

When there are no Dirichlet conditions for the pressure, but mixed bcs such as symmetry, the pressure can grow to a very high amplitude. This fixes this issue by ensuring the pressure in these cases always has zero mean.

- Updates the ortho function.
- Adds some functions to bc_list_t to check if no bcs are present.
- Adds the global number of unique and non-unique points to the fluid scheme.
- Add the calls to ortho in fluid_pnpn.


timofeymukha:
Interesting. Typically, when there are no pressure bcs one has to somehow select a reference pressure location, where the pressure will be the reference one. I wonder why the issue only occurs with symmetry, though.

src/math/operators.f90
```diff
@@ -202,16 +202,21 @@ end subroutine opgrad
 
   !> Othogonalize with regard to vector (1,1,1,1,1,1...,1)^T.
   !! @param x The vector to orthogonolize.
-  !! @param n The size of `x`.
-  !! @param glb_n The global number of elements of `x` across all MPI ranks. Be careful with overflow!
-  subroutine ortho(x, n, glb_n)
+  !! @param glb_n_points The global number of non-unique gll points in the grid.
```
timofeymukha, line 205:
```suggestion
  !! @param glb_n_points The global number of non-unique gll points in the grid.
  !! @note This is equivalent to subtracting the mean of `x` from each of its elements.
```

timofeymukha:
I guess I don't fully understand how this is done in PnPn or PnPn-2. Like, when you run channel flow, what determines your reference pressure?

timofeymukha (approved):

njansson (approved):

PR #1634, opened 2024-12-10
Feature/smagorinsky dyn smag device

Shiyu-Sandy-Du:
Device implementation for smagorinsky and dynamic smagorinsky is added.
And the intent of tnsr3d is modified to be strict.


njansson (approved):
Great stuff! 

But would be good if an LES person also gives it a look

muenschm (approved):
Well Done!

PR #1641, opened 2024-12-15
Improved bc_list type

timofeymukha:
This puts `bc_list_t` into its own file, and makes similar to the `field_list_t` type. Various bc_list-related subroutines are made methods of the type.

@tuananhdao I sadly had to revert the intents on the bclst arguments to inout in the ksp solvers. But maybe this can be averted? Could you maybe have a quick look at the time? Essentially, the routines where I run something for the `items`, it complained that it is incompatible with `intent(in)`.


tuananhdao:
I like this refactoring :). Probably, the bclst arguments have to be `intent(inout)` due to the new `bc_list_append(this, bc)` in place of `bc_list_add(bclst, bc)`, which is an improvement.

njansson (approved):
🍻

src/bc/bc_list.f90
```diff
@@ -0,0 +1,272 @@
...
+    type(c_ptr) :: x_d
+    type(c_ptr) :: y_d
+    type(c_ptr) :: z_d
+    integer :: i
+
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       x_d = device_get_ptr(x)
+       y_d = device_get_ptr(y)
+       z_d = device_get_ptr(z)
+       if (present(t) .and. present(tstep)) then
```
MartinKarp, line 215: I think we figured out with Tim that we can just always send down t and tstep. So the only case that is needed is the first one.

timofeymukha: Oh, really? Cool! I will change it then.

PR #1643, opened 2024-12-16
Add neko_type_error for reporting invalid type selection

timofeymukha:
Adds a routine for reporting when an invalid type has been attempted to allocate for some base pointer type.

The idea is to use this in factories. The main bonus is the convenience of getting a list of valid types.


timofeymukha:
The idea is to use this in factories, leveraging that we already have lists of valid types for all of them.

njansson (approved):

PR #1646, opened 2024-12-19
tamg updates and print fixes

pohm01:
Updated the printout from tamg.

Remove unused aggregation functions.

Added objects for easier mapping between levels.


src/multigrid/tree_amg.f90
```diff
@@ -266,9 +285,7 @@ recursive subroutine tamg_matvec_flat_impl(this, vec_out, vec_in, lvl_blah, lvl_
       n = size(vec_in)
       !> Call local finite element assembly
       call this%gs_h%op(vec_in, n, GS_OP_ADD)
-      do i = 1, n
-        vec_in(i) = vec_in(i) * this%coef%mult(i,1,1,1)
-      end do
+      call col2( vec_in, this%coef%mult(:,1,1,1), n)
```
njansson, line 288:
```suggestion
      call col2( vec_in, this%coef%mult(1,1,1,1), n)
```
I think we might get array slicing without a leading 1

njansson (approved):

src/multigrid/tree_amg.f90
pohm01: hmm ok. will fix

njansson:
Seems like ifx fails due to the new bc list, @timofeymukha ?

timfelle:
> Seems like ifx fails due to the new bc list, @timofeymukha ?

ld: ./.libs/libneko.a(tree_amg.o): in function `tree_amg_mp_tamg_matvec_flat_impl_':
ld: tree_amg.f90:(.text+0x6fd1): undefined reference to `bc_list_apply_'

src/multigrid/tree_amg.f90
```diff
@@ -277,35 +294,35 @@ recursive subroutine tamg_matvec_flat_impl(this, vec_out, vec_in, lvl_blah, lvl_
...
+
+      !> Average on overlapping dofs
+      call this%gs_h%op(wrk_in, n, GS_OP_ADD)
+      call col2( wrk_in, this%coef%mult(1,1,1,1), n)
+      !> Finest level matvec (Call local finite element assembly)
+      call this%ax%compute(wrk_out, wrk_in, this%coef, this%msh, this%Xh)
+      !>
+      call this%gs_h%op(wrk_out, n, GS_OP_ADD)
+      call this%blst%apply(wrk_out, n)
+      call bc_list_apply(this%blst, wrk_out, n)
```
timfelle, line 316:
```suggestion
      call this%blst%apply(wrk_out, n)
```

njansson: Ah wait it only needs to include the bc list module

src/multigrid/tree_amg.f90
```diff
@@ -277,35 +294,35 @@ recursive subroutine tamg_matvec_flat_impl(this, vec_out, vec_in, lvl_blah, lvl_
...
+
+      !> Average on overlapping dofs
+      call this%gs_h%op(wrk_in, n, GS_OP_ADD)
+      call col2( wrk_in, this%coef%mult(1,1,1,1), n)
+      !> Finest level matvec (Call local finite element assembly)
+      call this%ax%compute(wrk_out, wrk_in, this%coef, this%msh, this%Xh)
+      !>
+      call this%gs_h%op(wrk_out, n, GS_OP_ADD)
+      call this%blst%apply(wrk_out, n)
+      call this%blst%apply(wrk_out, n)
```
timfelle, line 316: Are we certain the BC's should be applied twice?

timfelle (approved):

timofeymukha:
Hi! I missed the party. Bit things look good now?

src/multigrid/tree_amg.f90
pohm01: Ah, git merge shenanigans... only need it once. Though to be honest I'm not even sure the BCs should be applied at all since I don't quite grasp the concept of them on the coarse grids

PR #1647, opened 2024-12-19
Feature/bc/safety_and_field

timfelle:
Addition of a field interface to bc_list apply.
Hiding the internals of bc_list with functions for list elements and the size. Capacity should not be available.


njansson (approved):

vbaconnet:
This will break the field_dirichlet example and documentation, if you want I push something later today :)

timfelle:
> Tim, really sorry, but can we pause with this one for the moment? I am gonna open a Godzilla PR with bcs latest tomorrow and merging in even this innocent change into my branch will be a bit hellish. However, I will implement at least some of the stuff you add here! 

Lets actually close this PR and I will target yours instead.

PR #1650, opened 2024-12-19
Boundary condition overhaul

timofeymukha:
# Boundary condition overhaul

## Introduction
This PR is probably the first in a series, aiming to have a better and more transparent implementation of boundary conditions as well as a better experience for the user.
The PR is still in *don't merge* due to a few outstanding issues.

- [x] Fixing all the examples. We need to remove the built-in boundaries for RBC, lid, and pipe examples.
- [x] The pressure residual for `hsmg` does not match current develop, when strong pressure boundary conditions are present. For example, in `hemi`. The solver still converges well, but there is something off. I would really love some help diagnosing that, I spent a few days and failed :(. @martinkarp you are perhaps my biggest hope...
- [x] Implementing the `output_boundary` functionality.
- [x] Minor things like docstring improvements.
- [x] There is some instrumentation with `write` and field dumps to help debug point 2. That should be removed.
- [x] Add documentation for the new JSON once we settle on the naming.

That being said, the PR is sadly really big, and in essense it is ready for review.
These are the main focus points of the PR:
- The user-side changes in JSON, with `boundary_types` being ditched.
- Interface changes in `bc_t`, forcing and `init` and `finalize` for all descendents, as well as some other changes to distinguish between strong and weak boundary conditions.
- Removing all boundary condition type components from `scalar_scheme_t` and `fluid_scheme_t` and instead having only `bc_list_t` components, that are populated by factories, based on the case file.

There are other things, which will be covered below, but I would say those three points are really the core.

### JSON-file boundary condition definitions
The PR removes the label-based definition of boundary conditions coupled with JSON objects that provide the configuration. For example, a `v` label + and `inflow_condition` object. Instead, a `boundary_conditions` array is introduced, where each object corresponds to one or several boundary zones, and all the configuration of the boundary is done in that object. Here is `hemi`

```json
        "boundary_conditions": [
        {
            "type": "symmetry",
            "zone_indices":  [1]
        },
        {
            "type": "no_slip",
            "zone_indices":  [2]
        },
        {
            "type": "outflow",
            "zone_indices":  [3]
        },
        {
            "type": "blasius_profile",
            "zone_indices":  [4]
            "approximation": "sin",
            "delta": 0.6,
            "freestream_velocity": [1.0, 0.0, 0.0]
        }
        ],
```

So, I had to make up a name for each of the boundaries. **This is something we should approve together**. Here is a table with all the boundary conditions for the fluid.

| Boundary Condition         | Description            |
|----------------------------|------------------------|
| symmetry                   | symmetry               |
| velocity_value             | A Dirichlet condition for velocity                       |
| no_slip                    | A no-slip wall. Didn't want to use "wall" because e.g. wall models are also at walls, etc.                       |
| outflow                    | The "o" boundary                       |
| normal_outflow             | The "on" boundary                       |
| outflow+dong               | The "o+dong"                       |
| normal_outflow+dong        | The "on+dong"                       |
| shear_stress               | The "sh"                       |
| user_velocity              | The `field_dirichlet_vector_t` user-defined bc for velocity                       |
| user_pressure              | The `field_dirichlet_t` user-defined bc for pressure                       |
| user_velocity_pointwise    | The old pointwise user-defined bc for velocity                       |
| blasius_profile    | A Blasius profile                       |
| wall_model                 | The "wm" boundary                       |

So, all of these are up for discussion. Just a few comments from my side: I was swaying between using "mathy" names like e.g. "velocity_dirichlet" and more "physical" ones, like "velocity_value". I went for the latter, because some of the bcs for the fluid are too complex to concicely define in a mathy way, which is also reflected in Nek5000's and ours way of naming the bc labels.

As seen below, for the scalar I went for "dirichlet" and "neumann" instead, since the physical meaning changes depending on what the scalar represents. However, maybe it is more consitent to go with something like "value" and "flux". Opinions welcome!

| Boundary Condition         | Description            |
|----------------------------|------------------------|
| dirichlet                  | A "d" bounary               |
| neumann             | An "n" boundary             |

All the boundary condition require a `zone_index` keyword to determine the index of the zone in the list of labeled zones. **Built-in zones are no longer supported**. The other keywords are pretty much unchanged, see e.g. the Blasius configuration for hemi--it just moved from its own object into the `boundary_conditions` item.

### The `field_dirichlet_t` user interface

Now no longer accepts a solver name, and the second argument is a `field_dirichlet_t` instead of `bc_list`. The names of the dummy fields one changes in the user routine match the underlying unknown fields for which the bc is actually set. So, by default for velocity it is `u`, `v`, `w`, and for the scalar it is `s`. So, the user should distinguish the solver based on the field name.
Please see that updated `cyl_bl.f90`.

As seen above, I split the condition for pressure and velocity in the JSON. This is reflected in the code. **So currently you cannot change both pressure and velocity in one go**, you will never get a field list containing both of them. **If this is really bad, it should be possible to fix**.

@vbaconnet 

#### Type construction
One of the big issues we had was the lack of deferred procedures for `init` and `finalize`. These are now added, which removes a lot of delicate handling of constuction etc, that we currently have in the code. For all types, include those with nested bcs, one should be able to `init`, mark the boundary, and `finalize`.

@martinkarp Could you please check that I didn't skew up some ordering in `symmetry_t` and `non_normal_t`? I remember that recent PR where you fixed some issues... I moved the whole procedure of finding the normal etc to `finalize`, and I think things should be ok.

The deferred constructor handles parsing the JSON. Most types also have an `init_from_components`, which avoids using JSON, when not needed.

#### Distinction between weak and strong boundary conditions

In order not to have huge `select type` statements in the `_scheme` types, I needed to be able to distinguis between strong and weak bcs, which are applied at different stages of the `_step()` routines. So, a `logical :: strong` component is added to `bc_t`. 

There is an issue however: the mixed conditions like `shear_stress`. They have both a strong and a weak component.
By convention, I currently mark them as `weak`. But it was still unclear on how to run `apply_scalar` / `apply_vector` on them. These conditions currently contain nested bcs, which are responsible for handling the weak and strong parts. For example, the `shear_stress_t` has a `symmetry_t` for no-penetration and `neumann_t` for setting the stress.

One option would be to put these nested bcs directly into the `bc_list`s in the `*_scheme`s. This is kind of what is done now. But I really wanted to have at most one `bc_t` for each boundary in the list. So, the solution I ended up with was adding a `logical :: strong` dummy argument to the `apply_scalar` / `apply_vector` routines. This way, the caller can signal which part of a mixed bc--weak or strong--should be applied. For example, in `fluid_pnpn_t`

```fortran
      ! Add Neumann bc contributions to the RHS
      call this%bcs_vel%apply_vector(f_x%x, f_y%x, f_z%x, &
           this%dm_Xh%size(), t, tstep, strong = .false.)
```

will trigger the Neumann part of the `shear_stress_t`:

```fortran
  !> Boundary condition apply for a generic shear_stress condition
  !! to vectors @a x, @a y and @a z
  subroutine shear_stress_apply_vector(this, x, y, z, n, t, tstep, strong)
    class(shear_stress_t), intent(inout) :: this
    integer, intent(in) :: n
    real(kind=rp), intent(inout),  dimension(n) :: x
    real(kind=rp), intent(inout),  dimension(n) :: y
    real(kind=rp), intent(inout),  dimension(n) :: z
    real(kind=rp), intent(in), optional :: t
    integer, intent(in), optional :: tstep
    logical, intent(in), optional :: strong
    logical :: strong_ = .true.

    if (present(strong)) strong_ = strong

    if (strong_) then
       call this%symmetry%apply_vector(x, y, z, n, t, tstep, .true.)
    else
       call this%neumann_x%apply_scalar(x, n, t, tstep, .false.)
       call this%neumann_y%apply_scalar(y, n, t, tstep, .false.)
       call this%neumann_z%apply_scalar(z, n, t, tstep, .false.)
    end if

  end subroutine shear_stress_apply_vector
```

I guess the minus of this design is that it is superflous for non-mixed bcs, we could of just checked `bc_t%strong` for them and either call the `apply` or not.

#### `bc_labels` is killed

Yep, no need for parsing labels any more.

### The changes in `fluid_scheme_t` and `scalar_scheme_t`

- I have removed all handling of pressure and put it into `fluid_pnpn_t`, which I think makes a lot of sense. 
- There is no longer need for 2 constructors (the second one just handled pressure), so there is now just one `init_base`.
- All boundary condition handling is done in `fluid_pnpn_t`. Hard to say how many bcs will be shared across schemes, since we only have one :-). But I imagine that e.g. for compressible flow the bcs will be somewhat different.
- All the individual bcs components are removed. 
- **The type has two lists for bcs: `bcs_vel` and `bcs_prs`**. These are populated in `fluid_pnpn`.

So, overall this type has shrinked quite a bit. Similar changes are applied to `scalar_scheme_t` altough there there is less work.

### The changes in `fluid_pnpn` and `scalar_pnpn_t`

* Both types have a `setup_bcs_` routine that populates the bc lists by calling factories defined in submodules. For the fluid there are two factories: one for pressure and one for velocity. This includes all the user-interface bcs. The factories always call the deferred `init`, mark a single labled zone, and `finalize`. *So, there is never a single bc that is applied on more than one zone* (except periodic, as usual). This is different from what we have now, where a single bc would "mark all zone with label `o`" and the like. I think this paves the way for better interaction with the bcs, i.e. adding or changing them on the fly, which has been requested. However, this is future work.

* The various `dp_`, `du_` etc lists are populated based on the `strong` property of the bcs. For example

```fortran
    ! Mark Dirichlet bcs for pressure
    call this%bclst_dp%init()
    call this%bc_dp%init_from_components(this%c_Xh)

    do i = 1, this%bcs_prs%size()
       if (this%bcs_prs%strong(i) .eqv. .true.) then
          call this%bc_dp%mark_facets(this%bcs_prs%items(i)%ptr%marked_facet)
       end if
    end do
    call this%bc_dp%finalize()
    call this%bclst_dp%append(this%bc_dp)
```
**But I guess it is here that something is wrong** which leads to the hsmg problem mentioned in the intro. I checked that I always get the correct `p_res`, but after the solve the `dp` is different from `develop`. So I think it has to be something with `bclst_dp` or similar. `hsmg` is the only what that uses this `bc_list` in the computation. 

Also, the nested boundary types are still handled via `select type`, I don't really see a way around it. But I hope we can minimize nesting when we try to support curved boundaries. An example of how things are now:

```fortran
   ! Add all strong velocity bcs.
    do i = 1, this%bcs_vel%size()

       ! We need to treat mixed bcs separately because they are by convention 
       ! marked weak and currently contain nested bcs, some of which are strong.
       select type (vel_bc => this%bcs_vel%items(i)%ptr)
       type is (symmetry_t)
          ! Symmetry has 3 internal bcs, but only one acutally contains
          ! markings. All 3 are zero_dirichlet, so the value is
          ! appropriate to use for the the du,dv,dw,vel_res.
          ! The apply_scalar doesn't do anything, so we need to add
          ! individual nested bcs to the du,dv,dw, whereas the vel_res can
          ! just get symmetry as a whole, because on this list we call
          ! apply_vector.
          write(*,*) "MARKING SYMMETRY IN VELOCITY BLISTS"
          call this%bclst_vel_res%append(vel_bc)
          call this%bclst_du%append(vel_bc%bc_x)
          call this%bclst_dv%append(vel_bc%bc_y)
          call this%bclst_dw%append(vel_bc%bc_z)
```

#### `zero_dirichlet_t`

This is one of the few actual changes to a boundary condition. I renamed the no-slip wall bc to `zero_dirichlet_t` and use that type for pretty much all auxillary helper bcs like `bc_vel_res` and similar. I think it is a very good way of signalling the intention of a component by the type name.

#### `field_dirichlet_t` changes

I had to make it so that the `update`, i.e. the user routine, runs automatically when `apply_scalar/vector` is called. Plus, since the bc is always applied on a single zone, there was no more reason to send a bc list to the user: one can just send one bc.

I also had to remove the mechanism where dummy fields were allocated by component, so e.g. a field for `w` would only be allocated if there was a `d_vel_w` in the bc label. Now the bc always allocates 3 fields. Which I think is OK to be honest. What is *worse* is that if you have several `user_velocity` conditions, that will create a corresponding amount of `field_dirichlet_t` objects, each creating the 3 fields. So this doesn't scale well at all, so should lump all the zones indices where one want sthe field_dirichlet into a single bc specification in the JSON.

THANKS if you made it this far :D, and sorry about the wall of text!


timfelle (commented):
Brilliant work. It is a really needed update at this point.

A  note for the whole strong vs weak thing.
Why not signify `weak` with a variable as well instead of infering that `not strong` means weak?
So you would have a variable, not signifying "is the bc strong or weak" but two each signifying "does this bc modify the strong and weak forms".

In regards to the list, I think we could have two generics of `bc_list`, `apply_stong` and `apply_weak` to make it completely clear at a glance when what is called, instead of someone just writing ` call bc_list%apply_vector(f_x, f_y, f_z, n, t, tstep, .false.)` and then we need to remember what did the logical signify, was it strong or weak? 

A note on inputs for apply.
How about getting rid of `get_device_ptr` from the bc_list?
We have everything nicely organised in the `field_t` and we are essentially always operating on that anyway. And, implicitly calling device code here, when we attempt to do so explicitly most other places feel a bit off.
This was actually the plan with the PR you asked me to wait with, to introduce the interface and slowly go through and replacing the calls before removing the old interface. But i guess we could consider just doing it here instead.

MartinKarp:
Yeah, really great Timofey! I will start checking the hsmg and mixed bcs part :)

src/bc/bc.f90
```diff
@@ -273,83 +411,10 @@ subroutine bc_mark_zone(this, bc_zone)
...
-  subroutine bc_mark_zones_from_list(this, bc_zones, bc_key, bc_labels)
-    class(bc_t), intent(inout) :: this
-    class(facet_zone_t), intent(in) :: bc_zones(:)
-    character(len=*) :: bc_key
-    character(len=100), allocatable :: split_key(:)
-    character(len=NEKO_MSH_MAX_ZLBL_LEN) :: bc_labels(:)
-    integer :: i, j, k, l, msh_bc_type
-
-    msh_bc_type = 0
-    if(trim(bc_key) .eq. 'o' .or. trim(bc_key) .eq. 'on' &
```
MartinKarp, line 291: This logic is currently needed for the fast-diagnaolization method in hsmg. The best solution for this would be to proabably pass down strong and weak bcs to the hsmg, schwarz, fdm combo directly to avoid having this flag in the mesh directly. Currently I realize this solution is also a bit hardcoded for only the fluid solve.

I can have a look at it, but I guess the timeline would not be right now.

examples/cylinder/cylinder.case
```diff
@@ -37,6 +37,29 @@
...
+            "value": [1, 0, 0],
+            "zone_index": 1
+        },
+        {
+            "type": "outflow+dong",
+            "zone_index": 2
+        },
+        {
+            "type": "normal_outflow",
+            "zone_index": 5
```
MartinKarp, line 52: Would it be easy to make it possible to mark several zonesin one line?

MartinKarp (commented):
Just some first few comments.

timofeymukha:
> Yeah, really great Timofey! I will start checking the hsmg and mixed bcs part :) It is only hsmg that differs and the solver works correctly with jacobi?

Thank you! Yes, as far as I could tell the jacobi always gave the same residuals.

examples/cylinder/cylinder.case
timofeymukha: I think it would. Essentially, the factory would loop over the indices and mark them.

src/bc/bc.f90
timofeymukha: Oh, ok! I guess I can add this in the setup_bc_ for the fluid. So the distinction here is basically a strong pressure bc gets a 1 and all the rest gets a 2, right?

src/fluid/fluid_pnpn_bc_fctry.f90
```diff
@@ -0,0 +1,195 @@
...
+          obj%update => user%user_dirichlet_update
+          call json%add("field_name", scheme%p%name)
+       end select
+    else
+      do i=1, size(FLUID_PNPN_KNOWN_BCS)
+         if (trim(type) .eq. trim(FLUID_PNPN_KNOWN_BCS(i))) return
+      end do
+      call neko_type_error("fluid_pnpn boundary conditions", type, &
+           FLUID_PNPN_KNOWN_BCS)
+    end if
```
timfelle, line 104: Since the if block is this large, we might be better off with a 'select case'.

```suggestion
    select case (trim(type))
      case ("outflow", "normal_outflow")
       allocate(zero_dirichlet_t::object)
      case ("outflow+dong", "normal_outflow+dong")
       allocate(dong_outflow_t::object)
      case ("user_pressure")
       allocate(field_dirichlet_t::object)
       select type(obj => object)
         type is(field_dirichlet_t)
          obj%update => user%user_dirichlet_update
          call json%add("field_name", scheme%p%name)
       end select
      case default
       do i = 1, size(FLUID_PNPN_KNOWN_BCS)
          if (trim(type) .eq. trim(FLUID_PNPN_KNOWN_BCS(i))) return
       end do
       call neko_type_error("fluid_pnpn boundary conditions", type, &
            FLUID_PNPN_KNOWN_BCS)
    end select
```

src/fluid/fluid_pnpn_bc_fctry.f90
```diff
@@ -0,0 +1,195 @@
...
+       type is(usr_inflow_t)
+          call obj%set_eval(user%fluid_user_if)
+       end select
+    else
+      do i=1, size(FLUID_PNPN_KNOWN_BCS)
+         if (trim(type) .eq. trim(FLUID_PNPN_KNOWN_BCS(i))) return
+      end do
+      call neko_type_error("fluid_pnpn boundary conditions", type, &
+           FLUID_PNPN_KNOWN_BCS)
+    end if
```
timfelle, line 176: Same goes here, the 'select case' block is easier to read, especially since we
only trim once.

```suggestion
    select case (trim(type))
      case ("symmetry")
       allocate(symmetry_t::object)
      case ("velocity_value")
       allocate(inflow_t::object)
      case ("no_slip")
       allocate(zero_dirichlet_t::object)
      case ("normal_outflow", "normal_outflow+dong")
       allocate(non_normal_t::object)
      case ("blasius_profile")
       allocate(blasius_t::object)
      case ("shear_stress")
       allocate(shear_stress_t::object)
      case ("wall_model")
       allocate(wall_model_bc_t::object)
       ! Kind of hack, but maybe OK? The thing is, we need the nu for
       ! initing the wall model, and forcing the user duplicate that there
       ! would be a nightmare.
       call json%add("nu", scheme%mu / scheme%rho)
      case ("user_velocity")
       allocate(field_dirichlet_vector_t::object)
       select type(obj => object)
         type is(field_dirichlet_vector_t)
          obj%update => user%user_dirichlet_update
       end select
      case ("user_velocity_pointwise")
       allocate(usr_inflow_t::object)
       select type(obj => object)
         type is(usr_inflow_t)
          call obj%set_eval(user%fluid_user_if)
       end select
      case default
       do i = 1, size(FLUID_PNPN_KNOWN_BCS)
          if (trim(type) .eq. trim(FLUID_PNPN_KNOWN_BCS(i))) return
       end do
       call neko_type_error("fluid_pnpn boundary conditions", type, &
            FLUID_PNPN_KNOWN_BCS)
    end select
```

src/fluid/fluid_pnpn_bc_fctry.f90
```diff
@@ -0,0 +1,195 @@
...
+
+    call json_get(json, "zone_index", zone_index)
+    call object%init(coef, json)
+    call object%mark_zone(coef%msh%labeled_zones(zone_index))
+    call object%finalize()
+
+    ! Exclude these two because they are bcs for the residual, not velocity
+    if (type .ne. "normal_outflow" .and. type .ne. "normal_outflow+dong") then
```
timfelle, line 184: Missing some trimming.

```suggestion
    if (trim(type) .ne. "normal_outflow" .and. &
         trim(type) .ne. "normal_outflow+dong") then
```

src/fluid/fluid_pnpn.f90
```diff
@@ -135,12 +188,50 @@ module fluid_pnpn
...
+     !> Constructor.
      procedure, pass(this) :: init => fluid_pnpn_init
+     !> Destructor.
      procedure, pass(this) :: free => fluid_pnpn_free
+     !> Perform a single time-step of the scheme.
      procedure, pass(this) :: step => fluid_pnpn_step
+     !> Restart from a previous solution.
      procedure, pass(this) :: restart => fluid_pnpn_restart
+     !> Set up boundary conditions.
+     procedure, pass(this) :: pnpn_setup_bcs => fluid_pnpn_setup_bcs
```
timfelle, line 200: How about lifting the bc setup to a deferred operator on the scheme itself?
Would all schemes not need to know how to deal with boundary conditions?

src/fluid/fluid_pnpn.f90
```diff
@@ -836,7 +895,180 @@ subroutine fluid_pnpn_step(this, t, tstep, dt, ext_bdf, dt_controller)
...
+             type is(non_normal_t)
+                ! This is a bc for the residuals and increments, not the
+                ! velocity itself. So, don't append to bcs_vel
+                call this%bclst_vel_res%append(bc_i)
+                call this%bc_du%mark_facets(bc_i%bc_x%marked_facet)
+                call this%bc_dv%mark_facets(bc_i%bc_y%marked_facet)
+                call this%bc_dw%mark_facets(bc_i%bc_z%marked_facet)
+             type is (shear_stress_t)
+                ! Same as symmetry
+                write(*,*) "MARKING SHEAR_STRESS IN VELOCITY BLISTS"
```
timfelle, line 982: Could we make these write statements part of the log. Possibly with the debug
level?

src/fluid/fluid_scheme.f90
```diff
@@ -1097,97 +776,7 @@ subroutine fluid_scheme_set_bc_type_output(this, params)
       call this%bdry%init(this%dm_Xh, 'bdry')
```
timfelle, line 776: Maybe there is no need to have this in a separate subroutine. Probably could be
with the other IO options.

timofeymukha:
Dear all, an update from today.

- Each boundary can now take several zones, so zone_index is renamed to zone_indices, which is an array.
- I have implemented the checks on zone_indices, that we discussed, as @Mr-2c suggested.
- I have updated all the example case files.
- I have 100% residual agreement on all the examples except
  - immersed_bunny
  - turb_pipe (this is really a surprise)

  This has to be investigated further.

- I removed most of the extra debug output.


Sadly, documentation and the output_boudary field handling I could not fit, so it is shifted to next week.

timofeymukha:
Removing "don't merge"!!

Added documentation, updated examples, and redon the functionality for `output_boundary`. The numbering is a little bit changed in the latter: periodic bcs are "marked" with 0 (by doing nothing), and outflow had to be lumped with normal_outflow because it is impossible to distinguish them from the instantiated types at the moment.

I have addressed almost all of your comments. A few things from @timfelle are missing, which I am postponing to next PRs.
- Potentially adding a "weak" attribute to bc_t to have mixed bcs marked as both strong and weak.
- Debug-level log output in bc setup.
- Make the setup_bc defferred in scalar_scheme (done in the fluid_scheme).

src/krylov/pc_hsmg.f90
```diff
@@ -367,6 +369,7 @@ subroutine hsmg_solve(this, z, r, n)
     type(ksp_monitor_t) :: crs_info
     integer :: thrdid, nthrds
```
timfelle, line 371: Why this extra empty line?

src/multigrid/phmg.f90
```diff
@@ -149,10 +150,11 @@ subroutine phmg_init(this, msh, Xh, coef, dof, gs_h, bclst)
...
 
        call this%phmg_hrchy%lvl(i)%bc%init_base(this%phmg_hrchy%lvl(i)%coef)
-       if (bclst%size .gt. 0 ) then
-          do j = 1, bclst%size
+       if (bclst%size() .gt. 0 ) then
+          do j = 1, bclst%size()
+             bc_j => bclst%get(j)
              call this%phmg_hrchy%lvl(i)%bc%mark_facets(&
-                  bclst%items(j)%ptr%marked_facet)
+                  bc_j%marked_facet)
```
timfelle, line 157: No need to break this line anymore.

```suggestion
             call this%phmg_hrchy%lvl(i)%bc%mark_facets(bc_j%marked_facet)
```

src/scalar/scalar_pnpn.f90
```diff
@@ -30,15 +30,15 @@
 ! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 ! POSSIBILITY OF SUCH DAMAGE.
 !
-!> Containts the scalar_pnpn_t type.
+!> Contains the `scalar_pnpn_t` type.
 
 module scalar_pnpn
+  use comm
```
timfelle, line 36: Missing `only`

src/scalar/scalar_pnpn.f90
```diff
@@ -428,5 +440,105 @@ subroutine scalar_pnpn_step(this, t, tstep, dt, ext_bdf, dt_controller)
...
+       write(log_buf,'(A,A,E15.7,A,E15.7,A,E15.7)') 'Scalar debug', &
+            ' l2norm s', glsc2(this%s%x, this%s%x, n), &
+            ' slag1', glsc2(this%slag%lf(1)%x, this%slag%lf(1)%x, n), &
+            ' slag2', glsc2(this%slag%lf(2)%x, this%slag%lf(2)%x, n)
+       call neko_log%message(log_buf)
+       write(log_buf,'(A,A,E15.7,A,E15.7)') 'Scalar debug2', &
+            ' l2norm abx1', glsc2(this%abx1%x, this%abx1%x, n), &
+            ' abx2', glsc2(this%abx2%x, this%abx2%x, n)
+       call neko_log%message(log_buf)
+    end if
```
timfelle, line 460: This can be done by adding the `NEKO_LOG_DEBUG` to the `message` commands
instead. No need to manually do the if statement

src/case.f90
```diff
@@ -170,10 +170,10 @@ subroutine case_init_common(this)
        call neko_log%section('Load Balancing')
        call parmetis_partmeshkway(this%msh, parts)
        call redist_mesh(this%msh, parts)
-       
+
        ! store the balanced mesh (for e.g. restarts)
        string_val = trim(string_val(1:scan(trim(string_val), &
-            '.', back=.true.) - 1))//'_lb.nmsh'
+            '.', back = .true.) - 1))//'_lb.nmsh'
```
timfelle, line 176:
```suggestion
            '.', back = .true.) - 1)) // '_lb.nmsh'
```

timfelle (approved):

MartinKarp (changes requested):
I think the field_update should be made sure to be called on all ranks, and preferrably not called too many times per time step, but otherwise I think it looks really great. I guess I will have a look at the update function later as well, but I think we should try to emulate the current functionality for now.

src/fluid/fluid_pnpn.f90
```diff
@@ -836,7 +857,351 @@ subroutine fluid_pnpn_step(this, t, tstep, dt, ext_bdf, dt_controller)
...
+    integer :: temp_index, i
+    class(bc_t), pointer :: bci
+    character(len=LOG_SIZE) :: log_buf
+
+    call neko_log%section("Fuid boundary conditions")
+    write(log_buf, '(A)') 'Marking using integer keys in bdry0.f00000'
+    call neko_log%message(log_buf)
+    write(log_buf, '(A)') 'Condition-value pairs: '
+    call neko_log%message(log_buf)
+    write(log_buf, '(A)') '  periodic                        = 0'
```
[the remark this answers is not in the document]
timofeymukha: There is not such thing as a zone without a bc anymore, you have to assign all zones a bc, or the solver will crash with an appropriate error. Periodic is the only exception.

But still perhaps I can grab from the mesh as you say.

src/scalar/scalar_pnpn.f90
timofeymukha: I think `comm` is one of those exceptions we have.

src/bc/symmetry.f90
```diff
@@ -131,62 +134,119 @@ subroutine symmetry_init(this, coef)
...
+  !! @details The normal direction is the one for which s is 0, so e.g
+  !! for a y normal, sx and sz will be unity, and sy will be 0.
+  subroutine symmetry_get_normal_axis(this, sx, sy, sz, facet, el)
+    class(symmetry_t), target, intent(inout) :: this
+    real(kind=rp), intent(out) :: sx, sy, sz
+    integer, intent(in) :: facet
+    integer, intent(in) :: el
+    integer :: i, m, j, l
+    type(tuple_i4_t), pointer :: bfp(:)
+    real(kind=rp), parameter :: TOL = 1d-3
```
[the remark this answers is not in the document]
timofeymukha: I add a TODO

src/bc/bc.f90
```diff
@@ -33,7 +33,7 @@
 !> Defines a boundary condition
 module bc
   use neko_config
-  use num_types
+  use num_types, only : rp
   use device
```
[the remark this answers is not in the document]
timofeymukha: `device` and `comm` are exceptions, right?

src/bc/symmetry.f90
```diff
@@ -38,44 +38,78 @@ module symmetry
...
+
+  !> Finalize.
+  !! Marks the appropriate faces for application of a homogeneous Dirchlet
+  !! condition based on the direction of the axis normal.
+  subroutine symmetry_finalize(this)
+    class(symmetry_t), target, intent(inout) :: this
+    integer :: i, m, j, l
     type(tuple_i4_t), pointer :: bfp(:)
     real(kind=rp) :: sx, sy, sz
     real(kind=rp), parameter :: TOL = 1d-3
```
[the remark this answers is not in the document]
timofeymukha: TOL is used!

src/bc/field_dirichlet.f90
```diff
@@ -143,14 +168,20 @@ end subroutine field_dirichlet_free
...
     real(kind=rp), intent(inout),  dimension(n) :: x
     real(kind=rp), intent(in), optional :: t
     integer, intent(in), optional :: tstep
+    logical, intent(in), optional :: strong
+    logical :: strong_ = .true.
+
+    if (present(strong)) strong_ = strong
 
-    if (this%msk(0) .gt. 0) then
+    if (strong_ .and. this%msk(0) .gt. 0) then
```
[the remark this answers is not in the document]
timofeymukha: Right! So, I just leave if(strong_), which is consistent across ranks.

src/scalar/scalar_pnpn.f90
timfelle: Yes `comm` was an exception to it. I have just never been able to reproduce any issues with it.

src/bc/bc.f90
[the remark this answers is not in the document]
timfelle: I don't think `device` was an exception no.

src/bc/symmetry.f90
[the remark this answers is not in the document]
timfelle: BTW, this tolerance is not used. The other one is. And please specify the precision `1.0e-3_rp`

src/bc/symmetry.f90
```diff
@@ -131,62 +134,119 @@ subroutine symmetry_init(this, coef)
...
+    integer, intent(in) :: facet
+    integer, intent(in) :: el
+    integer :: i, m, j, l
+    type(tuple_i4_t), pointer :: bfp(:)
+    real(kind=rp), parameter :: TOL = 1d-3
+    type(tuple_i4_t) :: bc_facet
+
+    associate(c => this%coef, nx => this%coef%nx, ny => this%coef%ny, &
+              nz => this%coef%nz)
+      sx = 0d0
```
timfelle, line 157:
```suggestion
      sx = 0.0_rp
```

src/bc/blasius.f90
```diff
@@ -58,12 +62,79 @@ module blasius
      procedure, pass(this) :: apply_scalar_dev => blasius_apply_scalar_dev
      procedure, pass(this) :: apply_vector_dev => blasius_apply_vector_dev
      procedure, pass(this) :: set_params => blasius_set_params
+     !> Constructor
+     procedure, pass(this) :: init => blasius_init
+     !> Constructor from components
+     procedure, pass(this) :: init_from_components => &
```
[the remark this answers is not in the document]
timfelle: Sure, I just checked our coding guidelines. There it also states `init_from_components` should be an initializer and we should have a generic `init` which chooses the json or attributes. Let's go back and fix this across other parts of the code in a new PR.

src/bc/symmetry.f90
```diff
@@ -38,44 +38,78 @@ module symmetry
...
-    type(coef_t), target, intent(in) :: coef
-    integer :: i, j, l
+    call this%init_from_components(coef)
+
+  end subroutine symmetry_init
+
+  !> Constructor from components
+  !! @param[in] coef The SEM coefficients.
+  subroutine symmetry_init_from_components(this, coef)
+    class(symmetry_t), intent(inout), target :: this
```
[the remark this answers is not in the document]
timfelle: my understanding is that if you do a `=>` to something, it must be a target. But we don't in this function. Question is whether it also is required if any of the subroutines downstream does the pointer assignment @njansson?

src/bc/field_dirichlet.f90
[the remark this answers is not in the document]
timofeymukha: I think 2 is really a very tiny corner case. I imagine that if s on the boundary depends on u on the boundary (which is already a pretty unusual thing, or?), it is the value at i+1 of u that you want typically, that is, to be synced in time.

Regarding 1, is the problem with some rank not calling the update function that if one putts some MPI-syncing code in it, it will get stuck?
I am not a strong proponent of that if statement for the bc_list, probably better to just consistently do things for zero-sized bcs. On the other hand, I think it would be very nice if the update can be inside the bc code like right now, unless we really cannot avoid splitting it, as we had before.

src/bc/bc.f90
```diff
@@ -32,9 +32,10 @@
...
 !> Defines a boundary condition
 module bc
-  use neko_config
-  use num_types
-  use device
+  use neko_config, only : NEKO_BCKND_DEVICE
+  use num_types, only : rp
+  use device, only : device_get_ptr, HOST_TO_DEVICE, c_associated, &
+       device_memcpy, device_free, device_map
   use dofmap, only : dofmap_t
```
timofeymukha, line 39: @timfelle I think I am getting one of those strage errors now: c_associated is in the `use`,  but `nvfortran` is complaining.

timfelle: `c_associated` is not part of `device`. It is part of the `iso_c_binding`. So add 
`use iso_c_binding, only: c_associated`

timfelle: This is exactly the kind of implicit inclusion we can avoid by always using only, everywhere.

src/bc/field_dirichlet.f90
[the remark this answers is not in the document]
MartinKarp: Yeah lets not think about 2 for now as I dont think it is in the scope of the PR really and I am not sure exactly what one wants.

I think another issue with allowing calls to bcs of length 0 might be similar to when we call device subroutines with length 0 and we get segfaults. If that doesnt happen we can just remove the check and do like we do here. The I think we can be happy.

In the end what I want to support is doing things like I did in the example for interpolation: https://github.com/ExtremeFLOW/neko/blob/2bb71482a3cf2392c80f9a12ac0f8ca24d8dac27/examples/interpolation/recycling.f90

If that is all good I am a happy camper

src/bc/bc.f90
timofeymukha: Oooh, ok. Yes, I totally agree. As much `only` as bad compilers allow!

src/bc/field_dirichlet.f90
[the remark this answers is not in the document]
timofeymukha: OK. Removing the statement on the CPU seems to work. I guess we need to check on a GPU somewhere.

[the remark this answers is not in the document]
timofeymukha: I pushed a version with the `if` commented out. I guess it needs to be tested on a reasonably big case so that one has multiple ranks with the GPU. @MartinKarp can you make such a test run? I don't really have good access to a machine right now, although I could perhaps try to get onto Dardel.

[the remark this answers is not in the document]
MartinKarp: https://stackoverflow.com/questions/19022725/running-zero-blocks-in-cuda

This is more or less the issue we have. I have made some functios where we check the size, but in most places we will get an error if we run device kernels of size 0.

My feeling for now is to make each bc in apply device check the size and act correctly.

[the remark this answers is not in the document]
timofeymukha: OK, so just not launch anything and return if the size is 0 for device_apply routines?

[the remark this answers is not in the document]
timofeymukha: Hmm, but for the field_dirichlet we cannot do that, right? Basically due to original reason why you added the comment. There could be both a kernel launch and some MPI communication in the user routine, and for the sake of the latter, we always have to launch the `update`, right? So I guess it will be up to the user to only launch kernels if the size is > 0?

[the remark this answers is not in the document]
timofeymukha: A reasonable solution is then maybe to add and `ignore_empty` argument to `bc_list.append`, defaulting to `.true.`, so everything is retained as is for most of the bcs. And for the `field_dirichlet` and `field_dirichlet_vector` we force-add it, and then the user code should be careful with kernel launches (which we will add to the already extensive documentation).

src/bc/bc.f90
```diff
@@ -76,49 +84,89 @@ module bc
...
+  end type
+
+
+  abstract interface
+     !> Constructor
+     subroutine bc_constructor(this, coef, json)
+       import :: bc_t, coef_t, json_file
+       class(bc_t), intent(inout), target :: this
+       type(coef_t), intent(in) :: coef
+       type(json_file), intent(inout) ::json
```
timfelle, line 134:
```suggestion
       type(json_file), intent(inout) :: json
```

timofeymukha:
As an update after the discussion with @MartinKarp (see also dialog above).
- The bc_list will add bcs with 0 mask size. So all ranks have the same bcs.
- In device_apply_* routines, kernels are only launched id this%mask > 0, because otherwise the kernel may crash.
- In the field_dirichlet*, the `update` will be called even with 0 mask in case the user added MPI calls. But the masked copy kernel will not be launched if the mask is 0.

MartinKarp (approved):

timofeymukha:
@njansson Merge? :-)

njansson (approved):

njansson:
> @njansson Merge? :-)

Go! 🍶

src/bc/wall_model_bc.f90
```diff
@@ -52,113 +53,152 @@ module wall_model_bc
     type, public, extends(shear_stress_t) :: wall_model_bc_t
        !> The wall model to compute the stress.
        class(wall_model_t), allocatable :: wall_model
+       !> The kinematic viscosity.
+       real(kind=rp) :: nu
+       !> Parameter dictionary for the wall model
+       type(json_file) params_
```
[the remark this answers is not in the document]
timofeymukha: Will address!

timfelle (approved):

PR #1651, opened 2024-12-20
Enhance bc_list functionality and code cleanup

timfelle:
Improve the safety and performance of the bc_list application routines, streamline variable usage, and ensure proper memory management. Additionally, implement range checks and utilize getter methods for better encapsulation.


src/scalar/scalar_pnpn.f90
```diff
@@ -476,14 +479,12 @@ subroutine scalar_pnpn_setup_bcs_(this, user)
...
           ! Create a new json containing just the subdict for this bc
           call json_extract_item(core, bc_object, i, bc_subdict)
-
-          call bc_factory(this%bcs%items(i)%ptr, this, bc_subdict, &
-                          this%c_Xh, user)
-
-          this%bcs%size_ = this%bcs%size_ + 1
+          bc_i => null()
+          call bc_factory(bc_i, this, bc_subdict, this%c_Xh, user)
+          call this%bcs%append(bc_i)
```
timofeymukha, line 487: You checked that this works, right? :D Even though we are pointing to a temporary in the `append`?
`this%items(this%size_)%ptr => bc`

I am always scared of this "in Fortran a pointer is not just an address like in C".

src/bc/bc_list.f90
```diff
@@ -201,7 +227,74 @@ subroutine bc_list_apply_vector(this, x, y, z, n, t, tstep, strong)
...
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       do i = 1, this%size_
+          call this%items(i)%ptr%apply_vector_dev(x%x_d, y%x_d, z%x_d, t, tstep)
+       end do
+    else
+       do i = 1, this%size_
+          call this%items(i)%ptr%apply_vector(x%x, y%x, z%x, n, t, tstep, &
+               strong)
+       end do
+    end if
```
timofeymukha, line 295: I remember now that I added something similar on the level of one bc in `bc_t` but never actually used it.
```fortran
    !> Apply the boundary condition to a scalar field. Dispatches to the CPU
     !! or the device version.
     procedure, pass(this) :: apply_scalar_generic => bc_apply_scalar_generic
     !> Apply the boundary condition to a vector field. Dispatches to the CPU
     !! or the device version.
     procedure, pass(this) :: apply_vector_generic => bc_apply_vector_generic
```
Maybe one can make use of those here, or maybe those can just be removed...

src/scalar/scalar_pnpn.f90
timfelle: Well I have not run a lot of examples to test it. TGV works 😅 And only with my local compiler.

src/bc/bc_list.f90
timfelle: Yea I think it should be at the bc level instead of the list level.
But I still want to get rid of the `get_device_ptr` call.

src/scalar/scalar_pnpn.f90
timofeymukha: Hemi seems to work. So hopefully we are good. I would say we merge this in, and then I will merge my with develop, so that compilation checks are run on the PR.

PR #1658, opened 2025-01-10
Remove unlabeled zones in examples

timofeymukha:
Concerns 3 examples

- The lid. I used the provided .box files to regenerate the meshes and rea2nbin them to get labeled zones.
- The pipe example.
  - Renamed to turb_pipe, with _, for consistency with turb_channel.
  - I regenerated the mesh from Saleh's .geo trying to match the paramters of the existing mesh as best I could. It is still not entirely identical, but close.
  - Added mesh generation to the README
  - *I could not use the gmsh2nek we have in the neko repo*, for some reason that crashed. I used one directly from Nek5000.
- RBC
  - Renamed to rayleigh_benard_cylinder, so _ instead of - for consistency.
  - New mesh, close to the original one. 
  - @adperezm is the "v" zone on the top of the cylinder actually a wall for velocity?

Also renamed rayliegh-benard to use _.

I am currently running the pipe and RBC to see that things look reasonable.


njansson (approved):

timofeymukha:
The pipe looks okay, i.e. low Re turbulence.

This is the temperature for the RBC at 60% of the simulation time. Is this ok :-)?
image

MartinKarp (approved):

MartinKarp:
Nice, I guess you need to fix make dist before merging!

njansson:
The `EXTRA_DIST` section in `Makefile.am` needs some love to pass the CI 🍻

timofeymukha:
Maybe it's a good time for me to ask what the `dist` stuff does, actually :-).

PR #1661, opened 2025-01-14
Added example and compilation instructions for Insitu

adperezm:
Hi,

I added a new variable to makeneko to include the ADIOS files during compilation. I’m not sure if this is the best approach. In the future we might have streaming as a simulation component, so it would be good if that file is taken into account during the linking of Neko itself (Maybe it already is). I’m happy to adjust this if needed—this is just one possible solution.

I’ve also added an example that demonstrates starting a simulation in Neko, transferring data to Python, and writing from Python. Do you think there’s a better way to name this example? I avoided integrating it directly into the lid example because requiring ADIOS to run it might be overkill.


examples/lid_insitu/README.md
```diff
@@ -0,0 +1,6 @@
+## Simulation of a lid-driven cavity
+In this case we simulate a lid-driven cavity with smoothened belt velocity to fulfil the continuity equaiton in the corners. There is a 3D example (with 3 elements in the spanwise $z$ direction) and a pseudo-2D example (with a 2D mesh that then generates a case with one element in the spanwise direction).
+
+The `lid.box` file needs to be converted to a .re2 file using the Nek5000 tool genbox, and then using rea2nbin into the `lid.nmsh` Neko mesh file.
```
MartinKarp, line 4: Looks good! Would it be possible to use genmeshbox instead? To avoid using rea2nbin.

Would it also be possible to provide a description of what the insitu part does?

examples/lid_insitu/run.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/bash
```
MartinKarp, line 1: Should this be here?

adperezm:
Hi, I agree with your comments. I will proceed with them.

adperezm:
This one is not needed, as #1837 does the same.

Issue #1623, opened 2024-12-02
LES eddy viscosities and wall model shear stresses must be computed based on velocity field at time step n+1 (the unknown one)


Mr-2c:
to the best of my understanding of the Pn-Pn method at the moment, we have
$\frac{\partial u_i}{\partial t}\vert_{n+1} = RHS_{n+1} $
which means all terms on the right hand side must be evaluated for t_{n+1} which is the velocity field we are solving for (that is why the convective term is extrapolated for instance). 
what this means for eddy viscosity (and wall models) is that, at least, the velocity must be first extrapolated to time n+1 (similar to and with the same method as how the convective term is extrapolated) and then the extrapolated velocity is the one that should be used for computing the eddy viscosity (and the wall shear stress if there is a wall model).

Shiyu-Sandy-Du:
From the implementation point of view, if it's the case (seems to be), then the LES simcomp should be isolated out from other simcomps as a pre-time-stepping simcomp.

Mr-2c:
the problem is that you never have the velocity at n+1 so you have to always strapolate, so you can still call it with others but make sure that instead of u,v,w, from registry the extrapolated velocity is computed and given to the compute subroutines.

Shiyu-Sandy-Du:
Isn't it calculated in pnpn already? Maybe there is no such field for it, but the computation has been done once, isn't it?

timofeymukha:
Yeah, isn't that what sumab does? (a nek5000 name..) 

```
      call sumab%compute_fluid(u_e, v_e, w_e, u, v, w, &
           ulag, vlag, wlag, ext_bdf%advection_coeffs, ext_bdf%nadv)

```

I guess methodologically you are right, Siavash. But not sure if it makes any tangible difference for the wall models at least.

Mr-2c:
not really sure what sumab does... where is it called?
I agree that probably it does not increase the tangible accuracy but the current implementation is only first order in time, and the velocity is not fully divergence free since the right hand side of the pressure Poisson equation does not match the velocity at n+1

timofeymukha:
It's in the fluid_pnpn's step. You get the u_e, v_e, w_e arrays, which are then used in the pressure residual to compute the "curl of the curl" term.

timofeymukha:
See rhs_maker_cpu.f90 for the implemenation.

Shiyu-Sandy-Du:
[...]

PR #1663, opened 2025-01-15
Fix/les evaluation

Shiyu-Sandy-Du:
This PR is regarding to Issue #1623 from @Mr-2c. One might argue the extrapolation is not accurate neither but discussion are very welcome!


Shiyu-Sandy-Du:
Noting the field for nut computation should be interpolated before the time stepping, I will issue another PR ...

PR #1666, opened 2025-01-19
Compressible solver

tuananhdao:
## Overview
This PR implements a **minimal** compressible solver in Neko. The solver:
- Solves the compressible Euler equations
- Is stabilized using vanishing artificial viscosity with a simple (but thermodynamically correct) viscous flux https://doi.org/10.1016/j.physa.2018.03.097. Now the viscosity is first-order and scaled with h.
- Explicit timestepping, given a Butcher's tableau, implemented for RK1 (forward Euler), RK2 (Heun's method), SSPRK3, and the regular RK4.
- CPU, HIP, CUDA

An example case file:
```
{
"version": 1.0,
"case": {
    "mesh_file": "box.nmsh",
    "output_at_end": true,
    "output_boundary": false,
    "output_checkpoints": false,
    "end_time": 0.2,
    "timestep": 0.0001,
    "numerics": {
        "time_order": 3,
        "polynomial_order": 3,
        "c_avisc_low": 1.0
    },
    "fluid": {
        "scheme": "compressible",
        "gamma": 1.4,
        "initial_condition": {
            "type": "user"
        },
        "output_control": "nsamples",
        "output_value": 100
    }
}
}
```
`c_avisc_low` is the first-order artificial viscosity scale. The value `1.0` should work most of the time unless one wants to play around. `gamma` is the gas constant "the specific heat ratio of a gas".

TODO List:
- [x] Separation of `time_scheme_controller_t` from the incompressible modules
- [x] Unit tests (RK only)

Missing essential features that will be covered in the next PR(s):
- Boundary conditions, after #1650 
- Fast estimation from above of the maximum wave speed (https://doi.org/10.1016/j.jcp.2016.05.054) to get correct time step size and guaranteed bound-preserving artificial viscosity. Now the time steps are fixed and the first-order viscosity is not scaled with wave speed.
- High-order artificial viscosity, for example https://doi.org/10.1016/j.jcp.2024.113146. This kind of methods doesn't lower high-order of smooth solutions.
- Restart

### Smooth wave propagation
`examples/euler_2d_smooth`

It's a known analytical solution of the compressible Euler equations. We will use it to measure accuracy.

### Sod Tube Problem
`examples/euler_1d_sod`

It's a common test for Euler solvers to see if shocks/discontinuities are captured correctly.

### Compressible Taylor-Green Vortex
`examples/euler_tgv`

We have one for the incompressible solver and here's the one for the compressible solver.

## Unit tests
- **Unit tests should be implemented before merged**. I want to wait until I get some comments on the new structure. It's highly probable that I will need to change the current structure. Then I will write unit tests.

## Tested on GPU clusters

Running CUDA on 4xGH200

```
    _  __  ____  __ __  ____
   / |/ / / __/ / //_/ / __ \
  /    / / _/  / ,<   / /_/ /
 /_/|_/ /___/ /_/|_|  \____/

 (version: 0.9.99)
 (build: 2025-01-16 on aarch64-unknown-linux-gnu using gnu)


 -------Job Information--------
 Start time: 20:30 / 2025-01-16
 Running on: 4 MPI ranks
 CPU type  : ARM Neoverse V2
 Bcknd type: Accelerator (CUDA)
 Dev. name : NVIDIA GH200 120GB
 Real type : double precision
```
Running HIP code on 8xMI250X
```
    _  __  ____  __ __  ____
   / |/ / / __/ / //_/ / __ \
  /    / / _/  / ,<   / /_/ /
 /_/|_/ /___/ /_/|_|  \____/

 (version: 0.9.99)
 (build: 2025-01-15 on x86_64-unknown-linux-gnu using cray)


 -------Job Information--------
 Start time: 15:39 / 2025-01-15
 Running on: 8 MPI ranks
 CPU type  : AMD EPYC 7A53 64-Core Processor
 Bcknd type: Accelerator (HIP)
 Dev. name : AMD Instinct MI250X
 Real type : double precision
```


pschlatt1:
Very nice indeed! One comment regarding the time stepping: I would say that using a  RK for this compressible solver is perfectly fine; then you do not have to deal with the issues of multistep methods. As I remember, there was this problem with DAE index-2 for NS, so we use a multistep scheme, but not RK. In your case, without Poisson equation, you are fine to use single-step method with substeps, to using those is better :-)

src/fluid/bcknd/sx/euler_res_sx.f90
```diff
@@ -0,0 +1,30 @@
...
+                coef, gs, h, c_avisc_low, rk_scheme, dt)
+    type(field_t), intent(inout) :: rho_field, m_x, m_y, m_z, E
+    type(field_t), intent(in) :: p, u, v, w, h
+    class(Ax_t), intent(inout) :: Ax
+    type(coef_t), intent(inout) :: coef
+    type(gs_t), intent(inout) :: gs
+    real(kind=rp) :: c_avisc_low
+    class(runge_kutta_time_scheme_t), intent(in) :: rk_scheme
+    real(kind=rp), intent(in) :: dt
+    call neko_error("SX not supported")
```
njansson, line 27: For this case, we can either call the CPU version here (it will work fine), or replicate it to keep some separation of concerns.

(Remember SX will be our RISC-V backend in the future)

njansson:
Hear hear, very nice! 

Regarding the structure, and namings. I think it is good enough as a start, and if needed we can do another refactoring

timofeymukha:
Hi Tuan!

Just a couple of comments / questions.
- For an explicit compressible solver, does it make sense to operate in terms of the residual of the governing equations?
- In any case, please add as many lines of documentation as possible to the CPU kernel of the residual computation, or the eventual equivalent.  This is usually the core of the scheme, and it is *really* useful if it is well-documented. We are somehow still unwrapping what it is we are doing in the incompressible case :-))).
- Add a README.md to examples. Take a critical look at which examples are really need to show off the new functionality. Updating examples is quite a bit of work whenever we have a braking change.
- Add documentation. Particularly to the governing equations page!
- Don't forget @param docstrings for subroutines.
- I think you better rename fluid_scheme to fluid_scheme_incompressible so that it is clear what is what.

Tell me once you merge in develop, and I will actually checkout the branch on my PC and take a closer look!

src/fluid/bcknd/sx/euler_res_sx.f90
tuananhdao: I replicated this part to the SX version. Will test it later.

tuananhdao:
Thanks for the comments!

> * For an explicit compressible solver, does it make sense to operate in terms of the residual of the governing equations?

I thought about this. Quite a big part of the literature refers the right hand side (-div(f)) of conservation laws to as the "residual", for example in residual distribution schemes. I initially named the class "rhs" but I thought I could reuse this class when calculating the residual (in the standard sense) for high-order viscosity. Put both rhs and residual evaluation in `_res_` classes maybe good/bad? We can decide.

> * In any case, please add as many lines of documentation as possible to the CPU kernel of the residual computation, or the eventual equivalent.  This is usually the core of the scheme, and it is _really_ useful if it is well-documented. We are somehow still unwrapping what it is we are doing in the incompressible case :-))).

Good point! I will add more comments.

> * Add a README.md to examples. Take a critical look at which examples are really need to show off the new functionality. Updating examples is quite a bit of work whenever we have a braking change.

Sure! README are surely missing.

> * Add documentation. Particularly to the governing equations page!

Regarding this, the method is not complete as it's missing quite a lot of features as I mentioned, and not yet usable as it's first-order. Maybe I should add this at least after implementing BCs and high-order artificial viscosity?

> * Don't forget @param docstrings for subroutines.

Yes!! This is missing.

> * I think you better rename fluid_scheme to fluid_scheme_incompressible so that it is clear what is what.

I will do that before merging to develop, just to avoid potential conflicts.

> 
> Tell me once you merge in develop, and I will actually checkout the branch on my PC and take a closer look!

I will ping you.

timofeymukha:
You are right, the documentation in the .md files is probably better to be postponed to when we are ready for users to pick up the solver.

njansson (approved):

src/fluid/flow_ic.f90
```diff
@@ -407,11 +442,11 @@ subroutine set_flow_ic_fld(u, v, w, p, file_name, &
 
           if (sample_mesh_idx .eq. -1) &
                call neko_error("Invalid file name for the initial condition. &
-&The file format must be e.g. 'mean0.f00001'")
+               &The file format must be e.g. 'mean0.f00001'")
```
timfelle, line 445: Please don't linebreak a single-line `if` statement. If it is too long, make it an `if`-block.
```suggestion
          if (sample_mesh_idx .eq. -1) then
             call neko_error("Invalid file name for the initial condition. &
                  &The file format must be e.g. 'mean0.f00001'")
          end if
```

src/fluid/fluid_plan1.f90
```diff
@@ -1,14 +1,14 @@
 !> Classic NEKTON formulation
 !! Compute pressure and velocity using consistent approximation spaces.
 module fluid_plan1
-  use fluid_scheme
+  use fluid_scheme_incompressible
```
timfelle, line 4: Please limit imports by adding `only`

src/fluid/fluid_scheme_base.f90
```diff
@@ -0,0 +1,289 @@
...
+       character(len=:), allocatable :: solver_type, precon_type
+       character(len=LOG_SIZE) :: log_buf
+       real(kind=rp) :: GJP_param_a, GJP_param_b
+     end subroutine fluid_base_init_all_intrf
+  end interface
+
+  !> Initialize common data for the current scheme
+  abstract interface
+     subroutine fluid_base_init_common_intrf(this, msh, lx, params, scheme, user, &
+          kspv_init)
```
timfelle, line 155: You added the linebreak, but still break 80 characters.
```suggestion
     subroutine fluid_base_init_common_intrf(this, msh, lx, params, scheme, &
          user, kspv_init)
```

src/fluid/fluid_scheme_compressible.f90
```diff
@@ -0,0 +1,256 @@
...
+contains
+  !> Initialize common data for compressible fluid scheme
+  !> @param this The compressible fluid scheme object
+  !> @param msh Mesh data structure
+  !> @param lx Polynomial order in x-direction
+  !> @param params JSON configuration parameters
+  !> @param scheme Name of the numerical scheme
+  !> @param user User-defined parameters and functions
+  subroutine fluid_scheme_compressible_init(this, msh, lx, params, scheme, user)
+    implicit none
```
timfelle, line 88: We keep `implicit none` statements at a module level.
```suggestion
```

src/fluid/fluid_scheme_compressible.f90
```diff
@@ -0,0 +1,256 @@
...
+  use json_module, only : json_file
+  use logger, only : LOG_SIZE
+  use num_types, only : rp
+  use mesh, only : mesh_t
+  use scratch_registry, only : scratch_registry_t
+  use space, only : space_t, GLL
+  use user_intf, only : user_t
+  use usr_inflow, only : usr_inflow_eval
+  use json_utils, only : json_get_or_default
+
```
timfelle, line 49: Please fill in which parts should be public from this module. I do not expect
you to intend allow others to call `fluid_scheme_compressible_free` directly
from a different module.
```suggestion
    implicit none
    private
    public: 
```

src/fluid/fluid_scheme_base.f90
```diff
@@ -0,0 +1,289 @@
...
+
+  interface
+     !> Initialise a fluid scheme
+     module subroutine fluid_scheme_base_factory(object, type_name)
+       class(fluid_scheme_base_t), intent(inout), allocatable :: object
+       character(len=*) :: type_name
+     end subroutine fluid_scheme_base_factory
+  end interface
+
+  public :: fluid_scheme_base_t, fluid_scheme_base_factory
```
timfelle, line 288: Please keep the public interface list together with the `implicit none` and
`private` statements at the top of the module. This will anyone to find the
information at once, instead of having to dig through the file.

src/fluid/fluid_source_term.f90
```diff
@@ -45,7 +45,7 @@ module fluid_source_term
 
   !> Wrapper contaning and executing the fluid source terms.
   !! @details
-  !! Exists mainly to keep the `fluid_scheme_t` type smaller and also as
+  !! Exists mainly to keep the `fluid_scheme_incompressible_t` type smaller and also as
   !! placeholder for future optimizations.
```
timfelle, line 49:
```suggestion
  !! Exists mainly to keep the `fluid_scheme_incompressible_t` type smaller and
  !! also as placeholder for future optimizations.
```

src/time_schemes/runge_kutta_scheme.f90
```diff
@@ -0,0 +1,179 @@
...
+! POSSIBILITY OF SUCH DAMAGE.
+!
+module runge_kutta_time_scheme
+  use neko_config
+  use num_types, only : rp
+  use time_scheme, only: time_scheme_t
+  use math, only : rzero
+  use utils, only : neko_error
+  use device, only : device_free, device_map, device_memcpy, HOST_TO_DEVICE
+  use, intrinsic :: iso_c_binding
```
timfelle, line 40: Please specify what you import form other modules by specifying it with `only`.

src/time_schemes/runge_kutta_scheme.f90
```diff
@@ -0,0 +1,179 @@
...
+! INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+! BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+! LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+!
+module runge_kutta_time_scheme
+  use neko_config
```
timfelle, line 34: Please specify what you import form other modules by specifying it with `only`.

src/time_schemes/runge_kutta_scheme.f90
```diff
@@ -0,0 +1,179 @@
...
+contains
+  !> Constructor for Runge-Kutta time scheme
+  !> @param this The Runge-Kutta scheme object
+  !> @param order Order of accuracy (1-4), determines coefficients:
+  !>              1: Forward Euler
+  !>              2: Heun's method
+  !>              3: SSPRK3
+  !>              4: Classic RK4
+  subroutine runge_kutta_scheme_coeffs_init(this, order)
+    implicit none
```
timfelle, line 68: `implicit none` should be done at module level

src/case.f90
```diff
@@ -210,22 +210,22 @@ subroutine case_init_common(this)
     !
     ! Set order of timestepper
     !
-    call json_get(this%params, 'case.numerics.time_order', integer_val)
-    call this%ext_bdf%init(integer_val)
+    !  call json_get(this%params, 'case.numerics.time_order', integer_val)
+    !  call this%ext_bdf%init(integer_val)
```
timfelle, line 214: Please delete this instead of leaving it as a commend

src/simulation.f90
```diff
@@ -274,18 +276,18 @@ subroutine simulation_restart(C, t)
...
     do i = 1, size(C%dtlag)
-       call C%ext_bdf%set_coeffs(C%dtlag)
+       call C%fluid%ext_bdf%set_coeffs(C%dtlag)
     end do
 
     call C%fluid%restart(C%dtlag, C%tlag)
     call C%fluid%chkp%previous_mesh%free()
     if (allocated(C%scalar)) &
-       call C%scalar%restart( C%dtlag, C%tlag)
+         call C%scalar%restart( C%dtlag, C%tlag)
```
timfelle, line 285: Why break such a short line?
```suggestion
    if (allocated(C%scalar)) call C%scalar%restart(C%dtlag, C%tlag)
```

src/simulation.f90
```diff
@@ -274,18 +276,18 @@ subroutine simulation_restart(C, t)
...
 
     t = C%fluid%chkp%restart_time()
     call neko_log%section('Restarting from checkpoint')
     write(log_buf, '(A,A)') 'File :   ', &
-      trim(restart_file)
+         trim(restart_file)
```
timfelle, line 290: Same here, no reason to break this line
```suggestion
    write(log_buf, '(A,A)') 'File :   ', trim(restart_file)
```

timfelle (approved):

src/fluid/flow_ic.f90
```diff
@@ -93,34 +93,34 @@ subroutine set_flow_ic_int(u, v, w, p, coef, gs, type, params)
        call json_get(params, 'case.fluid.initial_condition.value', uinf)
        call set_flow_ic_uniform(u, v, w, uinf)
 
-    !
-    ! Blasius boundary layer
-    !
+       !
+       ! Blasius boundary layer
+       !
     else if (trim(type) .eq. 'blasius') then
```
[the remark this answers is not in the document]
tuananhdao: uhmm. This is not from this PR

PR #1668, opened 2025-01-21
Compute eddy viscosity based on the extrapolated velocity fields

Shiyu-Sandy-Du:
This PR is associated with issue #1623 for the eddy viscosity part. The wall model part has not been changed yet


src/les/les_model.f90
```diff
@@ -148,6 +158,19 @@ subroutine les_model_init_base(this, dofmap, coef, nut_name, delta_type)
...
+    call neko_field_registry%add_field(dofmap, 'u_e')
+    call neko_field_registry%add_field(dofmap, 'v_e')
+    call neko_field_registry%add_field(dofmap, 'w_e')
+    this%u_e => neko_field_registry%get_field('u_e')
+    this%v_e => neko_field_registry%get_field('v_e')
+    this%w_e => neko_field_registry%get_field('w_e')
+
+    ! Setup backend dependent summation of AB/BDF
+    this%case => case
+    call rhs_maker_sumab_fctry(this%sumab)
```
timofeymukha, line 171: Is it possible to just let fluid_scheme register these fields?

The thing is, `u_e` etc are very much coupled to the fluid_scheme. If we now have a compressible solver, there will be no `u_e` involved and no argument to use it in SGS models. So I think the ideal solution is to have field pointers `u`, `v`, `w` in the les_model, and then (for now) just point them to `u_e` etc from the registry.

It would also be very good if we can avoid needing a case to init the model since that is a huge dependency.

Shiyu-Sandy-Du: now it's cleaned up.

Shiyu-Sandy-Du:
Okay it's ready to be merged.

src/les/dynamic_smagorinsky.f90
```diff
@@ -146,14 +147,37 @@ subroutine dynamic_smagorinsky_compute(this, t, tstep)
...
+         u => neko_field_registry%get_field_by_name("u")
+         v => neko_field_registry%get_field_by_name("v")
+         w => neko_field_registry%get_field_by_name("w")
+         u_e => neko_field_registry%get_field_by_name("u_e")
+         v_e => neko_field_registry%get_field_by_name("v_e")
+         w_e => neko_field_registry%get_field_by_name("w_e")
+
+         call this%sumab%compute_fluid(u_e, v_e, w_e, u, v, w, &
+              ulag, vlag, wlag, ext_bdf%advection_coeffs, ext_bdf%nadv)
+
```
timofeymukha, line 166: Do we still need a local sumab?

src/fluid/fluid_scheme_incompressible.f90
```diff
@@ -372,6 +376,15 @@ subroutine fluid_scheme_init_base(this, msh, lx, params, scheme, user, &
...
     end if
 
+    if (this%variable_material_properties) then
+       call neko_field_registry%add_field(this%dm_Xh, 'u_e')
+       call neko_field_registry%add_field(this%dm_Xh, 'v_e')
+       call neko_field_registry%add_field(this%dm_Xh, 'w_e')
+       this%u_e => neko_field_registry%get_field('u_e')
+       this%v_e => neko_field_registry%get_field('v_e')
+       this%w_e => neko_field_registry%get_field('w_e')
+    end if
```
timofeymukha, line 386: I think you can skip the `if` and just let it always register these.

My idea was that when you do the sumab in fuild_pnpn, you would use these fields. So that in the LES model, you can just take them from the registry and not have to do a sumab there.

src/les/sigma.f90
```diff
@@ -66,13 +66,11 @@ module sigma
 
 contains
   !> Constructor.
-  !! @param dofmap SEM map of degrees of freedom.
-  !! @param coef SEM coefficients.
+  !! @param fluid The fluid_scheme_base_t object.
   !! @param json A dictionary with parameters.
-  subroutine sigma_init(this, dofmap, coef, json)
+  subroutine sigma_init(this, fluid, json)
```
timofeymukha, line 71: Sorry for nagging about this stuff, but do you actually need the fluid? It is better than passing the case :-). But why were dofmap and coef not sufficient?

src/les/dynamic_smagorinsky.f90
Shiyu-Sandy-Du: Yes, because the extrapolation of u v w has not been done in the fluid solve.

src/fluid/fluid_scheme_incompressible.f90
Shiyu-Sandy-Du: Why don't we put sumab in this if-statement as well? Or maybe there is some potential risks?

src/les/dynamic_smagorinsky.f90
timofeymukha: But on line 690 in fluid_pnpn we do the sumab, right? It is needed for the pressure residual.

src/les/sigma.f90
Shiyu-Sandy-Du: The reason for now is that we do the extrapolation in les_model.

src/les/dynamic_smagorinsky.f90
timofeymukha: The way it is is that in the fluid_pnpn_step, the u_e, v_e and w_e are temporaries given by the scratch registry. But if add those fields to the registry in the fluid_scheme_incompressible, like you added now, we can use those fields in the _step, and then they will be their for LES model. That was my thinking.

Shiyu-Sandy-Du: Oh yes. I didn't realize that. But another reason is that the pre-stepping computation of the simcomp (line 128 in simulation.f90) is ahead of fluid step. We could probably make fluid_pnpn use the extrapolated uvw if LES is activated and keep it as it is now if not LES.

src/les/bcknd/cpu/dynamic_smagorinsky_cpu.f90
```diff
@@ -61,8 +62,9 @@ module dynamic_smagorinsky_cpu
   !! @param lij The Germano identity.
   !! @param num The numerator in the expression of c_dyn, i.e. <mij*lij>
   !! @param den The denominator in the expression of c_dyn, i.e. <mij*mij>
-  subroutine dynamic_smagorinsky_compute_cpu(t, tstep, coef, nut, delta, &
-                                             c_dyn, test_filter, mij, lij, num, den)
+  subroutine dynamic_smagorinsky_compute_cpu(if_ext, t, tstep, coef, nut, delta, &
+       c_dyn, test_filter, mij, lij, num, den)
+    logical :: if_ext
```
timfelle, line 67: No intent
```suggestion
    logical, intent(in) :: if_ext
```

src/les/bcknd/cpu/sigma_cpu.f90
```diff
@@ -53,13 +53,15 @@ module sigma_cpu
...
+  !! @param if_ext If extrapolate the velocity field to evaluate
   !! @param t The time value.
   !! @param tstep The current time-step.
   !! @param coef SEM coefficients.
   !! @param nut The SGS viscosity array.
   !! @param delta The LES lengthscale.
   !! @param c The Sigma model constant
-  subroutine sigma_compute_cpu(t, tstep, coef, nut, delta, c)
+  subroutine sigma_compute_cpu(if_ext, t, tstep, coef, nut, delta, c)
+    logical :: if_ext
```
timfelle, line 64: Missing intent
```suggestion
    logical, intent(in) :: if_ext
```

src/les/bcknd/cpu/sigma_cpu.f90
```diff
@@ -144,129 +152,129 @@ subroutine sigma_compute_cpu(t, tstep, coef, nut, delta, c)
...
+               g31%x(i,1,1,e)*g33%x(i,1,1,e)
           sigG23 = g12%x(i,1,1,e)*g13%x(i,1,1,e) + &
-                   g22%x(i,1,1,e)*g23%x(i,1,1,e) + &
-                   g32%x(i,1,1,e)*g33%x(i,1,1,e)
+               g22%x(i,1,1,e)*g23%x(i,1,1,e) + &
+               g32%x(i,1,1,e)*g33%x(i,1,1,e)
 
           !        If LAPACK compute eigenvalues of the semi-definite positive matrix G
           !        ..........to be done later on......
           !        ELSE use the analytical method as done in the following
```
timfelle, line 166: Long Line
```suggestion
          ! If LAPACK compute eigenvalues of the semi-definite positive matrix G
          ! ..........to be done later on......
          ! ELSE use the analytical method as done in the following
```

src/les/bcknd/cpu/smagorinsky_cpu.f90
```diff
@@ -49,13 +49,15 @@ module smagorinsky_cpu
...
+  !! @param if_ext If extrapolate the velocity field to evaluate
   !! @param t The time value.
   !! @param tstep The current time-step.
   !! @param coef SEM coefficients.
   !! @param nut The SGS viscosity array.
   !! @param delta The LES lengthscale.
   !! @param c_s The smagorinsky model constant
-  subroutine smagorinsky_compute_cpu(t, tstep, coef, nut, delta, c_s)
+  subroutine smagorinsky_compute_cpu(if_ext, t, tstep, coef, nut, delta, c_s)
+    logical :: if_ext
```
timfelle, line 60:
```suggestion
    logical, intent(in) :: if_ext
```

src/les/bcknd/cpu/vreman_cpu.f90
```diff
@@ -41,21 +41,23 @@ module vreman_cpu
...
+  !! @param if_ext If extrapolate the velocity field to evaluate
   !! @param t The time value.
   !! @param tstep The current time-step.
   !! @param coef SEM coefficients.
   !! @param nut The SGS viscosity array.
   !! @param delta The LES lengthscale.
   !! @param c The Vreman model constant
-  subroutine vreman_compute_cpu(t, tstep, coef, nut, delta, c)
+  subroutine vreman_compute_cpu(if_ext, t, tstep, coef, nut, delta, c)
+    logical :: if_ext
```
timfelle, line 60:
```suggestion
    logical, intent(in) :: if_ext
```

src/les/bcknd/device/dynamic_smagorinsky_device.f90
```diff
@@ -64,9 +65,10 @@ module dynamic_smagorinsky_device
   !! @param lij The Germano identity.
   !! @param num The numerator in the expression of c_dyn, i.e. <mij*lij>
   !! @param den The denominator in the expression of c_dyn, i.e. <mij*mij>
-  subroutine dynamic_smagorinsky_compute_device(t, tstep, coef, nut, delta, &
-                                             c_dyn, test_filter, mij, lij, &
-                                             num, den)
+  subroutine dynamic_smagorinsky_compute_device(if_ext, t, tstep, coef, nut, delta, &
+       c_dyn, test_filter, mij, lij, &
+       num, den)
+    logical :: if_ext
```
timfelle, line 71:
```suggestion
    logical, intent(in) :: if_ext
```

src/les/bcknd/device/sigma_device.f90
```diff
@@ -51,13 +51,15 @@ module sigma_device
...
+  !! @param if_ext If extrapolate the velocity field to evaluate
   !! @param t The time value.
   !! @param tstep The current time-step.
   !! @param coef SEM coefficients.
   !! @param nut The SGS viscosity array.
   !! @param delta The LES lengthscale.
   !! @param c The Sigma model constant
-  subroutine sigma_compute_device(t, tstep, coef, nut, delta, c)
+  subroutine sigma_compute_device(if_ext, t, tstep, coef, nut, delta, c)
+    logical :: if_ext
```
timfelle, line 62:
```suggestion
    logical, intent(in) :: if_ext
```

src/les/bcknd/device/smagorinsky_device.f90
```diff
@@ -50,28 +50,36 @@ module smagorinsky_device
...
+  !! @param if_ext If extrapolate the velocity field to evaluate
   !! @param t The time value.
   !! @param tstep The current time-step.
   !! @param coef SEM coefficients.
   !! @param nut The SGS viscosity array.
   !! @param delta The LES lengthscale.
   !! @param c_s The smagorinsky model constant
-  subroutine smagorinsky_compute_device(t, tstep, coef, nut, delta, c_s)
+  subroutine smagorinsky_compute_device(if_ext, t, tstep, coef, nut, delta, c_s)
+    logical :: if_ext
```
timfelle, line 61:
```suggestion
    logical, intent(in) :: if_ext
```

src/les/bcknd/device/vreman_device.f90
```diff
@@ -50,13 +50,15 @@ module vreman_device
...
+  !! @param if_ext If extrapolate the velocity field to evaluate
   !! @param t The time value.
   !! @param tstep The current time-step.
   !! @param coef SEM coefficients.
   !! @param nut The SGS viscosity array.
   !! @param delta The LES lengthscale.
   !! @param c The Vreman model constant
-  subroutine vreman_compute_device(t, tstep, coef, nut, delta, c)
+  subroutine vreman_compute_device(if_ext, t, tstep, coef, nut, delta, c)
+    logical :: if_ext
```
timfelle, line 61:
```suggestion
    logical, intent(in) :: if_ext
```

src/les/les_model.f90
```diff
@@ -214,9 +245,9 @@ subroutine les_model_compute_delta(this)
              volume_element = volume_element + this%coef%B(k, 1, 1, e)
           end do
           this%delta%x(:,:,:,e) = (volume_element / this%coef%Xh%lx &
-                                                  / this%coef%Xh%ly &
-                                                  / this%coef%Xh%lz) &
-                                                  **(1.0_rp / 3.0_rp)                
+               / this%coef%Xh%ly &
+               / this%coef%Xh%lz) &
+               **(1.0_rp / 3.0_rp)
```
timfelle, line 250:
```suggestion
               / this%coef%Xh%ly / this%coef%Xh%lz)**(1.0_rp / 3.0_rp)
```

src/les/dynamic_smagorinsky.f90
timofeymukha: I guess to be safe we should add the fields to the registry in both the fluid and the LES model, with the `ignore_existing` flag set to true. I don't really know what the right way of treating the pre-stepping is. I guess if you apply the model that way, you *want* it to lag behind the fluid in some sense...? I'm not sure why one would ever do that :-).

I think it is ok if the fluid always puts this in the registry. If our approach is to use these any time we evaluate some right-hand-side term, it makes sense to have them there.

Shiyu-Sandy-Du: It wouldn't be a problem for the registry if we only have it in the scheme (since the registry is done by the constructor). But the main issue here is to avoid double estimation of the same variables :-)

src/les/sigma.f90
Shiyu-Sandy-Du: Oh and another reason for that: we need the fluid scheme to decide whether we use the extrapolated velocity or not. That is, we use the u_e v_e w_e for fluid_pnpn, and use uvw for compressible

src/les/dynamic_smagorinsky.f90
timofeymukha: OK, good! But then it seems like we can have it in the scheme only, modify the `fluid_pnpn_step` to run `sumab` on them, and avoid `sumab` in the `les_model`. And then regarding passing `fluid` to the `les_model` constructor, we can probably avoid it by just passing the `if_ext` as a flag instead? Then this logic:
```
     if_ext = .false.
     select type (fluid)
      type is (fluid_pnpn_t)
         if_ext = .true.
      end select

     call les_model%init(..., if_ext)
```
can be put in the les_simcomp.

timofeymukha: But les_simcomp does not over load preprocess, it overloads compute. So it will be run on line 171 in simulation.f90, after the fluid. Or am I completely lost? :P

Shiyu-Sandy-Du: It was changed to overload preprocess in this PR. And if you want to make it based on u_e, you have to enter the next time step first to get the extrapolation coefficients.

timofeymukha: Oh, I see. What was the reason to change to preprocess_?

src/les/bcknd/cpu/sigma_cpu.f90
```diff
@@ -144,129 +152,129 @@ subroutine sigma_compute_cpu(t, tstep, coef, nut, delta, c)
...
+                sigma1 = sqrt(max(Invariant1/3.0_rp + 2.0_rp*sqrt(alpha1), &
+                     0.0_rp))
+                sigma2 = sqrt(Invariant1/3.0_rp - sqrt(alpha1))
+                sigma3 = sigma2
+             else
+                alpha3 = acos(tmp1)/3.0_rp
+
+                if (abs(Invariant3) .lt. eps) then
+                   ! In case of Invariant3=0, one or more eigenvalues are equal to zero
+                   ! Therefore force sigma3 to 0 and compute sigma1 and sigma2
```
timfelle, line 248:
```suggestion
                   ! In case of Invariant3=0, one or more eigenvalues are equal 
                   ! to zero. Therefore force sigma3 to 0 and compute sigma1 and
                   ! sigma2
```

src/les/bcknd/cpu/vreman_cpu.f90
```diff
@@ -134,25 +142,25 @@ subroutine vreman_compute_cpu(t, tstep, coef, nut, delta, c)
...
+               a31%x(i,1,1,e)*a33%x(i,1,1,e)
           beta23 = a12%x(i,1,1,e)*a13%x(i,1,1,e) + &
-                   a22%x(i,1,1,e)*a23%x(i,1,1,e) + &
-                   a32%x(i,1,1,e)*a33%x(i,1,1,e)
+               a22%x(i,1,1,e)*a23%x(i,1,1,e) + &
+               a32%x(i,1,1,e)*a33%x(i,1,1,e)
 
           b_beta = beta11*beta22 - beta12*beta12 + beta11*beta33 - beta13*beta13 &
-                  + beta22*beta33 - beta23*beta23
+               + beta22*beta33 - beta23*beta23
```
timfelle, line 155:
```suggestion
          b_beta = beta11*beta22 - beta12*beta12 + beta11*beta33 &
               - beta13*beta13 + beta22*beta33 - beta23*beta23
```

src/les/bcknd/device/dynamic_smagorinsky_device.f90
```diff
@@ -64,9 +65,10 @@ module dynamic_smagorinsky_device
   !! @param lij The Germano identity.
   !! @param num The numerator in the expression of c_dyn, i.e. <mij*lij>
   !! @param den The denominator in the expression of c_dyn, i.e. <mij*mij>
-  subroutine dynamic_smagorinsky_compute_device(t, tstep, coef, nut, delta, &
-                                             c_dyn, test_filter, mij, lij, &
-                                             num, den)
+  subroutine dynamic_smagorinsky_compute_device(if_ext, t, tstep, coef, nut, delta, &
+       c_dyn, test_filter, mij, lij, &
+       num, den)
```
timfelle, line 70:
```suggestion
  subroutine dynamic_smagorinsky_compute_device(if_ext, t, tstep, coef, nut, 
       delta, c_dyn, test_filter, mij, lij, num, den)
```

src/les/les_model.f90
```diff
@@ -123,31 +136,46 @@ end subroutine les_model_factory
...
-    this%delta_type = delta_type
+    associate(dofmap => fluid%dm_Xh, &
+         coef => fluid%c_Xh)
+
+      if (.not. neko_field_registry%field_exists(trim(nut_name))) then
+         call neko_field_registry%add_field(dofmap, trim(nut_name))
+      end if
+      if (.not. neko_field_registry%field_exists("les_delta")) then
+         call neko_field_registry%add_field(dofmap, "les_delta")
+      end if
```
timfelle, line 159: The "ignore_existing" keyword make it a lot simpler, instead of the if statements
```suggestion
      call neko_field_registry%add_field(dofmap, trim(nut_name), .true.)
      call neko_field_registry%add_field(dofmap, "les_delta", .true.)
```

src/les/dynamic_smagorinsky.f90
Shiyu-Sandy-Du: If the `nu_t` needs the extrapolated velocity field, then one needs to calculate it before the `fluid%step` of current time step (to have the extrapolation coefficients), instead of taking the one computed from the past time step.

timofeymukha: You are right.. jeez, what a mess. I hope this has some beneficial effects on the solution...

MartinKarp (approved):

PR #1670, opened 2025-01-21
Move filter base type under src/ and complete the documentation of filters

Shiyu-Sandy-Du:
This PR should contain the following stuffs

- [ ] Move all filter-related code to src/filter;
- [ ] Add documentation for filter base type and its derived types;
- [ ] Maybe we should change the name of filters.f90 into another name to avoid confusion?

Maybe @timfelle and @gloopydoop have some good ideas.


timfelle (commented):
`filters.f90` should probably be called `mappings` or possibly just added to the math library since they are just simple 1d functions.

src/filter/elementwise_filter.f90
```diff
@@ -52,11 +52,11 @@ module elementwise_filter
   implicit none
   private
 
-  !> Implements the explicit filter for SEM.
+  !> Implements the elementwise filter for SEM.
   type, public, extends(filter_t) :: elementwise_filter_t
      !> filter type:
      !> possible options: "Boyd", "nonBoyd"
-     character(len=:), allocatable :: filter_type
+     character(len=:), allocatable :: elementwise_filter_type
```
timfelle, line 59: It is part of the type `elementwise_filter`, so I would just keep the
`filter_type`.

src/source_terms/brinkman_source_term.f90
```diff
@@ -40,7 +40,7 @@ module brinkman_source_term
   use field_list, only : field_list_t
   use field_math, only : field_subcol3, field_copy
   use field_registry, only : neko_field_registry
-  use filters, only : smooth_step_field, step_function_field, permeability_field
+  use mappings, only : smooth_step_field, step_function_field, permeability_field
```
timfelle, line 43:
```suggestion
  use mappings, only : smooth_step_field, step_function_field, &
       permeability_field
```

src/les/dynamic_smagorinsky.f90
```diff
@@ -99,9 +100,23 @@ subroutine dynamic_smagorinsky_init(this, fluid, json)
 
       call this%free()
       call this%init_base(fluid, nut_name, delta_type, if_ext)
-
       call json_extract_object(json, "test_filter", json_subdict)
       call this%test_filter%init(json_subdict, coef)
+      if (json%valid_path('filter.transfer_function')) then
+         call neko_error("Dynamic Smagorinsky model does not support transfer &
+         &function specified in the json file. &
```
timfelle, line 107: As @timofeymukha mentioned elsewhere, you should break the string by `" // &`
instead of `& &`.

src/filter/filter_fctry.f90
```diff
@@ -0,0 +1,75 @@
...
+       allocate(elementwise_filter_t::object)
+    else if (trim(type_name) .eq. 'PDE') then
+       allocate(pde_filter_t::object)
+    else
+       type_string = concat_string_array(FILTER_KNOWN_TYPES, &
+            NEW_LINE('A') // "-  ", .true.)
+       call neko_error("Unknown filter type: " &
+            // trim(type_name) // ".  Known types are: " &
+            // type_string)
+       stop
```
timfelle, line 69: Stop is not needed as `neko_error` does that for you. Also please use the
`neko_type_error`, that way you do not have to do the concat stuff.

timofeymukha (approved):

PR #1671, opened 2025-01-22
Remove unnecessary parameters from fluid_scheme class

timfelle:
Reduce the dependency on JSON files by removing the parameters from the fluid_scheme class, simplifying the initialization and validation processes. This additionally limit the use of the json file to be only for initialization.


MartinKarp (approved):
GreatI guess params is always an arguments if thera are constructurs that use it.

njansson (approved):

PR #1672, opened 2025-01-22
Make it possible to change lambda and mu in time

MartinKarp:
I tried to change lmabda in time for multi phase, but noticed it didnt work. I think it is not only LES that needs to be able to modify these things.


src/scalar/scalar_scheme.f90
```diff
@@ -679,6 +679,8 @@ subroutine scalar_scheme_update_material_properties(this)
           call cfill(this%lambda_field%x, this%lambda, n)
           call add2s2(this%lambda_field%x, nut%x, lambda_factor, n)
        end if
+    else then
+      this%lambda_field = this%lambda
```
timofeymukha, line 683: Should you just pull the cfill out of the `if (this%variable_material_properties) then`? Same for the fluid.

A bit tricky semantically here, I guess what you do is also variable properties, but only in time, so you can still use the simplified stress formulation.

MartinKarp: cleaned it a bit

src/scalar/scalar_scheme.f90
```diff
@@ -668,17 +669,12 @@ subroutine scalar_scheme_update_material_properties(this)
     real(kind=rp) :: lambda_factor
 
     lambda_factor = this%rho*this%cp/this%pr_turb
+    this%lambda_field = this%lambda
```
timofeymukha, line 672: But should we take care of the case when the field is on the device?

MartinKarp: This occurs on the device we are currently running on, so should be equivalent to before. 
https://github.com/ExtremeFLOW/neko/blob/30aedb07ed64516b9a05a6441f6549fa112da6cf/src/field/field.f90#L223

timfelle (approved):

njansson (approved):

PR #1673, opened 2025-01-22
Change htable AoS -> SoA, and refactor to submodules

njansson:
This pr changes the hash table data structure to use SoA instad of a AoS, and move the implementations into submodules. Furthermore, we fix the `key` type to avoid too many `select type` lookups (making it more cache friendly) while inserting or retrieving elements. This unfortunately increases the code size with a lot of duplication, but better supports compilers with a flaky OO implementation. 

This pr also contains a workaround to fix a SIGSEGV for LLVM 18 based compilers (e.g. CCE 18.x.x) during `stable` rebuild


njansson:
This refactoring also significantly improves the hash table performance on A64FX

src/adt/htable.f90
timfelle, line 1: Any chance we could avoid using intrinsic fortran keywords as variable names?
`data` is a keyword in fortran. I am aware that they are not strictly reserved
words per say.

https://fortranwiki.org/fortran/show/Keywords

njansson: Sure that's a valid point, and good timing to change since we're ripping things apart atm.

Maybe that would fix gcc10

src/adt/htable_cptr.f90
timfelle, line 1: We are missing the copyright statement here

src/adt/htable_i4t2.f90
```diff
@@ -0,0 +1,296 @@
+! Copyright (c) 2019-2023, The Neko Authors
```
timfelle, line 1:
```suggestion
! Copyright (c) 2019-2025, The Neko Authors
```

MartinKarp:
Have we checked it works for very large problems? The htable has been one of the main bottlenecks and sources to crashes when running at scale (both in rea2nbin as well as when generating the connectivity in mesh.f90). I think generating a mesh on the order of 10M  elements or more and the time to build connectivity would be good to check.

src/adt/htable.f90
timfelle: Could we also limit the scope of the last two use statements for good measure?
`num_types` and `utils`

njansson:
> Have we checked it works for very large problems? The htable has been one of the main bottlenecks and sources to crashes when running at scale (both in rea2nbin as well as when generating the connectivity in mesh.f90). I think generating a mesh on the order of 10M  elements or more and the time to build connectivity would be good to check.

Time is reduced significantly (mileage varies depending on compiler). However the integer limits still remains wrt table size

src/adt/htable_i8.f90
```diff
@@ -0,0 +1,292 @@
...
+    end if
+
+    size = ishft(1, ceiling(log(dble(size)) / NEKO_M_LN2))
+
+    allocate(this%key(0:size))
+    if (present(data)) then
+       allocate(this%data(0:size), source=data)
+    else
+       allocate(integer(kind=i8)::this%data(0:size))
+    end if
```
timfelle, line 58: The gfortran-12 seem to report an allocation of an already allocated variable here.

njansson: Should be fixed now, it was a left over call to the wrong free method

timfelle (commented):
> I kind of have given up hope on getting this through with gcc-10. How about that we replace it with gcc-14 in our CI? (gcc-10 is a bit old imo)  

Well I don't blame you. These files also contain a plethora of implicit type casts, missing intents and other stuff that make it hard to figure out exactly what is what.

One thing to look at is the data array. In fortran you only have `private` and `public`. You do not have `protected` as you do in c++. This mean, either a variable is visible to everyone, or it is not visible to anyone. A derived class cannot access the private components of a parent, if it is defined in a seperate module. I am not sure if that causes some problems for the submodule approach here. It might explain why gcc-10 is not working, since the variables are accessed from a different, albeit sub-, module.

njansson:
> > I kind of have given up hope on getting this through with gcc-10. How about that we replace it with gcc-14 in our CI? (gcc-10 is a bit old imo)
> 
> Well I don't blame you. These files also contain a plethora of implicit type casts, missing intents and other stuff that make it hard to figure out exactly what is what.
> 

As long as intent is matching between interface and implementation these should push gcc-10 to go belly up. But these implicit type casts, where are those? (also the code in the submodules are the same as before, so I think the issue is "a not perfect" submodule implementation in gcc-10

> One thing to look at is the data array. In fortran you only have `private` and `public`. You do not have `protected` as you do in c++. This mean, either a variable is visible to everyone, or it is not visible to anyone. A derived class cannot access the private components of a parent, if it is defined in a seperate module. I am not sure if that causes some problems for the submodule approach here. It might explain why gcc-10 is not working, since the variables are accessed from a different, albeit sub-, module.

my understanding of the standard is that a submodule has the same access as the parent module, except the procedures of the parent module (since there's no definition visible to the submodule).

timfelle:
> As long as intent is matching between interface and implementation these should push gcc-10 to go belly up. But these implicit type casts, where are those? (also the code in the submodules are the same as before, so I think the issue is "a not perfect" submodule implementation in gcc-10

Regarding intent, there are some intents missing. All size parameters for the initializers and the `c` for the hash functions. In most of the setter functions there are no precision defined for the values used in the computation if `i` and the real typed output from the `log` functions are implicitly cast to an integer, as `i` is an integer.

> my understanding of the standard is that a submodule has the same access as the parent module, except the procedures of the parent module (since there's no definition visible to the submodule).

It would not surprise me if that is exactly the source of the imperfect submodule implementation you mentioned above.

src/adt/htable_i4.f90
```diff
@@ -0,0 +1,290 @@
...
+  !> Insert an integer into the hash table
+  recursive module subroutine htable_i4_set(this, key, data)
+    class(htable_i4_t), intent(inout) :: this
+    integer, intent(inout) :: key   !< Table key
+    class(*), intent(inout) :: data !< Data associated with @a key
+    type(htable_i4_t), allocatable :: tmp
+    integer index, i, c
+
+    c = 0
+    i = log(1.0/this%size)/log(0.6)
```
timfelle, line 89: Implicit type casting

njansson:
> > As long as intent is matching between interface and implementation these should push gcc-10 to go belly up. But these implicit type casts, where are those? (also the code in the submodules are the same as before, so I think the issue is "a not perfect" submodule implementation in gcc-10
> 
> Regarding intent, there are some intents missing. All size parameters for the initializers and the `c` for the hash functions. In most of the setter functions there are no precision defined for the values used in the computation if `i` and the real typed output from the `log` functions are implicitly cast to an integer, as `i` is an integer.

Not for the purpose of nitpicking but both `c` and `size` has intent since they are marked as `value`. I'll fix the casts in the `log` things now.

> 
> > my understanding of the standard is that a submodule has the same access as the parent module, except the procedures of the parent module (since there's no definition visible to the submodule).
> 
> It would not surprise me if that is exactly the source of the imperfect submodule implementation you mentioned above.

Also a note, `gcc-10` is end of life so we should probably consider moving on regardless (11 is still supported)

timfelle:
> Not for the purpose of nitpicking but both `c` and `size` has intent since they are marked as `value`. I'll fix the casts in the `log` things now.
Sure thing, I haven't worked that much with the value thing.

> Also a note, `gcc-10` is end of life so we should probably consider moving on regardless (11 is still supported)

Ahh by all means remove it then. I tried looking for the EoL date but couldn't really find it.

Issue #1675, opened 2025-01-24
Incompatible metadata files with paraview reader


Mr-2c:
The paraview reader has issues with the current metadata files written with the "type: binary" and exits. the error is:
Error parsing file.  Unknown tag type:
I would recommend removing the last line (type) from the metadata files.

timofeymukha:
Is it the new reader or the old one?

Mr-2c:
The new one, paraview 5.13.2

On 25. Jan 2025, at 10:54, Timofey Mukha ***@***.***> wrote:



Is it the new reader or the old one?

Mr-2c:
oh, and by metadata files I mean .nek5000 files. so we need to just remove the last line from those files

timofeymukha:
Very nice! It seems to work for me as well! In 5.11 it crashed regardless, so that's good improvement :-)

timofeymukha:
Can someone check if this would break Visit? @njansson @MartinKarp

MartinKarp:
Works locally on my laptop with visit 3.4.1

njansson:
🍻  I guess we can close this now when #1677 is merged?

timofeymukha:
Resolved!

PR #1677, opened 2025-01-28
Remove "type: binary" from .nek5000

timofeymukha:
Should enable the new Nek5000 reader in Paraview to read our .fld files. Works with Paraview 5.13.2


timofeymukha:
Solves #1675

timfelle (approved):

PR #1678, opened 2025-01-28
Remove redundant sweep over all grid points for nab.eq.3

njansson:
Besides the performance delta. This change also allows for `frtpx` to parallelise both loop nests.


timofeymukha (approved):

pohm01 (approved):
lgtm

PR #1681, opened 2025-01-29
Optimise the device computation of the eddy viscosity field

Shiyu-Sandy-Du:
In this PR, the device kernel is optimised by the following changes

1. Reducing the number of calling elements from arrays;
2. Using constant variable if possible;
3. Moving the multiplication of mult to the end of nut computation: we compute the nut from a "wrong" tensor and correct it in the end.


njansson (approved):

src/les/bcknd/device/cuda/dynamic_smagorinsky_nut_kernel.h
```diff
@@ -96,19 +95,26 @@ __global__ void lij_compute_part1(T * __restrict__ l11,
...
-    fww[i] = w[i] * w[i];
-    fuv[i] = u[i] * v[i];
-    fuw[i] = u[i] * w[i];
-    fvw[i] = v[i] * w[i];
+    const T u_r = u[i];
+    const T v_r = v[i];
+    const T w_r = w[i];
+    const T fu_r = fu[i];
+    const T fv_r = fv[i];
+    const T fw_r = fw[i];
```
timofeymukha, line 103: Is the idea that addressing the array elements one time less is worth introducing new variables? Just curious.

timfelle (approved):

Issue #1469, opened 2024-09-09
Incompatible linear solver in scalar scheme if stress formulation is activated


njansson:
The current scalar scheme is reusing the fluid's velocity solver settings for its linear solver.

If the stress formulation is used, a coupled CG is used for the fluid, which is then incompatible with a scalar solve

PR #1682, opened 2025-01-31
Let scalar have its own solver config

timofeymukha:
This changes the solver configuration for the scalar: instead of using the fluid's velocity solver, it is now configured separately.

This is necessary because if one uses the coupled solvers for the fluid, the scalar tries to pick it up and crashes.

### DIFF
--- a/examples/scalar_mms/scalar_mms.case
+++ b/examples/scalar_mms/scalar_mms.case
@@ -44,6 +44,13 @@
       "initial_condition": {
         "type": "user"
       },
+      "ssolver": {
+        "type": "cg",
+        "preconditioner": "jacobi",
+        "projection_space_size": 0,
+        "absolute_tolerance": 1e-7,
+        "max_iterations": 800
+      },
       "source_terms": [
         {
           "type": "user_vector"
--- a/src/scalar/scalar_scheme.f90
+++ b/src/scalar/scalar_scheme.f90
@@ -253,17 +253,17 @@ contains
     this%w => neko_field_registry%get_field('w')
 
     call neko_log%section('Scalar')
-    call json_get(params, 'case.fluid.velocity_solver.type', solver_type)
-    call json_get(params, 'case.fluid.velocity_solver.preconditioner', &
+    call json_get(params, 'case.scalar.solver.type', solver_type)
+    call json_get(params, 'case.scalar.solver.preconditioner', &
          solver_precon)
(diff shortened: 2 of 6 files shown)


timofeymukha:
Closes #1469

examples/scalar_mms/scalar_mms.case
```diff
@@ -44,6 +44,13 @@
       "initial_condition": {
         "type": "user"
       },
+      "ssolver": {
```
njansson, line 47:
```suggestion
      "solver": {
```

Typo?

njansson (commented):
Very nice! Thanks for fixing yet another thing from my endless todo list

timfelle (approved):

timfelle (commented):
We are slowly moving closer to a common interface where we can just pass a subkey of the json "case.fluid" or "case.scalar" and then everything is configured from the same subroutine and interface. The `scheme` parent seem to get close.

timofeymukha:
> We are slowly moving closer to a common interface where we can just pass a subkey of the json "case.fluid" or "case.scalar" and then everything is configured from the same subroutine and interface. The `scheme` parent seem to get close.

Yeah, some external init from the case constructor has also been removed in the bc PR and in some previous ones.

src/scalar/scalar_scheme.f90
```diff
@@ -533,28 +533,28 @@ subroutine scalar_scheme_set_material_properties(this, params, user)
...
-                                        this%cp, this%lambda, params)
+            this%cp, this%lambda, params)
     else
        if (params%valid_path('case.scalar.Pe') .and. &
-           (params%valid_path('case.scalar.lambda') .or. &
+            (params%valid_path('case.scalar.lambda') .or. &
             params%valid_path('case.scalar.cp'))) then
           call neko_error("To set the material properties for the scalar,&
-          & either provide Pe OR lambda and cp in the case file.")
+               & either provide Pe OR lambda and cp in the case file.")
```
timofeymukha, line 553: @timfelle I am using the --indent-ampersand, I copied the string from the documentation. And this *is* indenting it, right? So it seems like the CI is not indenting, or?

timfelle: Right I think I found the issue.
There seem to be an issue when reading the options from FINDENT_FLAGS rather than directly specifying them on commandline.

PR #1683, opened 2025-01-31
Introducing a C API for Neko

njansson:
This PR introduces a C API for Neko, allowing us to set up and execute simulation cases in Neko from C, Julia or Python.

A simple example in Python is provided, based on the refactored pyneko interface.


tuananhdao (approved):
Great! LGTM!

njansson:
> Nice! My only concern is that we are building a parallel solve function inside the API. Do you think there is theoretically a way we could tap into the normal routines? Maybe by somewhat restructuring the latter? I think if we now understand how to do the connection to Python properly, it would be wise to think how we can adapt the code to make that easy to do!

No no. `solve` will go directly to the normal Neko solve, which should be the intended uses. While `step` should be seen as a special case, which needs some minor duplication. 

I would argue rather strongly against adapting the code to make it easier to call _everything_, this will just make it a mess. Let's keep it as an API that can be called by various languages, and not morph this into some horrible, dedicated python + Fortran mess :D

examples/pyneko/hemi.case
```diff
@@ -0,0 +1,65 @@
+{
+"version": 1.0,
```
MartinKarp, line 2: Would it be ok to use another example than Hemi? Any other example except hemi would also be fine I think as it is the only case which shows a incorrectly set up mesh. Cylinder is quite similar.

src/api/neko_api.f90
```diff
@@ -0,0 +1,285 @@
...
+       call neko_error('Invalid Neko case')
+    end if
+
+  end subroutine neko_api_solve
+
+  !> Compute a time-step for a neko case
+  !! @param case_iptr Opaque pointer for the Neko case
+  !! @param time The time value
+  !! @param tstep The current time-stepper iteration
+  subroutine neko_api_step(case_iptr, time, tstep) bind(c, name="neko_step")
```
MartinKarp, line 154: Would it be possible to have this one just calling simulation_step or similar?

MartinKarp (commented):
Really cool! If possible, would it be possible to move away from hemi? (or fix the mesh and blasius bc)

examples/pyneko/hemi.case
njansson: any case that doesn't require `makeneko` should just work

.github/workflows/check_format.yml
```diff
@@ -70,7 +70,7 @@ jobs:
             echo "" >&2
             echo "Please run the following commands to fix the formatting:" >&2
             echo "pip install findent" >&2
-            echo "findent -Rr -i2 -d3 -f3 -s3 -w3 -t3 -j3 -k5 --ws_remred --indent_ampersand --openmp=0 <file.f90 >file.f90.tmp" >&2
+            echo "findent -Rr -i2 -d3 -f3 -w3 -t3 -j3 -k5 --ws_remred --indent_ampersand --openmp=0 <file.f90 >file.f90.tmp" >&2
```
timfelle, line 73:
```suggestion
            echo "findent -Rr -i2 -d3 -f3 -s3 -w3 -t3 -j3 -k5 --ws_remred --indent_ampersand --openmp=0 <file.f90 >file.f90.tmp" >&2
```

src/api/neko_api.f90
njansson: I was thinking about splitting our current simulation into a driver loop and a step function. But I somehow felt that there was a lot of extra things in the current loop, that would not work with the current setup. 

Maybe time to rethink that.

.github/workflows/check_format.yml
```diff
@@ -53,7 +53,7 @@ jobs:
 
       - name: Format sources
         env:
-          FINDENT_FLAGS: "-Rr -i2 -d3 -f3 -s3 -w3 -t3 -j3 -k5 --ws_remred --indent_ampersand --openmp=0"
+          FINDENT_FLAGS: "-Rr -i2 -d3 -f3 -w3 -t3 -j3 -k5 --ws_remred --indent_ampersand --openmp=0"
```
timfelle, line 56: This is fixed in a separate PR.
```suggestion
          FINDENT_FLAGS: "-Rr -i2 -d3 -f3 -s3 -w3 -t3 -j3 -k5 --ws_remred --indent_ampersand --openmp=0"
```

src/api/neko_api.f90
timfelle: I definitely think it would be worth rethinking that, since we (The Neko-Top) group, will be needing that very soon to build unsteady adjoints. 😁

njansson: Ok! But what do you guys needs for the top stuff? Or rather how do want to control each simulation?

The main reason why it is a standalone step function now is since we don't have a simulation object that keeps track of state (do we want that?). Such an object would be necessary for calling a step function from the outside.

timfelle: It is still early so I do not have the full overview yet. 

But we will need some finegrain control of the simulation. To a point where it is trivial to restart the simulation to a given point and do single steps from there. It is possibly too early to make any decisions on it, but in general the more modular the solvers and disconnected each scheme can be will make it easier for us to add the required pieces later. 

Currently we have essentially a full copy of the fluid scheme, to serve as the adjoint scheme. And a full copy of the case object to handle the calling of said adjoint scheme. Not object copies, but code duplication.

timfelle: > Actually it might not be necessary with such an object, the only thing necessary is a `time_step_controller_dt`. Either way I can have a look

Also i think the recent idea of moving the time_step stuff inside each scheme might be sufficient. I think that was in relation to the compressible solver and having a separate time scheme for the scalar scheme.

njansson: Fixed in the latest commits

examples/pyneko/hemi.case
njansson: fixed

timofeymukha (approved):
I like it. A good bit of extra structure in simulation.f90 is a bonus!

examples/api/pyneko/cylinder.case
```diff
@@ -0,0 +1,98 @@
+{
+"version": 1.0,
+"case": {
+    "mesh_file": "cylinder.nmsh",
+    "output_at_end": true,
```
timofeymukha, line 5: Perhaps a better way is to add a prepare.sh that copies over the case and nmsh from the cylinder example. This way we have less files to deal with when we update case file structure.

njansson: Yes but, all json parsers aren't as forgiving as json Fortran and goes belly up with the // comments ;)

PR #1686, opened 2025-02-03
Change use device

MartinKarp:
Changing to be consistent with how we use device. Also makes non-device related things private in device.F90


timfelle:
Most likely due to the `c_loc`, `c_associated`, `c_null` and other constructs which are not explicitly loaded in. To debug it, put only on every single use entry and find the commands missing that way.

njansson:
Projection does have `iso_c_binding` by it self. So it's not a straightforward missing `use/only` statement.

njansson:
Not sure if this is a gfortran thing or that we have stretched the standard a bit. Looking at the error message:
```bash
common/projection.f90:169:27:

  169 |        ptr_size = c_sizeof(C_NULL_PTR) * this%L
      |                           1
Error: ‘x’ argument of ‘c_sizeof’ intrinsic at (1) must be an interoperable data entity: Expression is a noninteroperable derived type

```

By changing the argument of `c_sizeof` to an actual data entry, it works without any other `only` fixes.

timfelle:
> Projection does have `iso_c_binding` by it self. So it's not a straightforward missing `use/only` statement.

My point is more that since removing the only statements made the issue go away, it is clear that something is not imported or is somehow shared from the device module.

njansson:
> > Projection does have `iso_c_binding` by it self. So it's not a straightforward missing `use/only` statement.
> 
> My point is more that since removing the only statements made the issue go away, it is clear that something is not imported or is somehow shared from the device module.

Hear hear.

It would be good to have language lawyer now, such that we can understand if we or gfortran is interpreting the standard differently. 

Ps. with gfortran 14.2.1, it seems to compile fine

timfelle:
> Hear hear.
> 
> It would be good to have language lawyer now, such that we can understand if we or gfortran is interpreting the standard differently.
> 
> Ps. with gfortran 14.2.1, it seems to compile fine

Yea we may be at that point soon. `device` was also one of the libraries we could not make the internals private of for some reason, i forget why.

timfelle:
> So what do we prefer here? Change the entry or remove only?
> 
> The issue with device I think is that when we do only on device we import a bunch of things that are not in device. For example in vector.f90 we do:
> 
> ```fortran
> use device, only: device_map, device_free, c_ptr, C_NULL_PTR
> ```
> 
> where neither c_ptr or C_NULL_PTR are defined in device.f90. This type of use of only seeems to be quite common in the code. To me it looks like if we do not make the internals private in a class we should avoid using only on that module to avoid this kind of imports (where we implicitly import from a module that is used in device, but not defined in device itself).
> 
> We also quite often import glb_cmd_queue from device for example, which are also not defined in device, but in the hip/cuda/opencl_intf modules.
> 
> I would overall be a proponent of removing only on all classes which do not have private internals (or make everything private, but then one would need to import a lot of redundant stuff from the intf for example).

Explicitly importing things from a module which does not define it is in my mind a bug and should not be allowed by compilers.
So in all cases I would say we should fix the imports in `vector` (and everywhere else). Either we import things correctly, (`c_ptr` from `iso_c_binding`) or we do not specify the `only` from device and let the imports probagate.

Regarding the `glb_cmd_queue`, it should probably be defined in `device` in any case since all of them define it.

MartinKarp:
Yeah I also think that is a bug when importing things from a module which do not define them. For now I was thinking then to just remove only on device everywhere.

njansson:
This seems to be long standing gfortran issue (T_T), https://gcc.gnu.org/bugzilla/show_bug.cgi?id=103496

timofeymukha:
Has this been superseded by @njansson PR?

njansson:
> Has this been superseded by @njansson PR?

Yes as far as I know develop works again with prgenv gnu, but would be good if someone else verifies it

MartinKarp:
This is fixed with Niclas work.

Issue #1696, opened 2025-02-05
Overwriting of fluid_stats files after restart from joblimit


vbaconnet:
EDIT: I think @Mr-2c mentioned something like this before, or? 

# Context

When a simulation is terminated due to the `job_timelimit` being reached, and if `output_at_end` is set to `true`, the output controller triggers sampling of a field file snapshot `fld`, a checkpoint, and if activated a `fluid_stats` file. 

```f90
! code taken from simulation.f90
do while (t .lt. c%end_time .and. (.not. jobctrl_time_limit()))
   ! ... simulation loop
end do

call json_get_or_default(c%params, 'case.output_at_end',&
                             output_at_end, .true.)
call c%output_controller%execute(t, tstep, output_at_end)
```

`fluid_stats` files are written with a numbering similar to `fld` files: `fluid_stats0.f00000`, `fluid_stats0.f00001`, etc.

# Issue

Let's say a simulation is running and statistics files are output at every 1 second of simulation time, and the `job_timelimit` interrupts the simulation at `1.3` seconds. We would be left with the following set of fluid stats files:

 `fluid_stats0.f00000` written at `t=0`,
 `fluid_stats0.f00001` written at `t=1`, 
`fluid_stats0.f00002` written at `t=1.3`.

Restarting now from `t = 1.3`, as soon as the simulation time reaches `t = 2` the output controller for `fluid_stats` triggers sampling for the averaged fields taken in the interval `[1.3, 2]`. This creates a new file `fluid_stats0.f00002`, overwriting the one that was output at `t = 1.3`. This means that instead of having:

 `fluid_stats0.f00000` written at `t=0`,
 `fluid_stats0.f00001` written at `t=1`,
`fluid_stats0.f00002` written at `t=1.3`,
`fluid_stats0.f00003` written at `t=2`,

we have

 `fluid_stats0.f00000` written at `t=0`,
 `fluid_stats0.f00001` written at `t=1`,
`fluid_stats0.f00002` written at `t=2`. <--- sampling is in `[1.3, 2]` but  `average_fields_in_time` thinks it's `[1, 2]`

When `average_fields_in_time` goes through the fluid_stats files for total averaging, it will wrongfully assume that the sampling time interval between `fluid_stats0.f00001` and `fluid_stats0.f00002` is [1, 2] when it is actually [1.3, 2], and the weighted average will be wrong.

This is fine for `fld` files since they are just snapshots. 

So the issue is:
[...]

PR #1699, opened 2025-02-06
Change fluid output file name when restarting

vbaconnet:
Proposed solution for issue #1696.

Whenever we restart, we add the timestamp of the first file in the series to the file name itself to not overwrite any fluid statistics files and to make it easier to organize the files. The behavior now is (taking the examples from the issue):

`fluid_stats0_0.f00000` at t = 0
`fluid_stats0_0.f00001` at t = 1
`fluid_stats0_0.f00002` at t = 1.3 <--- sample triggered by job_limit 
`fluid_stats2_0.f00002` at t = 2 <--- instead of `fluid_stats0.f00002`

We do it by calling the init_base of `fluid_output` but it technically boils down to just changing the name of the file. 

Also added a getter for the `counter` attribute for convenience. 

Corner case is that will still overwrite files if the simulation writes out less than 1 file per run. Hence it is still recommended to make consecutive runs in different directories.

TODO:
- [x] Documentation: recommend to run in different output folders when doing big runs as "best practice" just in case.


adperezm:
Hej, why don't you change instead the one with the job limit, is it more complicated?

It is the one at t=2 that complies with your sampling time of t=1, although of course is incomplete.

Is your reasoning that all the files with the same f0000X combine to form the requested statistics time or something like that?

pschlatt1:
I put a comment in the original issue. I would really suggest to avoid too much complication in naming schemes. This is user dependent and can be easily done outside using your job script.

timofeymukha:
I guess if we rely on external scripts, there is no issue here to solve really. But my concern is that, as Philipp also mentioned in the issue discussion, for large runs on clusters stopping due to joblimit is the default, not the exception. So it would be nice if that case is streamlined.

adperezm:
I think there is an issue to solve with regards to consistency. Right now we are not following @pschlatt1 approach either, as the file sequence is not restarted every time. So it gives a user a false sense of security.

A user will very likely make a mistake and then realize that they need a script, and that is just mean :). I vote for now to change the name of the file output with the job limit.

The more destructive way is to again write every file from 0, as nek5000 does, but I somehow feel that is a step backwards.

vbaconnet:
> Is your reasoning that all the files with the same f0000X combine to form the requested statistics time or something like that?

Yea, the idea is that each run produces its own independent batch of statistics.

vbaconnet:
> I vote for now to change the name of the file output with the job limit.

That was my idea in the beginning as well. Only thing is that with the way I did it in the PR, you can `average_fields_in_time` each batch independently, without additional renaming and re-indexing, etc. I personally prefer that but I guess that's user preference :)

timofeymukha:
Let's discuss this today at the meeting.

MartinKarp:
This will be merged and we can undo it at a later point if this totally ruins the experience of using stats for someone.

doc/pages/user-guide/case-file.md
```diff
@@ -86,6 +86,23 @@ but also defines several parameters that pertain to the simulation as a whole.
 | `cfl_deviation_tolerance`  | The tolerance of the deviation from the target CFL number                                             | Positive real less than `1`                     | `0.2`         |
 | `end_time`                 | Final time at which the simulation is stopped.                                                        | Positive reals                                  | -             |
 | `job_timelimit`            | The maximum wall clock duration of the simulation.                                                    | String formatted as HH:MM:SS                    | No limit      |
+| `output_at_end`            | Whether to always write all enabled output at the end of the run.                                    | `true` or `false`                               | `true`        |
+
+### Restarts and joblimit
+Restarts will restart the simulation from the exact state at a given time that 
+the checkpoitn was written. This means that the flow field and potential scalars
```
MartinKarp, line 93:
```suggestion
the checkpoint was written. This means that the flow field and potential scalars
```

MartinKarp (approved):

PR #1703, opened 2025-02-10
Update runner to nvhpc 25.1, and add arm64 runners and gfortran 14 (cpu)

njansson:


timfelle:
Maybe a `static_cast<MPI_Request *>(vreqs)` will be the solution to this?

```
"gs/bcknd/device/device_mpi.c", line 60: error: a value of type "void *" cannot be used to initialize an entity of type "MPI_Request *" (aka "ompi_request_t **")
    MPI_Request *reqs = vreqs
```

src/gs/bcknd/device/device_mpi.c
```diff
@@ -68,7 +68,7 @@ void device_mpi_isend(void *buf_d, int offset, int nbytes, int rank,
 
 void device_mpi_irecv(void *buf_d, int offset, int nbytes, int rank,
 		      void *vreqs, int i) {
-  MPI_Request *reqs = vreqs;
+  MPI_Request *reqs = (MPI_Reuqest *) vreqs;
```
timfelle, line 71:
```suggestion
  MPI_Request *reqs = (MPI_Request *) vreqs;
```

njansson: Sigh, this is what happens when one tries to debug several things in parallel :(

timfelle: Yea, its not always clear what works and what does not.

PR #1721, opened 2025-02-20
Add NCCL comm backend for gather-scatter

njansson:


src/gs/bcknd/device/gs_device_nccl.F90
```diff
@@ -0,0 +1,363 @@
...
+     procedure, pass(this) :: nbrecv => gs_device_nccl_nbrecv
+     procedure, pass(this) :: nbwait => gs_device_nccl_nbwait
+  end type gs_device_nccl_t
+
+#ifdef HAVE_HIP
+  interface
+     subroutine hip_gs_pack(u_d, buf_d, dof_d, offset, n, stream) &
+          bind(c, name='hip_gs_pack')
+       use, intrinsic :: iso_c_binding
+       implicit none
```
timfelle, line 82: any particular reason why you do not do a `import c_ptr, c_int` instead of
importing the entire module in every single interface?

njansson: No reason, it's simply copy pasta of `gs_device_mpi` since it shares a lot of code

timfelle: Alright, feel free to resolve this if you do not want to keep track of the suggestion.

timfelle (approved):

src/gs/bcknd/device/gs_device_nccl.F90
njansson: I've a NVSHMEM pr in progress as well. With that I will revisit this, and also update the documentation, adding some text about the new env. variable `GS_COMM`, which can be used to force the use of a certain backend

timofeymukha (approved):

PR #1724, opened 2025-02-21
Comment on the trailing 0 in .f/fld files

MartinKarp:


src/io/fld_file.f90
```diff
@@ -520,6 +520,11 @@ subroutine fld_file_write(this, data, t)
        open(unit = 9, &
             file = trim(this%fname(1:suffix_pos-1)) // &
             trim(adjustl(start_field)), status = 'replace')
+       !This will specify that each files is the filename followed by a 0.
```
timofeymukha, line 523: This sentence is a little poor. Do you mean to say that the name of each file has a 0 appended to it?

MartinKarp: I changed it a bit. Yeah, all filenames has a 0 appended.

src/io/fld_file.f90
```diff
@@ -520,6 +520,12 @@ subroutine fld_file_write(this, data, t)
        open(unit = 9, &
             file = trim(this%fname(1:suffix_pos-1)) // &
             trim(adjustl(start_field)), status = 'replace')
+       !The following string will specify that the files in the file series
+       !are defined by the filename followed by a 0.
+       !This 0 is necessary as it specifies the index of number of files
+       !the output file is split across.
+       !In the past, many .f files were generated for each write.
+       !To be consistent with this the trailing 0 is still necessary today.
```
timfelle, line 528:
```suggestion
       ! The following string will specify that the files in the file series
       ! are defined by the filename followed by a 0.
       ! This 0 is necessary as it specifies the index of number of files
       ! the output file is split across.
       ! In the past, many .f files were generated for each write.
       ! To be consistent with this the trailing 0 is still necessary today.
```

timfelle (approved):

vbaconnet (approved):

PR #1727, opened 2025-02-22
Fix/dt to controller

pschlatt1:
added a grace 10% for the time-based controllers to enable a perfect hit for the controller. This allows for instance to write out fields at a perfect interval for fixed time steps. 10% should be sufficient. Main change is in subroutine time_based_controller_check.


timofeymukha:
Just as a side-note, the reason why I had to add dt to compute_ and preprocess_ for all simcomps, is because probes handles the output in its compute_ routine, so it needed the dt there to pass to the controller. I think we will soon use @MartinKarp time_struct for this everywhere, it is handy to just pass the whole time info usually.

timofeymukha:
Alright, we have a regression test fail, which means this change actually changes the way the CI cylinder simulation progresses. I guess this may be expected given the nature of the change. Then, we need to update the reference logs.

pschlatt1:
that is interesting, and we should look at in more details. Let me first have a look at the lid example and make that one final.

MartinKarp:
I think the unit test fails for FP64 since the update arguments have not been added threre.

For FP32 I guess it is quite interesting that the reference log differs. I guess we should have a look at that.

pschlatt1:
regarding the CI regression, indeed, the single precision of the old run was actually restarting like this:

   --Restarting from checkpoint--  
   File :   fluid00001.chkp
   Time :   0.5999999E-01
   
   and the new one like this:

   --Restarting from checkpoint--
   File :   fluid00001.chkp
   Time :   0.5000000E-01

which is the correct value for checkpointing every 0.05. So indeed, we need to update the log.

What goes wrong with the double precision run I cannot understand why another source is taken (without the dt). In any case, I guess I have reached the end of my knowledge here... just that I think the new version is actually correct.

pschlatt1:
please! I am not sure I would be able to fix the tests, so please help!

MartinKarp:
I ahve update the PR now, I guess the only weird behavior I see from this is upon restart where the number of executions of the time_based_controller after restart no longer necessarily is aligned with the previous number of exectuions - leading to that one may overwrite a file directly after restart (see the reference logs for the cylinder case).

MartinKarp:
I think that is unwanted behavior, so I will look into making some fixes for it today.

MartinKarp:
I fix this beahoir and incorporate time_t in PR #1803

PR #1730, opened 2025-02-24
Fix/lidexample

pschlatt1:
Updated the example with the lid-driven cavity. Adjusted the Reynolds number to be suitable for the given mesh, removed the simulation component to compute the vorticity, but added a direct computation into the user file, read the json parameter for the compuation interval from the case file. Removed also a factor 0.5 from the enstrophy definition.

What may be done for the example is to include the compuation of the streamfunction. Later it may be an excellent example for linear stability and adjoint calculcation, and optimal initial conditions.


timofeymukha:
You branched this off the branch which adds the dt as a controller argument. So, we need to merge that one in first.

timofeymukha:
@pschlatt1 I think the 0.5 is needed for enstrophy. Here is the definition in Frisch
image


OpenFOAM also uses 0.5
https://www.openfoam.com/documentation/guides/v2206/api/classFoam_1_1functionObjects_1_1enstrophy.html#details

pschlatt1:
yes, I realised that that branch was missing.

pschlatt1:
To me, the 1/2 in the energy makes sense as it comes from the integration of the momentum along a trajectory. Enstrophy however is not defined in that way, so the 1/2 seems to be there only due to some consistency to the kinetic energy. I don't mind.

pschlatt1:
so, I guess the final decision regarding enstrophy: 
image
we keep the factor 1/2, but I add a comment.

njansson:
> so, I guess the final decision regarding enstrophy: (this is from Hillewaert et al. for TGV) image we keep the factor 1/2, but I add a comment.

so we can merge this?

pschlatt1:
I guess you cannot merge as it was done on my dt branch.

But I guess I will close that one and do a completely new one which includes the comments and the 2D fix.

PR #1733, opened 2025-02-26
Change unnecessary intent(inout) into intent(in) in Ax

Shiyu-Sandy-Du:


timofeymukha (approved):

njansson:
I'm in favour of our code style rules, but for these kernels, the line break style we have adopted is not looking great, and makes it difficult to work with these kernels I would say

timfelle:
> I'm in favour of our code style rules, but for these kernels, the line break style we have adopted is not looking great, and makes it difficult to work with these kernels I would say

Well that is the backside of requiring a fixed style everywhere.

MartinKarp:
Would it be possible to not enforce an indentation level after line breaks?

timfelle:
> Would it be possible to not enforce an indentation level after line breaks?

Yes. By changing the indentation flag `-k5 -> -k-`  in the workflow script. That will stop the tool formatting the continuation lines.

MartinKarp:
From my perspective that would be fine to update out guidelines to that. As there are quite a few places where it increases readability and editing to start the lines at a different point.

timofeymukha:
Perhaps we can relax the k5 in the CI. But I would add the formatting script similar to the one I posted on zulip somewhere in the repo for devs to use, and *include* the -k5 there by default. People will likely use the script to fix formatting, and so in most cases k5 will be applied. Then, for files where we don't want it, one can manually run findent with -k-

timfelle:
> Perhaps we can relax the k5 in the CI. But I would add the formatting script similar to the one I posted on zulip somewhere in the repo for devs to use, and _include_ the -k5 there by default. People will likely use the script to fix formatting, and so in most cases k5 will be applied. Then, for files where we don't want it, one can manually run findent with -k-

Sure i can do that

MartinKarp:
If we could have both commands I think that would be great. With a suggestion to use the one with -k5

timfelle:
> If we could have both commands I think that would be great. With a suggestion to use the one with -k5

I will make a script which by default runs with the indentation of continuation lines. And then have an optional argument to skip that rule.

Shiyu-Sandy-Du:
I use the latest script but it still have some indent for line breaks in fld_file.f90...

MartinKarp:
What option did you use?

Shiyu-Sandy-Du:
-k-

timfelle:
So what the `-k-` does is that it will not modify the indentation of continuations which do not start with an `&`. So it will not undo you indentations if you do some yourself.

timfelle:
I guess for the `fld_file` it is the former of the two.

Shiyu-Sandy-Du:
Yes it is. And I don't think it's a big issue here because a line start with a & is usually the continuation of a string.

PR #1738, opened 2025-02-28
JSON subdictionary for setting initial conditions

gloopydoop:
This PR introduces a json sub-dictionary instead of hard coding `case.fluid` and `case.scalar` to allow more re-usability of `flow_ic` and `scalar_ic`.

**NOTE**
There is a change to how the Blasius parameters enters case file, so I hope this doesn't cause an inconvenience to people already using this BC.


timofeymukha:
Harry, there is probably a problem with the hemi in reframe. You have to change the case file template there.

timofeymukha:
So you are just changing where the config for blasius is, right? I think that is good, since it is no longer shared between the bcs and the ics.

pschlatt1:
btw, regarding Blasius in general: Should one not just add a correct Blasius profile? I would never use any of these approximations for a serious simulation.

njansson:
> btw, regarding Blasius in general: Should one not just add a correct Blasius profile? I would never use any of these approximations for a serious simulation.

Yes please do!

gloopydoop:
> Harry, there is probably a problem with the hemi in reframe. You have to change the case file template there.

Thanks :+1:

tests/reframe/src/minihemi.case.template
```diff
@@ -17,16 +17,14 @@
...
-        "inflow_condition": {
-            "type": "blasius"
-        },
-        "blasius": {
+            "type": "blasius",
             "approximation": "sin",
             "delta": 0.6,
             "freestream_velocity": [1.0, 0.0, 0.0]
         },
+        "inflow_condition": {
```
timofeymukha, line 25: By the way, this can be removed, because it is not read anymore. This stuff should be directly in the boundary_conditions dict.

src/case.f90
```diff
@@ -264,13 +265,15 @@ subroutine case_init_common(this)
     !
     call json_get(this%params, 'case.fluid.initial_condition.type',&
          string_val)
+    call json_extract_object(this%params, 'case.fluid.initial_condition', &
+       json_subdict)
```
timofeymukha, line 269: Unrelate to this PR, we should reasses as to why the case has to take care of fluids ics, and not fluid itself.

timofeymukha (approved):

timfelle (approved):

tests/reframe/src/minihemi.case.template
timfelle: Have this been resolved?

PR #1748, opened 2025-03-04
fixed the lid example to avoid 3D instabilities, read postprocessing …

pschlatt1:
…interval from case file and other cosmetic changes


timofeymukha:
Thanks, Philipp! I guess this goes back to whether we welcome python notebooks in the repo, due to their poor git-ability and eventual heavy weight. This was brought up some time ago in the context of @adperezm nice RBC tutorial PR (which we should merge in somewhere, by the way).

pschlatt1:
Yes, indeed. I saw there is a notebook also in the channel_TS case (ok, one by me), which is merged, so I thought it was ok. In fact, I think it is helpful for the user. What one could do regarding size is to remove the figures before committing, or just use plain Python. 

I think it is an important questions, as it has implications to other examples as well.

njansson:
As long as the notebooks aren't too big I think it's fine.

MartinKarp:
@pschlatt1 is this ok to review and merge?

MartinKarp (approved):

pschlatt1:
Thanks, looks good!

adperezm (approved):

PR #1753, opened 2025-03-06
Update smagorinsky_cpu.f90

muenschm:
As I can see s11, s22 ... are the components of the strain rate tensor basing on the classical definition: $$\frac{1}{2}(\frac{\partial u_i}{\partial x_j}+\frac{\partial u_j}{\partial x_i})$$. Thus, it is not double of the strain rate tensor (as considered in the Lilly paper for example) and the former comment was misleading.


Shiyu-Sandy-Du (approved):
Oh it was due to the lack of 1/2 in the old implementation of `subroutine strain_rate`. Thank you for correction here.

timofeymukha (approved):

PR #1757, opened 2025-03-07
Enable custom type injection to factories and a plugin architecture

timofeymukha:
I think I finally had a breakthrough in extending neko factories to user-defined types. In this demonstrator, I show how it works and also propose a prototype of a plugin architecture. Here, everything is shown using LES models as an example.

## Injecting types

The main idea is as follows.
* For built-in types the factories work exactly as before, with if-else statements.
* For user types, we introduce an registry of (name, allocator) pairs, here `les_model_registry`. An allocator is a simple subroutine that takes a pointer to a base type and allocates it to a custom derived type. Inside the factory, if the type_name is not among the KNOWN_TYPES (i.e. the built-in ones) it starts searching in this registry instead.
* The user is provided a `register_les_model` subroutine, which adds a (typename, allocator) a pair to the registry.
* **The tricky part** is, of course, performing this registration, i.e. calling the user routine that calls register_les_model ahead of time, before the factories are used in the code. This is where we have to rely on a convention.
We say that a user module called `my_module` should always add a `my_module_register_types` subroutine, in which the calls to `register_...` routines defined by the extendable types happen. We then modify `makeneko` to grep the .f90 files to figure out the module names and automatically inject `use` statements and calls to the `....register` routines. So we end up with something like.

```fortran
program usrneko
  use neko
  use user
  use test_module !!! Injected

  type(case_t), target :: C

  ! Register all user-defined types
  call user_setup(C%usr)
  ! Injection, custom type is now in the les_model_registry and ready to be allocated by the 
  ! factory
  call test_module_register_types() factory
  call neko_init(C)
  call neko_solve(C)
  call neko_finalize(C)

end program usrneko
```

and on the side of the custom module we have

```fortran
  !> The allocator for my_model_t
  subroutine custom_allocator(obj)
    class(les_model_t), allocatable, intent(inout) :: obj

    allocate(my_model_t::obj)
  end subroutine custom_allocator

  !> module name + register_types routine
  !! Can register all the custom types from the module here!
  subroutine test_module_register_types()
    procedure(les_model_allocate), pointer :: allocator_ptr

    allocator_ptr => custom_allocator  ! Assign procedure pointer

    write(*,*) "Registering the my_model_t SGS model"
    call register_les_model("my_model", allocator_ptr)
    write(*,*) "Done"

  end subroutine test_module_register_types
```

If a module does not inject any types, the routine is still used, but it should just be empty. I think this is ok for the huge benefit this provides.

## Plugins
Once I was there, it was easy to extend this to a primitive plugin architecture. 

* We define a NEKO_PLUGINS env var, where each folder will contain some .f90 and a *modules.txt*. The latter contains a list of all type-injecting modules inside the plugin.
* In the case directory we add a plugins.txt, with a list of plugins, just directory names inside NEKO_PLUGINS.
* Based on this, makeneko grabs all the .f90s from the requested plugins, and then does the type registration routine injection in exactly the same way as for the .f90s in the case directory. 

In the PR I just added a NEKO_PLUGINS right to src to have it somewhere, and there is an SGS model inside. 

I originally thought of going one step forward and have a way of compiling pluings into libraries, but I realized there are too many factors to consider and I suck at this stuff. Also, custom GPU code is not considered here and we should discuss this. But even as is, this is a huge step forward and I am very keen on getting this in in some form soon. The plugins and the local .f90 can be treated as separate tracks later, one a natural continuation of the other.

Feel free to play with the provided SGS model example in `tgv`!

### DIFF
--- a/NEKO_PLUGINS/my_model2/modules.txt
+++ b/NEKO_PLUGINS/my_model2/modules.txt
@@ -1,0 +1,1 @@
+my_model2
--- a/NEKO_PLUGINS/my_model2/my_model2.f90
+++ b/NEKO_PLUGINS/my_model2/my_model2.f90
@@ -1,0 +1,121 @@
+module my_model2
+  use num_types, only : rp
+  use les_model, only : les_model_t, register_les_model, les_model_allocate
+  use field, only : field_t
+  use fluid_scheme_base, only : fluid_scheme_base_t
+  use json_utils, only : json_get_or_default
+  use json_module, only : json_file
+  use field_registry, only : neko_field_registry
+  use logger, only : LOG_SIZE, neko_log
+  use field_math, only : field_cfill
+
+  implicit none
+
+  !> Implements a dummy user model
+  type, public, extends(les_model_t) :: my_model2_t
+     !> Model constant, defaults to 0.07.
+     real(kind=rp) :: c
+   contains
+     !> Constructor from JSON.
+     procedure, pass(this) :: init => my_model2_init
+     !> Constructor from components.
+     procedure, pass(this) :: init_from_components => &
+          my_model2_init_from_components
+     !> Destructor.
+     procedure, pass(this) :: free => my_model2_free
+     !> Compute eddy viscosity.
+     procedure, pass(this) :: compute => my_model2_compute
+  end type my_model2_t
+
+contains
+  !> Constructor.
+  !! @param fluid The fluid_scheme_base_t object.
+  !! @param json A dictionary with parameters.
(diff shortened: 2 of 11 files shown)


njansson:
Nice! Can't we make a separate example for this? to keep tgv rather clean

timofeymukha:
> Nice! Can't we make a separate example for this? to keep tgv rather clean

Nevermind the TGV, this is just to demonstrate things and people to play with. Once we agree on the approach, I will start working on a clean PR.

timofeymukha:
I will close, but we can look at this later when we think of plugins/libs

PR #1759, opened 2025-03-07
BCs for compressible solver

tuananhdao:
## Changes

- Add support for various boundary conditions in the compressible solver. All are strongly imposed. These should also work on GPUs
    - `velocity_value`
    - `density_value`
    - `pressure_value`
    - `no_slip`
    - `symmetry`
    - `outflow`/`normal_outflow`
- Add 2 examples
    - Wind tunnel with a forward-facing step
    - 2D flow over cylinder

These changes shouldn't affect any incompressible modules.

## Case file

Boundary conditions are defined in `fluid` similarly to the incompressible solver. There are some differences to allow flexibility:
- Users don't have to define BCs on all boundaries; and
- Multiple BCs can be defined on the same zone.

```json
...
        "boundary_conditions": [
            {
                "type": "velocity_value",
                "zone_indices": [1],
                "value": [1, 0, 0]
            },
            {
                "type": "density_value",
                "zone_indices": [1],
                "value": 1
            },
            {
                "type": "no_slip",
                "zone_indices": [7]
            },
            {
                "type": "symmetry",
                "zone_indices": [3,4,5,6]
            }
        ]
...
```

## Mach-3 wind tunnel with a forward-facing step

Pressure
Wind tunnel with a forward-facing step

## 2D Flow over cylinder

Velocity magnitude
Cylinder


njansson (approved):
🍻 🎉 🍻

njansson:
Isn't this ready to be merged (asking due to the don't merge label?)

tuananhdao:
> Isn't this ready to be merged (asking due to the don't merge label?)

I want to write a bit more details to the readme files of the cases.

njansson:
> > Isn't this ready to be merged (asking due to the don't merge label?)
> 
> I want to write a bit more details to the readme files of the cases.

Ok! I'll disable auto merge then

njansson:
BTW very nice! 🍻

timofeymukha:
The code itself is fine. But we need to add the docs, at least at some point, when you want the users to try the solver :-).

I have a question: how do you choose the variables to set the bcs for? Is it good to have pressure / density bcs, or is it better to have e.g. temperature and density? I guess there is some freedom here based the fact that you have the equation of state coupling? In the PR there are no temperature bcs, right?

tuananhdao:
> The code itself is fine. But we need to add the docs, at least at some point, when you want the users to try the solver :-).
> 
> I have a question: how do you choose the variables to set the bcs for? Is it good to have pressure / density bcs, or is it better to have e.g. temperature and density? I guess there is some freedom here based the fact that you have the equation of state coupling? In the PR there are no temperature bcs, right?

Hi,

Thanks! I have that in mind. I want to make 1 more PR to implement entropy viscosity + adaptive time-stepping. Then I think it's good to ship.

Good question! I think we should add support for temperature BCs as well. For the toy problems I have in mind, it’s easier to set density, velocity, and pressure at the inlet, letting temperature follow from the equation of state. Since I’m assuming an ideal gas, temperature and pressure are interchangeable.

timofeymukha (approved):

PR #1735, opened 2025-02-27
Add linting and formatting scripts with README updates

timfelle:
Introduce a linting script for Fortran files and a formatter script. Update the README to provide usage instructions for both tools.

Additionally we stop forcing people to format continuation lines to 5 spaces

PR #1762, opened 2025-03-10
See if we can simplify the ksp factory

### LINKED ISSUE #1313: The ksp factory has a redundant select type
Uses an extra `select type` to call `init`, whereas we should in principle be able to just call `%init` on the `object` variable without a `select type`. We should see if we can get away with that.

timofeymukha:
See issue #1313

* I introduce a deferred init for ksp_t. Some constructors had `target` on this, and some didn't. I chose to have `target` in the shared interface.
* Remove the `krylov_solver_destroy` subroutine. We should be able to just call %free on the object.
* Remove the `select` type in the factory and use the shared `init`.

Caveat:
GMRES and CACG had additional paramters to their init, which had to be removed :(. I just set the default values in the type definition. The code never used any values but the defaults though, since I didn't have to change a single constructor call. 
We could add additional constructors for those type, which account for the extra paramters, if we want to.


timofeymukha:
@timfelle I think the formatter is counting my correct 5 space indentation (produced by our formatting script) as wrong, or?

timfelle:
> @timfelle I think the formatter is counting my correct 5 space indentation (produced by our formatting script) as wrong, or?

That is indeed strange. I will look a bit into it

timfelle:
> > @timfelle I think the formatter is counting my correct 5 space indentation (produced by our formatting script) as wrong, or?
> 
> That is indeed strange. I will look a bit into it

Alright, this quirk come from the change in the end of february, #1735 

In that, we stopped checking the indentation of continuation lines. However, it seems the continuation lines preceded by a `&` still gets checked and now we cannot control how deep that indentation should be, since we do not check the continuation lines in general. 

It comes from the combination flags `-k- --indent_ampersand`. I have not been able to adjust them to resolve your issue, without turning it back on, which we clearly did not want as strictly as this check is.

njansson:
Why do we need target now?

timofeymukha:
> Why do we need target now?

You tell me :-). It was there in some of the constructors. And for some types, we also have it, for example in fluid. You have added those at some point after going thourgh with NAG.

```
 abstract interface
     subroutine fluid_base_init_all_intrf(this, msh, lx, params, kspv_init, &
          kspp_init, scheme, user)
       import fluid_scheme_base_t
       import mesh_t
       import json_file
       import user_t
       import rp
       import LOG_SIZE
       class(fluid_scheme_base_t), target, intent(inout) :: this
```

timofeymukha:
> > > Why do we need target now?
> > 
> > 
> > You tell me :-). It was there in some of the constructors. And for some types, we also have it, for example in fluid. You have added those at some point after going thourgh with NAG.
> > ```
> >  abstract interface
> >      subroutine fluid_base_init_all_intrf(this, msh, lx, params, kspv_init, &
> >           kspp_init, scheme, user)
> >        import fluid_scheme_base_t
> >        import mesh_t
> >        import json_file
> >        import user_t
> >        import rp
> >        import LOG_SIZE
> >        class(fluid_scheme_base_t), target, intent(inout) :: this
> > ```
> 
> Sure for fluid and friends (and several other places) they are needed. But for `ksp_t` derived type I don't recall that we need it. We never assign anything to these objects using a pointer

This
```
    if (present(M)) then
       this%M => M
    else
```
appears to want the target on `this`!

src/scalar/scalar_scheme.f90
```diff
@@ -540,21 +539,21 @@ subroutine scalar_scheme_set_material_properties(this, params, user)
...
        call user%material_properties(0.0_rp, 0, dummy_rho, dummy_mu, &
             this%cp, this%lambda, params)
     else
        if (params%valid_path('case.scalar.Pe') .and. &
             (params%valid_path('case.scalar.lambda') .or. &
             params%valid_path('case.scalar.cp'))) then
-          call neko_error("To set the material properties for the scalar,&
-               & either provide Pe OR lambda and cp in the case file.")
+          call neko_error("To set the material properties for the scalar, // &
+               either provide Pe OR lambda and cp in the case file.")
```
timfelle, line 552:
```suggestion
          call neko_error("To set the material properties for the scalar, " // &
               "either provide Pe OR lambda and cp in the case file.")
```

src/scalar/scalar_scheme.f90
```diff
@@ -540,21 +539,21 @@ subroutine scalar_scheme_set_material_properties(this, params, user)
...
           ! Non-dimensional case
        else if (params%valid_path('case.scalar.Pe')) then
-          write(log_buf, '(A)') 'Non-dimensional scalar material properties &
-               & input.'
+          write(log_buf, '(A)') 'Non-dimensional scalar material properties // &
+               input.'
```
timfelle, line 556:
```suggestion
          write(log_buf, '(A)') 'Non-dimensional scalar material properties ' // &
               'input.'
```

src/scalar/scalar_scheme.f90
```diff
@@ -540,21 +539,21 @@ subroutine scalar_scheme_set_material_properties(this, params, user)
 
     if (.not. associated(user%material_properties, dummy_mp_ptr)) then
 
-       write(log_buf, '(A)') "Material properties must be set in the user&
-            & file!"
+       write(log_buf, '(A)') "Material properties must be set in the user // &
+            file!"
```
timfelle, line 543:
```suggestion
       write(log_buf, '(A)') "Material properties must be set in the user " // &
            "file!"
```

timofeymukha:
Hmm, what happened to the CI :).

But I guess for the target the conclusion is that we need it. And it is indeen present in ksp_t currently on develop.

timofeymukha:
I think we can merge. Or are there any reservations?

MartinKarp (approved):
I think this is nice. I guess if we want to start supporting more options for the different KSP (restart size in GMRES, number of s-steps in CACG, etc) we can solve that later. I really thought we could choose lgmres in the case file though, but apprantly this has not been possible for quite some time.

timfelle (approved):

PR #1763, opened 2025-03-10
Fix 0 length math calls on GPU and add gather and scatter masked ops

MartinKarp:
- Adds checks in device_math to avoid launching kernels with size 0 which yields segfaults.
- Adds the masked_scatter and masked_gather subroutines to conveniently gather an array into a shorter one according to a mask and then scatter it back again.
- Refactoring of the kernels issuing collective communication calls as they need special treatment as not to deadlock in allreduce.


src/math/bcknd/device/opr_device.F90
```diff
@@ -651,10 +651,10 @@ subroutine opr_device_curl(w1, w2, w3, u1, u2, u3, work1, work2, c_Xh, event)
...
 
     call device_opcolv(w1%x_d, w2%x_d, w3%x_d, c_Xh%B_d, gdim, n)
-    call c_Xh%gs_h%op(w1, GS_OP_ADD, event)
-    call c_Xh%gs_h%op(w2, GS_OP_ADD, event)
-    call c_Xh%gs_h%op(w3, GS_OP_ADD, event)
-    call device_stream_wait_event(glb_cmd_queue, event, 0)
+    call c_Xh%gs_h%op(w1, GS_OP_ADD)
+    call c_Xh%gs_h%op(w2, GS_OP_ADD)
+    call c_Xh%gs_h%op(w3, GS_OP_ADD)
+    !call device_stream_wait_event(glb_cmd_queue, event, 0)
```
njansson, line 657: Was there a reason for removing the event? If so we need to find out how to fix it without omitting event to gs

src/math/bcknd/device/cuda/math.cu
```diff
@@ -567,42 +539,97 @@ extern "C" {
       CUDA_CHECK(cudaMalloc(&bufred_d, nb*sizeof(real)));
 #endif
     }
+  }
+
+  /**
+   * Global additive reduction
+   */
+  void cuda_global_reduce_add(real * bufred, void * bufred_d, int n, const cudaStream_t stream) {
```
njansson, line 547: Is `n` here between 0 and 1? Otherwise something seems to be a bit strange after the warp shuffle things?

MartinKarp:
There are a few things missing in the PR, will push when I have time. global_interp branch has the most updated changes.

timofeymukha (approved):

adperezm (approved):

PR #1764, opened 2025-03-11
update the documentation of the option of extrapolation in les

Shiyu-Sandy-Du:
It's simply an update of the documentation


pschlatt1:
regarding: Computation of subgrid scale velocity via an LES model \ref simcomp_les_model
a comment: I don't think that there is a "subgrid scale velocity", these are subgrid-scale stresses. Also, I would not call the model used in an LES to be an "LES model" but rather an "SGS model".

Shiyu-Sandy-Du:
regarding the stresses, maybe using "eddy viscosity" is a better term? because it's what we actually compute in simcomp. Regarding the name of "les_model" or "sgs_model", doesn't it include a huge amount of code modification regarding the name?

Shiyu-Sandy-Du:
Yes, maybe changing the document is of minor pain for now. Otherwise it would kill all users' runs

MartinKarp (approved):

vbaconnet (approved):

PR #1765, opened 2025-03-11
Add ADIOS2 as transport layer and second file I/O to enable fluid data compression.

gregorweiss:
This PR adds ADIOS2 as a second file format and transport layer for I/O. It was introduced to enable usage of the ADIOS2 supported compressors, such as ZFP, SZ, and BigWhoop. Using compressors based on image compressors the local structure from higher order elements has to be exploited. Hence, there are three data layouts that can be used.

Accordingly, the case file now takes two more parameters
```
  "output_format": "adios2",
  "output_layout": 2,
```

The default for `output_format` is `nek5000` and opts for the classic fld files. If `adios2` is chosen the fld files are replaced by adios' bp files which contains data fields depending on the layout.

The `output_layout` takes three cases:
`1` = 1D layout for every field (just like in MPI-IO for fld-files)
`2` = 4D layout for every field
`3` = 4D layout with field/parameter packing; all fields are packed together into one variable called `fields` in the bp-file. BigWhoop balance and leverage compression efficiency and loss between parameters for a given target compression rate.

The 4D layout will have dimensionality 8x8x8x#elements if polynomial_order was set to 7 to represent the local structure in the initial three dimensions and the fourth dimension lists the number of elements. A 4D layout should be used for image compression based compressors, while the field packing is only supported by BigWhoop.

The compression algorithm and settings are currently controlled by a `adios2.xml` file. Moving such settings into the Neko case file is possible but would add a lot more additional compressor dependent options.

Additionally, a converter tool `adios2_to_nek5000` is added and a tool `psnr` can be used to calculate and estimate the compression efficiency.


adperezm:
Hi!

Very nice work! Before I go in depth into the files, I want to ask a couple of things:

- I understand that Paraview has a reader that can directly take BP files from nek type codes. Is this header and file arrangement compatible with that?
- How has your experience been with ADIOS2-fortran on clusters? In the past, I always had trouble compiling it on cray computers. Just asking since I do not like that once we merge this, we will have two ADIOS2 workflows that do not support each other (The other one uses C++ bindings). If ADIOS2 fortran is reliable, then I would change the other object to also use it.

gregorweiss:
Both are very interesting questions.

I know that Paraview has and grows support for BP file reading and writing. But I didn't know that nek-type bp-files would be readable right away. Does this use [Fides](https://gitlab.kitware.com/vtk/fides) or does Paraview natively support this nek-type bp-file reading? If you can point me to resources that describe the data structure and layout thereof, I can check if this PR could be made conformant.

Right now it conforms with the feel of fld output. There are bp-files for each time step and a descriptor file `field.adios2` file that is equivalent to `field.nek5000`.
```
field0.00000.bp
field0.00001.bp
field0.00002.bp
field0.00003.bp
field0.adios2
```
```
filetemplate:         field%01d.%05d.bp
firsttimestep:     0
numtimesteps:     4
type: adios2-bp                
```

Regarding CPE and ADIOS2-Fortran I can check. So far I have deployed this branch with HPE-MPT compiler on Hawk at HLRS. But with the HLRS's new system Hunter, I will check for compatibilities with its native Cray compilers. 

In fact,  I encountered some difficulties deploying and using the readily available ADIOS2-C++ wrappers and bindings in Neko on Hawk, which is why I added the second ADIOS2-Fortran thirdparty dep in the hopes of having less hassle with the C++ to Fortran interfacing.

adperezm:
> I know that Paraview has and grows support for BP file reading and writing. But I didn't know that nek-type bp-files would be readable right away. Does this use [Fides](https://gitlab.kitware.com/vtk/fides) or does Paraview natively support this nek-type bp-file reading? If you can point me to resources that describe the data structure and layout thereof, I can check if this PR could be made conformant.

Oh, now that you mentioned it, the example I am thinking of used Fides. We had communications with people at ORNL and Kitware and I know that ONE person was able to visualize directly from bp adding the FIDES schema as attributes. Since maybe this was just a test on their part, this is perhaps not crucial right now. I think we can add this and then update if needed.

> In fact, I encountered some difficulties deploying and using the readily available ADIOS2-C++ wrappers and bindings in Neko on Hawk, which is why I added the second ADIOS2-Fortran thirdparty dep in the hopes of having less hassle with the C++ to Fortran interfacing.

That is funny :smiley: . It seems there will always be problems haha. Then I really appreciate if you check for compiling problems in other platforms. But I think that anyway I will go through the files and we merge when the tests pass and all reviewers approve.

gregorweiss:
I can compile ADIOS2 with CPE where one has to disable Fortran submodules like this:
```
CC=cc CXX=CC FC=ftn cmake .. -DADIOS2_USE_Fortran=ON -DADIOS2_HAVE_FORTRAN_SUBMODULES=False
```
That still gives me warnings because of implicit module inclusions but it compiles completely. So I could test it in Neko.

Though I am having troubles compiling Neko with CPE. It tries to look for gcc compilers:
```
./configure FC=ftn MPIFC=ftn CC=cc MPICC=cc --prefix=...
...
checking for gcc... cc
checking whether the C compiler works... no
configure: error: in `/zhome/academic/HLRS/hlrs/hpcgweis/code/neko/source':
configure: error: C compiler cannot create executables
```
How do you compile Neko with CPE compilers?

njansson:
> I can compile ADIOS2 with CPE where one has to disable Fortran submodules like this:
> 
> ```
> CC=cc CXX=CC FC=ftn cmake .. -DADIOS2_USE_Fortran=ON -DADIOS2_HAVE_FORTRAN_SUBMODULES=False
> ```
> 
> That still gives me warnings because of implicit module inclusions but it compiles completely. So I could test it in Neko.
> 
> Though I am having troubles compiling Neko with CPE. It tries to look for gcc compilers:
> 
> ```
> ./configure FC=ftn MPIFC=ftn CC=cc MPICC=cc --prefix=...
> ...
> checking for gcc... cc
> checking whether the C compiler works... no
> configure: error: in `/zhome/academic/HLRS/hlrs/hpcgweis/code/neko/source':
> configure: error: C compiler cannot create executables
> ```
> 
> How do you compile Neko with CPE compilers?

Hmm, what does `config.log` say the error is?

gregorweiss:
Good point.

config.log states problems with pkg-config
```
Error invoking pkg-config!
Package dsmml was not found in the pkg-config search path.
Perhaps you should add the directory containing `dsmml.pc'
to the PKG_CONFIG_PATH environment variable
Package 'dsmml', required by 'virtual:world', not found
Package 'xpmem', required by 'virtual:world', not found
Package 'dsmml', required by 'virtual:world', not found
configure:4075: $? = 1
configure:4113: result: no
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "neko"
| #define PACKAGE_TARNAME "neko"
| #define PACKAGE_VERSION "0.9.99"
| #define PACKAGE_STRING "neko 0.9.99"
| #define PACKAGE_BUGREPORT ""
| #define PACKAGE_URL ""
| #define PACKAGE "neko"
| #define VERSION "0.9.99"
| /* end confdefs.h.  */
|
| int
| main ()
| {
|
|   ;
|   return 0;
| }
configure:4118: error: in `/zhome/academic/HLRS/hlrs/hpcgweis/code/neko/source':
configure:4120: error: C compiler cannot create executables
```

gregorweiss:
[config.log](https://github.com/user-attachments/files/19208577/config.log)

gregorweiss:
Ok. I had my PKG_CONFIG_PATH messed up as obviously stated in the config.log. After fixing it, everything went smoothly.

To summarize: ADIOS2 with Fortran bindings works in Neko using the CPE on Hunter. The TGV case could be switched to the bp-file format of this PR. Only thing to take care is to disable Fortran submodules `-DADIOS2_HAVE_FORTRAN_SUBMODULES=False` during ADIOS2 configuration.

I hope that helps to merge Neko features with the Fortran interface of ADIOS that currently use the C++ interface.

timfelle:
@njansson and @gregorweiss I have now brought it up to date with the latest develop. Can you verify that is till works as intended. There was a single conflict on how the filenames were passed to the output controller.

adperezm:
I will review this very soon and hopefully we can merge by next week. Apologies for the delay!

gregorweiss:
Thank you @timfelle for your efforts!

> ... Can you verify that it still works as intended. 

I checked this at HLRS, and it still works.

njansson (approved):

Issue #1766, opened 2025-03-12
Segfault in connection to the re-implementation boundary conditions


muenschm:
Currently neko segfaults before entering the first time-step. This issue is connected to the re-implementation of the boundary condition, i.e., the new definition of the boundary conditions in the case-file and probably the initialization during startup. The problem was detected basing on the periodic hill case but can be reproduced in the same way by considering /examples/turb_channel. In the turb_channel (or the periodic hill case), the no-slip b.c. are now set via:

 "boundary_conditions": [
            {
                "type": "no_slip",
                "zone_indices": [3, 4]
            }
          ],

Checking zone_indices0.f00000 shows the correct zone indices (3 and 4) for the no-slip surfaces. When not defining b.c. for these surfaces by uncommenting the upper lines in the case file, neko enters the time-loop. But of course, bdry0.f00000 shows a value of 0 for the surfaces, i.e., no b.c. is active here. For the surfaces connected to periodic b.c. bdry0.f00000 shows the right value.

P.S.: Why it is called zone_indice and not surface or face_indice?

timofeymukha:
Yeah, so if you comment out like this

```
"boundary_conditions": [
//{
//"type": "no_slip",
//"zone_indices": [3, 4]
//}
],
```
then you would get an error that zone 3 has received no boundary condition.
But if you comment out the whole boundary_conditions entry, then you are allowed to proceed, with the implicit assumption that everything is periodic, like in tgv.

I guess what we need to do is check that when no boundary_conditions entry is provided, that the entire boundary is marked in periodic zones. Not sure how easy that is, should try. Another alternative is to demand "boundary_conditions" to always be present, and left empy in cases like tgv.

njansson:
Fixed in #1767

PR #1767, opened 2025-03-12
Check for boundary_conditions entry when labled zones present

timofeymukha:
Fixes issue #1766 

I guess we only have two options now: labels or periodic zones. If there are labled zones, boundary conditions must be provided.

So, this checks that there are no labeled zones in the mesh if the boundary_conditions keyword is not provided, and then throws an error.

### DIFF
--- a/src/fluid/fluid_pnpn.f90
+++ b/src/fluid/fluid_pnpn.f90
@@ -1034,9 +1034,7 @@ contains
        ! Check that there are no labeled zones, i.e. all are periodic.
        do i = 1, size(this%msh%labeled_zones)
           if (this%msh%labeled_zones(i)%size .gt. 0) then
-             write(error_unit, '(A, A, I0)') "*** ERROR ***: ", &
-                  "No boundary_conditions entry in the case file!"
-             error stop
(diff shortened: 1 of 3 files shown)


src/fluid/fluid_pnpn.f90
```diff
@@ -1030,6 +1030,16 @@ subroutine fluid_pnpn_setup_bcs(this, user, params)
           end if
 
        end do
+    else
+       ! Check that there are no labeled zones, i.e. all are periodic.
+       do i = 1, size(this%msh%labeled_zones)
+          if (this%msh%labeled_zones(i)%size .gt. 0) then
+             write(error_unit, '(A, A, I0)') "*** ERROR ***: ", &
+                  "No boundary_conditions entry in the case file!"
+             error stop
```
timfelle, line 1039: Why not use `neko_error`?

timofeymukha: Oh, true.. this is due to a copy-paste..

MartinKarp (approved):
Tested on cylinde ron 8 GPUs.

njansson (approved):

PR #1768, opened 2025-03-12
Fix SEGV in vorticity simcomps due to missing event

njansson:


src/math/operators.f90
```diff
@@ -383,7 +383,12 @@ subroutine curl(w1, w2, w3, u1, u2, u3, work1, work2, coef, event)
     else if (NEKO_BCKND_XSMM .eq. 1) then
        call opr_xsmm_curl(w1, w2, w3, u1, u2, u3, work1, work2, coef)
     else if (NEKO_BCKND_DEVICE .eq. 1) then
-       call opr_device_curl(w1, w2, w3, u1, u2, u3, work1, work2, coef, event)
+       if (present(event)) then
```
MartinKarp, line 386: You can remove this check for present and just pass the event down. It will be checked in the device routine correctly.

MartinKarp (approved):

tuananhdao (approved):

PR #1771, opened 2025-03-12
Some imporvements to scalar

timofeymukha:
- Fixes the scalar_mms example (need to set bcs for the fluid now).
- Adds check for boundary_conditions keywords, similar to the fluid fix earlier today.
- The scalar no longer searches in case.scalar for parameters, so we extract the case.scalar subdict. This is not needed now, but I am prepping for the possibility to init a scalar from another object, making the case.scalar hardcoded pathing incorrect.
- The advection factory needs the numerics subdict of the parameter file, so we pass that as an additional argument to the scalar constructor. In turn, the advection factory no longer searches in case.numerics. This change is also reflected in `fluid_pnpn`. @tuananhdao I tried to see if I need to fix the compressible solver, but it seems like the `adv` component is not used?


njansson (approved):
🍶 🍣  🍶

tuananhdao (approved):

timofeymukha:
This needs some love because the params are passed to some user routines.

PR #1774, opened 2025-03-13
Add user_startup routine, make case public in neko_simcomps

timofeymukha:
The PR contains some improvements for user-file lovers.

- I make the neko_simcomps%case object public again, so that the user file is in "god mode" via it.
- I add docstrings to the user file .f90 and do a bit of refactoring and reformatting.
- A new user routine is added, called `user_startup`. This is run right at the beginning of the case init and only gets the case file JSON as input. The purpose is to either hack into the params before they are passed to constructors, or define user-defined variables of built-in type (constants in particular). A case in point is the material property constants in the rayleigh and RBC examples. I guess this will be a welcome addition. What prompted me to add this is that in the PR with scalar improvements, the material_properties routine no longer received the full case params, so doing things as before in the examples above did not work.

### DIFF
--- a/doc/pages/user-guide/user-file.md
+++ b/doc/pages/user-guide/user-file.md
@@ -215,12 +215,9 @@ example](https://github.com/ExtremeFLOW/neko/blob/564686b127ff75a362a06126c6b23e
     integer, intent(in) :: tstep
     real(kind=rp), intent(inout) :: rho, mu, cp, lambda
     type(json_file), intent(inout) :: params
-    real(kind=rp) :: Re
 
-    call json_get(params, "case.fluid.Ra", Ra)
-    call json_get(params, "case.scalar.Pr", Pr)
+    ! Re and Pr computed in `user_startup`
 
-    Re = sqrt(Ra / Pr)
     mu = 1.0_rp / Re
     lambda = mu / Pr
     rho = 1.0_rp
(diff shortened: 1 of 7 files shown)


doc/pages/user-guide/user-file.md
```diff
@@ -78,28 +78,50 @@ executable `bin/neko` in your local neko installation folder.
...
+  dirichlet values e.g. `d=1`. For more information on the scalar, see the
+  [relevant section of the case file](@ref case-file_scalar).
 
 ### Initializing and finalizing {#user-file_init-and-final}
 
-The two subroutines `user_init_modules` and `user_finalize_modules` may be used
-to initialize/finalize any user defined variables, external objects, or
-processes. They are respectively executed right before/after the simulation time
-loop.
+Threeo subroutines `user_startup`, `user_init_modules` and
```
vbaconnet, line 100:
```suggestion
Three subroutines `user_startup`, `user_init_modules` and
```

njansson (approved):

timfelle (approved):

MartinKarp:
Nice! Would it be possible to have this merged witht the init modules or they need to happen at different times somewhat?

timofeymukha:
> Nice! Would it be possible to have this merged witht the init modules or they need to happen at different times somewhat?

@MartinKarp  Yes, at different times. The new routine is called very early, nothing is initialized, so you cannot have a dofmap for example. But you can define constants etc that are needed during init of objects (e.g. in material properties, which is run in the constructor). Or you can completely hack the json file in-memory, if you so wish.  The init_modeles is ran after the case is fully constructed, so you can define fields etc, based on the data in the case at hand.

PR #1776, opened 2025-03-13
Fix missing weak_grad in simcomp factory

timofeymukha:
And the example in the .md.

### DIFF
--- a/src/simulation_components/simulation_component_fctry.f90
+++ b/src/simulation_components/simulation_component_fctry.f90
@@ -44,6 +44,7 @@ submodule (simulation_component) simulation_component_fctry
   use weak_grad, only : weak_grad_t
   use derivative, only : derivative_t
   use spectral_error, only: spectral_error_t
+  use utils, only : neko_type_error
 
   ! List of all possible types created by the factory routine
   character(len=20) :: SIMCOMPS_KNOWN_TYPES(9) = [character(len=20) :: &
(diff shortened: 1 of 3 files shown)


njansson (approved):

src/simulation_components/simulation_component_fctry.f90
```diff
@@ -97,11 +98,11 @@ module subroutine simulation_component_factory(object, json, case)
...
-       type_string =  concat_string_array(SIMCOMPS_KNOWN_TYPES, &
-            NEW_LINE('A') // "-  ",  .true.)
+       type_string = concat_string_array(SIMCOMPS_KNOWN_TYPES, &
+            NEW_LINE('A') // "-  ", .true.)
        call neko_error("Unknown simulation component type: " &
-                       // trim(type_name) // ".  Known types are: " &
-                       // type_string)
+            // trim(type_name) // ".  Known types are: " &
+            // type_string)
        stop
```
timfelle, line 106: How about using your neat little type error function instead?
Remember the use statement
```suggestion
       call neko_type_error("simulation component", trim(type_name), &
            SIMCOMPS_KNOWN_TYPES)
```

src/simulation_components/simulation_component_fctry.f90
```diff
@@ -97,11 +98,11 @@ module subroutine simulation_component_factory(object, json, case)
     else if (trim(type_name) .eq. "spectral_error") then
```
timfelle, line 98: Could we move this to select now we are at it? 

```.f90
select case (trim(type_name))
case ("vorticity")
    allocate(vorticity_t::object)
! ....
case default
    call neko_type_error(!...
```

timfelle (approved):
Great 🥳

PR #1779, opened 2025-03-14
Redesign material_properties user routine and their handling in schemes

timofeymukha:
The following changes are introduces to the material properties handling
- No components in `_scheme` for constant values, both scalar and fluid have `field_t` for material properties.
- A helper `material_properties` field list is used to collect pointers to all the material properties of a scheme together. Used in the user routine, see below.
- The `pnpn` schemes currently hard-code using a constant `rho` and it is also used in the `rhs_maker`. Addressing the latter is left for a future PR, currently we just pass e.g. `this%rho%x(1,1,1,1)`. This is not different from what we had before in terms of functionality.
- The schemes call their own update routines for material properties, which calls the user routine, and, when needed, adds the turbulent component to diffusivity.
- simulation.f90 is cleaned up.
- The density of the scalar is changed to be a pointer to a field, now pointing to the fluid, so updates in the fluid density are respected by the scalar.
- A new keyword for fluid is introduced: `full_stress_formulation`, which changes the governing equations to have the full form of the viscous stress tensor. False by default.

The following changes are introduced to the user material_properties scheme
* Adds `name` to distinguish between callees. The fluid and scalar send their `name` component.
* A properties field list is received and manipulated. Can be of arbitrary size in case new `schemes` use it. For incompressible fluid, the first field is density and the second is mu. For the scalar, the first is cp, and the second lambda.

Finally, the rayleigh_bendard and rayleigh_benard_cylinder examples are updated to showcase the new stuff.


doc/pages/user-guide/user-file.md
```diff
@@ -210,13 +210,27 @@ example](https://github.com/ExtremeFLOW/neko/blob/564686b127ff75a362a06126c6b23e
...
+    ! If we are called from the scalar
+    else if (field_name .eq. "s") then
+       ! cp
+       properties(1) = 1.0_rp
+       ! lambda
+       properties(2) = mu / Pr
+    end if
 
     mu = 1.0_rp / Re
     lambda = mu / Pr
```
timfelle, line 236: should this not be above the if statements?

timfelle (commented):
Why use `u` to pass from the incompressible fluid? How about using `fluid` or something more descriptive, since I assume you would want the same parameters for the other directions of the fluid. Or do you set them per direction, so that u, v, w are all set independently?

src/scalar/scalar_scheme.f90
```diff
@@ -511,12 +514,21 @@ subroutine scalar_scheme_precon_factory(pc, ksp, coef, dof, gs, bclst, &
...
+    integer, intent(in) :: tstep
     type(field_t), pointer :: nut
     integer :: n
     ! Factor to transform nu_t to lambda_t
     real(kind=rp) :: lambda_factor
+    real(kind=rp) :: properties(2)
+
+    properties(1) = this%cp
+    properties(2) = this%lambda
+    call this%user_material_properties(t, tstep, this%u%name, properties)
```
timfelle, line 529: How about introducing a `this%name`?

MartinKarp (commented):
This is really great Timofey!

Should we also accomodate for setting the mu_field to a variable property in space or should we do this in another PR?

src/scalar/scalar_scheme.f90
timofeymukha: Yep, this is a bug :P.

I actually like the idea of this%name, especially for simcomps.

timfelle: It will help keep track of multiple schemes as well, multiple scalars and fluids interacting.

timofeymukha: Added!

timofeymukha:
(dont merge, work still in progress)

timofeymukha:
The scope of the PR is expanding after a discussion with @MartinKarp, so one can ignore it for now.

timofeymukha:
I am re-opening this one for review.

timfelle (approved):

timofeymukha:
Are there? I should take a look then, it should have been OK.

timofeymukha:
I could reproduce with CUDA. The CPU backened seems to work fine.

timofeymukha:
@timfelle @MartinKarp @njansson  OK, things should be working now hopefully.

The issue with GPUs was the usage of `rho%x(1,1,1,1)` and such in some of the residual interfaces, and in the force_torque simcomp. Essentially, we have to copy-over some material properties back to the host at each timestep. For `lambda` things are already clean, but not yet for `mu`, but should be fixed later. For `rho` and `cp` things are a bit less clear. I guess one could set the size to 1 in calls like, since only the first element is used as a stand-in for a const value. Is that valuable?

```
       call device_memcpy(this%cp%x, this%cp%x_d, this%cp%size(), &
            DEVICE_TO_HOST, sync=.false.)
```

I also removed all hand-holding for setting the full stress Ax as per discussion with @pschlatt1 . You have `full_stress_formulation` as a keyword now, that you have to set to true if you want it. Will crash if you don't do that and have a `nut_field` though, since that is just wrong.

njansson:
> @timfelle @MartinKarp @njansson OK, things should be working now hopefully.
> 
> The issue with GPUs was the usage of `rho%x(1,1,1,1)` and such in some of the residual interfaces, and in the force_torque simcomp. Essentially, we have to copy-over some material properties back to the host at each timestep. 

Great! this copy back and forth was the reason for those ugly lines in simulation before.

> For `lambda` things are already clean, but not yet for `mu`, but should be fixed later. For `rho` and `cp` things are a bit less clear. I guess one could set the size to 1 in calls like, since only the first element is used as a stand-in for a const value. Is that valuable?
> 
> ```
>        call device_memcpy(this%cp%x, this%cp%x_d, this%cp%size(), &
>             DEVICE_TO_HOST, sync=.false.)
> ```
> 

Perfectly fine to pass `1` as size without breaking anything. But take care, with only one scalar, you might have to sync to avoid data race.


> I also removed all hand-holding for setting the full stress Ax as per discussion with @pschlatt1 . You have `full_stress_formulation` as a keyword now, that you have to set to true if you want it. Will crash if you don't do that and have a `nut_field` though, since that is just wrong.

njansson (approved):

PR #1782, opened 2025-03-18
changed the output of time step from t to t+dt

pschlatt1:


njansson (approved):

njansson:
My guess would be the check it self, it's a diff between the gold file and the simulation output

pschlatt1:
yes, but then no pR should go through for now...

MartinKarp:
I can check it later today

MartinKarp (approved):

PR #1792, opened 2025-03-21
Only execute field_dirichlet_update() once per timestep

vbaconnet:
The fact that we apply boundary conditions twice per timestep (see [here](https://github.com/ExtremeFLOW/neko/blob/24660f35a432d43dc911371585531c41c0375977/src/fluid/fluid_scheme_incompressible.f90#L513-L541)) can cause unwanted behaviors with `field_dirichlet` and `field_dirichlet_vector` when calling the `field_dirichlet_update` subroutine.

Specifically, it adds extra overhead when, for example, there is reading of files involved or some time-dependent computation at every timestep, which is then done twice per timestep and can add redundant overhead. Technically, **all** BCs are applied twice, but field_dirichlet struck me as the most critical one since user stuff is involved.

I don't know if this also applies to the `wall_model_bc` since it has a `compute` and an `apply`? I have no idea if that is generally an expensive computation or not @timofeymukha.

The proposed solution involves adding an attribute `tstep_applied` to the `field_dirichlet_t` and `field_dirichlet_vector_t` objects, which stores the value of the time step it has latest called `this%update()` at. This logic is pretty simple and at quite a low level in the `field_dirichlet` objects, but it can serve as basis for discussions (hence the draft).


timofeymukha:
I guess the double-apply came in as the fix for unstructured meshes, right @MartinKarp?

So the solution is that you apply twice, but run the user routine once, right @vbaconnet ? That sounds quite reasonable to me.

MartinKarp:
Yeah, good we point out this behavior. I guess we just need to decide what the intended behavior is. For my application with moving geometries for example I will need to update this more than once per time step. I think I do this logic in the field_dirichlet update myself instead.

Also, how will this work for scalars for example?

MartinKarp:
One way would be to break it out and be called just before apply_bc_vel/prs (which I think we did originally). 

As you say, scalars will probably be fine.

timofeymukha:
Keep in mind that the whole thing can in principle be configured in the JSON now, in the bc's `{}`. If we want to support bouth double-update and single-update, we could have an option for that and pass it to the field_dirichet type.

vbaconnet:
Let's discuss this on Tuesday and decide how to do it! From what Martin wrote above it seems if someone needs to call user dirichlet more than once per timestep it can be done manually in the user file?

vbaconnet:
Redid a bit of the logic, I put `update` outside of `apply_vec` and `apply_scalar`, I believe it makes more sense since in essence `updating` and `applying` are different things.

This mimics a bit the way it was before, with `update` being explicitly used in `fluid_scheme`. The way it is now it just loops through the BCs and if I find a `field_dirichlet` or `field_dirichlet_vector` I just call its `update`.

Tested this on the cyl_boundary_layer example and it seems to work.

vbaconnet:
So indeed it is a lot cleaner now :)

src/fluid/fluid_scheme_incompressible.f90
```diff
@@ -533,6 +531,12 @@ subroutine fluid_scheme_bc_apply_vel(this, t, tstep, strong)
     call this%gs_Xh%op(this%w, GS_OP_MAX, glb_cmd_event)
     call device_event_sync(glb_cmd_event)
 
+    do i = 1, this%bcs_vel%size()
```
vbaconnet, line 534: This is here to not clutter `step` but can be moved if people have preferences

timofeymukha: This will work OK right now, I think. If @MartinKarp plans to call this routine a few times per step, then it will have to be moved.

MartinKarp (approved):
I think this will work well.

vbaconnet:
can someone auto-merge this please?

PR #1797, opened 2025-03-24
Make chkp not output at time 0

MartinKarp:
As pointed out by @pschlatt1 my fix for start_time to avoid writing chkps at time 0 did not work, they only got outputted at the second time step instead 🙃 

This is a fix for that.

A note is that working with the output is quite confusing and in the future we should give the counter, nexecution logic some love.


src/common/time_based_controller.f90
```diff
@@ -91,21 +89,15 @@ module time_based_controller
   !! @param control_value The value definining the execution frequency.
   !! @param start_time Value for when to start the time based controller.
```
MartinKarp, line 90: ``
```suggestion
```

adperezm (approved):

PR #1799, opened 2025-03-25
Fix user simcomp setup bug

timfelle:
Resolve an issue where the user simcomp setup would invalidate the json_object after use by adjusting the way simulation components are extracted from the parameters.

### DIFF
--- a/examples/tgv/tgv.f90
+++ b/examples/tgv/tgv.f90
@@ -67,9 +67,9 @@ contains
     real(kind=rp) :: ux, uy, uz
     real(kind=rp) :: uvw(3)
 
-    uvw(1)   = sin(x)*cos(y)*cos(z)
-    uvw(2)   = -cos(x)*sin(y)*cos(z)
-    uvw(3)   = 0._rp
+    uvw(1) = sin(x)*cos(y)*cos(z)
+    uvw(2) = -cos(x)*sin(y)*cos(z)
+    uvw(3) = 0._rp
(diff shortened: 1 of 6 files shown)


MartinKarp (dismissed):

timofeymukha (dismissed):

timfelle:
@timofeymukha and @MartinKarp 

I have updated the TGV example, I expect that it is good, but i dismissed your reviews for an extra look. I can see the regression test worked so we should be good I guess

examples/tgv/tgv.f90
```diff
@@ -88,12 +88,20 @@ subroutine user_initialize(t, u, v, w, p, coef, params)
     ! initialize work arrays for postprocessing
     call w1%init(u%dof, 'work1')
```
gloopydoop, line 89: unrelated to this PR, but I suppose this could be replaced by the scratch registry

gloopydoop (approved):

adperezm (approved):

MartinKarp (changes requested):
So what was the underlying problem?

timfelle:
> So what was the underlying problem?

It seem the use of the json_core and json_value pointer caused issues and invalidated the json object going forward. I am unsure exactly what the issue was. But moving to our json_utils functions, which extract sub objects by reading them into a string and building a new object, seem to be safer.

MartinKarp (approved):

PR #1801, opened 2025-03-26
intent for vector and matrix getter functions

gloopydoop:
to allow the size of vector with intent(in) to be retrieved.

### DIFF
--- a/src/math/matrix.f90
+++ b/src/math/matrix.f90
@@ -133,7 +133,7 @@ contains
 
   !> Returns the number of entries in the matrix.
   function matrix_size(m) result(s)
-    class(matrix_t), intent(inout) :: m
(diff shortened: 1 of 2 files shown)


src/math/matrix.f90
```diff
@@ -133,7 +133,7 @@ end subroutine matrix_free
 
   !> Returns the number of entries in the matrix.
   function matrix_size(m) result(s)
-    class(matrix_t), intent(inout) :: m
+    class(matrix_t), intent(in) :: m
```
gloopydoop, line 136: aside from formatting, there were only two small changes:
1) this

timfelle (approved):

vbaconnet (approved):

PR #1803, opened 2025-03-27
Continuation of #1727 with time_state_t to track the time

MartinKarp:
In order to not swamp #1727 with tons of changes if we do not like this approach I start this PR.
 
# Changes 
- Introduces the module time_state with the type time_state_t (formerly time_struct_t) which encapsulates everything about the state of the time.
- Most arguments in the code which made use of some combination of tstep, t, dt now takes the type(time_state_t) as an argument instead (But there are places which I have not yet changed the t, tstep, dt combo).
- Builds on #1727 so the changes to allow for a "perfect hit" are still there. This PR fixes so that the correct nexecutions is set after restart.

No other changes (such as how the time is actually updated) are introduced as to not make this a mega-PR and avoid potential bugs (for example with the logic for the bdf_ext coupling). User routines are kept the same as before.

# Suggestion
My suggestion if we think this interface looks good (using time_t instead of tstep, t, etc.) is to close #1727 and merge this instead.


MartinKarp:
I was thinking it might good to avoid the module name time somehow.

MartinKarp:
It might be an idea to rename the time_t to neko_time_t instead. The main thing is that I dont want it to be confused to the wall-clock time.

pschlatt1:
what about "timestepping" or similar as name

MartinKarp:
time_state and time_state_t

timofeymukha (approved):

timfelle (approved):
LGTM

Issue #1806, opened 2025-03-31
cyl_boundary_layer/cyl_bl_user: *** ERROR: Invalid user defined inflow condition ***


timfelle:

njansson:
> neko/cyl_boundary_layer/cyl_bl_user_test________________________________________ *** ERROR: Parameter solver.preconditioner.type missing from the case file ***

Seems to be an issue with the scalar part

PR #1807, opened 2025-03-31
Initialize BC pointers and fix validation issue

timfelle:
Pointers for boundary condition components are now initialized to null to prevent potential errors. The validation function has been corrected to ensure it can find the coefficient after initialization. Formatting improvements have also been made for better readability.

Closes: #1805, #1806

### DIFF
--- a/src/bc/usr_inflow.f90
+++ b/src/bc/usr_inflow.f90
@@ -83,7 +83,7 @@ module usr_inflow
      !! @param t Current time
      !! @param tstep Current time-step
      subroutine usr_inflow_eval(u, v, w, x, y, z, nx, ny, nz, &
-                                ix, iy, iz, ie, t, tstep)
+          ix, iy, iz, ie, t, tstep)
        import rp
        real(kind=rp), intent(inout) :: u
        real(kind=rp), intent(inout) :: v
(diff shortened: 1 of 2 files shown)


MartinKarp (approved):
What validation issues did you find?

src/fluid/fluid_pnpn_bc_fctry.f90
```diff
@@ -146,41 +146,40 @@ module subroutine velocity_bc_factory(object, scheme, json, coef, user)
...
        end select
 
-      case ("user_velocity_pointwise")
+    case ("user_velocity_pointwise")
        allocate(usr_inflow_t::object)
        select type (obj => object)
-         type is (usr_inflow_t)
+       type is (usr_inflow_t)
           call obj%set_eval(user%fluid_user_if)
-          call obj%validate()
```
gloopydoop, line 180: Shouldn't this end up being called later on somewhere, after the initialization?

timfelle: Well it is only defined for the user type. And it checks for coef, which does not get set untill the initializer gets assigned in line 192

gloopydoop: maybe it could be squeezed into `usr_inflow_finalize`?

I feel like I'm not the right person to review this PR :sweat_smile: , let's wait for @timofeymukha

timofeymukha (approved):

PR #1816, opened 2025-04-04
Enhance option parsing for macOS compatibility

timfelle:
Improve option parsing in scripts to support long options on macOS using gnu-getopt, with a fallback to BSD getopt. Additionally, implement reporting for unknown options.

### DIFF
--- a/contrib/lint_format/format.sh
+++ b/contrib/lint_format/format.sh
@@ -52,7 +52,8 @@ if [ $(uname) == "Darwin" ]; then
     if command -v gnu-getopt >/dev/null 2>&1; then
         PARSED=$(gnu-getopt --options=$OPT --longoptions=$OPTIONS --name "$0" -- "$@")
     else
-        echo "Warning: macOS uses BSD getopt, long options are not supported." >&2
+        echo "GNU-getopt not found." >&2
+        echo "Falling back to BSD getopt, long options are not supported." >&2
         PARSED=$(getopt $OPT "$@")
     fi
 else
--- a/contrib/lint_format/format.sh
+++ b/contrib/lint_format/format.sh
@@ -47,7 +47,18 @@ CONTINUATION="5"
 TARGET_BRANCH="develop"
 
 # Parse the inputs for options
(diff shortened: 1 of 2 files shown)


contrib/lint_format/format.sh
```diff
@@ -47,7 +47,18 @@ CONTINUATION="5"
 TARGET_BRANCH="develop"
 
 # Parse the inputs for options
-PARSED=$(getopt --options=$OPT --longoptions=$OPTIONS --name "$0" -- "$@")
+if [ $(uname) == "Darwin" ]; then
+    # Use gnu-getopt on macOS
+    if command -v gnu-getopt >/dev/null 2>&1; then
+        PARSED=$(gnu-getopt --options=$OPT --longoptions=$OPTIONS --name "$0" -- "$@")
+    else
+        echo "Warning: macOS uses BSD getopt, long options are not supported." >&2
```
njansson, line 55: I would suggest to reword this, something along the line that gnu getup not found. There are many more systems than macOS that doesn't have gnu getopt as default

njansson (approved):

contrib/lint_format/format.sh
timfelle: This is only checked for the "Darwin" uname, so it is only checked for macOS though

MartinKarp (approved):

MartinKarp:
hope I was not too fast!

timfelle:
> Great, hope I was not too fast in merging!

Well did you try it 😅 I don't have a mac around to test it myself

MartinKarp:
I mean, not extensively, but seems to work like a charm (modified a file and the formatter fixed it).

MartinKarp:
Now I tested the linter and the -P option for grep is missing on Mac app

njansson:
> Now I tested the linter and the -P option for grep is missing on Mac app

Correct that is GNU grep specific, these lines has to be changed to some `perl` thing

timfelle:
> > Now I tested the linter and the -P option for grep is missing on Mac app
> 
> Correct that is GNU grep specific, these lines has to be changed to some `perl` thing

Im on it

PR #1821, opened 2025-04-07
Combine user_velocity and user_pressure into one user_dirichlet

vbaconnet:
As the title says, the idea here is to allow for some mixed BCs where the user can set dirichlet BCs on selected velocity components (say `u` and `v`) AND a pressure profile `p`.

# Usage

The new logic suggest that the user should specify a "fields" array in the json, for example:

```json
{
 "type": "user_dirichlet",
"fields": ["u", "v", "p"],
"zone_indices": [1]
}
```

Of course the previous behavior of `user_velocity` and `user_pressure` can be emulated by setting `fields : ["u","v","w"]` and `fields : ["p"]` respectively.

# Implementation

To achieve this, there is now an array of `components` in `field_dirichlet_vector_t` that stores the names of the fields on which dirichlet should be applied. In `apply_vector`, only those fields are copied to the solution fields. However, all the `bc_u, bc_v, bc_w` fields are still initialized such that the user does not run into segfaults if they accidentally try to manipulate a field that is not intended for dirichlet (in the example above, `w`). 

Pressure remains a `field_dirichlet_t`, while only the velocity is affected by this change. In the `velocity_bc_factory` there is some logic that makes sure to only keep the velocity components if we specify for example `u,v,p`. I manipulate the json object a bit for the `field_dirichlet_vector_t` constructor to pick it up.

# Todo

- [ ] I have modified the `cyl_boundary_layer` example but the fluid solver is blowing up, i think it has to do with the velocity residual bcs etc, I think I need some help on this.
- [ ] Do it also on the scalar
- [ ] Test on GPUs
- [ ] Cleanup print statements

# Dependencies

Depends on PR #1792, therefore is in draft for the time being.


pschlatt1:
maybe I dont fully understand the case, but can one (perhaps should one) set the pressure as a boundary condition?

vbaconnet:
> maybe I dont fully understand the case, but can one (perhaps should one) set the pressure as a boundary condition?

Well, the case that started the whole `field_dirichlet` stuff almost a year ago is this one (flat plate with a leading edge): https://pubs.aip.org/aip/pof/article/36/7/074105/3302120/Direct-numerical-simulation-of-transition-under. In this case there are two `'o'` outflows and one `'on'` horizontal boundary over the flat plate. Pressure values from a RANS simulation are imposed on these boundaries to maintain a specific pressure gradient at the leading edge (+ tangential velocity values on the `'on'` boundary).

But in general this is meant to mimic the 'on' and 'o' boundary conditions in Nek5000: https://nek5000.github.io/NekDoc/problem_setup/BCs.html#pressure-outlet-normal-on. If I understand correctly in Nek5000 the pressure dirichlet is be done [here](https://github.com/Nek5000/Nek5000/blob/d6ae08b6ee6bfe1d6e26003a5c94d8c41a6250d3/core/bdry.f#L956-L962).

As Martin pointed out I'm missing some surface terms to have the implementation correct but that's the idea...

pschlatt1:
ok, so it is the total normal stress, the pa ("ambient pressure"), that you want to set.

vbaconnet:
After some discussions with Martin I think we will leave the user dirichlet stuff as it is for now

PR #1825, opened 2025-04-09
Fix Coriolis term bug

timofeymukha:
Wasn't using the velocity field where it should have, using instead the source term fields.

Lorenzo is going to test it, so don't merge for now.


loretet:
The behaviour has improved with the fix and it's now much more physical, but the implementation still has something wrong when specifying Coriolis using the rotation vector or omega and phi. Works fine when specifying just f. 
diff_coriolis
Will look into this, don't merge for now.

timofeymukha:
OK, great, thanks @loretet !

In the end, all three inputs populate this%omega in the type. This is done in coriolis_source_term_init_from_json

After the constructor call on line 119, the type should be ready to use. Could you maybe add a 
`write(*,*) "OMEGA", this%omega, this%u_geo` on the line below, and just experiment quickly with what gets printed for various inputs? We should be able to track this down quickly, I hope.

timofeymukha:
I played a bit as well and did not find anything criminal yet. What are the exact settings you are testing? I think if the rotation vectors are the same, we should be getting the same result.

loretet:
Yes, I agree, and the rotation vectors are indeed correct. I have tried re-setting the cases from scratch and re-building Neko (just because I literally cannot find anything wrong in the code) and now I get a slightly different result (see attachment). Still slightly wrong (the v profile above the inversion doesn't exactly go to zero), but it's better than before.
What bugs me is that I did not change anything in the source code and I don't see how I could get two different results with the same settings. I can try running at a higher resolution (time resolution should be ok, 15000 seconds with CFL=0.5), perhaps the combination of very coarse resolution + turbulence initialisation with random fluctuations could give some weird results....? (turbulence is initialised with a small controlled perturbation as described in Linnea's paper draft)

Just FYI,
```
f_in      :  f = 1e-4
om_phi    :  om = 7.3e-5 , phi = 45
rot_vec_1 : [0, 0, om*sin(radians(45))]
rot_vec_2 : [0, om*cos(radians(45)), om*sin(radians(45))]
```
I am hence expecting the results of `(NekTrad, f_in, rot_vec_1)` and `(NekFull, om_phi, rot_vec_2)` to be the same (`NekTrad` has the traditional Coriolis implementation with only the sine terms, while `NekFull` has all).

njansson (approved):

timfelle (approved):

PR #1827, opened 2025-04-12
CUDA implementation of the Neumann bc

timofeymukha:
Implements the Neumman boundary condition for CUDA GPUs.


njansson:
Not sure if this was the problem, but if one place the non-linear  index function in a common heder for both Neumann and e.g. normal facet, then this header should be included in the .h file where it is called. We use a similar pattern for the warp shuffle reduce kernels. These are in `math.h` and reused in several places like GMRES etc.

timofeymukha:
> Not sure if this was the problem, but if one place the non-linear index function in a common heder for both Neumann and e.g. normal facet, then this header should be included in the .h file where it is called. We use a similar pattern for the warp shuffle reduce kernels. These are in `math.h` and reused in several places like GMRES etc.

This is a bit tricky.. Typically one would only gave the definition for the function the .h and the implementation elsewhere, thus including the .h does not trigger the "multiple definition" error. However, for these __device__ kernels it seems one cannot do this common split. However, adding `__inline__` seem to have made things compile... Let's see if CI is happy.

njansson (approved):

MartinKarp (commented):
I think this looks very good, it would not be possible to merge this and the facet_normal bc? To me it seems they are more or less identical, only that this one is coded for one flux vector and facet_normal in the corner cases where the flux for each component is u, v, and w.

timofeymukha:
> I think this looks very good, it would not be possible to merge this and the facet_normal bc? To me it seems they are more or less identical, only that this one is coded for one flux vector and facet_normal in the corner cases where the flux for each component is u, v, and w.

It could be possible down the road. For now, Neumann only implements apply_scalar. The potential apply_vector would be close to facet_normal as you say, but some diffs do exist, for example in the Neumann you add to x, y, z, whereas facet_normal overwrites them. But one should still look at the possibility.

MartinKarp:
However, I am 90% it is a bug that facet normal overwrites the components. But sounds good!

src/bc/neumann.f90
```diff
@@ -228,29 +235,33 @@ pure function neumann_flux(this) result(flux)
...
     class(neumann_t), intent(inout) :: this
-    real(kind=rp), intent(in) :: flux(this%msk(0))
+    type(vector_t), intent(in) :: flux
     integer :: i
 
-    call copy(this%flux_, flux, this%msk(0))
+    this%flux_ = flux
+
+    ! ONLY WORKS ON THE CPU
     this%uniform_0 = .true.
```
MartinKarp, line 262: What is the intended behavior here? Maybe we could add a comment.

MartinKarp (changes requested):

src/bc/neumann.f90
timofeymukha: Yeah, I guess ideally it should be implemented on the GPU as well. The consequence is that the GPU kernel will run even if the flux value is zero. So it doesn't really break anything. I wasn't sure if it is worth the time fixing it since the uniform thing was kind of a patch until we have a GPU kernel? It will still work in the constructor from the scalar flux, so this should not change the case when neumann is constructed from json, and so one should still be able to run on AMD, etc.

src/bc/neumann.f90
```diff
@@ -213,12 +219,13 @@ subroutine neumann_finalize(this, only_facets)
 
     call this%finalize_base(only_facets_)
```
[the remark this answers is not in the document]
MartinKarp: I guess the question one can ask is, should this BC be applied on an edge along a boundary with no associated facet? I would say in this case the normal of this edge is not well defined and the area is 0 so it should not be applied.

src/bc/neumann.f90
timofeymukha: I guess the first thing to decide is whether we want the uniform check going forward if we were to have all the kernels in place.

timofeymukha: @MartinKarp @njansson How bad would this be? Is a loop on the host acceptable?

```fortran
       call device_memcpy(this%flux_%x, this%flux_%x_d, this%flux_%n, &
            DEVICE_TO_HOST, sync=.true.)
    end if
    do i = 1,this%msk(0)
       this%uniform_0 = abscmp(this%flux_%x(i), 0.0_rp) .and. this%uniform_0
    end do
```

timofeymukha: No, just the constructor.

timofeymukha: Wait, no, it *could* be every timestep if one uses this routine... Sorry. Then we cannot have this. But I am now thinking that in this routine we can just remove this check. If someone is passing an entire `vector_t` to set the flux, it is really unlikely that this is a case of a homogenous bc.

MartinKarp: Sounds good to me

timofeymukha: @MartinKarp alright, removed and added a comment.

MartinKarp (approved):

PR #1828, opened 2025-04-13
Make non-convergence warning only print once

timofeymukha:
Dominates the log completely otherwise. Also fix missing space (it was killed by `trim`).


timfelle (approved):

vbaconnet (approved):
A much needed fix especially for larger jobs on clusters, thanks! 

(funny, I was literally doing the exact same thing on friday but forgot to push it)

A silly question since I don't know that much about the logic in the solvers but could it ever happen that one or a few rank do not converge while the other ones do?

vbaconnet:
> The residuals and iterations, etc are global (There are reductions and such). So without looking at the code I would say that it should not happen.

Aha okay thanks!

PR #1835, opened 2025-04-23
Update PkgConfig CFlags

timfelle:
We currently force any dependent program to use the identical list of compiler flags used to compile Neko.

This is probably not really what we intended so they are removed from the package here in order to improve portability

### DIFF
--- a/neko.pc.in
+++ b/neko.pc.in
@@ -4,10 +4,10 @@ libdir=@libdir@
 includedir=${prefix}/include
 includedir_pkg=${prefix}/include/neko
 compiler=@FC@
- 
+
 Name: NEKO
 Version: @PACKAGE_VERSION@
 Description:
-Requires: 
+Requires:
 Libs: -L${libdir} -lneko @LDFLAGS@ @LIBS@
-Cflags: @FCFLAGS@ -I${includedir} -I${includedir_pkg} 
+Cflags: -I${includedir} -I${includedir_pkg}


njansson:
Wouldn't this cause issues for certain configurations, e.g. if openmp is enabled

timfelle:
> Wouldn't this cause issues for certain configurations, e.g. if openmp is enabled

That is possible I guess 🤔
But stuff like `-O` and `-g` should definitely not propagate up to dependent programs.

njansson:
> > Wouldn't this cause issues for certain configurations, e.g. if openmp is enabled
> 
> That is possible I guess 🤔 But stuff like `-O` and `-g` should definitely not propagate up to dependent programs.

True, optimisation flags etc should not, but fixes (like -Mblackslash for nvhpc) should. But now we stash everything into FCFLAGS, which also contains paths to e.g. json-fortran, hdf5 and suchs

njansson:
> > > Wouldn't this cause issues for certain configurations, e.g. if openmp is enabled
> > 
> > 
> > That is possible I guess 🤔 But stuff like `-O` and `-g` should definitely not propagate up to dependent programs.
> 
> True, optimisation flags etc should not, but fixes (like -Mblackslash for nvhpc) should. But now we stash everything into FCFLAGS, which also contains paths to e.g. json-fortran, hdf5 and such. 

How a

timfelle:
> > > Wouldn't this cause issues for certain configurations, e.g. if openmp is enabled
> > 
> > 
> > That is possible I guess 🤔 But stuff like `-O` and `-g` should definitely not propagate up to dependent programs.
> 
> True, optimisation flags etc should not, but fixes (like -Mblackslash for nvhpc) should. But now we stash everything into FCFLAGS, which also contains paths to e.g. json-fortran, hdf5 and suchs

Are you sure about -Mbackslash? the code is already compiled so is it really required for a dependent user to also include that flag?

njansson:
Another concern is search paths for dependencies, e.g. location of json-fortran and hdf5, I think using a module that depends on these would fail at compile since it can't find these modules files

timfelle:
> Another concern is search paths for dependencies, e.g. location of json-fortran and hdf5, I think using a module that depends on these would fail at compile since it can't find these modules files

Yes, we need to make sure all include paths are done correctly.

njansson:
> > True that one should be fine, but openmp flags will be necessary (unless we requires a user to provide the openmp link flags them self, not really recommended)
> 
> From my understanding of this, the _linking_ of openmp should be done under libs (`-lopenmp`). That way we make sure openmp is available for neko, but we do not force a user to have the _compiler_ flag `-fopenmp` (or similar).

Which library (or libraries) are compiler dependent afaik, which are added by the compiler flag.

timfelle:
> Which library (or libraries) are compiler dependent afaik, which are added by the compiler flag.

Is it then not a thing that should be in the `Requires` section of the pc file?

njansson:
> > Which library (or libraries) are compiler dependent afaik, which are added by the compiler flag.
> 
> Is it then not a thing that should be in the `Requires` section of the pc file?

That would be nice, but there's no generic openmp.pc file to put under the required section, and compiler vendors generally don't ship one

timfelle:
Well I can report that OpenMP does not correctly propagate on develop as is. So we clearly need to do something 😅

njansson:
> Well I can report that OpenMP does not correctly propagate on develop as is. So we clearly need to do something 😅

That is strange, if I configure with `--enable-openmp` the flags propagates correctly to the pc file

timfelle:
> That is strange, if I configure with `--enable-openmp` the flags propagates correctly to the pc file

Yes, I see them as well. But when I then try to build the topopt library it fails when linking. 
`neko-top/external/neko/src/comm/comm.F90:89:(.text+0x5): undefined reference to 'omp_get_thread_num'`

njansson:
> > That is strange, if I configure with `--enable-openmp` the flags propagates correctly to the pc file
> 
> Yes, I see them as well. But when I then try to build the topopt library it fails when linking. `neko-top/external/neko/src/comm/comm.F90:89:(.text+0x5): undefined reference to 'omp_get_thread_num'`

Well, I would check if Neko-top is mangling something then, since both the poisson example, which uses pkg-config, and other external benchmark linking against neko works fine with the generated pc (with OpenMP)

njansson:
Cleaning up CFLAGS from internal things I fully agree with, but I don't think we should merge this before OpenMP is settled (we have enough benchmark and tests at PDC using libneko, which needs OpenMP).

Regarding where to place OpenMP, after looking at other packages e.g. PETSc and such, all are passing the openmp flag in their cflags (which often is a clean var unless openmp is activated). Yes this breaks builds that uses `pkg-config --libs`, but that is fine, since that also breaks if one tries the same with e.g. PETSc or some other package.

The correct pkg-config usage for Neko I would argue must be `pkg-config --variable=compiler --libs --cflags neko`,  to capture all dependencies.

timfelle (commented):
> Cleaning up CFLAGS from internal things I fully agree with, but I don't think we should merge this before OpenMP is settled (we have enough benchmark and tests at PDC using libneko, which needs OpenMP).
> 
> Regarding where to place OpenMP, after looking at other packages e.g. PETSc and such, all are passing the openmp flag in their cflags (which often is a clean var unless openmp is activated). Yes this breaks builds that uses `pkg-config --libs`, but that is fine, since that also breaks if one tries the same with e.g. PETSc or some other package.
> 
> The correct pkg-config usage for Neko I would argue must be `pkg-config --variable=compiler --libs --cflags neko`,  to capture all dependencies.

As it stands, the flags for OpenMP are still added, but user defined compiler flags for Neko such as -O and -g or whatever else is not added to the package. This seem like the happy medium,

timfelle:
Another issue I have noticed is that the CFLAGS variable in the PKG-Config file applies to your device compilers as well. Meaning if you work on lumi you cannot add any flags which are not identical between your compilers. Working with Cray for CPU and the ROCm compilers on lumi, you cannot set "-m4" since it does not exist for clang++

timfelle:
In any case, I believe this PR is ready for merge as it is. Openmp is still added to the package, but the user defined ones are not.

njansson:
> Another issue I have noticed is that the CFLAGS variable in the PKG-Config file applies to your device compilers as well. Meaning if you work on lumi you cannot add any flags which are not identical between your compilers. Working with Cray for CPU and the ROCm compilers on lumi, you cannot set "-m4" since it does not exist for clang++

No, that sounds more like a configuration issue. Device specific flags have another variable (`CUDA_CFLAGS`, HIP_HIPCC_CFLAGS`), which is not propagated to FCLAGS or CFLAGS.

njansson:
> In any case, I believe this PR is ready for merge as it is. Openmp is still added to the package, but the user defined ones are not.

No we can't merge it as is, since this will break `makeneko`, especially for openmp builds

njansson (approved):

pohm01 (approved):
lgtm

PR #1837, opened 2025-04-23
Feature/insitu pod example

adperezm:
I add an example of insitu POD on the turbPipe. 

I added the files to the same folder as the vanilla example and provided some description of them.


adperezm:
After this is merged, I will close pending pull requests that also attempt to exemplify in-situ processing.

examples/turb_pipe/insitu_turb_pipe.f90
```diff
@@ -0,0 +1,153 @@
...
+  end subroutine user_ic
+
+  function pipe_ic(x, y, z) result(uvw)
+    real(kind=rp) :: x, y, z
+    real(kind=rp) :: uvw(3)
+    real(kind=rp) :: rand, ux, uy, uz, xr, yr, rr, zo
+    real(kind=rp) :: amp_z, freq_z, freq_t, amp_tht, amp_clip, blt
+    real(kind=rp) :: phase_z, arg_tht, amp_sin, pi, th
+
+    pi = 4d0 * atan(1d0)
```
timfelle, line 114: In the declaration you have defined the precision of `rp` for the reals in this
subroutine. However, you define many of the constants with double precision.
Please use `_rp` instead of `d0` to avoid implicit changes to the precision.

examples/turb_pipe/insitu_turb_pipe.f90
```diff
@@ -0,0 +1,153 @@
...
+    xr = x
+    yr = y
+    rr = xr*xr + yr*yr
+    if (rr .gt. 0) rr = sqrt(rr)
+    th = atan2(y,x)
+    zo = 2*pi*z/25d0
```
timfelle, line 120: Please do not mix integers into the computation of reals. It will cause hard to
catch bugs when you move to systems that have different standards for how to
deal with implicit conversion from integer to real.
```suggestion
    zo = 2.0_rp * pi * z / 25.0_rp
```

examples/turb_pipe/insitu_turb_pipe.f90
adperezm: Fair point! I changed that here and in the original turb_pipe.f90

adperezm:
Thanks for the comments. I implemented them now. @pschlatt1, let me know if there is something that we should add to this example at this stage. Otherwise let's merge and change the rest later.

timofeymukha (approved):

timfelle (approved):

adperezm:
I actually suggest that we don't wait for @pschlatt1, since I do not see any PR open for the pipe. Future changes can incorporate this instead. So if someone with merge permission sees this, then please accept.

PR #1839, opened 2025-04-24
Cleanup our factories.

timfelle:
I think I got all of them to use select case when it makes sense.
Also the manual construction of the error message with concat_string is now removed and using the neko type error instead.


njansson (approved):
Happy days 🍻

timofeymukha (approved):

PR #1842, opened 2025-04-24
Don't allocate unecessary zero sized arrays

njansson:
We should avoid allocating zero sized arrays, especially on devices, which might not support it

### DIFF
--- a/src/common/projection.f90
+++ b/src/common/projection.f90
@@ -120,7 +120,8 @@ contains
   subroutine projection_init(this, n, L, activ_step)
     class(projection_t), target, intent(inout) :: this
     integer, intent(in) :: n
-    integer, optional, intent(in) :: L, activ_step
+    integer, intent(in) :: L
+    integer, optional, intent(in) :: activ_step
     integer :: i
     integer(c_size_t) :: ptr_size
     type(c_ptr) :: ptr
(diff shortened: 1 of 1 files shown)


src/common/projection.f90
```diff
@@ -156,7 +162,6 @@ subroutine projection_init(this, n, L, activ_step)
 
        call device_map(this%xbar, this%xbar_d, n)
        call device_alloc(this%alpha_d, int(c_sizeof(dummy)*this%L, c_size_t))
-
        call device_rzero(this%xbar_d, n)
```
njansson, line 165:
```suggestion

       call device_rzero(this%xbar_d, n)
```

timfelle (commented):
Should we change the defaul `L` to 0 in this function as well? Currently the default here are 20 as well.

njansson:
> Should we change the defaul `L` to 0 in this function as well? Currently the default here are 20 as well.

Or should we make `L` none optional? I think that makes more sense

timfelle:
> > Should we change the defaul `L` to 0 in this function as well? Currently the default here are 20 as well.
> 
> Or should we make `L` none optional? I think that makes more sense

Sure, that seem way more sensible. If you manually call the function you probably know how many projections you want.

njansson:
I'll update the pr

timfelle (approved):

timofeymukha (approved):

PR #1845, opened 2025-04-24
Type injection improvements

timofeymukha:
This is mainly an update to makeneko.in, but first two small improvements:

- Added a check to all `register_type` routines to see if a type with the same name already exists. A helpful error message is then provided.
- Added all the routines for type injection and creation to neko.f90

Now to the main topic of the PR. These changes follow the discussion and requests during the type injection demo, hoping to arrive to a convenient and flexible makeneko.

- makeneko now detects which .f90s have the `modulename_register_types` routines, and which don't. Thus you don't have to add any dummy routine anymore if you import your modules in the user file and use them there, without registering any new types with factories. Essentially, this just allows the default makeneko behavior before I started meddling with type injection. **Note**: we rely on the convention that `mymodule.f90` contains `module mymodule`. This is generally the convention for neko itself, so I would actually like to enforce it.
- One can now use makeneko to compile a list of type-injection modules without having a `user module`. This way you can add the custom types to any case, and have no need to provide a redundant user file. If makeneko detects that there is no user file and none of the modules inject types it throws an error because in this case makeneko is useless.
- I added an `rm` for `usr_driver.f90` in the beginning of makeneko, because a remaining one sometimes caused issues during my testing.
- Finally, for purists (like @adampep and I suspect also @pschlatt1  :-] ) that want full control over type injections, I have added a flag to the user interface, called `supress_type_injection`. Defaults to `false`, but if set to `true` in the user file in `user_setup`, the type injection from modules is suppressed, and you have to do it yourself. The appropriate user routine is `user_startup`.
- In the same spirit, @timfelle, you can define types you want to inject directly in the user file and use `user_startup` to register them.


njansson (approved):

timofeymukha:
@njansson I added a check that ignores non-.f90 files and throws a warning about ignored files. If no valid files are provided, an error is thrown. So, won't get stuck you just give it a .case or so anymore.

timfelle (approved):

PR #1851, opened 2025-04-25
Use a json subdict to extract and pass down (potential) precon params

### LINKED ISSUE #1849: Coarse grid solver can't be selected from case file
#1839 introduced a regression which removed the functionality to specify the hsmg coarse grid solver via a `+` e.g. `hsmg+tamg`

njansson:
This add the functionallty of providing options to the preconditioner. A use case is for example to specify a different coarse grid solver for HSMG, tune the number of iterations and such.

The current implementation extract a subdict from the json file and pass that down to precon init.

```json
        "velocity_solver": {
            "type": "cg",
            "preconditioner": {
                "type": "jacobi"
            },
            "projection_space_size": 4,
            "absolute_tolerance": 1e-7,
            "max_iterations": 800
        },
        "pressure_solver": {
            "type": "gmres",
            "preconditioner": {
                "type": "hsmg",
                "coarse_grid":  {
                    "solver": "cheby",
                    "iterations": 50
                },
            },
            "projection_space_size": 4,
            "absolute_tolerance": 1e-4,
            "max_iterations": 800
        },
```

This is a *breaking change* all case files needs to be updated to include at least `preconditioner.type` 

fixes #1849


src/krylov/pc_hsmg.f90
```diff
@@ -138,19 +141,21 @@ module hsmg
 contains
 
   !> @note I do not think we actually use the same grids as they do in the original!
-  subroutine hsmg_init(this, msh, Xh, coef, dof, gs_h, bclst, crs_pctype)
+  subroutine hsmg_init(this, msh, Xh, coef, dof, gs_h, bclst, hsmg_params)
```
timofeymukha, line 144: Could you please add an `init_from_components` that this constructor calls after extract stuff from the JSON? This way it will also be able to init the type without having to rely on parsing a JSON.

njansson: Not sure if I understand what that means ;)

timofeymukha (commented):
A necessary evil I guess :-). The ideal thing would to also have a common init interface for all the precons. I think it isonly hsmg that stands out, right?

Note that msh, Xh, dof, and gs are actually all accesible from coef, that you also pass. For better or worse :-). So a minimalistic common interface would be (coef, bclst, pcparams)

njansson:
> A necessary evil I guess :-). The ideal thing would to also have a common init interface for all the precons. I think it isonly hsmg that stands out, right?
> 
> Note that msh, Xh, dof, and gs are actually all accesible from coef, that you also pass. For better or worse :-). So a minimalistic common interface would be (coef, bclst, pcparams)

For now it's only `hsmg` and potentially `phmg`

src/krylov/pc_hsmg.f90
timofeymukha: Like for most bcs, for example. You have a constructor that takes in a json, and parses it for all the comonents that are needed to construct the case. Then, instead of actually constructing the type, that constructor calls `init_from_components`, which takes in as arguments all the components necessary for constuction. 

```fortran
subroutine blasius_init(this, coef, json)
    class(blasius_t), intent(inout), target :: this
    real(kind=rp) :: delta
    real(kind=rp), allocatable :: uinf(:)
    character(len=:), allocatable :: approximation

    call json_get(json, 'delta', delta)
    call json_get(json, 'approximation', approximation)
    call json_get(json, 'freestream_velocity', uinf)

    if (size(uinf) .ne. 3) then
       call neko_error("The uinf keyword for the blasius profile should be an &
       & array of 3 reals")
    end if

    call this%init_from_components(coef, delta, uinf, approximation)

  end subroutine blasius_init

  !> Constructor from components
  !! @param[in] coef The SEM coefficients.
  !! @param[in] delta The boundary layer thickness.
  !! @param[in] uinf The freestream velocity.
  !! @param[in] approximation The type of approximation for the profile.
  subroutine blasius_init_from_components(this, coef, delta, uinf, &
       approximation)
```

src/fluid/fluid_pnpn.f90
```diff
@@ -375,7 +377,8 @@ subroutine fluid_pnpn_init(this, msh, lx, params, user, chkp)
     call this%solver_factory(this%ksp_prs, this%dm_Xh%size(), &
          solver_type, solver_maxiter, abs_tol, monitor)
     call this%precon_factory_(this%pc_prs, this%ksp_prs, &
-         this%c_Xh, this%dm_Xh, this%gs_Xh, this%bcs_prs, precon_type)
+         this%c_Xh, this%dm_Xh, this%gs_Xh, this%bcs_prs, &
+         precon_type, precon_params)
```
timfelle, line 381: If you pass the params in, please remove the type from here since that is
redundant then.

src/source_terms/brinkman/PDE_filter.f90
```diff
@@ -308,17 +308,17 @@ subroutine filter_precon_factory(pc, ksp, coef, dof, gs, bclst, &
...
+ !      if (len_trim(pctype) .gt. 4) then
+ !         if (index(pctype, '+') .eq. 5) then
+ !            call pcp%init(dof%msh, dof%Xh, coef, dof, gs, bclst, &
+ !                 trim(pctype(6:)))
+ !         else
+ !            call neko_error('Unknown coarse grid solver')
+ !         end if
+ !      else
+ !         call pcp%init(dof%msh, dof%Xh, coef, dof, gs, bclst)
+ !      end if
```
timfelle, line 321: Should we not remove this now then?

timfelle: The comments?

examples/immersed_bunny/immersed_bunny.case
```diff
@@ -26,14 +26,18 @@
...
                 "projection_space_size": 0,
                 "absolute_tolerance": 1e-8,
                 "max_iterations": 800
             },
             "pressure_solver": {
                 "type": "gmres",
-                "preconditioner": "hsmg",
+                "preconditioner": {
+                    "type": "hsmg"
+                }.
```
timfelle, line 40:
```suggestion
                },
```

examples/cyl_boundary_layer/cyl_bl_basic.case
```diff
@@ -22,14 +22,18 @@
...
             "projection_space_size": 3,
             "absolute_tolerance": 1e-8,
             "max_iterations": 800
         },
         "pressure_solver": {
             "type": "gmres",
-            "preconditioner": "hsmg",
+            "preconditioner": {
+                "type": "hsmg"
+            }
```
timfelle, line 36:
```suggestion
            },
```

examples/scalar_mms/scalar_mms.case
```diff
@@ -22,14 +22,18 @@
...
         "projection_space_size": 0,
         "absolute_tolerance": 1e-7,
         "max_iterations": 800
       },
       "pressure_solver": {
         "type": "gmres",
-        "preconditioner": "hsmg",
+        "preconditioner": {
+            "type": "hsmg"
+        },
```
timfelle, line 36:
```suggestion
        },
```

tests/regression/cylinder/cylinder_part2.case
```diff
@@ -25,14 +25,18 @@
         },
         "velocity_solver": {
             "type": "cg",
-            "preconditioner": "jacobi",
+            "preconditioner": {
+                "type": "jacobi",
```
timfelle, line 29:
```suggestion
                "type": "jacobi"
```

examples/immersed_zones/block.case
```diff
@@ -3,7 +3,7 @@
     "case": {
         "mesh_file": "box.nmsh",
         "output_directory": "block_fields",
-        "output_at_end": false,
+x        "output_at_end": false,
```
timfelle, line 6:
```suggestion
        "output_at_end": false,
```

src/source_terms/brinkman/PDE_filter.f90
njansson: ah true that can be removed

src/fluid/fluid_pnpn.f90
timfelle: Well the precon_type can always be read from the precon_parameters, so we are passing the information twice. What he talks about is having an initializer for the individual types, which can be called with no json_file at all.

src/source_terms/brinkman/PDE_filter.f90
```diff
@@ -308,17 +308,6 @@ subroutine filter_precon_factory(pc, ksp, coef, dof, gs, bclst, &
        call pcp%init(coef, dof, gs)
     type is (device_jacobi_t)
        call pcp%init(coef, dof, gs)
-    type is (hsmg_t)
```
timfelle, line 311: Did we intend to stop supporting `hsmg` here?

njansson: Is it needed?

timfelle: @gloopydoop Is it? I think you know most about these filtering things.

gloopydoop: I tried to include everything that the velocity/pressure gets, but in earnest, I've never used `hsmg` it and the default is `jacobi` which seems to work fine in all tests we've run.

src/krylov/pc_hsmg.f90
njansson: Should be in place now

njansson:
> A necessary evil I guess :-). The ideal thing would to also have a common init interface for all the precons. I think it isonly hsmg that stands out, right?
> 
> Note that msh, Xh, dof, and gs are actually all accesible from coef, that you also pass. For better or worse :-). So a minimalistic common interface would be (coef, bclst, pcparams)

very true, `hsmg` is ancient. We can (and probably should) cleanup the other PC inits as well. But I suggest todo that in a separate PR

njansson:
Could we get this in as a small PR, after that we can start to clean up things, and propagate this usage down to e.g. `tamg`, and `cheby`.

timofeymukha (approved):

timfelle (approved):

PR #1856, opened 2025-04-27
Add a poor mans' device implementation for Coriolis

timofeymukha:
This would benefit from writing actual kernels quite a bit. But better than nothing.


adperezm (approved):

adperezm:
Why poor mans'  version? This looks perfectly fine to me haha. Is it that one would prefer to fuse more stuff to reduce the multiple launches?

Shiyu-Sandy-Du:
Maybe we could keep the implementation as it is now since it's only once per time step so you don't lose a lot of performance if you only use device_math? And then one could fuse it into a single kernel in another PR?

timofeymukha:
> Maybe we could keep the implementation as it is now since it's only once per time step so you don't lose a lot of performance if you only use device_math? And then one could fuse it into a single kernel in another PR?

Yeah, that's the idea, hence the PR :-)

njansson (approved):

PR #1857, opened 2025-04-27
Add masked_scatter_copy and masked_gather_copy for OpenCL

timofeymukha:
- Adds the OpenCL kernel for masked_red_copy.
- Adds a new math routine, and the device kernels for it, called `masked_scatter_copy`, which is just like `masked_red_copy`, but applies the mask to the destination array. So you copy an array into a bigger array, with the mask telling to which elements to scatter the contiguous array. The need came in wall modelling code where I want to scatter the $\tau_w$ predictions stored in a vector to the boundary nodes of a field, which is used for processing.

### DIFF
--- a/src/math/bcknd/device/cuda/cuda_math.f90
+++ b/src/math/bcknd/device/cuda/cuda_math.f90
@@ -64,14 +64,6 @@ module cuda_math
        integer(c_int) :: n, m
      end subroutine cuda_masked_scatter_copy
 
-
-     subroutine cuda_masked_scatter_copy(a_d, b_d, mask_d, n, m) &
-          bind(c, name = 'cuda_masked_scatter_copy')
-       use, intrinsic :: iso_c_binding, only: c_int, c_ptr
-       type(c_ptr), value :: a_d, b_d, mask_d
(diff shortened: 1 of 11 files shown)


MartinKarp:
Could we perhaps merge these two in some way? #1763

njansson (approved):

MartinKarp:
Perhaps most intuitive is that this PR adds the scatter_copy and renames the red_copy to gather_copy and #1763 only does the fixes for 0 length calls.

timofeymukha:
> Perhaps most intuitive is that this PR adds the scatter_copy and renames the red_copy to gather_copy and #1763 only does the fixes for 0 length calls.

Sorry martin, I completely missed that you added it there. But to me, it looks like this PR is not needed aside from the OpenCL? Then, we can merge yours first, and I will add the OpenCL stuff on top.

MartinKarp:
Yeah, I think the main difference is the OpenCL. I will update my PR now then :)

MartinKarp:
Merged #1763, so now we can add all improvements such as OpenCL in this branch.

PR #1860, opened 2025-04-28
Fix bounds in masked gather/scatter copies

timofeymukha:
I think this is correct?

We should always loop across the size of the contiguous array. For the gather mask, this is `a`, so we loop until `m`. For the scatter mask, this is `b`, so we loop until `n`. We assume that the array that is addressed via the mask is at least the same size. The size of the mask is also the same as that of the contiguous array.


njansson (commented):
Please take care, m and n is not directly what one thinks. Think of number of data to recv vs data to send 

Do you have a case that breaks the current indices?

njansson:
> Please take care, m and n is not directly what one thinks. Think of number of data to recv vs data to send 
> 
> Do you have a case that breaks the current indices?

Disregard, I clearly didn't read the pr before writing the comment

njansson (approved):

src/math/math.f90
```diff
@@ -297,16 +298,17 @@ end subroutine masked_gather_copy
...
   subroutine masked_scatter_copy(a, b, mask, n, m)
     integer, intent(in) :: n, m
     real(kind=rp), dimension(n), intent(in) :: b
     real(kind=rp), dimension(m), intent(inout) :: a
-    integer, dimension(0:m) :: mask
+    integer, dimension(0:n) :: mask
     integer :: i, j
 
-    do i = 1, m
+    do i = 1, n
```
MartinKarp, line 311: I think this should be m. To me the mask is always of the same size as the gathered array.

MartinKarp (changes requested):
Good catch, the bounds are incorrect in some places I think. My thinking was to always loop over `m` with `m` being the length of the mask array and gathered array, and that the length of the mask array is always the last argument. And `n` then being the length of the usual field or array.

I guess my main use case are things like this to follow the convention of the first argument being the output.

``` fortran
  call masked_gather_copy(a, b, mask, n, m)
  Carry out things with a
  Scatter the result back into b
  call masked_scatter_copy(b, a, mask, n, m)
```

src/math/math.f90
timfelle: Should the size of the mask not correspond to the size of the output?

timofeymukha: OK, that could be a convention to follow. Then depending on the routine, the size of `a` and `b` will correspond to different variables.

timofeymukha: Yeah, I think this will be good. I am renaming `m` to `n_mask` for clarity, and also `mask_size` to `n_mask` in the `cfill_mask`, for consistency.

timofeymukha: Alright, updated everything, please take a look. Always looping to n_mask!

MartinKarp (approved):
Looks great!

PR #1865, opened 2025-04-30
Add point-wise division and related operations

timfelle:
Introduce point-wise division and multiplication for vectors, along with new matrix operations. Implement device kernels and Fortran wrappers for these operations. Clean up and fix issues in existing functionality.

I am aware these new operations might create temporary arrays, however, for less performance critical sections, it allows for very simple coding.

### DIFF
--- a/src/math/vector.f90
+++ b/src/math/vector.f90
@@ -105,8 +105,8 @@ module vector
           vector_pointwise_div
      generic :: operator(**) => vector_pointwise_power
 
-     ! Private procedures.
-     procedure, private, nopass :: alloc => vector_allocate
+     ! Private interfaces
+     procedure, pass(a), private :: alloc => vector_allocate
 
   end type vector_t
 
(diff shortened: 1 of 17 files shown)


njansson (approved):

src/math/vector.f90
```diff
@@ -323,4 +338,96 @@ function vector_cmult_right(c, a) result(v)
...
+
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_col3(v%x_d, a%x_d, b%x_d, v%n)
+    else
+       call col3(v%x, a%x, b%x, v%n)
+    end if
+
+  end function vector_pointwise_mult
+
+  !> Pointwise vector multiplication \f$ v = a*b \f$.
```
MartinKarp, line 358:
```suggestion
  !> Pointwise power \f$ v = a^b \f$. OBS integer b
```

src/math/vector.f90
```diff
@@ -323,4 +338,96 @@ function vector_cmult_right(c, a) result(v)
...
+  !! Todo: Incredibly poor performance, needs to be optimized.
+  function vector_pointwise_power(a, b) result(v)
+    class(vector_t), intent(in) :: a
+    integer, intent(in) :: b
+    type(vector_t) :: v
+    integer :: i
+
+    call v%alloc(a%n)
+
```
MartinKarp, line 367:
```suggestion
    if (b .eq. 0) then 
       v = 1.0_rp
       return
    end if
```
Something like this is needed I think

MartinKarp (changes requested):
Nice!

MartinKarp (approved):

PR #1868, opened 2025-04-30
Check bounding box and right-handedness in the mesh

timofeymukha:
- Adds the mesh bounding box to the output of `mesh_checker`
- Adds a check for the right-handedness of each hex element in the mesh. Called during the finalization of `mesh_t`. Follows the implementation in Nek5000 based on the signed volume of a parallelepiped build on the rst vectors.


adperezm (approved):

MartinKarp (approved):

src/mesh/mesh.f90
```diff
@@ -1868,4 +1874,82 @@ function mesh_is_shared_facet(this, f) result(shared)
...
+               )
+
+          if (v(1) .le. 0 .or. &
+               v(2) .le. 0 .or. &
+               v(3) .le. 0 .or. &
+               v(4) .le. 0 .or. &
+               v(5) .le. 0 .or. &
+               v(6) .le. 0 .or. &
+               v(7) .le. 0 .or. &
+               v(8) .le. 0 ) then
```
timfelle, line 1948:
```suggestion
          if (v(1) .le. 0.0_rp .or. &
               v(2) .le. 0.0_rp .or. &
               v(3) .le. 0.0_rp .or. &
               v(4) .le. 0.0_rp .or. &
               v(5) .le. 0.0_rp .or. &
               v(6) .le. 0.0_rp .or. &
               v(7) .le. 0.0_rp .or. &
               v(8) .le. 0.0_rp ) then
```

timfelle (changes requested):
remember precision. I have seen some funky problems when comparing a real to an integer.

timofeymukha:
Ha, there is an error, because points are always stored in FP64. But I think it would be good to keep the math routine in kind=rp? @MartinKarp what should I do :D?

src/mesh/mesh.f90
```diff
@@ -1868,4 +1874,82 @@ function mesh_is_shared_facet(this, f) result(shared)
 
   end function mesh_is_shared_facet
 
+  !> Check the correct orientation of the rst coordindates.
+  subroutine mesh_check_right_handedness(this)
```
vbaconnet, line 1878: Amazing! Do we need to add that this is "inspired" from Nek5000?

timfelle (commented):
> Ha, there is an error, because points are always stored in FP64. But I think it would be good to keep the math routine in kind=rp? @MartinKarp what should I do :D? 

`real(var, kind=rp)`

But since they are inputs you have to do it in `mesh.f90`

src/mesh/mesh.f90
```diff
@@ -1868,4 +1874,82 @@ function mesh_is_shared_facet(this, f) result(shared)
...
+
+          if (v(1) .le. 0 .or. &
+               v(2) .le. 0 .or. &
+               v(3) .le. 0 .or. &
+               v(4) .le. 0 .or. &
+               v(5) .le. 0 .or. &
+               v(6) .le. 0 .or. &
+               v(7) .le. 0 .or. &
+               v(8) .le. 0 ) then
+             call neko_error("Invalid mesh element orientation")
```
vbaconnet, line 1949: Could we also output the element id in the error message? I found that very helpful in Nek to see where the problem is coming from.

vbaconnet (changes requested):

timfelle (commented):
> > Ha, there is an error, because points are always stored in FP64. But I think it would be good to keep the math routine in kind=rp? @MartinKarp what should I do :D? 
> 
> `real(var, kind=rp)`
> 
> But since they are inputs you have to do it in `mesh.f90` 

But the proper solution is to add the interface in math and do a generic function.

src/mesh/mesh.f90
timofeymukha: Yeah, we can add a comment that it is similar to nek5000 `verify`

timofeymukha:
Alright guys, I moved the routine into mesh.f90 and made it dp for now. Probably the most elegant solution would be to do this stuff directly as a routine in `hex_t` and `quad_t`. But I think what we have now is "good enough", and then we can let it eventually evolve.

@vbaconnet It will now list all the wrong elements with the centroid location and then die.

vbaconnet (approved):

timfelle (approved):

timofeymukha:
Somehow the bounding box checking disappeared here and I can't find it... *sigh*

PR #1869, opened 2025-04-30
Update file unit handling to use newunit

timfelle:
Refactor the code to replace manually assigned file units with the newunit feature, improving file handling consistency and reducing potential errors.

### DIFF
--- a/src/common/log.f90
+++ b/src/common/log.f90
@@ -93,8 +93,7 @@ contains
 
     call get_environment_variable("NEKO_LOG_FILE", log_file, envvar_len)
     if (envvar_len .gt. 0) then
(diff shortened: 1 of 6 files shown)


timofeymukha (approved):

timofeymukha:
Nice, didn't know this existed!

njansson (commented):
Newunit is nice but f2018 afaik. So I guess the question if we would like to enforce f2018 or not

timfelle:
> My bad seems to be f2008 after all

https://fortranwiki.org/fortran/show/newunit
😅

njansson (approved):

PR #1871, opened 2025-05-01
Refactor time step controller and update case files

timfelle:
Enhance the time step controller by introducing minimum time step and update frequency options, restructuring time-related parameters, and improving documentation. Adjust case files to align with the new time structure.

OBS: This will break case files for users.


timfelle (commented):
> Very much breaking as you say, but I like it quite a lot honestly. Is there a use case for `min_update_frequency`? 

I do not have one in mind per say, but, I thought it might be useful to someone. We can remove it if you believe it is not useful.

src/common/time_state.f90
```diff
@@ -34,32 +34,90 @@
...
   private
 
   !> A struct that contains all info about the time, expand as needed
   type, public :: time_state_t
-     real(kind=rp), dimension(10) :: tlag = 0.0_rp!< Old times
-     real(kind=rp), dimension(10) :: dtlag = 0.0_rp!< Old dts
-     real(kind=rp) :: dt = 0.0_rp!< Current dt
+     real(kind=rp), dimension(10) :: tlag = 0.0_rp !< Old times
+     real(kind=rp), dimension(10) :: dtlag = 0.0_rp !< Old dts
      real(kind=rp) :: t = 0.0_rp !< Current time
```
MartinKarp, line 46:
```suggestion
     ! @note Might be good to make these of kind=xp in the future.
     ! Would ensure that the time is not subject to stagnation.
     real(kind=rp) :: t = 0.0_rp !< Current time
```
I hope it is ok to add a memo.

src/common/time_state.f90
```diff
@@ -34,32 +34,90 @@
...
+
   !> Write status banner
   subroutine time_state_status(this)
     class(time_state_t), intent(in) :: this
     character(len=LOG_SIZE) :: log_buf
     character(len=38) :: log_fmt
     real(kind=rp) :: t_prog
 
-    t_prog = 100.0_rp * this%t / this%end_time
+    t_prog = 100.0_rp * (this%t - this%start_time) / this%end_time
```
MartinKarp, line 120: How does the start_time impact things like calculation of mean quantities? Will the weight of the first sample be correct?

MartinKarp (changes requested):
A good direction I think, nice job! Main thing I am wondering is regarding the start_time and how it interacts with various outputs and especially calculation of mean quantities.

src/common/time_state.f90
timfelle: Well if they assume the start time to be 0, then it would indeed not work. But I did not think to go through that. Are there any places other than the statistics code where that would be relevant?

timfelle: I have added a warning print in the statistics simcomp to tell the user if their statistics tried to start earlier than the simulation. After the warning is printed, the start time is adjusted such that the statistics start time is the simulation start.

MartinKarp: Cool! I think we need something similar for the output controller as well? I think, now one would output at every time step for a while until the counter has "catched" up.

timfelle: right, I have changed the time based controller to be a relative time instead. That should fix the issue. Can you check if i got all of the places it would be relevant

src/common/time_based_controller.f90
```diff
@@ -199,7 +199,8 @@ subroutine time_based_controller_set_counter(this, time)
     type(time_state_t) :: time
 
     if (this%nsteps .eq. 0) then
-       this%nexecutions = int((time%t+0.1_rp*time%dt) / this%time_interval) + 1
+       this%nexecutions = int(((time%t - time%start_time) + 0.1_rp*time%dt) &
```
MartinKarp, line 202: How does it work for nsamples and the like? I think we need to adjust their frequency as well.

src/common/time_state.f90
MartinKarp: I guess we need to check the nsamples and such as well at the init.

src/common/time_based_controller.f90
timfelle: Yes I am exactly just looking at the init thing now

timfelle: I am quite certain it behaves now

timofeymukha:
@timfelle Is the material properties thing still an issue? I can't quite pinpoint where it is.

doc/pages/user-guide/case-file.md
```diff
@@ -60,34 +61,46 @@ The three following options are possible.
...
+| `min_timestep`            | Minimum time-step size when variable time step is activated                                 | Positive reals                    | `0.0`         |
+| `target_cfl`              | The desired CFL number                                                                      | Positive real                     | `0.4`         |
+| `max_update_frequency`    | The minimum interval between two time-step-updating steps in terms of time steps            | Integer                           | `0`           |
+| `min_update_frequency`    | The maximum interval between two time-step-updating steps in terms of time steps            | Integer                           | `huge`        |
+| `running_avg_coeff`       | The running average coefficient `a` where `cfl_avg_new = a * cfl_new + (1-a) * cfl_avg_old` | Positive real between `0` and `1` | `0.5`         |
+| `max_dt_increase_factor`  | The maximum scaling factor to increase time step                                            | Positive real greater than `1`    | `1.2`         |
```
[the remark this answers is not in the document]
timfelle: I think what we actually need for that is a more detailed description of how the
variable time stepping is done, and there it would make sense to have some
guidelines on how the variables should be chosen.

timfelle:
> @timfelle Is the material properties thing still an issue? I can't quite pinpoint where it is.

I think it is fine yes.

I have rebuild the branch due to quite a lot of conflicts after the step function refactor

PR #1873, opened 2025-05-01
Refactor time-step interfaces

timfelle:
Streamline the interfaces for step functions by directly using the time state object.

Also moves some temporary material property management for clarity and later gif diff simplification.


timofeymukha (approved):

src/fluid/fluid_pnpn.f90
```diff
@@ -621,9 +618,8 @@ subroutine fluid_pnpn_step(this, t, tstep, dt, ext_bdf, dt_controller)
          rho => this%rho, mu => this%mu, oifs => this%oifs, &
          rho_field => this%rho_field, mu_field => this%mu_field, &
          f_x => this%f_x, f_y => this%f_y, f_z => this%f_z, &
-         if_variable_dt => dt_controller%if_variable_dt, &
-         dt_last_change => dt_controller%dt_last_change, &
-         event => glb_cmd_event)
+         t => time%t, tstep => time%tstep, dt => time%dt, &
+         ext_bdf => this%ext_bdf, event => glb_cmd_event)
```
MartinKarp, line 622: Nice! There was no problem using using `this%ext_bdf` here? Just checking the update  of it in simulation.f90 and such propagates as it should.

MartinKarp (commented):
Really nice I think!

src/fluid/fluid_pnpn.f90
timfelle: Well the call from simulation was made with `C%fluid%ext_bdf` so I would certainly hope it is the same as calling `this%ext_bdf` inside the function. I have not detected any changes but I am also not exactly sure which examples use that feature.

MartinKarp (approved):

PR #1875, opened 2025-05-02
Feature/wall model device

Shiyu-Sandy-Du:
In this PR, device implementation is done for the following wall models.

- [x] Spalding
- [x] Rough log law

I wrote something but I am not 100% sure it's correct... But anyway the Spalding's model and the rough log law works in the same way as the CPU implementation now.


timofeymukha:
Great work @Shiyu-Sandy-Du !! I've actually been working on this as well, but thankfully I have not reached much further than wall_model.f90. I will take a closer look on Sunday!

src/wall_models/bcknd/cpu/spalding_cpu.f90
```diff
@@ -0,0 +1,141 @@
...
+    integer, intent(in), dimension(n_nodes) :: ind_r, ind_s, ind_t, ind_e
+    real(kind=rp), dimension(n_nodes), intent(in) :: n_x, n_y, n_z, h
+    real(kind=rp), intent(in) :: nu
+    real(kind=rp), dimension(n_nodes), intent(inout) :: tau_x, tau_y, tau_z
+    real(kind=rp), intent(in) :: kappa, B
+    integer :: i
+    real(kind=rp) :: ui, vi, wi, magu, utau, normu, guess
+
+    do i=1, n_nodes
+       ! Sample the velocity
```
timofeymukha, line 60: What I think would be great is factoring out sampling into a separate routine, because this will be done for essentially any wall model. Moreover, when the new interpolation code is completed by Martin, we will probably switch to probe-based sampling, so using index-based addressing will only be a backup. It would be good, to only have to change the logic in one place. But perhaps we can keep it like this in this PR to just have things ported to the device.

Shiyu-Sandy-Du: I think you are right. We could shift to probe-based sampling and separate it out once new interpolation code is merged :)

njansson (approved):

timofeymukha (approved):

PR #1876, opened 2025-05-04
Disable right-handedness checker

timofeymukha:
It claims hemi has an illegal mesh, which I guess has to be wrong. I tried on a bunch of other examples, and they all pass (pip, cylinder, channel, RBC). A little strange, but I guess there must be some bug.


timfelle (approved):

src/mesh/mesh.f90
```diff
@@ -402,7 +402,7 @@ subroutine mesh_finalize(this)
     end do
     call this%curve%finalize()
 
-    call this%check_right_handedness()
+    !call this%check_right_handedness()
```
vbaconnet, line 405: can we just add an extra comment that explains why this line is commented? To avoid accidents :)  
```suggestion
    ! Due to a bug, right handedness check disabled for the time being.
    !call this%check_right_handedness()
```

MartinKarp:
I wouldnt be surpirised if hemi is wrong either tbh.

MartinKarp (approved):

PR #1877, opened 2025-05-05
Remove generic power and fix chsign

MartinKarp:
Only works with -O2 or lower on LUMI still.


gloopydoop (approved):
looks good for me so I'll approve, but I'm guessing Tim will want to do some tinkering with the `**`

:+1:

timfelle (approved):
LGTM.

PR #1880, opened 2025-05-08
Add possibility to install with pixi

timofeymukha:
Adds a pixi.toml for installing Neko with pixi. For now, CPU only, though I think we can extend this later, since conda-forge can install cuda and rocm I believe.

I think this is now the fastest and easiest way to get a CPU build with all deps, and also a convenient isolated dev environment.
Please try it out on your machines as per the instructions.


MartinKarp:
How does this differ from installing with spack?

timofeymukha:
> How does this differ from installing with spack?

The obvious diff is of course that it doesn't use spack, which is not everybody's cup of tea. This also gives you an isolated environment, spack does not by default. It is also faster, because none of the stuff you see under `dependencies` in the pixi.toml is compiled. Spack will generally compile everything from scratch. My goal with this is getting a fully charged neko in under 5 mins with minimal pain.

I tried to use spack quite a lot over the last 2 years, and I find it to be a highly-configurable but quite difficult tool to use. Very often some dep would fail to compile, some weird interactions would occur with the outside environment, etc. It is probably my fault, but I would like to have something there for the users that are actually *worse* than me at handling this type of stuff.

timofeymukha:
> I see! Sounds convenient :) Especially the not compiling thing. On a side note it also difficult with spack for Neko? Havent tried installing neko with it for a while and I have had issues with spack for things with lots of deps, but just Neko-cpu is not that much or?

Give it a try, it is very unintrusive besides the pixi install. I think the neko-cpu with spack is not so bad, but I never tried it with also hdf5, parmetis and gslib.

timfelle (approved):

PR #1882, opened 2025-05-08
Programming examples

timofeymukha:
I've started preparing some programming examples in user files not attached to a particular case. I wanted to have a .f90 as the format so that one can actually check that they compile as the code gets updated. I hope you like the format.

@Mr-2c You have requested something like this a long while back, I think it would be good to have for v1.0. What topics would you think be good to cover?


Mr-2c:
I did not really have too much of an opinion on what the program itself should or could do, but rather that there is a step by step tutorial of how to create it. I was inspired by the 'poisson' example that was a completely new program using Neko, and make me realize how powerful and flexible neko can be, only to then realize that when I don't understand how it was created, there was no real benefit to knowing it could be done... 

so what I thought was best was a tutorial video (nothing formal, something like what you did for us in the Neko meeting a while back, which by the way I have already forgotten) that goes though the creation of that program step by step.

if you decide to do it, my suggestion is to start by something simple and silly and keep adding stuff to it until it becomes something like 'poison' so that every user level from beginner to advance can follow it and watch a portion to do what they want. and the more informal it is the better in my opinion. for example, if it does not compile the first time, etc., just leave it, because that also teaches people a lot! I know it might take a couple of hours to create such a tutorial, but I think it would save at least 5 hours from 100 people each in the end 😃

timofeymukha:
> I did not really have too much of an opinion on what the program itself should or could do, but rather that there is a step by step tutorial of how to create it. I was inspired by the 'poisson' example that was a completely new program using Neko, and make me realize how powerful and flexible neko can be, only to then realize that when I don't understand how it was created, there was no real benefit to knowing it could be done...
> 
> so what I thought was best was a tutorial video (nothing formal, something like what you did for us in the Neko meeting a while back, which by the way I have already forgotten) that goes though the creation of that program step by step.
> 
> if you decide to do it, my suggestion is to start by something simple and silly and keep adding stuff to it until it becomes something like 'poison' so that every user level from beginner to advance can follow it and watch a portion to do what they want. and the more informal it is the better in my opinion. for example, if it does not compile the first time, etc., just leave it, because that also teaches people a lot! I know it might take a couple of hours to create such a tutorial, but I think it would save at least 5 hours from 100 people each in the end 😃

I guess for now I am doing it in written form instead of a video :). I should see what is inside Poisson.

timfelle (commented):
Have a look at the doxygen command [`include`](https://www.doxygen.nl/manual/commands.html#cmdinclude)

Which allow you to grab snippets of code and such out of files like the following line:
\include{} "sources/drivers/topopt-user.f90"

This would allow the source code to be directly part of the documentation you are writting with out copypasting stuff arround.

njansson (approved):
🍰 !

njansson:
Very nice! Definitely needed for 1.0

timfelle (approved):

PR #1884, opened 2025-05-11
Removing the redundant rzero from time_interpolator_scalar

Kamyarmvn:
The redundant rzero in time_interpolator_scalar is removed and the loop is optimized.


njansson (approved):

src/common/time_interpolator.f90
```diff
@@ -139,18 +139,18 @@ subroutine time_interpolator_scalar(this, t, f_interpolated, f_n, tlag, n)
...
-                      &of stored convecting fields")
+       call neko_error("lwtmax is smaller than the number &
+       &of stored convecting fields")
     end if
 
     no = this%order - 1
     call fd_weights_full(t, tlag, no, 0, wt) ! interpolation weights
-    call rzero(f_interpolated, n)
 
     do concurrent (i = 1:n)
```
adperezm, line 149: I do not get it, why is rzero redundant?

src/common/time_interpolator.f90
```diff
@@ -139,18 +139,18 @@ subroutine time_interpolator_scalar(this, t, f_interpolated, f_n, tlag, n)
     wt = 0
 
     if (this%order .gt. lwtmax) then
-      call neko_error("lwtmax is smaller than the number &
-                      &of stored convecting fields")
+       call neko_error("lwtmax is smaller than the number &
+       &of stored convecting fields")
     end if
```
adperezm, line 144: This error is there, but lwtmax is hard coded. isn't there a way to directly infer how many fields exist when you initialize the object or something like that?

src/common/time_interpolator.f90
```diff
@@ -139,18 +139,18 @@ subroutine time_interpolator_scalar(this, t, f_interpolated, f_n, tlag, n)
...
 
     no = this%order - 1
     call fd_weights_full(t, tlag, no, 0, wt) ! interpolation weights
-    call rzero(f_interpolated, n)
 
     do concurrent (i = 1:n)
-       do l = 0, no
-        f_interpolated(i) = f_interpolated(i) + wt(l) * f_n(i,l)
-      end do
+       f_interpolated(i) = wt(0) * f_n(i, 0)
```
adperezm, line 150: This only works on CPUs, right?.

We should have an error message saying that this is not implemented on device.

src/common/time_interpolator.f90
Kamyarmvn: In the new revision, instead of initializing f_interpolated with zeros first, the initialization is done directly in the loop, followed by the accumulation. This will reduce the number of writes for f_interpolated.

src/common/time_interpolator.f90
Kamyarmvn: The value of lwtmax will always be more than the number of convecting fields. The number of convecting fields used for interpolation is equal to the order of the time integration scheme. So I think the case of the error will probably never happen.

src/common/time_interpolator.f90
Kamyarmvn: Yes, this function as a part of the OIFS scheme works only with CPUs. I can add the error.

src/common/time_interpolator.f90
adperezm: I see! good!

src/common/time_interpolator.f90
adperezm: aha!, I see.

src/common/time_interpolator.f90
adperezm: Okay! I understand. However, given that this type can be used by anyone, not just OIFS. I would really suggest to put that message.  Should not be too much of an overhead to do so.

Kamyarmvn: Sure! I will put the error.

adperezm: Thanks!

adperezm:
Someone with merging permission should now accept this one.

PR #1889, opened 2025-05-13
Add singleton for user access

timofeymukha:
The core of the PR is this

```fortran
  !> Helper type to give users global access to the simulation case.
  type, public :: user_access_t
     type(case_t), pointer :: case
   contains
     !> Constructor.
     procedure, pass(this) :: init => user_access_init
     !> Destructor.
     procedure, pass(this) :: free => user_access_free
  end type user_access_t

  !> The singleton object.
  type(user_access_t), target, public :: neko_user_access
```
plus a corresponding addition to makeneko.in

The premise is that users need to have access to everything controlling the simulation, including the case and simcomps. However,  we don't want a global singleton case_t, because that, among other things, might promote using it in neko itself. Similarly, we may want to rethink how simcomps and case interact in terms of ownership. So the current (somewhat akward) way of using `neko_simcomps%case` in the user file is not very good.

The above type aims to solve this by
- Being a singleton that is *only init'ed by makeneko*. You really *can't* use it in normal code, you'll get a segfault.
- Serving essentially as a container where we can add any arbitrary stuff useful to the user. For example, if we remove `neko_simcomps`, we could add the simcomps as another component to `user_access_t`. So, while the contents of the type might fluctuate a bit until we really settle on everything, we never have to take exposing stuff to the user into consideration when designing our core types: we have `user_access_t` specifically for that purpose.

In essense, for users @pschlatt1 @Mr-2c, you will use the `user_access_t` variable to grab all possible stuff, in particular, the case and its contents (mesh, parameters, solvers, etc). Think of it as a modern `include TOTAL` that does not polute your namespace.

As a bonus, the PR harmonizes calling the user variable `user` everywhere (e.g. in examples we had `u`, `usr` and `user`).


src/common/user_access_singleton.f90
```diff
@@ -0,0 +1,73 @@
...
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+
+! Defines a singleton object available in the user file. Intended to allow
+! unrestricted access to the entire simulation case and enable all sorts of
+! hacking, while keeping neko proper clear.
+! The object is only initialized by makeneko, so trying to use it
+! outside user code will result in a segfault---intentionally.
```
timfelle, line 37: Lets have doxygen see this
```suggestion
!> User access singleton
!! Defines a singleton object available in the user file. Intended to allow
!! unrestricted access to the entire simulation case and enable all sorts of
!! hacking, while keeping neko proper clear.
!! The object is only initialized by makeneko, so trying to use it
!! outside user code will result in a segfault---intentionally.
```

src/simulation.f90
```diff
@@ -96,7 +96,7 @@ subroutine neko_solve(C)
     call neko_log%section('Postprocessing')
     call C%output_controller%execute(C%time)
 
-    call C%usr%user_init_modules(C%time%t, C%fluid%u, C%fluid%v, C%fluid%w,&
+    call C%user%user_init_modules(C%time%t, C%fluid%u, C%fluid%v, C%fluid%w,&
```
timfelle, line 99:
```suggestion
    call C%user%user_init_modules(C%time%t, C%fluid%u, C%fluid%v, C%fluid%w, &
```

njansson (approved):

PR #1891, opened 2025-05-14
Add user-specified pressure for `outflow` and `normal_outflow` BCs

vbaconnet:
As the title says. 

This is done by allowing the user to specify pressure profiles through the field dirichlet interface, with the same logic as for `user_pressure`.


timofeymukha:
Looks good. At some point I guess we will add the possibility to set the velocity and pressure bcs separately and that will allow to have these type of combos more naturally

timofeymukha (approved):

njansson (approved):

PR #1892, opened 2025-05-14
Differential operator simcomps

timofeymukha:
- Adds div and grad simcomps.
- Renames vorticity to curl. Can compute based on an arbitrary list of 3 registered fields.
- Fixes non-json constructors.
- Adds JSON control for the name of the fields the simcomps register in field_registry (`registered_name` keyword).
- Docs.

Bonus: field_registry lists its field before dying to due to requesting an non-existent field.


src/field/field_registry.f90
```diff
@@ -220,8 +221,14 @@ function get_field_by_name(this, name) result(f)
...
 
     if (.not. found) then
-       call neko_error("Field " // name // &
-            " could not be found in the registry")
+       if (pe_rank .eq. 0) then
+          write(*,*) "Current field_registry contents:"
+
+          do i=1, this%n_fields()
+             write(*,*) "- ", this%fields(i)%name
+          end do
```
timfelle, line 229: Could we write to stderr instead of stdout to have the error message in one
place rather than split?

Especially since this will not even be in the log.

examples/tgv/tgv.case
```diff
@@ -45,7 +45,9 @@
   "simulation_components": 
   [
     {
-      "type": "vorticity",
+      "type": "curl",
+      "fields": ["u", "v", "w"],
+      "registered_name": "omega",
```
timfelle, line 50: Maybe call it input fields and output fields or something. "registered_name"
does not really explain that the result will be stored as a field.

timofeymukha: Yeah, I thought about that. I realized `output_` must be avoided because all the parameters starting with that are about stuff we save to disk, like `output_filename`, saw it might introduce confusion. Maybe `registered_field` or `computed_field` or even `registered_field_name`?

timfelle: I think `computed_field` is the better one if we retain the input just being `fields`.

src/simulation_components/derivative.f90
timfelle, line 1: Why are all the other ones called "operator"_simcomp, and this one called "derivative"
We should probably align that naming as well.

src/simulation_components/curl_simcomp.f90
```diff
@@ -0,0 +1,272 @@
...
+    character(len=20), allocatable :: field_names(:)
+    character(len=:), allocatable :: registered_name
+
+
+    call json_get_or_default(json, "registered_name", registered_name, "curl")
+    call json_get(json, "fields", field_names)
+
+    if (size(field_names) .ne. 3) then
+       call neko_error("The curl simcomp requires exactly 3 entries in " // &
+            "field_names.")
```
timfelle, line 114: The parameter in the json is called `fields` not `field_names`.

src/simulation_components/derivative.f90
timofeymukha: Because I hate the name dudxyz and wanted to save the users from it :-)).

timofeymukha: Actually weak_grad also wraps opgrad, but weak_grad is just so much more telling.

src/field/field_registry.f90
timofeymukha: Outputting to error_unit now.

src/simulation_components/derivative.f90
timfelle: I agree, I would probably just change grad to gradient, div to divergence and so on. Spell out the names of the operators instead of contracting them. And adding the _simcomp in order to distinguish them from the math modules.

src/simulation_components/div_simcomp.f90
```diff
@@ -0,0 +1,243 @@
...
+    character(len=20), allocatable :: field_names(:)
+    character(len=:), allocatable :: computed_field
+
+
+    call json_get_or_default(json, "computed_field", computed_field, "div")
+    call json_get(json, "fields", field_names)
+
+    if (size(field_names) .ne. 3) then
+       call neko_error("The div simcomp requires exactly 3 entries in " // &
+            "field_names.")
```
timfelle, line 107: Please let the json entry match the error message

src/simulation_components/derivative.f90
timofeymukha: Done!

timfelle (approved):
I like i

MartinKarp:
This is breaking some examples

timofeymukha:
> This is breaking some examples

I will check

PR #1893, opened 2025-05-15
Fix install target

njansson:
Fix installation target such that we only install proper headers


njansson:
Regarding the `.cl` files, These should not be installed given our current JIT approach in Neko

tuananhdao (approved):

PR #1904, opened 2025-05-22
Add missing documentation page to index

timofeymukha:
I think this is all what is needed to make it show up, right?

### DIFF
--- a/doc/DoxygenLayout.xml
+++ b/doc/DoxygenLayout.xml
@@ -4,8 +4,6 @@
   <!-- Navigation index tabs for HTML output -->
   <navindex>
     <tab type="mainpage" visible="yes" title=""/>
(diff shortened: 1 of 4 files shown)


timfelle:
I think you also want to add it as a subpage in the user guide matkdown file. Otherwise it will appear as a top level page.

You can chevk the documentation locally by running 'make html' and open the docs/html/index.html file

timfelle (approved):

njansson (approved):

timofeymukha:
> I think you also want to add it as a subpage in the user guide markdown file. Otherwise it will appear as a top level page.
> 
> You can chevk the documentation locally by running 'make html' and open the docs/html/index.html file

You are right, that was missing. I have troubles building the html on my machine for some reason. The menu items become broken links etc.

PR #1905, opened 2025-05-23
Fix/0-iterations

timfelle:
Warnings were printed about non-convergence, when returning early.

Additionally we now return if the initial residual is converged, it does not have to be exactly 0.


njansson:
Are we good to merge this or do we need more testing?

timfelle:
> Are we good to merge this or do we need more testing?

From my part, it is done. Unless someone else wants more testing.

MartinKarp (changes requested):
I have a strong hunch this will lead to surprise divergence for various cases where one currently only has 1 pressure iteration. As such I would really suggest this to not be a default setting.

src/krylov/bcknd/cpu/gmres.f90
```diff
@@ -214,7 +215,7 @@ function gmres_solve(this, Ax, x, f, n, coef, blst, gs_h, niter) &
             ksp_results%res_start = gam(1) * norm_fac
          end if
 
-         if (abscmp(gam(1), 0.0_xp)) return
+         if (abscmp(gam(1), 0.0_xp)) exit
```
MartinKarp, line 218: Sorry, for gmres we do not do early exit?

timofeymukha (approved):

njansson:
I share the same concern as @MartinKarp, unless I've missed something this will fail when we have for example a single pressure iteration, and the initial residual is very small.

njansson:
> > The issue as I see it is that the early termination checks (before the iteration starts) does not involve the solution, nor the initial guess. These checks for the trivial solution, thus we can easily find a right hand side that is small enough, then we will have false early convergence, with no update on the solution vector
> 
> That makes a lot of sense.
> 
> I only added an early return for CG. We can remove that if you want. All the rest of the solvers already had them, i just made sure the `ksp_results` was updated before returning early.

That is definitely needed, if the check can be modified to check against 0.0 instead of `.gt. abstol` then all should be fine

timfelle:
> > > The issue as I see it is that the early termination checks (before the iteration starts) does not involve the solution, nor the initial guess. These checks for the trivial solution, thus we can easily find a right hand side that is small enough, then we will have false early convergence, with no update on the solution vector
> > 
> > 
> > That makes a lot of sense.
> > I only added an early return for CG. We can remove that if you want. All the rest of the solvers already had them, i just made sure the `ksp_results` was updated before returning early.
> 
> That is definitely needed, if the check can be modified to check against 0.0 instead of `.gt. abstol` then all should be fine

done

njansson (approved):

PR #1906, opened 2025-05-23
Add  material properties to the field registry

timofeymukha:
Added as `this%name`, plus `_rho`, `_mu`, `_lambda`, `_cp`.

I am mainly doing this to have easy access to this in wall models and also force_torque simcomp.

I guess we are kid of drifting towards registering more and more things, but maybe that's not too bad as long as the naming is solidly unique.

### DIFF
--- a/examples/rayleigh_benard_cylinder/rayleigh.f90
+++ b/examples/rayleigh_benard_cylinder/rayleigh.f90
@@ -39,11 +39,11 @@ contains
     type(field_list_t), intent(inout) :: properties
 
     if (name .eq. "fluid") then
-       call field_cfill(properties%get_by_name("rho"), 1.0_rp)
-       call field_cfill(properties%get_by_name("mu"), mu)
+       call field_cfill(properties%get_by_name("fluid_rho"), 1.0_rp)
+       call field_cfill(properties%get_by_name("fluid_mu"), mu)
     else if (name .eq. "scalar") then
-       call field_cfill(properties%get_by_name("cp"), 1.0_rp)
-       call field_cfill(properties%get_by_name("lambda"), mu / Pr)
+       call field_cfill(properties%get_by_name("scalar_cp"), 1.0_rp)
+       call field_cfill(properties%get_by_name("scalar_lambda"), mu / Pr)
     end if
   end subroutine set_material_properties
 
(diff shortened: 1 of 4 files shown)


Shiyu-Sandy-Du:
Considering multiple scalar equations,  is there anyway to register a field list instead of fields?

Shiyu-Sandy-Du:
> > Considering multiple scalar equations, is there anyway to register a field list instead of fields? Maybe @tuananhdao has some idea?
> 
> Not really since the registry is for individual fields only. But each scalar will register its own fields, with unique names based on scalart_scheme_t%name.

So their material property should also be unique, right?

timofeymukha:
> Are we getting to a point where it makes more sense to make a dedicated material property class, and associate a name to that one. That would pave the way toward having multiple material properties for different regions of the simulation.
> 
> With this we are associating a material property to each solver. Combine that with our work toward multiple scalars, that could turn out to be a challenge memory wise. Whereas having a registry of material property objects, would allow multiple solvers to share their materials.

So, we actually had that before, but it was just a struct with constant mu, rho, lambda and cp. We could revive it as a field registry that sits in the case. Should we then have a dedicated JSON object in the case for the properties? it is a bit tricky with also the non-dimentional input support, but one could try..

timofeymukha:
> > So, we actually had that before, but it was just a struct with constant mu, rho, lambda and cp. We could revive it as a field registry that sits in the case. Should we then have a dedicated JSON object in the case for the properties? it is a bit tricky with also the non-dimentional input support, but one could try..
> 
> I am not talking about the field registry. I am talking about a new registry for material components. And yes a struct or class would be needed for such a thing, the base material could be quite simple and have two accessors, one returning a pointer to a field of mu's and one returning just a value. That way we only allocate the field whenever it is needed.

What I meant is that it would be *a* field registry, separate from the neko_field_registry. But I see what you mean, you want to also be able to have mu/lambda as a constant. I think that would be very hard to manage though. If we look at the fluid and scalar, we are more or less forced to have a field, because we pass mu/lambda to a bunch of stuff (residual calculations, volflow, etc.) So one would then have to maintain two interfaces for all of those, one for a scalar mu and one for a field. So I think the best thing we can do is to have fields, but try to use pointers when the properties are the same. That said, I don't think it will save much in practice, because it will probably be rare that you have a lot of scalars with the same material properties.

timofeymukha:
@njansson @timfelle What I suggest is that once we have multiple scalars, we can have material properties in an object in JSON and support something like

```json5
"material_properties":
{
    "copy_from": "temperature" // The name of the scalar to copy from
}
```
and then lambda and cp will point to `temperature_lambda` and `temperature_cp` in the neko_field_registry.

njansson (approved):

PR #1910, opened 2025-05-26
Add bounding box output to the mesh_checker

timofeymukha:
Somehow it disappeared from a previous PR, so I reimplemented.

### DIFF
--- a/contrib/mesh_checker/mesh_checker.f90
+++ b/contrib/mesh_checker/mesh_checker.f90
@@ -122,7 +122,6 @@ program mesh_checker
      end if
   end do
 
-
   if (write_zone_ids) then
      if (pe_rank .eq. 0) write(*,*) 'Writing zone ids to zone_indices0.f00000'
      call gs%init(dofmap)


contrib/mesh_checker/mesh_checker.f90
```diff
@@ -77,20 +78,35 @@ program mesh_checker
...
      write(*,*) 'Number of faces:    ', msh%glb_mfcs
      write(*,*) 'Number of edges:    ', msh%glb_meds
-     write(*,*) ''
+     write(*,*) 'Bounding box:'
+     write(*,'(A,2(g10.3,1X))') '    x', xmin, xmax
+     write(*,'(A,2(g10.3,1X))') '    y', ymin, ymax
+     write(*,'(A,2(g10.3,1X))') '    z', zmin, zmax
+  end if
+
+  if (pe_rank .eq. 0) then
```
timfelle, line 109: Why did you split the if statement in two?

Also you removed the empty line in the log that was printed before the zone header.

contrib/mesh_checker/mesh_checker.f90
```diff
@@ -108,10 +124,9 @@ program mesh_checker
      end if
   end do
 
+
```
timfelle, line 127:
```suggestion
```

njansson (approved):

PR #1914, opened 2025-06-02
Add possibility to add aliases to field_register_t

timofeymukha:
- Adds the possibility to add alieases to an exsiting field in the registry. For, example if there is a `field1` in the registry, we can add an alias for it, say `alias1`. Then if we call `get_field`, we can retrieve the field `field1` by both names: `field1` and `alias1`. I think this may be useful in contexts where we want have a field register under a particular name, but not take up extra memory. Material properties was an inspiration. If we have many scalars with the same properties, we still want fields like `saclar1_lambda` and `scalar2_lambda` to be in the registry. Otherwise, external objects somehow have to know that e.g. `scalar1_lambda` exists and `scalar2` uses it. With aliases, `scalar2` can just add an alias to the registry.

   Internally the aliasing is done using an array of small jsons containing
  ```json5
  {
  "alias": "alias_name",
  "target": "name_of_actual_field"
  }
  ```

- Refactored internal routine names to follow the standard conventions.
- Added some unit tests.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -398,6 +398,7 @@ AC_CONFIG_FILES([tests/unit/stack/Makefile\
                  tests/unit/vector/Makefile\
                  tests/unit/matrix/Makefile\
                  tests/unit/scratch_registry/Makefile\
+                 tests/unit/field_registry/Makefile\
                  tests/unit/fast3d/Makefile\
                  tests/unit/time_scheme/Makefile\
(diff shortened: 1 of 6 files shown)


timfelle (approved):
I am a bit confused about why a list of json_files would be better than two lists of strings or something :sweat_smile:

PR #1928, opened 2025-06-04
Some more fixes for gs and avoid accidently deallocating things too early

MartinKarp:


MartinKarp:
Are we changing the format of space_t?

timfelle (commented):
I guess stuff could be changed during the restructure of the mesh and such.

njansson (approved):

timfelle (approved):

PR #1931, opened 2025-06-05
Add documentation for non-MPI based comm. backends

### LINKED ISSUE #1729: Documentation of non-MPI based gs comm. backends
Provide documentation how to compile and use the new communication backends (NCCL, RCCL and NVSHMEM) in Neko.

njansson:
Fixes #1729

Installation instructions for NVSHMEM we need to defer for now.


njansson:
Regarding NVSHMEM, it's currently no easy, generic way to link Neko with NVHSMEM, since the linker needs to be changed to `nvcc` and all MPI libraries have to be figured out and added manually to the link.

PR #1932, opened 2025-06-05
Fix/hsmg stuff and zero v_d

MartinKarp:
Some bcs in HSMG that were not quite ok.

Found that one sometimes NaNs on the first device_gather and thought it is good to ensure v_d is 0 just in case. Dont think it is the root problem though.

### DIFF
--- a/src/gs/bcknd/device/gs_device.F90
+++ b/src/gs/bcknd/device/gs_device.F90
@@ -36,6 +36,7 @@ module gs_device
   use num_types
   use gs_bcknd
   use device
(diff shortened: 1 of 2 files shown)


njansson (commented):
Not too happy about the syncs, not sure how that would affect multistream runs. 

How about adding a memset in device_alloc instead? That would also fix similar issues in other places of the code

MartinKarp:
This will only be called on the first invocation right?

njansson:
Fixed in #1957 and #2017.

PR #1945, opened 2025-06-14
Refactor file_t constructor to a normal procedure

timofeymukha:
Removes the `= file_t()` constructor and adds the standard `%init` subroutine. As @MartinKarp pointed out, the = don't work well, and we use `%init` for everything else.

### DIFF
--- a/src/io/file.f90
+++ b/src/io/file.f90
@@ -55,6 +55,8 @@ module file
   type file_t
      class(generic_file_t), allocatable :: file_type
    contains
+     !> Constructor
+     procedure, pass (this) :: init => file_init
      !> Writes data to a file.
      procedure :: write => file_write
      !> Read @a data from a file.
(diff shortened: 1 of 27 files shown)


src/io/file.f90
```diff
@@ -75,20 +77,16 @@ module file
...
-  !> File reader/writer constructor.
+  !> Constructor.
   !! @param fname Filename.
-  function file_init(fname, header, precision, layout) result(this)
+  subroutine file_init(this, fname, header, precision, layout)
+    class(file_t), intent(inout) :: this
     character(len=*) :: fname
     character(len=*), optional :: header
     integer, optional :: precision
     integer, optional :: layout
```
timfelle, line 89: Could we add intents to all of these? I guess they are `intent(in)`

timofeymukha:
> Doesn't the programming examples also need to be modified?

I think not, because we never use file_t directly. And the other types have a proper %init!

njansson (approved):

PR #1953, opened 2025-06-18
Added tanh Blasius profile

nick-morse:
Adding the tanh-based Blasius profile approximation from [Ö. Savaş (2012)](https://doi.org/10.1016/j.cnsns.2012.02.002) (equations 2 and 11), which has good accuracy compared to the existing approximations. 

The only notable difference to existing approximations is that `delta` in this case corresponds to the 99\% thickness, such that flow for $y > \delta$ takes the slowly varying value of the approximation instead of an imposed constant $U_\infty$. `delta` could be changed to correspond to a different percent thickness if you think that would be more appropriate.

blasius_profiles

### DIFF
--- a/doc/pages/user-guide/case-file.md
+++ b/doc/pages/user-guide/case-file.md
@@ -436,6 +436,7 @@ file documentation.
       - `cubic`, cubic approximation.
       - `quartic`, quartic approximation.
       - `sin`, sine function approximation.
+      - `tanh`, hyperbolic tangent approximation of Savaş (2012). In this case `delta` is the 99\% thickness.
 4. `point_zone`, the values are set to a constant base value, supplied under the
    `base_value` keyword, and then assigned a zone value inside a point zone. The
    point zone is specified by the `name` keyword, and should be defined in the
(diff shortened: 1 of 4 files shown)


njansson (approved):
Nice!

njansson:
> All good, I think.
> 
> But keep in mind that Blasius is not very well implemented in general, if I recall. @MartinKarp used to always remind us of that. There are some strict limitations to the mesh, if I remember correctly.

Yeah the issues are related to the shape and orientation of the element. It currently assumes a nicely axis aligned element.

nick-morse:
> Do you think the tanh approximation is good enough? We have plenty of blasius solvers do we can also implement the right one, if needed. We should however not use the quadratic or other polynomials (pohlhausen)

I think the tanh profile is sufficient for many problems, but perhaps a Blasius solver could be implemented if more accuracy is needed (transition?) or if we'd just like to always use the most accurate method. Alternatively, the user could specify a user-defined inflow BC with a Blasius solver or interpolation of the profile in the user file.

PR #1542, opened 2024-10-11
Global interpolation without GSLIB

MartinKarp:
Global interpolation without gslib. Same functionality as previously + GPU support, but no need to build with gslib any more. Removes gslib entirely from the codebase. I am now happy enough with it that I think we can start tying to get it into develop. It will continue to be optimized during the spring. 

# Features

- Machinery for finding rst coordinates on the CPU/GPU without GSLIB
- Options to execute interpolation on the host rather than device
- Functions to copy data back and forth between host and device for vectors, fields and matrices
- Unit tests for global interpolation

# Todo before merge

- [x] Clean code a bit more and double check for comments.
- [x] Understand issue with Mac double precision test.

PR #1957, opened 2025-06-19
Add a device_memset before any gs operations (for extra safety)

njansson:
Add a memset after gs buffers are allocated. This is not needed for normal gather-scatter operations since they operate on the full length (m) for each call, however async. operations (as the ones done by global interpolation #1542, could be different).

This is similar to what was done in #1932, but here we use memset in the operating stream, thus reduce the numbers of global sync. points.


njansson:
The additional memset for the MPI recv buffer fixes divergence (NaN) when devices has very few elements

timfelle (approved):

PR #1959, opened 2025-06-20
Fix a bunch use statements 

timofeymukha:
Adds `only`.

Also changed the `map_t` type a bit to have a normal constructor structure with TBPs.

### DIFF
--- a/src/comm/parmetis.F90
+++ b/src/comm/parmetis.F90
@@ -108,10 +108,10 @@ contains
 
   !> Compute a k-way partitioning of a mesh @a msh
   subroutine parmetis_partmeshkway(msh, parts, weights, nprts)
-    type(mesh_t), intent(inout) :: msh                !< Mesh
-    type(mesh_fld_t), intent(inout) :: parts          !< Partitions
+    type(mesh_t), intent(inout) :: msh !< Mesh
+    type(mesh_fld_t), intent(inout) :: parts !< Partitions
     type(mesh_fld_t), intent(in), optional :: weights !< Weights
-    integer, intent(in), optional :: nprts            !< Number of partitions
+    integer, intent(in), optional :: nprts !< Number of partitions
     integer(kind=M_INT), target :: wgtflag, numflag, ncon, ncommonnodes
     integer(kind=M_INT), target :: nparts, options(3), edgecut, rcode
     real(kind=M_REAL), allocatable, target, dimension(:) :: tpwgts, ubvec
@@ -179,7 +179,7 @@ contains
   !> Compute a k-way partitioning of a mesh @a msh using
   !! a coordinated-based space-filing curves method
   subroutine parmetis_partgeom(msh, parts)
-    type(mesh_t), intent(inout) :: msh       !< Mesh
+    type(mesh_t), intent(inout) :: msh !< Mesh
     type(mesh_fld_t), intent(inout) :: parts !< Partitions
     integer(kind=M_INT), target :: ndims
     real(kind=M_REAL), allocatable, target, dimension(:) :: xyz
@@ -285,17 +285,17 @@ contains
 
   !> Compute a k-way partitioning of a mesh @a msh
   subroutine parmetis_partmeshkway(msh, parts, weights, nprts)
-    type(mesh_t), intent(inout) :: msh                !< Mesh
-    type(mesh_fld_t), intent(inout) :: parts          !< Partitions
+    type(mesh_t), intent(inout) :: msh !< Mesh
+    type(mesh_fld_t), intent(inout) :: parts !< Partitions
     type(mesh_fld_t), intent(in), optional :: weights !< Weights
-    integer, intent(in), optional :: nprts            !< Number of partitions
+    integer, intent(in), optional :: nprts !< Number of partitions
     call neko_error('NEKO needs to be built with ParMETIS support')
   end subroutine parmetis_partmeshkway
 
   !> Compute a k-way partitioning of a mesh @a msh using
   !! a coordinated-based space-filing curves method
   subroutine parmetis_partgeom(msh, parts)
-    type(mesh_t), intent(inout) :: msh       !< Mesh
+    type(mesh_t), intent(inout) :: msh !< Mesh
     type(mesh_fld_t), intent(inout) :: parts !< Partitions
     call neko_error('NEKO needs to be built with ParMETIS support')
   end subroutine parmetis_partgeom
--- a/src/comm/mpi_types.f90
+++ b/src/comm/mpi_types.f90
@@ -40,30 +40,30 @@ module neko_mpi_types
   implicit none
   private
 
-  type(MPI_Datatype) :: MPI_NMSH_HEX    !< MPI derived type for 3D Neko nmsh data
-  type(MPI_Datatype) :: MPI_NMSH_QUAD   !< MPI derived type for 2D Neko nmsh data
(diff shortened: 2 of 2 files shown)


timofeymukha:
The formatter here is not doing a very good job with line-up

```
<! Comment
```

strings. We don't really use that often anymore, but here we do. I think we can just force-merge.

timfelle (commented):
> The formatter here is not doing a very good job with line-up
> 
> ```
> <! Comment
> ```
> 
> strings. We don't really use that often anymore, but here we do. I think we can just force-merge. 

Well It does exactly what we asked about for. The line-ups are not really a special construct and we asked the formatter to remove multiple spaces if they are not indentation :sweat_smile:

src/common/time_interpolator.f90
```diff
@@ -39,7 +39,7 @@ module time_interpolator
   use math, only : add3s2, rzero
   use utils, only : neko_error
   use, intrinsic :: iso_c_binding
```
timfelle, line 41: I see no pointers or anything here. Is this used?

src/comm/parmetis.F90
```diff
@@ -33,11 +33,11 @@
...
-  use utils
-  use num_types
-  use mesh_field
-  use mesh
+  use point, only : point_t
+  use utils, only : neko_error
+  use num_types, only : rp
+  use mesh_field, only : mesh_fld_t
+  use mesh, only : mesh_t
   use, intrinsic :: iso_c_binding
```
timfelle, line 41: why not add only to this as well `, only: c_ptr`

src/io/buffer/buffer.F90
```diff
@@ -32,8 +32,8 @@
 !
 !> Generic buffer that is extended with buffers of varying rank
 module buffer
-  use num_types
-  use vector
+  use num_types, only : rp
+  use vector, only : vector_t
 #ifdef HAVE_ADIOS2_FORTRAN
   use adios2
```
timfelle, line 38: Why not for adios as well?

src/io/buffer/buffer_1d.F90
```diff
@@ -32,12 +32,12 @@
 !
 !> Generic buffer that is extended with buffers of varying rank
 module buffer_1d
-  use num_types
-  use vector
+  use num_types, only : dp, sp, i8, rp
+  use vector, only : vector_t
 #ifdef HAVE_ADIOS2_FORTRAN
   use adios2
```
timfelle, line 38: Here we can add some only as well

src/io/buffer/buffer_4d.F90
```diff
@@ -32,12 +32,12 @@
 !
 !> Generic buffer that is extended with buffers of varying rank
 module buffer_4d
-  use num_types
-  use vector
+  use num_types, only : dp, sp, i8, rp
+  use vector, only : vector_t
 #ifdef HAVE_ADIOS2_FORTRAN
   use adios2
```
timfelle, line 38: Keeps on going

src/io/buffer/buffer_4d_npar.F90
```diff
@@ -32,12 +32,12 @@
 !
 !> Generic buffer that is extended with buffers of varying rank
 module buffer_4d_npar
-  use num_types
-  use vector
+  use num_types, only : dp, sp, i8, rp
+  use vector, only : vector_t
 #ifdef HAVE_ADIOS2_FORTRAN
   use adios2
```
timfelle, line 38: and here as well

src/io/bp_file.F90
```diff
@@ -34,23 +34,25 @@
...
+  use vector, only : vector_t
+  use space, only : space_t
+  use mesh, only : mesh_t
+  use structs, only : array_ptr_t
+  use fld_file_data, only : fld_file_data_t
+  use utils, only : neko_error
+  use datadist, only : linear_dist_t
   use comm
 #ifdef HAVE_ADIOS2_FORTRAN
   use adios2
```
timfelle, line 50: Why not for adios

src/math/fdm.f90
```diff
@@ -60,22 +60,23 @@
 !> Type for the Fast Diagonalization connected with the schwarz overlapping solves.
 module fdm
   use neko_config
```
timfelle, line 62: why not neko_config

timfelle (changes requested):
Any chance we can settle on whether we want, it is kinda mixed.
 - use `module, only : function`
 - use `module, only: function`

I would say, `module, only: function`, is the most readable.

src/io/re2_file.f90
```diff
@@ -34,19 +34,22 @@
...
+  use map, only : map_t
+  use mesh, only : NEKO_MSH_MAX_ZLBLS
   use logger, only: neko_log, LOG_SIZE
+  use utils, only : neko_error, filename_chsuffix
+  use re2, only : Re2_ENDIAN_TEST, RE2_HDR_SIZE, re2v1_xyz_t, re2v1_xy_t, &
+       re2v1_curve_t, re2v1_bc_t, re2v2_xyz_t, re2v2_xy_t, &
+       re2v2_curve_t, re2v2_bc_t
+  use htable, only : htable_pt_t
+  use comm
+  use mpi_f08
```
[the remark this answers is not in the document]
njansson: we have seen at least one instance of using `only` for MPI modules causing a segfault (with cray-mpich)

src/math/fdm.f90
timofeymukha: The answer here and above is that because I was lazy and ran out of time to sink on this :-). So, this is meant as an iterative improvement rather than a full fix. So I left out some modules that were a bit questionable or external, like adios, and mostly fixed easy stuff.

timfelle:
> > Any chance we can settle on whether we want, it is kinda mixed.
> > 
> > * use `module, only : function`
> > * use `module, only: function`
> > 
> > I would say, `module, only: function`, is the most readable.
> 
> I have incidentally been adding the space before : here, but I don't really care about it. I agree that we should do it the same everywhere.

@njansson, Do you have an oppinion?

njansson:
> > > Any chance we can settle on whether we want, it is kinda mixed.
> > > 
> > > * use `module, only : function`
> > > * use `module, only: function`
> > > 
> > > I would say, `module, only: function`, is the most readable.
> > 
> > 
> > I have incidentally been adding the space before : here, but I don't really care about it. I agree that we should do it the same everywhere.
> 
> @njansson, Do you have an oppinion?

Not to be that person, but would prefer `module, only : function` since it's more consistent with how we name variables `type, modifier :: name`

timfelle:
> > > > Any chance we can settle on whether we want, it is kinda mixed.
> > > > 
> > > > * use `module, only : function`
> > > > * use `module, only: function`
> > > > 
> > > > I would say, `module, only: function`, is the most readable.
> > > 
> > > 
> > > I have incidentally been adding the space before : here, but I don't really care about it. I agree that we should do it the same everywhere.
> > 
> > 
> > @njansson, Do you have an oppinion?
> 
> Not to be that person, but would prefer `module, only : function` since it's more consistent with how we name variables `type, modifier :: name`

Right, can anything follow `only` that is not the `:`. For the rypes we do so because the order of modifiers are mutable so the `::` symbol does not attach to a specific one. Whereas I would argue that the `only:` is a symbol itself, and it would not make sense to split it.

timfelle (approved):

Issue #1620, opened 2024-11-27
Consistent naming in case file


MartinKarp:
Consistent naming of different components of the case file. Originally raised in PR #1617 where the Krylov solvers are a good example of things that could be improved to comply more with the rest of the file.

Key questions are perhaps to settle on one whitespace character for example.

timfelle:
Additionally it might be worth mentioning that the discussion here are not how the keys of the json file should be named, but more how should the options them selves be formatted. Do we want to enforce lowercase, or do we want capitalization similar to how it would be presented in a paper. Below is a list of variants we could consider. My personal favorite is the bottom one, since that is very close to how it would be written in a paper, meaning it might be simpler to parse for a non-expert user.

- `"velocity_solver": { "type": "fusedcoupledcg", ...`
- `"velocity_solver": { "type": "fused_coupled_cg", ...`
- `"velocity_solver": { "type": "fused-coupled-cg, ...`
- `"velocity_solver": { "type": "FusedCoupledCG", ...`
- `"velocity_solver": { "type": "Fused_Coupled_CG", ...`
- `"velocity_solver": { "type": "Fused-Coupled-CG", ...`

timofeymukha:
I think we have de facto converged on options 1 and 2 in Tim's list. Where to put a _ is more or less keyword-dependent.

PR #1961, opened 2025-06-20
Add _ to ksp names

timofeymukha:
Closes issue #1620 by adding _ to some of the KSP type names. None used in examples, as far as I can see. Will break LES cases.

### DIFF
--- a/doc/pages/user-guide/case-file.md
+++ b/doc/pages/user-guide/case-file.md
@@ -320,15 +320,15 @@ A more detailed description of each boundary condition is provided below.
   ```
 * `outflow+user`. Same as `outflow`, but with user-specified
   pressure. The pressure is specified via the same interface as `user_pressure`,
-  see the 
+  see the
   [relevant section](#user-file_field-dirichlet-update) for more information.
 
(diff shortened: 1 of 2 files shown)


njansson:
Looking at other packages, e.g. PETSc, this seems to go against "common" practices, e.g. having `pipecg` and not `pipe_cg`

njansson:
Maybe we should check with some solver people, like @pohm01

njansson:
> > Looking at other packages, e.g. PETSc, this seems to go against "common" practices, e.g. having `pipecg` and not `pipe_cg`
> 
> 
> 
> To me, the internal consistency within neko weighs much heavier. However,  If there is an established term in the scientific literature, like bicgstab, then it is another story.

pipecg is a good example of an established name, while the fused, coupled are not

timfelle:
> > I have to agree with @njansson here that pipecg is an established name
> 
> Also cacg, so I think it's a good idea to not write them as pipe_cg and ca_cg

done

njansson (commented):
This would make doc and factory consistent

njansson (approved):

PR #1972, opened 2025-06-27
Fix out of local mem. issues with OpenCL, verified on Mi250X

njansson:
The current kernel sizes thrown an out of local memory


timfelle (commented):
They now support varying kernel sizes. is that intended? Some cap out on 11 others at 13

njansson:
> They now support varying kernel sizes. is that intended? Some cap out on 11 others at 13

Yes it depends on the amount of shared memory they use.

timfelle (approved):

PR #1976, opened 2025-06-27
Fix zero-sized array c_loc

njansson:
Fixes an issue when using a device backend, and the pnpn pressure surface terms are zero since no symmetry conditions are present. In this case, no facets will be marked, and `facet_normal_t` will, during init map a zero sized host array to the device, causing an invalid zero-sized `c_loc` call.  

To reproduce the issue, run e.g. the `turb_pipe` example

This is probably the cause of some of the face_normal related issues we have been observing with the device backend

### DIFF
--- a/src/bc/facet_normal.f90
+++ b/src/bc/facet_normal.f90
@@ -260,10 +260,14 @@ contains
           call unique_point_idx%set(this%msk(i), j)
        end if
     end do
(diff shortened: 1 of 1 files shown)


timfelle (approved):

pohm01 (approved):
lgtm

PR #1980, opened 2025-07-01
Change fld_generate_interpolator to a subroutine instead of a function

vbaconnet:
As the title says. This follows Martin's recommendations when it comes to functions returning objects containing pointers etc.


timofeymukha (approved):

njansson (approved):
Will move the target to master, and cherry-pick back to the release branch such that we have it in both places

njansson:
> Will move the target to develop, and cherry-pick back to the release branch such that we have it in both places

Ok that didn't work so we cherry pick to develop instead

Issue #1979, opened 2025-06-30
Dynamic Smagorinsky model diverges for TGV test


Shiyu-Sandy-Du:
It used to pass the test but not any more. The reasons are unclear yet. Debugging is in progress. Smagorinsky, Sigma and Vreman works well.

Shiyu-Sandy-Du:
Develop branch and release/0.9 gives the same eddy viscosity field, but DS model works on release/0.9. Current suspicion is on the equation solving.

pschlatt1:
there has been a similar issue with some wing cases (not examples) that also suddently do not work anymore. I will get more info on that on Friday.

Shiyu-Sandy-Du:
Another observation is that the solving process divergence starts with pressure, and the momentum's start residual raises up right after that.

pschlatt1:
yet, that is a bit scary as the dynamic smagorinsky model with spanwise/wall-parallel averaging is known to work in a channel.

Shiyu-Sandy-Du:
I don't think it's the issue with the averaging. I might figured it out already but have to check it in detail tomorrow.

timofeymukha:
Now that we have both relatively easy-to-add integration tests with pytest and also the nightly test suite developed by @njansson , please think what kind of test would catch the regression once it is identified.

njansson:
> Now that we have both relatively easy-to-add integration tests with pytest and also the nightly test suite developed by [@njansson](https://github.com/njansson) , please think what kind of test would catch the regression once it is identified.

Indeed if you have a case (not too large) @Shiyu-Sandy-Du , we can add it to the nightly regression suite.

Shiyu-Sandy-Du:
> > Now that we have both relatively easy-to-add integration tests with pytest and also the nightly test suite developed by [@njansson](https://github.com/njansson) , please think what kind of test would catch the regression once it is identified.
> 
> Indeed if you have a case (not too large) [@Shiyu-Sandy-Du](https://github.com/Shiyu-Sandy-Du) , we can add it to the nightly regression suite.

Yes I have one. Is a TGV case with 8^3 elements and P=7 a good one to add into the suite?

njansson:
> > > Now that we have both relatively easy-to-add integration tests with pytest and also the nightly test suite developed by [@njansson](https://github.com/njansson) , please think what kind of test would catch the regression once it is identified.
> > 
> > 
[...]

PR #1984, opened 2025-07-02
Fix/dynamic smagorinsky

Shiyu-Sandy-Du:
This PR is aim at issue #1979. The first commit (deleted) is to make the simulation run without diverging. And the following up fix some bugs which did not crash the simulation.


timofeymukha:
We can't have a const_mu in the base type for fluid schemes. It is a very special case. 

I need to fix @timfelle comments in https://github.com/ExtremeFLOW/neko/pull/1908 and we can work from there. Could tell what exactly is the issue with mu that caused the issues?

timofeymukha (approved):

njansson (approved):

Issue #1924, opened 2025-06-03
neko/2d_cylinder *** ERROR: Fluid solver diverged ***


timfelle:

timofeymukha:
No, this is because this examples uses a mesh with built-in boundary conditions. I would say we can just remove it, it does not provide anything compared to the ordinary cylinder.

njansson:
> No, this is because this examples uses a mesh with built-in boundary conditions. I would say we can just remove it, it does not provide anything compared to the ordinary cylinder.

I tend to agree, unless we want to show that we can do quasi 2d? (but this might be show cased by other examples, like the step example)

njansson:
I guess we could reuse the Gmsh file from `examples/cylinder` to generated a one element thick mesh

PR #1994, opened 2025-07-03
Timfelle/issue1924

timfelle:
Fixes ExtremeFLOW/neko#1924


timofeymukha:
Good, if we get a new mesh we can restore the example later.

PR #1996, opened 2025-07-04
Merge old cylinder test into the pytest suite

timofeymukha:
What I actually test

- All residuals, CFL, etc. for all time steps in the first part of the run.
- Same for the 2nd part of the run that restarts from a checkpoint.
- Same between the two runs for the time-values where they overlap.

I use relative tolerance 1e-15 for DP and 1e-6 for SP. Reasonable? Looking at the logs even for SP the values seem to be exactly the same on my machine.

RP handling and log parser invocation added to the test library for convenience.


.github/workflows/check_gnu.yml
```diff
@@ -217,30 +217,15 @@ jobs:
...
-           cd ../..
-
       - name: Integration tests
         if: matrix.backend == 'cpu'
         run: |
            source myenv/bin/activate
            chmod +x $MAKENEKO_EXEC
            cd tests/integration
-           pytest --max_nprocs=2
+           pytest --max_nprocs=2 --real-precision=${RP}
```
timofeymukha, line 226:
```suggestion
           pytest --max_nprocs=2 --real_precision=${RP}
```

timofeymukha:
Alright, DP finally passes at least. For SP, it seems the diff we get in pressure start residuals is consistenly larger than 1e-6, and differs somewhat between the different runs in the CI. @njansson @timfelle Should one simply lower the expected tolerance? I was naively expected that the tolerance should roughly satisfy the epsilon of the precision - 1 digit. But it seem the numbers we get are larger and compiler-dependent.

njansson (approved):
🐈

timfelle:
> Alright, DP finally passes at least. For SP, it seems the diff we get in pressure start residuals is consistenly larger than 1e-6, and differs somewhat between the different runs in the CI. @njansson @timfelle Should one simply lower the expected tolerance? I was naively expected that the tolerance should roughly satisfy the epsilon of the precision - 1 digit. But it seem the numbers we get are larger and compiler-dependent.

Right, what I find interesting is that the error states:

`Column 'pressure_start_residual' does not match reference data.`

Which I would not say you should compare at all. I do not think you can expect the residuals will match before you start solving across compilers, the goal is that the converged solution does match.

timofeymukha:
Alrighty, we can merge this in now I think.

timfelle (approved):

PR #2000, opened 2025-07-06
Disable rho output for pnpn (added in #1981)

njansson:
Since we have a `rho` field in pnpn, the current check is a bit misleading. However, @timofeymukha, with variable material properties, maybe one wants to dump rho from pnpn as well?


Shiyu-Sandy-Du (approved):

Shiyu-Sandy-Du:
The rho field in pnpn is a constant field, so I don't feel like we should output the rho field.

timfelle:
> The rho field in pnpn is a constant field, so I don't feel like we should output the rho field.

I agree, the fields are available to be saved by the field writer simcomp, if the user wish to look at them. It might be worth upgrading the field writer with a "set" or something to allow the user to specify that "materials" should be saved.

timfelle (approved):

PR #2001, opened 2025-07-08
Add the possibility to compile .cu and .hip files with makeneko

njansson:
This allows `makeneko` to compile user provided device files (`.cu/.hip`), to be called from the user provided `.f90` file. Callable routines in the device files must allow for C linkage, hence be inside a `extern "C"` block.


njansson:
I'll add some docs in a couple of minutes

njansson:
Would be interesting if one could easily inject for example a custom `ax_t` object

PR #2003, opened 2025-07-08
Add possibility to not set default values

timofeymukha:
Inspired by previous discussion with @pschlatt1 on Zulip. Adds a `case.no_defaults` parameter to the case file. If true, json_get_or_default will act like json_get and throw an error if the requested parameter is not there. So, one can setup the case in full explicit control mode, which perhaps @pschlatt1 will like. It is also pretty useful in debugging, I think.

As a tangent, not using defaults is really difficult to do at the moment. For example there is stuff like

```fortran
    call json_get_or_default(params, 'variable_timestep', &
         this%is_variable_dt, .false.)
    call json_get_or_default(params, 'target_cfl', &
         this%cfl_trg, 0.4_rp)
    call json_get_or_default(params, 'max_timestep', &
         this%max_dt, huge(0.0_rp))
    call json_get_or_default(params, 'min_timestep', &
         this%min_dt, 0.0_rp)
    call json_get_or_default(params, 'max_update_frequency',&
         this%max_update_frequency, 0)
    call json_get_or_default(params, 'min_update_frequency',&
         this%min_update_frequency, huge(0))
    call json_get_or_default(params, 'cfl_running_avg_coeff', &
         this%alpha, 0.5_rp)
    call json_get_or_default(params, 'max_dt_increase_factor', &
         this%max_dt_increase_factor, 1.2_rp)
    call json_get_or_default(params, 'min_dt_decrease_factor', &
         this%min_dt_decrease_factor, 0.5_rp)
    call json_get_or_default(params, 'cfl_deviation_tolerance', &
         this%dev_tol, 0.2_rp)
```

So, I think we should aim to improve some logic in the code and just not read parameters when they are definitely not needed. With the no_defaults parameter we can test this in a good way.

### DIFF
--- a/src/common/json_utils.f90
+++ b/src/common/json_utils.f90
@@ -445,7 +445,7 @@ contains
     type(json_file), intent(inout) :: object
 
     type(json_value), pointer :: ptr
-    type(json_core)  :: core
+    type(json_core) :: core
     logical :: found
     character(len=:), allocatable :: buffer
 


timfelle (approved):
I like it 👍

PR #2014, opened 2025-07-10
Use POSIX-compliant test command in backend compilation script

timfelle:
Replace the `[[` test command with `[` for better compatibility across different shell environments.


njansson (approved):
Thanks!

PR #2019, opened 2025-07-11
Add missing memset in field init, make it consistent with host

njansson:


njansson:
> How does the cfill command in the `vector.f90` differ from this, and should we change it there and for `matrix_t`as well.

Cfill will also call memset via some extra steps in device math, so no difference. The issue here was that we had places were we simply initialised fields, and never cleared them

njansson:
> > > How does the cfill command in the `vector.f90` differ from this, and should we change it there and for `matrix_t`as well.
> > 
> > 
> > Cfill will also call memset via some extra steps in device math, so no difference. The issue here was that we had places were we simply initialised fields, and never cleared them
> 
> Is it worth aligning them to eachother for consistency still?

Not sure it is worth it. But I always find it more consistent if you have a section of "memory mgmt" code, like alloc to have a memset, than instead having to call device_rzero, likewise in more mathy code sections, throwing in a memset might confuse

timfelle (approved):

src/field/field.f90
```diff
@@ -110,9 +110,9 @@ end subroutine field_init_internal_dof
 
   !> Initialize a field @a this on the mesh @a msh using an internal dofmap
   subroutine field_init_external_dof(this, dof, fld_name)
-    class(field_t), intent(inout) :: this      !< Field to be initialized
-    type(dofmap_t), target, intent(in) :: dof  !< External dofmap for the field
-    character(len=*), optional :: fld_name     !< Name of the field
+    class(field_t), intent(inout) :: this !< Field to be initialized
```
timfelle, line 113: An alternative could be to add the parameter description in the doxygen block

```Fortran
!> Initialize a field @a this on the mesh @a msh using an internal dofmap
!! @param this Field to be initialized
```

PR #2022, opened 2025-07-22
Add missing header to distribution files

timfelle:
Include the `device/device_config.h` header in the distribution files to ensure it is available for builds.

### DIFF
--- a/.github/actions/setup_json-fortran/action.yml
+++ b/.github/actions/setup_json-fortran/action.yml
@@ -34,7 +34,7 @@ inputs:
   version:
     description: "The version of the JSON-Fortran library to use."
     required: false
-    default: "8.3.0"
+    default: "8.4.0"
   arch:
     description: "The architecture to use."
     required: false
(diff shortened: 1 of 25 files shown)


njansson:
What is the issue, doesn't the file gets included in a make install or not in a make dist?

Since the file is generated by autotools it should be included in both

timfelle:
> What is the issue, doesn't the file gets included in a make install or not in a make dist?
> 
> Since the file is generated by autotools it should be included in both

Well I guess it is a question of convention for our API. Where do we intend a user to get the file? It is not under `#include <neko/device/device_config.h>`, but rather `#include <neko/device_config.h>`.

timfelle:
> > What is the issue, doesn't the file gets included in a make install or not in a make dist?
> > Since the file is generated by autotools it should be included in both
> 
> Well I guess it is a question of convention for our API. Where do we intend a user to get the file? It is not under `#include <neko/device/device_config.h>`, but rather `#include <neko/device_config.h>`.

An interesting not is that the PKGConfig exposes both the parent include directory, but also the internal neko directory. Leading to 2 interfaces both being valid `#include <device_config.h>` and `#include <neko/device_config.h>`

timfelle:
@njansson, so, do we want to mirror the folder structure from the source folder for our inclusion interface, or do we want to move the `device_config.h` header to the top level of our include folder upon install?

timofeymukha (approved):

Issue #2023, opened 2025-07-22
Supported precision


timfelle:
The documentation lists the following supported precision types:
- `sp` for single precision
- `dp` for double precision
- `qp` for quad precision

However, compiling with quad precision (`qp`) results in an error:
```
common/time_state.f90:74:70:

   74 |     call json_get_or_default(params, 'start_time', start_time, 0.0_rp)
      |                                                                      1
Error: There is no specific subroutine for the generic 'json_get_or_default' at (1)
common/time_state.f90:75:47:

   75 |     call json_get(params, 'end_time', end_time)
      |                                               1
Error: There is no specific subroutine for the generic 'json_get' at (1)
common/time_state.f90:78:51:

   78 |        call json_get(params, 'timestep', time_step)
      |                                                   1
Error: There is no specific subroutine for the generic 'json_get' at (1)
```

timofeymukha:
Yeah, `qp` had not really been supported for a while now, I guess. Is there generally a motivation to have it at such a global level?

njansson:
Related to #2024, isn't it easer to add a quad version of `json_get_or_default` in `json_utils`?

njansson:
> > Related to [#2024](https://github.com/ExtremeFLOW/neko/pull/2024), isn't it easer to add a quad version of `json_get_or_default` in `json_utils`?
> 
> Sure, just sounded like it was not needed.

All issues with lacking support in compilers and various MPI implementations, there might be need for quad pretty soon

PR #2024, opened 2025-07-23
Remove unsupported precision from documentation

timfelle:
Update code to re-introduce `qp`

Closes: #2023

### DIFF
--- a/.github/workflows/check_gnu.yml
+++ b/.github/workflows/check_gnu.yml
@@ -28,7 +28,7 @@ jobs:
         os: [ubuntu-24.04, ubuntu-24.04-arm, macos-13, macos-15]
         compiler: [gfortran-12, gfortran-13, gfortran-14, gfortran-15]
         backend: [cpu, cuda, hip, opencl]
-        precision: [ssp, sp, dp]
+        precision: [sp, dp]
         exclude:
           - os: ubuntu-24.04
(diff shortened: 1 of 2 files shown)


timfelle:
> We should aslo remove the `qp` part in the help string of `--enable-real` in `configure.ac`

What about ssp?

njansson:
> > We should aslo remove the `qp` part in the help string of `--enable-real` in `configure.ac`
> 
> What about ssp?

`ssp` is actually true single precision, `sp` does compute in single, and reductions in double, don't ask me about the naming

timfelle:
> > > We should aslo remove the `qp` part in the help string of `--enable-real` in `configure.ac`
> > 
> > 
> > What about ssp?
> 
> `ssp` is actually true single precision, `sp` does compute in single, and reductions in double, don't ask me about the naming

and it is not in the documentation, should it be?

njansson (approved):

timofeymukha:
> @timofeymukha Any chance you can look a bit at the ssp integration test results? I am not certain how to best update them

Yes. I guess it boils down to adding ssp as an option alongisde dp and sp, and setting the tolerance for the residual comparison in such a way that it passes. I wasn't even aware of ssp :).

tuananhdao (approved):

PR #2021, opened 2025-07-20
New user file interface

timofeymukha:
Generally follows the discussion here on GH. Still work in progress.

- Adds a preprocess routine, ran at the beginning of the timestep.
- Kills all pointwise interfaces.
- Merges source term and user condition interfaces of the fluid and scalar (passes scheme name and a list of fields)
- Generally passes time_state_t instead of t and tstep. This spills over to some other types. 

One can see the main changes by just looking at user_intf.f90

Additionally, merges the user source term types into one, instead of having one for fluid and scalar. Now possible since they use the same interface for the user routine. Harmonizes all the "variable_name", "scalar_name", "field_name", etc into "scheme_name", i.e. the `%name` of the scheme that calls the routine or owns the object (e.g. a source term).

PR #2025, opened 2025-07-23
Pass time_state_t to boundary condition apply routines

timofeymukha:
In line with the same changes in other routines.

I mainly did this now for the sake of updating the user Dirichlet routine to a similar pattern as in the #2021, which is done here as well.

Killed quite a bit if(present) stuff, which is nice


src/bc/shear_stress.f90
```diff
@@ -120,22 +119,21 @@ subroutine shear_stress_apply_vector(this, x, y, z, n, t, tstep, strong)
...
     if (strong_) then
-       call this%symmetry%apply_vector(x, y, z, n, t, tstep, .true.)
+       call this%symmetry%apply_vector(x, y, z, n, time, .true.)
     else
-       call this%neumann_x%apply_scalar(x, n, t, tstep, .false.)
-       call this%neumann_y%apply_scalar(y, n, t, tstep, .false.)
-       call this%neumann_z%apply_scalar(z, n, t, tstep, .false.)
+       call this%neumann_x%apply_scalar(x, n, time, .false.)
+       call this%neumann_y%apply_scalar(y, n, time, .false.)
+       call this%neumann_z%apply_scalar(z, n, time, .false.)
```
njansson, line 126: Needs a `present` check here for time

src/bc/shear_stress.f90
```diff
@@ -165,11 +162,11 @@ subroutine shear_stress_apply_vector_dev(this, x_d, y_d, z_d, t, tstep, &
...
        call this%symmetry%apply_vector_dev(x_d, y_d, z_d, &
-            t, tstep, .true., strm)
+            time, .true., strm)
     else
-       call this%neumann_x%apply_scalar_dev(x_d, t, tstep, .false., strm)
-       call this%neumann_y%apply_scalar_dev(y_d, t, tstep, .false., strm)
-       call this%neumann_z%apply_scalar_dev(z_d, t, tstep, .false., strm)
+       call this%neumann_x%apply_scalar_dev(x_d, time, .false., strm)
+       call this%neumann_y%apply_scalar_dev(y_d, time, .false., strm)
+       call this%neumann_z%apply_scalar_dev(z_d, time, .false., strm)
```
njansson, line 169: Missing `present` check for `time`

njansson (commented):
It for sure cleans up all the `present` checks in the bc modules. However, isn't this tying boundary condition implementations too much to our current solvers, `time_state_t` contains a lot of pnpn specific things. 

I would argue that at the apply level there shouldn't be any objects/structs etc, except pure arrays or variables

src/bc/bc_list.f90
timfelle, line 1: We do not need the present cascades here.
Since all we do is pass the parameters on to the next level, where they are also optional.

`bc_list_apply_scalar_array` can be reduced to this:
```.f90
    x_d = device_get_ptr(x)
    do i = 1, this%size_
        call this%items(i)%ptr%apply_scalar_dev(x_d, time = time, &
            strong = strong, strm = strm)
    end do
else
    do i = 1, this%size_
        call this%items(i)%ptr%apply_scalar(x, n, time = time, &
            strong = strong)
    end do
end if
```

src/bc/shear_stress.f90
timofeymukha: I think here we are fine due to passing an optional to an optional. Right @timfelle ?

src/bc/wall_model_bc.f90
```diff
@@ -108,7 +107,7 @@ subroutine wall_model_bc_apply_vector(this, x, y, z, n, t, tstep, strong)
 
     if (.not. strong_) then
        ! Compute the wall stress using the wall model.
-       call this%wall_model%compute(t, tstep)
+       call this%wall_model%compute(time%t, time%tstep)
```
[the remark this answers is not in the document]
timofeymukha: Here things are more tricky since the wall model interface does require t and tstep (not optional), so I guess we have to through an error if not present. Should also be converted to time_state_t in the wall model later.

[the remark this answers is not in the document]
timfelle: You might still need to pack it in a present statement and then have the else throw the error.

njansson:
> > It for sure cleans up all the `present` checks in the bc modules. However, isn't this tying boundary condition implementations too much to our current solvers, `time_state_t` contains a lot of pnpn specific things.
> > I would argue that at the apply level there shouldn't be any objects/structs etc, except pure arrays or variables
> 
> Hmm, I don't think so, what is pnpn-specific?
> 
> ```fortran
>      real(kind=rp), dimension(10) :: tlag = 0.0_rp !< Old times
>      real(kind=rp), dimension(10) :: dtlag = 0.0_rp !< Old dts
>      real(kind=rp) :: t = 0.0_rp !< Current time
>      real(kind=rp) :: dt = 0.0_rp !< Current dt
>      real(kind=rp) :: start_time = 0.0_rp !< Start time
>      real(kind=rp) :: end_time = 0.0_rp !< End time
>      integer :: tstep = 0 !< Current timestep
> ```
> 
> I think any time-stepper can make use of this, or at least provide the contents. I think passing a structure like this should be fine, it only contains innocent built-in types after all, which can be extracted when you pass stuff to a kernel.

True, it is only `tlag` and `dtlag` that is a bit pnpn specific. But still, feels a bit strange to pass down a ~20x larger structure instead of one float and one integer. Do we foresee that boundary conditions might need more info than simply `t` or `tstep`?

timofeymukha:
> > > It for sure cleans up all the `present` checks in the bc modules. However, isn't this tying boundary condition implementations too much to our current solvers, `time_state_t` contains a lot of pnpn specific things.
> > > I would argue that at the apply level there shouldn't be any objects/structs etc, except pure arrays or variables
> > 
> > 
> > Hmm, I don't think so, what is pnpn-specific?
> > ```fortran
> >      real(kind=rp), dimension(10) :: tlag = 0.0_rp !< Old times
> >      real(kind=rp), dimension(10) :: dtlag = 0.0_rp !< Old dts
> >      real(kind=rp) :: t = 0.0_rp !< Current time
> >      real(kind=rp) :: dt = 0.0_rp !< Current dt
> >      real(kind=rp) :: start_time = 0.0_rp !< Start time
> >      real(kind=rp) :: end_time = 0.0_rp !< End time
> >      integer :: tstep = 0 !< Current timestep
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think any time-stepper can make use of this, or at least provide the contents. I think passing a structure like this should be fine, it only contains innocent built-in types after all, which can be extracted when you pass stuff to a kernel.
> 
> True, it is only `tlag` and `dtlag` that is a bit pnpn specific. But still, feels a bit strange to pass down a ~20x larger structure instead of one float and one integer. Do we foresee that boundary conditions might need more info than simply `t` or `tstep`?

Since this was mostly done for the sake of the user bc routine, it is a bit difficult to say what people will come up with :-). But overall, it will probably be rare. I was mostly riding on the idea that we now pass time_state_t, whenever we pass time information. This is why I also pass it to all the other user routines in #2021 It feels kind of nice to have that standardized, and to always have just one dummy argument. But yeah, most of the time you definitely pass more stuff than needed in practice.

timofeymukha:
> Sounds good, I'm just being annoying, and nitpick to avoid us passing down complex structures to lower levels of Neko

This is for sure good to guard for. A lot of complexity with bcs right now comes from the fact that they double as masks for arbitrary dirichlet boudnaries. The masks are rather specific things, and act in low-level routines like KSP. So I think this is why we have the `time` and `strong` as optional, and also why the interface is split at `bc_t` level into one for CPU and one for device. Ordinarily, we would just have one interface and then inside one would dispatch into kernels in `bcknd`, and then it would be no issues having something a bit "fat" as a dummy argument. 

So I think going forward, I would like to try to decouple the boundary conditions and the masking operations, particularly also in light of the fact that for curved boundaries we will need a new type of mask. Then, the masks can have a clean low-level interface, and the bcs a fatter one, and without the `optional`.

src/bc/shear_stress.f90
timofeymukha: What I realize is that we can simply not pass time to these apply routines, because it is not used there anyway. Then, no need for a check and things should be safe.

vbaconnet:
I think that looks great, and if it doesn't affect performance in any way it's nice to have everything passed down to the user. As you say Timofey you never know what someone is going to need.

examples/programming/user_file_template.f90
```diff
@@ -117,13 +117,11 @@ subroutine scalar_user_bc(scalar_name, s, x, y, z, nx, ny, nz, ix, iy, iz, ie, t
...
-       tstep)
-    type(field_list_t), intent(inout) :: dirichlet_field_list
-    type(field_dirichlet_t), intent(in) :: dirichlet_bc
+  subroutine user_dirichlet_update(fields, bc, coef, time)
+    type(field_list_t), intent(inout) :: fields
+    type(field_dirichlet_t), intent(in) :: bc
     type(coef_t), intent(inout) :: coef
-    real(kind=rp), intent(in) :: t
-    integer, intent(in) :: tstep
+    type(time_state_t), intent(in) :: time
```
vbaconnet, line 124: Should we add some comments/doc here in case we refer new users to the user file template? Or is there already a description of `time_state_t` and its attributes in the documentation?

vbaconnet: I can already see someone getting confused with `time` thinking it is the simulation time

timofeymukha: Yeah, for sure. The whole doc for the user file in general will need quite a bit of update. My plan is to consolidate everything in #2021 after we merge this in!

src/bc/shear_stress.f90
njansson: Great, but if only the boolean is passed, one still need to add `strong = .false.` since there it's not the first `optional` argument.

njansson:
Forgot to ask before, but does the poisson demo still work with this new interface?

timofeymukha:
> Forgot to ask before, but does the poisson demo still work with this new interface?

It appears so, I could run the example in #2021, which merges this in.

I will add an integration test that runs that example as a separate PR, so we don't have to worry in the future.

njansson (approved):

src/bc/bc.f90
```diff
@@ -329,23 +320,14 @@ subroutine bc_apply_vector_generic(this, x, y, z, n, t, tstep, strm)
...
-               t = t, tstep = tstep, strm = strm_)
-       else if (present(t)) then
-          call this%apply_vector_dev(x_d, y_d, z_d, t = t, strm = strm_)
-       else if (present(tstep)) then
-          call this%apply_vector_dev(x_d, y_d, z_d, tstep = tstep, strm = strm_)
+       if (present(time)) then
+          call this%apply_vector_dev(x_d, y_d, z_d, time = time, strm = strm_)
        else
           call this%apply_vector_dev(x_d, y_d, z_d, strm = strm_)
        end if
```
timfelle, line 327: No need to do the present detection here. You only pass it on.
```suggestion
       call this%apply_vector_dev(x_d, y_d, z_d, time = time, strm = strm_)
```

src/bc/bc.f90
```diff
@@ -329,23 +320,14 @@ subroutine bc_apply_vector_generic(this, x, y, z, n, t, tstep, strm)
...
-          call this%apply_vector(x, y, z, n, t = t, tstep = tstep)
-       else if (present(t)) then
-          call this%apply_vector(x, y, z, n, t = t)
-       else if (present(tstep)) then
-          call this%apply_vector(x, y, z, n, tstep = tstep)
+       if (present(time)) then
+          call this%apply_vector(x, y, z, n, time = time)
        else
           call this%apply_vector(x, y, z, n)
        end if
```
timfelle, line 333:
```suggestion
       call this%apply_vector(x, y, z, n, time = time)
```

src/bc/bc.f90
```diff
@@ -383,22 +363,14 @@ subroutine bc_apply_scalar_generic(this, x, n, t, tstep, strm)
...
-          call this%apply_scalar_dev(x_d, t = t, tstep = tstep, strm = strm_)
-       else if (present(t)) then
-          call this%apply_scalar_dev(x_d, t = t, strm = strm_)
-       else if (present(tstep)) then
-          call this%apply_scalar_dev(x_d, tstep = tstep, strm = strm_)
+       if (present(time)) then
+          call this%apply_scalar_dev(x_d, time = time, strm = strm_)
        else
           call this%apply_scalar_dev(x_d, strm = strm_)
        end if
```
timfelle, line 370:
```suggestion
       call this%apply_scalar_dev(x_d, time = time, strm = strm_)
```

src/bc/bc.f90
```diff
@@ -383,22 +363,14 @@ subroutine bc_apply_scalar_generic(this, x, n, t, tstep, strm)
...
-          call this%apply_scalar(x, n, t = t, tstep = tstep)
-       else if (present(t)) then
-          call this%apply_scalar(x, n, t = t)
-       else if (present(tstep)) then
-          call this%apply_scalar(x, n, tstep = tstep)
+       if (present(time)) then
+          call this%apply_scalar(x, n, time = time)
        else
           call this%apply_scalar(x, n)
        end if
```
timfelle, line 376:
```suggestion
       call this%apply_scalar(x, n, time = time)
```

timfelle (commented):
Could we do away with the implicit device interfacing soon?
Right now we define a lot of code to run on a cpu array, but internally the cpu array is never touched since we run a `get_device_ptr` and just call it on the device instead.

At this point we have fields, vectors and matrices that support explicit handling, so I think it is about time to follow through with that syntax.

src/bc/bc.f90
njansson: Time can still be optional so this is not safe , e.g. The poisson example

timfelle: Exactly why it is safe. Time is optional for apply_vector so the present check here is redundant as it is also in apply_vector

src/bc/bc.f90
njansson: Wait but you can't pass it down unless it is present. that is what caused issues in January

vbaconnet (approved):
LGTM

PR #2036, opened 2025-08-03
Adding Device Implementation of OIFS

Kamyarmvn:
Hi, 
In this PR I have added the device implementation of the OIFS scheme.


njansson:
> > > Hmm, that nvhpc issue is quite strange...I would guess it's something related to submodules, which nvhpc might not support. Any ideas @timfelle @timofeymukha
> > 
> > 
> > It looks exactly like the one in #2021 but for a different oprator
> 
> This issue, is "solved" if one applies the patch we have in `develop`
> 
> https://github.com/ExtremeFLOW/neko/blob/develop/patches/nvhpc_opgrad.patch
> 
> I tried, and it works, but it for sure is strange

However, this issue is not seen anymore on `develop`, so before adding that patch, I would suggest to merge this branch with develop. That will hopefully fix things

src/fluid/bcknd/advection/adv_oifs.f90
```diff
@@ -39,13 +39,16 @@ module adv_oifs
   use coefs, only: coef_t
   use math, only: copy, rzero
   use operators, only: runge_kutta, set_convect_rst
-  use neko_config, only: NEKO_BCKND_SX, NEKO_BCKND_XSMM
+  use neko_config, only: NEKO_BCKND_DEVICE, NEKO_BCKND_SX, NEKO_BCKND_XSMM, &
+    NEKO_BCKND_OPENCL, NEKO_BCKND_CUDA, NEKO_BCKND_HIP
```
njansson, line 43:
```suggestion
  use neko_config, only: NEKO_BCKND_DEVICE, NEKO_BCKND_SX, NEKO_BCKND_XSMM
```
Isn't it enough adding just the DEVICE ? (CUDA/HIP/OpenCL isn't checked specifically or?)

njansson (approved):
Nice!

src/math/operators.f90
```diff
@@ -619,53 +631,116 @@ subroutine runge_kutta(phi, c_r1, c_r23, c_r4, Xh_GLL, Xh_GL, coef, &
...
-         coef_GL, GLL_to_GL)
-    call col2(r4, coef%B, n)
-
-    c1 = -dtau/6.
-    c2 = -dtau/3.
-    do i = 1, n
-       phi(i) = phi(i) + c1 * (r1(i) + r4(i)) + c2 * (r2(i) + r3(i))
-    end do
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       phi_d = device_get_ptr(phi)
```
timfelle, line 652: Phi was only a clean CPU Array because this was a cpu only implementation.
Please change the type of phi if you need a device array.

src/fluid/bcknd/advection/adv_oifs.f90
```diff
@@ -291,86 +390,152 @@ subroutine adv_oifs_compute(this, vx, vy, vz, fx, fy, fz, Xh, coef, n, dt)
...
+               call dtime%interpolate_scalar(th, cr_k23, convr_GL, ctlag, n_GL)
+               call dtime%interpolate_scalar(th, cs_k23, convs_GL, ctlag, n_GL)
+               call dtime%interpolate_scalar(th, ct_k23, convt_GL, ctlag, n_GL)
+               call dtime%interpolate_scalar(tau1, cr_k4, convr_GL, ctlag, &
+                                             n_GL)
+               call dtime%interpolate_scalar(tau1, cs_k4, convs_GL, ctlag, &
+                                             n_GL)
+               call dtime%interpolate_scalar(tau1, ct_k4, convt_GL, ctlag, &
+                                             n_GL)
+               call runge_kutta(fx%x, conv_k1, conv_k23, conv_k4, Xh, Xh_GL, &
```
timfelle, line 460: Please do not touch host arrays with device code like this. Pass `fx%x_d` or the
entire `fx` instead

timfelle (changes requested):

src/fluid/bcknd/advection/adv_oifs.f90
Kamyarmvn: I fixed the issue.

src/math/bcknd/device/opr_device.F90
```diff
@@ -550,6 +630,57 @@ subroutine opr_device_conv1(du, u, vx, vy, vz, Xh, coef, nelv, gdim)
...
+    real(kind=rp) :: ud(Xh_GL%lx*Xh_GL%lx*Xh_GL%lx)
+    type(c_ptr) :: du_d, ud_d, u_d, cr_d, cs_d, ct_d, dx_d, dy_d, dz_d
+    integer :: n_GLL
+
+    n_GLL = coef_GL%msh%nelv * Xh_GLL%lxyz
+
+    du_d = device_get_ptr(du)
+    ud_d = device_get_ptr(ud)
+    u_d = device_get_ptr(u)
+
```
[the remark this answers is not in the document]
Kamyarmvn: I tried to follow the same pattern as other subroutines. I tried to fix the issue in this iteration by removing all the unnecessary device operations.

src/math/operators.f90
Kamyarmvn: Fixed!

njansson:
Interesting, suddenly nvfortran doesn't thrown an ICE anymore...

PR #2041, opened 2025-08-06
Feature/update neko api

njansson:
Update and extend API with the possibility to provide user-provided callbacks connected to the new user interface.

This also adds a Julia example, which requires Neko.jl, (unfortunetly) not distributed with Neko (T_T).


timofeymukha (approved):
Looks like great progress!

PR #2042, opened 2025-08-06
Significant improvements to out module use

timfelle:
MPI introduced most places where it was missing.
Reduced explict imports of MPI through the comm module.

This is not quite a replacement of #1706 But it is a step on the way there


njansson:
Great stuff. However, with a nonprivate `comm` and mpi included in other files, this will cause LOADS of warnings on LUMI. :(

timfelle:
> I don't think we can avoid to make comm private

I know. This is a step along the way

timofeymukha (approved):
Amazing!

njansson (approved):

njansson:
Since nvhpc is sad, I'll merge this manually

PR #2047, opened 2025-08-07
Reduce reliance on implicit device pointers

timfelle:
Unified the interfaces for BC's a bit and started work which should make it easier to phase out the get_device_ptr from BC system.


timofeymukha (approved):
lgtm

njansson (approved):

PR #2049, opened 2025-08-07
Cleanup module use

njansson:
More cleanup of `use` statements, including changing legacy types to tbp, and making `comm` private (rendering #1706 redundant)


.github/workflows/develop.yml
```diff
@@ -152,7 +152,7 @@ jobs:
       - depend
       - GNU
       - Intel
-      - Nvidia
+#      - Nvidia
```
timfelle, line 155: We should still wait for it to complete, we should just not use it to determine the return value of the final check
```suggestion
      - Nvidia
```

PR #2054, opened 2025-08-08
Function returns Vector/matrix

timfelle:
Updates to our vector and matrix types.
Should now reduce temporary arrays.

It was not really used internally, but still good to get updated.


njansson:
Nice! My questions would be if any users are using this already?

PR #2034, opened 2025-08-01
Add multi-scalar support to checkpointing system

tuananhdao:
## Background

Multi-scalar support was added in PR #1795 but checkpointing only works for the first scalar.

## Add multi-scalar support to checkpointing system

Example:

```json
{
  "version": 1.0,
  "case": {
    "mesh_file": "box.nmsh",
    # "restart_file": "fluid00120.chkp",
    "output_at_end": false,
    "output_boundary": false,
    "output_checkpoints": true,
    "time": {
        "end_time": 0.5,
        "timestep": 0.001
    },
    "checkpoint_control": "nsamples",
    "checkpoint_value": 500,
    "numerics": {
      "time_order": 2,
      "polynomial_order": 5,
      "dealias": true
    },
    "fluid": {
      "scheme": "pnpn",
      "mu": 1,
      "rho": 1,
      "freeze": true,
      "initial_condition": {
        "type": "user"
      },
      "velocity_solver": {
        "type": "cg",
        "preconditioner": {
            "type": "jacobi"
        },
        "projection_space_size": 0,
        "absolute_tolerance": 1e-07,
        "max_iterations": 800
      },
      "pressure_solver": {
        "type": "gmres",
        "preconditioner": {
            "type": "hsmg"
        },
        "projection_space_size": 4,
        "absolute_tolerance": 1e-07,
        "max_iterations": 800
      },
      "output_control": "nsamples",
      "output_value": 100,
      "output_scalars": true
    },
    "scalars": [
      {
        "name": "s1",
        "lambda": 1e-16,
        "cp": 1.0,
        "solver": {
          "type": "cg",
          "preconditioner": {
            "type": "jacobi"
          },
          "projection_space_size": 0,
          "absolute_tolerance": 1e-07,
          "max_iterations": 800
        },
        "initial_condition": {
          "type": "user"
        },
        "source_terms": [
          {
            "type": "user_vector"
          },
          {
            "type": "user_pointwise"
          },
          {
            "type": "constant",
            "values": [
              "1.0"
            ]
          }
        ],
      },
      {
        "name": "s2",
        "lambda": 1e-16,
        "cp": 1.0,
        "solver": {
          "type": "cg",
          "preconditioner": {
            "type": "jacobi"
          },
          "projection_space_size": 0,
          "absolute_tolerance": 1e-07,
          "max_iterations": 800
        },
        "initial_condition": {
          "type": "user"
        }
      },
      {
        "name": "s3",
        "lambda": 1e-16,
[...]

PR #2055, opened 2025-08-10
Multi scalars checkpoint hdf5

tuananhdao:
Solve PR #2034

## Case file

When multi-scalars fields are used, users are required to use

```json
"case": {
    ...
    "output_checkpoints": true,
    "checkpoint_format": "hdf5",
    ...
```

Restart from a .h5 file

```json
"case": {
    ...
    "restart_file": "fluid00250.h5",
    ...
```

Checkpointing with `.chkp` format only restarts the first scalar (so single-scalar still works).


src/io/hdf5_file.F90
```diff
@@ -444,26 +446,52 @@ subroutine hdf5_file_determine_data(data, msh, dof, fp, fsp, dtlag, tlag)
...
+          ab_count = 0
+          do i = 1, scalar_count
+             associate(slag => data%scalar_lags%get(i))
+               write(scalar_name, '(A,A)') trim(slag%f%name), '_abx1'
+               if (neko_field_registry%field_exists(scalar_name)) ab_count = ab_count + 1
+               write(scalar_name, '(A,A)') trim(slag%f%name), '_abx2'
+               if (neko_field_registry%field_exists(scalar_name)) ab_count = ab_count + 1
+             end associate
+          end do
+          fp_size = fp_size + ab_count
```
njansson, line 470: Nothing against the field registry, but can't this be fetched from the checkpoint object, to be consistent with how we deal with fluids? As far as I can see, the data is already in checkpoint?

src/io/hdf5_file.F90
```diff
@@ -444,26 +446,52 @@ subroutine hdf5_file_determine_data(data, msh, dof, fp, fsp, dtlag, tlag)
...
-       if (associated(data%s)) then
+       if (allocated(data%scalar_lags%items) .and. data%scalar_lags%size() > 0) then
+          scalar_count = data%scalar_lags%size()
+       else if (associated(data%s)) then
+          scalar_count = 1
+       else
+          scalar_count = 0
+       end if
+
+       if (scalar_count > 1) then
```
njansson, line 457:
```suggestion
       if (scalar_count .gt. 1) then
```

src/io/hdf5_file.F90
```diff
@@ -444,26 +446,52 @@ subroutine hdf5_file_determine_data(data, msh, dof, fp, fsp, dtlag, tlag)
...
-
        allocate(fp(fp_size))
 
        fsp_size = 0
        if (associated(data%ulag)) then
           fsp_size = fsp_size + 3
        end if
 
-       if (associated(data%slag)) then
+       if (scalar_count > 1) then
```
njansson, line 490:
```suggestion
       if (scalar_count .gt. 1) then
```

src/io/hdf5_file.F90
```diff
@@ -481,9 +509,39 @@ subroutine hdf5_file_determine_data(data, msh, dof, fp, fsp, dtlag, tlag)
        fp(4)%ptr => data%p
 
        fp_cur = 5
-       if (associated(data%s)) then
+
+       if (scalar_count > 1) then
```
njansson, line 513:
```suggestion
       if (scalar_count .gt. 1) then
```

src/io/hdf5_file.F90
```diff
@@ -496,20 +554,22 @@ subroutine hdf5_file_determine_data(data, msh, dof, fp, fsp, dtlag, tlag)
...
        if (associated(data%ulag)) then
           fsp(fsp_cur)%ptr => data%ulag
           fsp(fsp_cur+1)%ptr => data%vlag
           fsp(fsp_cur+2)%ptr => data%wlag
           fsp_cur = fsp_cur + 3
        end if
 
-       if (associated(data%slag)) then
+
+       if (scalar_count > 1) then
```
njansson, line 565:
```suggestion
       if (scalar_count .gt. 1) then
```

src/field/field_series_list.f90
```diff
@@ -0,0 +1,125 @@
...
+    if (.not. allocated(this%items)) then
+       call neko_error('Field series list not initialized')
+    end if
+
+    if (this%n_items >= size(this%items)) then
+       call neko_error('Field series list is full')
+    end if
+
+    this%n_items = this%n_items + 1
+    this%items(this%n_items)%ptr => field_series
```
timofeymukha, line 98: You could do a move_alloc to support an expanding size.

Looking and field_list_t, the method is called append, and it seems our logic there is that you allocate an initial size, then you are expected to fill it using the assign_ routines, and then you can append to that one by one.

```fortran
  subroutine field_list_append(this, f)
    class(field_list_t), intent(inout) :: this
    class(field_t), intent(in), target :: f
    type(field_ptr_t), allocatable :: tmp(:)
    integer :: len

    len = size(this%items)

    allocate(tmp(len+1))
    tmp(1:len) = this%items
    call move_alloc(tmp, this%items)
    this%items(len+1)%ptr => f

```

I am actually not entirely sure that is optimal either... but it would be good if all _list types behaved similar. I think the field_registry actually does it best, it both keeps tab on the number of fields (which does not have to match the allocated size of the items array), and expands when needed. And now that I write this, I also recall that there was some issue with the scratch_registry regarding the expansion...

I feel like maybe what you have here is OK, and then we should harmonize that to some degree across all these types.

src/io/hdf5_file.F90
```diff
@@ -444,26 +446,52 @@ subroutine hdf5_file_determine_data(data, msh, dof, fp, fsp, dtlag, tlag)
...
+       else if (associated(data%s)) then
+          scalar_count = 1
+       else
+          scalar_count = 0
+       end if
+
+       if (scalar_count > 1) then
+          fp_size = fp_size + scalar_count
+
+          ! Add Adams-Bashforth fields for each scalar if they exist
```
timofeymukha, line 460: It is not really Adams-Bashforth despite the confusing `ab` names.  That is some Nek5000 legacy in the naming, similar to `sumab`, which actually uses EXT extrapolation schemes... So better not to use  Adams-Bashforth in the comment, because it will breed confusion.

src/io/hdf5_file.F90
```diff
@@ -481,9 +500,35 @@ subroutine hdf5_file_determine_data(data, msh, dof, fp, fsp, dtlag, tlag)
...
+               write(scalar_name, '(A,A)') trim(slag%f%name), '_abx1'
+               fp(fp_cur)%ptr => neko_field_registry%get_field(scalar_name)
+               fp_cur = fp_cur + 1
+               write(scalar_name, '(A,A)') trim(slag%f%name), '_abx2'
+               fp(fp_cur)%ptr => neko_field_registry%get_field(scalar_name)
+               fp_cur = fp_cur + 1
+             end associate
+          end do
+       else if (associated(data%s)) then
+          ! Single scalar support
```
njansson, line 523: Can't we point to the item in the `data%scalar_lags` instead of going through the field registry?

njansson (approved):

PR #2066, opened 2025-08-26
Workaround for compiling the API with NAG in single precision

### LINKED ISSUE #2063: Issues with the API part of Neko when using real as sp or ssp
Various issues with the API when Neko is compiled using single precision

njansson:
This fixes #2063


njansson:
This does not solve the issue, just allow for compiling the API code

PR #2068, opened 2025-08-27
Entropy viscosity for the compressible solver

tuananhdao:
## Changes

This PR implements "entropy viscosity" for the compressible module. There has been some literature which proves that the resulting solution is entropy-consistent. And ultimately, the solution converges to the entropy solution.

The main idea is
- To use the entropy residual as a distributor of the artificial viscosity
- No accuracy is lost when solution is smooth (entropy residual is essentially zero)
- Entropy residual tracks the shocks/discontinuities
- The method adds just enough viscosity to stabilize the solution, guided by local wave speeds
- The CFL number doesn't change, totally depending on the advective flux
- The parameters `c_avisc_low=0.5` and `c_avisc_entropy=1.0` should work for all cases, unless user wants to fine-tune them to make the solution sharper/smoother.

**Reference**: Guermond, J. L., Pasquetti, R., & Popov, B. (2011). Entropy viscosity method for nonlinear conservation laws. Journal of Computational Physics, 230(11), 4248-4267. [[Link](https://www.sciencedirect.com/science/article/pii/S0021999110006583)]

## Use

Insert `c_avisc_low` and `c_avisc_entropy` to the `numerics` block. If not set, `c_avisc_low=0.5` and `c_avisc_entropy=1.0` are applied by default.

```json
        "numerics": {
            "time_order": 3,
            "polynomial_order": 5,
+           "c_avisc_low": 0.5,
+           "c_avisc_entropy": 1.0
        }
```

## Supported backends

(ticked means initial tests done)

- [x] CPU
- [x] CUDA
- [x] HIP
- [x] OpenCL

## To-do list

- [x] Test strong scaling. Known scaling issue from Niclas: "too much MPI traffic per node that would eat up the memory". Will be fixed in another PR.
- [x] Add documentation about the compressible solver

## Simulation Results

<p align="center"><b>Before and after runs use the SAME MESH and POLYNOMIAL ORDER</b></p>
<p align="center">The settings below are chosen so that they run <5 min on a personal laptop using 4 threads</p>

| Before | After |
| ------ | ----- |
| Forward-facing step ~1K elements, polynomial_order=5 | Forward-facing step ~1K elements, polynomial_order=5 |
| LO | EV |
| Cylinder 6939 elements, polynomial_order=5 | Cylinder 6939 elements, polynomial_order=5 |
| cylinder_before | cylinder_after |
| TGV 10x10x10 mesh, polynomial_order=5, t=0.2 | TGV 10x10x10 mesh, polynomial_order=5, t=0.2 |
| tgv_before | tgv_after |


tuananhdao:
> Maybe it's not the best time to comment, but maybe we could later expand this into a general type of artificial viscosity? And since the time stepping is explicit. Could we further expand it into a general "regularisation" type? I am asking that because I am doing some regularisation stuffs similar so maybe we can later make them under the same umbrella.

I agree. It make total sense. I will incorporate that in this PR.

njansson:
To avoid the PR to be come too big, could we start to consider merging it into develop?

njansson (approved):
🍻

njansson:
A comment regarding MPI and memory. This issue is specific to Fugaku.

timfelle (approved):

PR #2075, opened 2025-09-01
add the missing link

tuananhdao:
## Problem

This is not currently overriding the csv file but appending

```fortran
call file%init("output.csv", overwrite=.true.)
```

## Changes

Fix the problem.


timfelle:
How will this affect a restarted run? We probably need to make sure that on a restart it still appends rather than overwriting, since information would otherwise be lost.

tuananhdao:
> How will this affect a restarted run? We probably need to make sure that on a restart it still appends rather than overwriting, since information would otherwise be lost.

It shouldn't affect anything else since `overwrite=.false.` by default.

timfelle (approved):

adperezm (approved):

PR #2077, opened 2025-09-03
Checkpoint file counter in reader

timfelle:
A few relatively deep changes, however, the impact should be quite limited.

The file checker of the generic file can now work on a specific file name. which made it useful for checking counter based files, without modifying the class name.

The checkpoint files can have a name specified with the suffix.

The checkpoint counter can be used when reading files.

One possible situation of unexpected behaviour which might arrise here.
Say you set the checkpoint file name to be "file" but do not overwrite files, and then specify to load from checkpoint "file.chkp", you will actually load "file00000.chkp" instead of throwing an error. So a user might be confused why that point is loaded and not the one with the highest number.


src/io/chkp_file.f90
```diff
@@ -384,8 +384,21 @@ subroutine chkp_file_read(this, data)
...
+       fname = trim(this%fname)
+    else
+       suffix_pos = filename_suffix_pos(this%fname)
+       write(id_str, '(i5.5)') this%counter
+       fname = trim(this%fname(1:suffix_pos-1)) // id_str // '.chkp'
+       inquire(file = fname, exist = file_exists)
+       if (.not. file_exists) fname = trim(this%fname)
+    end if
+
+    call this%check_exists(fname)
```
njansson, line 401: Should we be nice to the filesystem, and do as in `check_exists` , and only issue inquire on rank 0 (+ broadcast)?

src/io/chkp_file.f90
```diff
@@ -384,8 +384,26 @@ subroutine chkp_file_read(this, data)
...
+
+    if (file_exists) then
+       fname = trim(this%fname)
+    else
+       suffix_pos = filename_suffix_pos(this%fname)
+       write(id_str, '(i5.5)') this%counter
+       fname = trim(this%fname(1:suffix_pos-1)) // id_str // '.chkp'
+
+       if (pe_rank .eq. 0) inquire(file = this%fname, exist = file_exists)
+       call MPI_Bcast(file_exists, 1, MPI_LOGICAL, 0, NEKO_COMM, ierr)
```
timofeymukha, line 401: Should we have some log message for this branch?

njansson (approved):

njansson:
> > Should this also go into 1.0?
> 
> Maybe, it is a relatively small thing so I guess it would be fine. But not critical. It would be nice for our TopOpt thing to be able to just say we depend on 1.0, but I expect more changes will come later, so that is unrealistic.

Ok! I'll cherry-pick it into 1.0

PR #2079, opened 2025-09-05
Fix broken build with parmetis

timofeymukha:
Some imports were missing, I guess nobody checked in a while. Caught it by using our pixi install, that includes parmetis.


timofeymukha:
Oops, I missclicked and asked copilot to review. No idea what it will do)))

njansson (approved):

timfelle (approved):

PR #2080, opened 2025-09-05
File names

timfelle:
Updates to the filename structure.
Counters are now handled in the generic_file_t while still allowing filetypes to define their own patterns.

See fld files for an example of how the filenames are still custom to that file alone.


timfelle:
> Should we pick this into 1.0 as well? (I think so)

I think that would be good, if you believe it is solid enough. I am not sure I managed to test all file types, as I don't think all of them are represented in out examples.

PR #2094, opened 2025-09-16
Make region ID optional in profiler and improve runtime statistics

timfelle:
Allow the region ID to be optional for profiling, enhancing flexibility in usage. Update runtime statistics to include new control integers and improve accessibility of internal components. 

Clean up imports for better code organization.


src/common/runtime_statistics.f90
```diff
@@ -153,63 +163,65 @@ end subroutine runtime_stats_start_region
...
     call neko_log%section('Runtime statistics')
     call neko_log%newline()
-    write(log_buf, '(A,A,1x,A,1x,A)') '                  ',&
-         '      Total time  ','     Avg. time ','     Range  +/-'
+    write(fmt, '(A,I0,A)') '(', RT_STATS_MAX_NAME_LEN, 'x,1x,A15,2x,A15,2x,A15)'
+    write(log_buf, fmt) 'Total time', 'Avg. time', 'Range +/-'
     call neko_log%message(log_buf)
-    write(log_buf, '(A)') &
-         '--------------------------------------------------------------------'
+    write(log_buf, '(A)') repeat('-', RT_STATS_MAX_NAME_LEN + 50)
```
njansson, line 211: Are we still within our 80 cols ? I think so, even with the initial left sided indent

src/common/runtime_statistics.f90
```diff
@@ -265,4 +279,28 @@ subroutine runtime_stats_report(this)
 
   end subroutine runtime_stats_report
 
+  !> Find or allocate a region id for the named region @a name
+  subroutine runtime_stats_find_region_id(this, name, region_id)
```
njansson, line 283: Would it make sense to also include a lookup table for our reserved (region, names) ? I do have the names on a piece of paper, somewhere...

njansson (commented):
Looks good, my only concern was if we don't have enough checks (or lack of error checks) for profiler backends that requires both region id and name. But so far it looks like we should be good

src/common/runtime_statistics.f90
timfelle: Yes the max name length was increased to 25 characters, along with the 50 sits at 75 without the indentations, which as far as I can see is just a single one here.

timfelle: Exactly, just allowed longer names and made it simpler to modify later

PR #2097, opened 2025-09-17
Add caching feature for the IBM method using meshes

timfelle:
Implement a caching feature for the mesh IBM method and update documentation to reflect these changes.

Additionally,
- Reworked the point-wise max and min operations.
- Changed the destructor of file_t to a %free function for files.


njansson (approved):

timofeymukha (approved):

timfelle:
> Should this be picked for 1.0?

Sure. I think it is a really good thing with how freaking slow my SDF computation actually is 😅

PR #2106, opened 2025-09-24
Update of fluid output file documentation; fixes #1982

### LINKED ISSUE #1982: Documentation of our fluid output files.
We currently have no documentation for the fluid ouput files. 
We should add a section to the `User guide/Input-output` documenting which
fields the file can contain and what they mean.

Could possibly be complimented by adding the list of fields to the log. That way
the user can see in the log, the name of the fields added to the files. Which is
mostly a help for large numbers of scalar fields, as these are just called `si`
when loaded in Paraview.

adampep:
Corrected output file description and added link to Nek5000 documentation

### DIFF
--- a/doc/pages/user-guide/io.md
+++ b/doc/pages/user-guide/io.md
@@ -31,15 +31,15 @@ periodic boundaries are directly encoded into the mesh file, and this will
 remain so in the future.
 
 ## Three-dimensional field output
-Neko stores the 3D fields with results in the `.fld` format, which is the same
+Neko stores the 3D fields with results in the `.f#####` format, which is the same
 as in Nek5000. The advantage of adopting this format, is that there is a reader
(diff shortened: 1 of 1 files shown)


adampep:
added

timofeymukha:
> .fld is an old nek5000 format. nek5000 could output both and visit could read both. As far as I understand neko supports the newer binary format .f only. Is this correct? It is true .fld shows in number of places in documentation. I didn't notice this, but I could change it to be consistent.

I see, I was not aware of the old format at all :-). I guess we need to decide. The issue is that f#### is not easy to pronounce, and fld is. I think we just have to decide what we mean by what and put into the docs.

adampep:
That would be good solution. I've realised, however, that in some prat of the code the file name with extension .fld is actually created. We can clarify this at the next meeting

PR #2107, opened 2025-09-24
Add copy to device in dirichlet_conditions

timofeymukha:
Adds sync to device for user Dirichlet conditions in the examples.

### DIFF
--- a/examples/cyl_boundary_layer/cyl_bl.f90
+++ b/examples/cyl_boundary_layer/cyl_bl.f90
@@ -146,6 +146,15 @@ contains
           w%x(msk_ind,1,1,1) = sin(th)*u_rho + cos(th)*u_th
        end if
     end do
+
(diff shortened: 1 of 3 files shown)


njansson (commented):
Isn't this already done by the user dirichlet apply?

timofeymukha:
> Isn't this already done by the user dirichlet apply?

No. There is only a device_masked_copy from the dummy fields set inside the user routine to the actual fields. I guess the idea is that you can perform GPU-only operations based on math routines in the user code, and then you do not need a copy. Or I guess nowadays one could even write a custom kernel?

njansson (approved):

timfelle (approved):
Great. It should be up to the user to understand that device code is needed if one intend to execute on devices.
I know I had some issues based on that earlier

PR #2117, opened 2025-10-01
Fix CI for nvhpc 25.9

njansson:
Fix CI builds for cuda using latest nvhpc 25.9 with cuda 13


njansson:
We can't fit onto the runners limited disk space

PR #2127, opened 2025-10-07
enable initialisation and destruction of a neko field registry from api

Shiyu-Sandy-Du:
The purpose of this PR is to allow for loops to run cases in API. When the users are intended to launch runs by for loops, they need to initialise the field registry at the begining and destruct it at the end of each for-loop iteration.

### DIFF
--- a/src/api/neko.h.in
+++ b/src/api/neko.h.in
@@ -17,6 +17,12 @@ void neko_finalize();
 /* Display job information */
 void neko_job_info();
 
+/* Initialise a Neko field registry */
+void neko_field_registry_init();
+
+/* Destroy a Neko field registry */
+void neko_field_registry_free();
+
 /* Allocate memory for a Neko case */
 void neko_case_allocate(int **case_iptr);
 
--- a/src/api/neko_api.f90
+++ b/src/api/neko_api.f90
@@ -108,6 +108,20 @@ contains
 
   end subroutine neko_api_job_info
 
+  !> Initialise a Neko field registry
+  subroutine neko_api_field_registry_init() bind(c, name="neko_field_registry_init")
+
+    call neko_field_registry%init()
+
+  end subroutine neko_api_field_registry_init
+
+  !> Destroy a Neko field registry
+  subroutine neko_api_field_registry_free() bind(c, name="neko_field_registry_free")
+
+    call neko_field_registry%free()
+
+  end subroutine neko_api_field_registry_free
+
   !> Allocate memory for a Neko case
   !! @param case_iptr Opaque pointer for the created Neko case
   subroutine neko_api_case_allocate(case_iptr) &
--- a/src/python/pyneko/intf.py.in
+++ b/src/python/pyneko/intf.py.in
@@ -36,6 +36,10 @@ libneko.neko_init.resType = None
 libneko.neko_finalize.resType = None
 libneko.neko_job_info.resType = None
 
+libneko.neko_field_registry_init.restype = None
+libneko.neko_field_registry_free.restype = None
+
+
 libneko.neko_case_allocate.argtypes = [POINTER(c_int)]
 libneko.neko_case_allocate.restype = None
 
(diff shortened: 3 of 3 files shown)


Shiyu-Sandy-Du:
A simpler way is to merge this two subroutines into one called flush and called at the end of the loop iteration, but it might be somehow inconsistent with the code design of Neko.

timofeymukha:
Are you actually using the API for your runs?

Shiyu-Sandy-Du:
> Isn't this kind of dealt with using `neko_init` + `neko_finalize` or is that not compatible with the loop thingy

It will lead to MPI errors.

Shiyu-Sandy-Du:
And another advantage brought by running all the iterations with the same MPI context and the same Neko is to avoid the overhead to initialise Neko (which is significantly time consuming if each interation only contains a few seconds' run).

njansson:
> > > > Isn't this kind of dealt with using `neko_init` + `neko_finalize` or is that not compatible with the loop thingy
> > > 
> > > 
> > > It will lead to MPI errors.
> > 
> > 
> > Since `neko_finalize` also calls `MPI_finalize`?
> > I guess what you are trying is to run all the iterations, with the same MPI context.
> > Not sure this is the right solution, since you will not clean things, e.g. devices etc not cleared
> 
> What do I need to clean in devices apart from fields%x_d?

There's a lot of infrastructure in the device layer that is separate from just the pointers, e.g. the address lookup table, and streams, and for some backends a lot of dead code laying around and eating resources.

timfelle:
> > In the TopOpt extension to Neko. We wrote a reset funtion that allowed us to reset the entire state of neko to allow a new run. We have not tried to use it to swap to a different case, but it shows what should be overridden atleast.
> > https://github.com/ExtremeFLOW/neko-top/blob/a7c81030d835cd219004d1f0c7bbddf34b0e8bf8/sources/neko_ext/neko_ext.f90#L52
> 
> Thanks for the information! So if I understand it correctly, they are basically u,v,w,p,s and neko_case related stuffs?

That was sufficient for our purposes yes. We do not swap the mesh or something crazy like that. But we do run the same case with modified source terms and such many times.

Shiyu-Sandy-Du:
> > > In the TopOpt extension to Neko. We wrote a reset funtion that allowed us to reset the entire state of neko to allow a new run. We have not tried to use it to swap to a different case, but it shows what should be overridden atleast.
> > > https://github.com/ExtremeFLOW/neko-top/blob/a7c81030d835cd219004d1f0c7bbddf34b0e8bf8/sources/neko_ext/neko_ext.f90#L52
> > 
> > 
> > Thanks for the information! So if I understand it correctly, they are basically u,v,w,p,s and neko_case related stuffs?
> 
> Not exactly, it relates to the things setup before the case in `neko_init`. I think a minimal working example would be to also add the `device_init` and `device_finalize` calls to the API. Then you can do what you like without hogging resources

and I guess it will not destroy the mpi context, right?

Shiyu-Sandy-Du:
> > > > In the TopOpt extension to Neko. We wrote a reset funtion that allowed us to reset the entire state of neko to allow a new run. We have not tried to use it to swap to a different case, but it shows what should be overridden atleast.
> > > > https://github.com/ExtremeFLOW/neko-top/blob/a7c81030d835cd219004d1f0c7bbddf34b0e8bf8/sources/neko_ext/neko_ext.f90#L52
> > > 
> > > 
> > > Thanks for the information! So if I understand it correctly, they are basically u,v,w,p,s and neko_case related stuffs?
> > 
> > 
> > That was sufficient for our purposes yes. We do not swap the mesh or something crazy like that. But we do run the same case with modified source terms and such many times.
> 
> As long as the case is fixed, that would solve it. @Shiyu-Sandy-Du is your case fixed? (Minus the change in the params)

Nope, the case is initialised once the params are changed. But I guess it could be somehow fixed?

Shiyu-Sandy-Du:
I would say if we could modify params in API, then it is fine for me

PR #2139, opened 2025-10-17
Sponge source term

vbaconnet:
My suggestion for the sponge source term :) 

I copy pasted what I wrote in the documentation, just so you don't have to dig for information. The code snippet is modified a bit to be shorter.

I added the sponge to the cylinder example as a temporary example, will remove it if/when everyone is happy with this.

# TL;DR:

- Source term definition for this implementation: 

$$ \mathbf{\lambda} F(\mathbf{x}) ( \mathbf{u}^{bf} - \mathbf{u}) $$

- $\lambda$ is the amplitude of the sponge forcing in each Cartesian direction, given in the case file `"amplitudes": [1.0, 1.0, 1.0]`. If someone wants to only force in say x,y they would set `amplitudes: [1.0, 1.0, 0.0]`. 
- $u^{\text{bf}}$ is a reference (baseflow) velocity field, that the user can specify to:
    - constant values, 
    - imported from an `fld` file (with all the interpolation features), or 
    - user defined.
- $F(\mathbf{x})$ is a user-defined sponge mask field (a scalar value), defining where the sponge is active. The user has to add the fringe field to the registry and fill it themselves in the user file. The name of the fringe field in the registry can be chosen as well. 

# Full description (from case-file.md)

The sponge source term adds a term to each of the momentum equations of the form

$$ \mathbf{\lambda} f(\mathbf{x}) ( \mathbf{u}^{bf} - \mathbf{u}) $$

where:

- \f$ \mathbf{\lambda} \f$ is a 3-element vector of amplitudes of the sponge forcing in each Cartesian direction,
- \f$ \mathbf{u}^{\text{bf}} \f$ is a reference (baseflow) velocity field,
- \f$ f(\mathbf{x}) \f$ is a user-defined sponge mask field, defining where the sponge is active.

Amplitudes are specified using the `amplitudes` keyword with an array of 3 reals. Any of those values can be set to 0 to suppress the forcing in that particular direction. For example `[1.0, 1.0, 0.0]` will multiply the fringe field by 1, 1, and 0 in the `x`, `y` and `z` directions respectively, effectively removing the forcing in the `z` direction.
 
The reference velocity field, or `baseflow` can be set from three methods:
1. `constant`, applies constant values according to the `values` keyword:

   ```.json
   "source_terms": [
      {
         "type": "sponge",
         "amplitudes": [1.0, 1.0, 1.0],
         "baseflow": {
             "method": "constant",
             "value": [2.0, 0.0, 0.0]
         }
      }
   ]
   ```

2. `field`, where the velocity fields are retrieved from an `fld` file. 
   Uses the same parameters as the field initial condition.
   @note The same parameters as the `field` initial condition apply here.
   
   ```json
   "source_terms": [
      {
         "type": "sponge",
         "amplitudes": [1.0, 1.0, 1.0],
         "baseflow": {
             "method": "field",
             "file_name": "my_field0.f00016",
             "mesh_file_name": "my_field0.f00000",
             "interpolate": true,
             "tolerance": 1e-6
         }
      }
   ]
   ```

3. `user`, where the velocity field is set according to what 
   is defined by the user. In that case the base flow fields must be added to the registry by the user.

   ```json
   "source_terms": [
      {
         "type": "sponge",
         "amplitudes": [1.0, 1.0, 1.0],
         "baseflow": {
             "method": "user"
         }
      }
   ]
   ```

Finally, the fringe function field must be filled by the user. This must be done through the user file by adding the fringe field to the `neko_field_registry` in either `user_init_modules` or `fluid_user_ic` (more specifically, before the first call to compute the sponge source term).

The fringe field must be set by adding a field to the `neko_field_registry` under a specific name that can be retrieved internally. By default, Neko will search for the field `"sponge_fringe"` in the registry, but this can be changed by setting the parameter `fringe_registry_name`, which is important when using more than one sponge source term. The same applies to the base flow fields, which will be fetched from the registry based on a prefix + `_uvw`. This prefix can be changed by setting `baseflow_registry_prefix`, which is  by default `sponge_bf` (e.g. for u velocity it is stored as `sponge_bf_u`)

```.f90
  ! User-defined initialization called just before time loop starts
  subroutine user_initialize(t)
    type(time_state_t), intent(in) :: t

    type(field_t), pointer :: u, ubf, vbf, wbf, fringe
    real(kind=rp) :: x
    integer :: i

    u => neko_field_registry%get_field("u")
    call neko_field_registry%add_field(u%dof,"sponge_fringe")
    fringe => neko_field_registry%get_field("sponge_fringe")

    ! Initialize the base flows
    call neko_field_registry%add_field(u%dof,"sponge_bf_u")
    ubf => neko_field_registry%get_field("sponge_bf_u")
    call neko_field_registry%add_field(u%dof,"sponge_bf_v")
    vbf => neko_field_registry%get_field("sponge_bf_v")
    call neko_field_registry%add_field(u%dof,"sponge_bf_w")
    wbf => neko_field_registry%get_field("sponge_bf_w")

    fringe = 0.0_rp
    do i = 1, fringe%size()
        x = fringe%dof%x(i,1,1,1)

        if (x .gt. 10.0_rp) then
           fringe%x(i,1,1,1) = 1.0_rp
        end if
        
       ! Set ubf,vbf to something random
       ubf%x(i,1,1,1) = sin(3.1415926_rp*2.0_rp/10.0_rp * x)
       vbf%x(i,1,1,1) = cos(3.1415926_rp*2.0_rp/10.0_rp * y)

    end do

    wbf = 0.0_rp
       call device_memcpy(ubf%x, ubf%x_d, ubf%size(), &
            HOST_TO_DEVICE, .false.)
       call device_memcpy(vbf%x, vbf%x_d, vbf%size(), &
            HOST_TO_DEVICE, .false.)
       call device_memcpy(fringe%x, fringe%x_d, fringe%size(), &
            HOST_TO_DEVICE, .false.)
    end if

    ! NOTE: You can dump the fringe field to file using the `dump_fields`
    ! parameter. The fringe field will be stored under `pressure`.

    nullify(fringe)
    nullify(u)
nullify(ubf)
nullify(vbf)
nullify(wbf)

  end subroutine user_initialize
```

In order to visualize your baseflow and fringe field, you may set `dump_fields` to `true`. An `fld` file will be written to disk as  `spng_fields.fld`(note, not in `output_directory`) with the fringe field stored as `pressure`. You may change the name of the field file by setting `dump_file_name` (must have the extension `fld`).

The parameters for the sponge source term are summarized in the table below: 

| Name                     | Description                                                                 | Admissible values                     | Default value       |
|--------------------------|-----------------------------------------------------------------------------|---------------------------------------|---------------------|
| `amplitudes`             | Sponge forcing strength in each Cartesian direction                         | Array of 3 reals                      | -                   |
| `baseflow.method`        | Method to define the reference (baseflow) velocity                          | `"constant"`, `"field"`, `"user"` | -          |
| `baseflow.value`        | Velocity vector for constant baseflow                                       | Array of 3 reals                      | -                   |
| `baseflow.file_name`     | File containing baseflow velocity field                                     | String                                | -                   |
| `baseflow.mesh_file_name`| Mesh file corresponding to the baseflow field                               | String                                | -                   |
| `baseflow.interpolate`   | Whether to interpolate field values to current mesh                         | Boolean                               | `false`             |
| `baseflow.tolerance`     | Tolerance for interpolation convergence                                     | Real                                  | -                   |
| `fringe_registry_name`   | Name of the fringe mask field in `neko_field_registry`                      | String                                | `"sponge_fringe"`   |
| `dump_fields`            | If `true`, dumps the fringe and baseflow fields for visualization           | Boolean                               | `false`             |
| `dump_file_name`         | Name of the `fld` file in which to dump the base flow and fringe fields     | String ending with `fld`              | `spng_fields.fld`   |

### DIFF
--- a/src/source_terms/sponge_source_term.f90
+++ b/src/source_terms/sponge_source_term.f90
@@ -143,7 +143,7 @@ contains
 
     call json_get(json, "amplitudes", amplitudes)
     if (size(amplitudes) .ne. 3) then
-         call neko_error("(SPONGE) Expected 3 elements for 'amplitudes'")
+       call neko_error("(SPONGE) Expected 3 elements for 'amplitudes'")
     end if
 
     call json_get_or_default(json, "start_time", start_time, 0.0_rp)
@@ -187,7 +187,7 @@ contains
 
        call json_get(baseflow_subdict, "value", constant_value)
        if (size(constant_value) .lt. 3) then
-            call neko_error("(SPONGE) Expected 3 elements for 'value'")
+          call neko_error("(SPONGE) Expected 3 elements for 'value'")
        end if
 
        call this%init_constant(fields, coef, start_time, end_time, &


src/source_terms/sponge_source_term.f90
```diff
@@ -0,0 +1,497 @@
...
+
+
+    call json_get(json, 'baseflow', &
+         baseflow_subdict)
+    call json_get(baseflow_subdict, "method", baseflow_method)
+
+    select case (trim(baseflow_method))
+
+       ! Import the base flow from an fld file. The same parameters as field
+       ! initial condition apply since we use the same subroutine to read adn
```
timofeymukha:
```suggestion
       ! initial condition apply since we use the same subroutine to read and
```

src/source_terms/sponge_source_term.f90
```diff
@@ -0,0 +1,536 @@
...
+       v_name = trim(this%bf_rgstry_pref) // "_v"
+       w_name = trim(this%bf_rgstry_pref) // "_w"
+
+       ! Check if all the base flow fields exist in the registry
+       this%baseflow_set = neko_field_registry%field_exists(trim(u_name)) &
+               .and. neko_field_registry%field_exists(trim(v_name)) .and. &
+               neko_field_registry%field_exists(trim(w_name))
+
+       if (.not. this%baseflow_set) call neko_error("SPONGE: No baseflow set &
+       &(searching for " // trim(this%bf_rgstry_pref) // "_u)")
```
timfelle: Please don't make single line if statements which do not fit a single line.
```suggestion
       if (.not. this%baseflow_set) then
          call neko_error("SPONGE: No baseflow set (searching for " // &
               trim(this%bf_rgstry_pref) // "_u)")
       end if
```

src/source_terms/sponge_source_term.f90
```diff
@@ -0,0 +1,536 @@
...
+
+       ! Check if the user has added the fringe field in the registry
+       if (.not. &
+            neko_field_registry%field_exists(trim(this%fringe_registry_name))) &
+            call neko_error("SPONGE: No fringe field set (" // &
+                this%fringe_registry_name // " not found)")
```
timfelle: Same here

```suggestion
       if (.not. neko_field_registry%field_exists( &
            trim(this%fringe_registry_name))) then
          call neko_error("SPONGE: No fringe field set (" // &
               this%fringe_registry_name // " not found)")
       end if
```

src/source_terms/sponge_source_term.f90
```diff
@@ -0,0 +1,536 @@
...
+
+    call wk%free()
+
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_memcpy(this%u_bf%x, this%u_bf%x_d, this%u_bf%size(), &
+            HOST_TO_DEVICE, .false.)
+       call device_memcpy(this%v_bf%x, this%v_bf%x_d, this%v_bf%size(), &
+            HOST_TO_DEVICE, .false.)
+       call device_memcpy(this%w_bf%x, this%w_bf%x_d, this%w_bf%size(), &
+            HOST_TO_DEVICE, .false.)
```
timfelle: In general the last copy of a block should sync.
```suggestion
            HOST_TO_DEVICE, .true.)
```

src/source_terms/sponge_source_term.f90
```diff
@@ -0,0 +1,536 @@
...
+
+    real(kind=rp), allocatable :: amplitudes(:)
+    real(kind=rp), allocatable :: ramp(:), pad(:)
+
+    character(len=:), allocatable :: baseflow_file, baseflow_method
+    character(len=:), allocatable :: read_str, fringe_registry_name, &
+            bf_registry_pref, dump_fname
+    character(len=NEKO_FNAME_LEN) :: fname, mesh_fname
+    logical :: interpolate, dump_fields
+    real(kind=rp), allocatable :: constant_value(:)
```
timfelle: This is also not deallocated

timfelle (changes requested):

src/source_terms/sponge_source_term.f90
timofeymukha: This is local tothe subroutine, right? Then, Fortran should deallocate automatically.

src/source_terms/sponge_source_term.f90
```diff
@@ -0,0 +1,539 @@
...
+
+    type(json_file) :: json_subdict
+    character(len=:), allocatable :: string_val
+
+    character(len=1024) :: u_name, v_name, w_name
+    type(fld_file_output_t) :: fout
+    integer :: tmp_index
+    class(case_t), pointer :: case
+    type(field_t), pointer :: wk
+    case => null()
```
timfelle: The case is never used
```suggestion
    type(field_t), pointer :: wk
```

examples/cylinder/cylinder.f90
```diff
@@ -0,0 +1,89 @@
...
+  ! User-defined initialization called just before time loop starts
+  subroutine user_initialize(t)
+    type(time_state_t), intent(in) :: t
+
+    type(field_t), pointer :: u, fringe
+    real(kind=rp) :: x, y, xmin1, delta_rise1, xmin2, delta_rise2
+    integer :: i, imask
+
+    fringe => null()
+    u => null()
```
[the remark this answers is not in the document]
timofeymukha: But keep it somewhere, maybe we can use that in a test later on?

vbaconnet:
Yea! All changes should be in now :)

PR #2149, opened 2025-10-29
Feature/scalar statistics

nick-morse:
Adding a scalar statistics simcomp, which is computed in the same manner as the fluid statistics. Basic statistics includes $\langle \theta \rangle$, $\langle u'_i \theta' \rangle$, and $\langle \theta' \theta' \rangle$, while the full statistics includes quantities to calculate skewness and kurtosis of $\theta'$, as well as quantities to calculate budgets of $\langle \theta' \theta' \rangle$ and $\langle u'_i \theta' \rangle$.

To compute some terms in the budgets, you would also need values from the fluid statistics (so as to not duplicate terms), which I assume the user would already be collecting. 

Right now I have an optional json entry for the name of the statistics field in case of multiple scalars, but I would be interested to hear your thoughts on what should be done in this case.

@linneahuusko tested the basic statistics for an ABL, and I tested the basic statistics and budgets of $\langle \theta' \theta' \rangle$, $\langle u' \theta' \rangle$, and $\langle v' \theta' \rangle$ in a turbulent channel flow at $Re_\tau = 180$ and $Pr = 0.71$ based on the `turb_channel` example, which I compare to the reference data from [https://www.rs.tus.ac.jp/t2lab/db/poi/poi.html](https://www.rs.tus.ac.jp/t2lab/db/poi/poi.html) and [https://jaxa-dns-database.jaxa.jp/channelflow.html](https://jaxa-dns-database.jaxa.jp/channelflow.html). These show reasonable agreement to suggest that the statistics are implemented correctly (despite noise in some of the terms). See the comparison of my data (colored) vs the reference data (black) below.

### Velocity statistics (for reference):
image
image

### Scalar profiles:
image
image

### Budget of $\langle \theta' \theta' \rangle$:
image

### Budgets of $\langle u' \theta' \rangle$ and $\langle v' \theta' \rangle$:
image
image

### DIFF
--- a/src/scalar/scalar_stats.f90
+++ b/src/scalar/scalar_stats.f90
@@ -1,4 +1,4 @@
-! Copyright (c) 2022-2024, The Neko Authors
+! Copyright (c) 2025, The Neko Authors
 ! All rights reserved.
 !
 ! Redistribution and use in source and binary forms, with or without


linneahuusko (approved):

src/simulation_components/scalar_stats_simcomp.f90
```diff
@@ -0,0 +1,279 @@
+! Copyright (c) 2024, The Neko Authors
```
njansson:
```suggestion
! Copyright (c) 2025, The Neko Authors
```

njansson (approved):

PR #2153, opened 2025-10-31
Stabilized-convergence logging

timfelle:
Add initial grace period for convergence warnings and checks.

This should help reduce superflous "Solver did not converged" messages as the flow stabilizes during the initial part of the simulation.

Here we declare a flow "stabilized" on the first iteration for which the solver converged for all components.

To facilitate programatic use, we also introduce a subroutine to reset the value.

### DIFF
--- a/doc/pages/user-guide/case-file.md
+++ b/doc/pages/user-guide/case-file.md
@@ -791,7 +791,8 @@ the prefix `"sponge_bf_"`, meaning that `u` will be in `sponge_bf_u`, etc.
 This prefix can be changed by setting the parameter `bf_registry_prefix`.
 
 <details>
-<summary><b><u>Example using `user_init_modules`</u></b></summary> 
+<summary><b><u>Example using `user_init_modules`</u></b></summary>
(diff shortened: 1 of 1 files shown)


timofeymukha:
@timfelle We just discussed this at the user meeting, and it seems people are generally happy with always having the warning there, to be on the safe side. Is there any particular reason / workflow where this becomes very annoying?

timfelle:
> > > Can't we reuse some of the verbosity logging settings we have to implement both things?
> > 
> > 
> > Well our default log level is 1. So there is not really room to add an extra log level between the default and the quiet mode. And it seems everyone want the default level to be the current verbosity.
> > So we should then move our default log level to something else, so that we can add a new level. Not sure how intuitive it would be for the default to be 2, 5 or something else.
> 
> But warnings doesn't respect the log levels, or does they?

No, a warning is a message printed to `stdout`. So they are not part of the logging system.

Shiyu-Sandy-Du (approved):

PR #2154, opened 2025-10-31
Always report runtime statistics if enabled

timfelle:
Ensure runtime statistics are logged even when quiet mode is enabled.


timofeymukha (approved):
Makes sense!

njansson (approved):

PR #2155, opened 2025-10-31
Add neko_vector_registry

timofeymukha:
Adds a registry for vector_t objects, similar to the field_registry_t. Has essentially the same capabilities, and has an associated singleton, neko_vector_registry_t. 

I wanted the possibility to retrieve by name, but vectors store no name unlike fields. So, I added a private named_vector type, which is what is stored in the registry. However, the user never sees this type, and the registry returns pointers to raw vector_t.


src/math/vector_registry.f90
```diff
@@ -0,0 +1,379 @@
...
+  !! A registry of named `vector_t` objects.
+  !! @details Same as field_registry_t but for vectors. Stores named
+  !! vectors, but raw vectors are returned by the `get_` methods.
+  !! Supports adding aliases to vectors as well, so that the same vector can be
+  !! accessed by different names.
+  type, public :: vector_registry_t
+     !> List of vectors stored.
+     type(named_vector_t), private, allocatable :: vectors(:)
+     !> List of aliases to vectors stored.
+     type(json_file), private, allocatable :: aliases(:)
```
timfelle: Since the names here are not really associated to the vectors, does it make
sense to have the alias components here?

src/math/vector_registry.f90
```diff
@@ -0,0 +1,379 @@
...
+    if (allocated(this%aliases)) then
+       deallocate(this%aliases)
+    end if
+
+    this%n_vectors_ = 0
+    this%n_aliases_ = 0
+    this%expansion_size = 0
+  end subroutine vector_registry_free
+
+  !> Expand the vectors array so as to accomodate more vectors.
```
timfelle:
```suggestion
  !> Expand the vectors array so as to accommodate more vectors.
```

src/math/vector_registry.f90
```diff
@@ -0,0 +1,379 @@
...
+  !> Get pointer to a stored vector by index.
+  function vector_registry_get_vector_by_index(this, i) result(f)
+    class(vector_registry_t), target, intent(in) :: this
+    integer, intent(in) :: i
+    type(vector_t), pointer :: f
+
+    if (i < 1) then
+       call neko_error("Vector index must be > 1")
+    else if (i > this%n_vectors()) then
+       call neko_error("Vector index exceeds number of stored fields")
```
timfelle:
```suggestion
       call neko_error("Vector index exceeds number of stored vectors")
```

src/math/vector_registry.f90
```diff
@@ -0,0 +1,379 @@
...
+  !! @param ignore_existing If true, will do nothing if the vector is already in
+  !! the registry. If false, will throw an error. Optional, defaults to false.
+  subroutine vector_registry_add_vector(this, n, name, ignore_existing)
+    class(vector_registry_t), intent(inout) :: this
+    integer, intent(in) :: n
+    character(len=*), target, intent(in) :: name
+    logical, optional, intent(in) :: ignore_existing
+
+    if (this%vector_exists(name)) then
+       if (present(ignore_existing) .and. ignore_existing .eqv. .true.) then
```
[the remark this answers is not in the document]
timfelle: Also, I am quite sure we decided against the using the equivalence operator earlier.

src/math/vector_registry.f90
```diff
@@ -0,0 +1,379 @@
...
+          vector_registry_get_expansion_size
+     !> Get total allocated size of `vectors`.
+     procedure, pass(this) :: get_size => vector_registry_get_size
+     !> Check if a vector with a given name is already in the registry.
+     procedure, pass(this) :: vector_exists => vector_registry_vector_exists
+     generic :: get_vector => get_vector_by_index, get_vector_by_name
+  end type vector_registry_t
+
+  !> Global vector registry
+  type(vector_registry_t), public, target :: neko_vector_registry
```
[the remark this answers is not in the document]
timfelle: Disregard, initialization does not work it seems

src/math/vector_registry.f90
```diff
@@ -0,0 +1,379 @@
...
+  !> Get the expansion size.
+  pure function vector_registry_get_expansion_size(this) result(n)
+    class(vector_registry_t), intent(in) :: this
+    integer :: n
+
+    n = this%expansion_size
+  end function vector_registry_get_expansion_size
+
+  !> Get pointer to a stored vector by index.
+  function vector_registry_get_vector_by_index(this, i) result(f)
```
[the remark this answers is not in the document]
timofeymukha: I think in practice it is probably not. At least not if you use the neko_... registry. Potentially if you use a local registry for some reason, but even then probably not so useful, to be honest. I guess it was just easy to add the routine to the field registry, from which it got carried over here.

src/math/vector_registry.f90
timfelle: Well it was more a question of whether this is ever useful. But feel free to resolve this conversation if you think it is.

timfelle (approved):

njansson (approved):

PR #2156, opened 2025-10-31
Fix pointer assignments for safety and clarity

timfelle:
Ensure all pointer variables are initialized to null to prevent potential issues during destruction and improve memory management.

### DIFF
--- a/src/math/schwarz.f90
+++ b/src/math/schwarz.f90
@@ -92,11 +92,11 @@ module schwarz
      type(gs_t) :: gs_schwarz !< We are only interested in the gather-scatter!
      type(dofmap_t) :: dm_schwarz !< needed to init gs
      type(fdm_t) :: fdm
-     type(space_t), pointer :: Xh
-     type(bc_list_t), pointer :: bclst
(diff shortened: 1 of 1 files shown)


timofeymukha (approved):

src/math/schwarz.f90
```diff
@@ -142,8 +143,9 @@ subroutine schwarz_init(this, Xh, dof, gs_h, bclst, msh)
     ! If we are running multithreaded, we need a local gs object,
     ! otherwise we can reuse the external one
     if (nthrds .gt. 1) then
-       allocate(this%gs_h)
-       call this%gs_h%init(this%dof)
+       allocate(this%gs_h_local)
+       call this%gs_h_local%init(this%dof)
+       this%gs_h => this%gs_h_local
```
njansson: Out of curiosity, why issues did the previous construct cause? I have been running that on both NVIDIA and AMD devices for some time now

njansson: Puuurfect, I was a bit afraid that something was messed up

njansson (approved):

PR #2165, opened 2025-11-17
Introduce matrix version of scratch registry

timfelle:
This matrix registry allow temporary matrices to be reused.

When requesting a matrix, the desired size is specified. A new matrix will be allocated if no matrices of the given size is currently available.

### DIFF
--- a/tests/unit/matrix_scratch_registry/test_matrix_scratch_registry.pf
+++ b/tests/unit/matrix_scratch_registry/test_matrix_scratch_registry.pf
@@ -21,6 +21,7 @@ contains
     @assertEqual(scratch%get_size(), 10)
     @assertEqual(scratch%get_nmatrices_inuse(), 0)
 
+    call scratch%free()
   end subroutine test_matrix_scratch_registry_init_default
 
   @test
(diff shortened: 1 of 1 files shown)


src/math/matrix_scratch_registry.f90
```diff
@@ -0,0 +1,268 @@
...
+     !> Getter for expansion_size
+     procedure, pass(this) :: get_expansion_size
+     !> Return size of allocated matrices
+     procedure, pass(this) :: get_size
+     !> Get value of inuse for a given index
+     procedure, pass(this) :: get_inuse
+     !> Get a new scratch matrix
+     procedure, pass(this) :: request_matrix
+     procedure, pass(this) :: relinquish_matrix_single
+     procedure, pass(this) :: relinquish_matrix_multiple
```
timofeymukha: What I wonder now that we have sevral of these registries is whether it would not be better to drop "matrix", "field", and "vector" from the procedure names, to just have a unified interface.

Had the same thoughts with the non-scratch registries as well...

timfelle: That is exactly the next thing I would go through

PR #2184, opened 2025-11-22
NEKO v1.0.0

njansson:
/ᐠ. ｡.ᐟ\ᵐᵉᵒʷˎˊ˗


timfelle (commented):
> **Important that we don't auto-merge this with a squash** 

I guess it should be a rebase instead right?
You can set specific rules on the github side to make sure squash is not allowed onto master.

PR #2187, opened 2025-11-24
Stop printing empty sections

timfelle:
Simple solution to the empty sections in the log.

Starting a section now stores the header in the class.
First message after the section call will print the header and clear the variable.
Ending a section will also clear the variable.


timofeymukha (approved):

njansson:
Nice! Finally getting rid of pre/post processing headers if not done.

However, we still get preprocessing printed, since we have CFL and dt info printed after. Should we move them before, such that we only get the preprocessing section if we have simcomps and or user routines (@timfelle @timofeymukha )?

PR #2188, opened 2025-11-24
Add vector kernels to neumann_t

timofeymukha:
Adds support for the vector apply for neumann_t.

As a consequence, the flux is an array of vector_t (should be 1 or 3), holding the fluxes.

I removed the badly implemented flux getter and renamed the flux_ component to just flux. It is not private, so a getter is not really needed.

### DIFF
--- a/src/bc/bcknd/device/hip/neumann_kernel.h
+++ b/src/bc/bcknd/device/hip/neumann_kernel.h
@@ -117,8 +117,8 @@ void neumann_apply_vector_kernel(const int * __restrict__ msk,
   const int str = blockDim.x * gridDim.x;
 
   for (int i = (idx + 1); i < m; i += str) {
-    const int k = (msk[i] - 1);
-    const int f = (facet[i]);
+    const int k = msk[i] - 1;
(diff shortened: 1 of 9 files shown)


src/bc/bcknd/device/hip/neumann_kernel.h
```diff
@@ -96,4 +96,64 @@ void neumann_apply_scalar_kernel(const int * __restrict__ msk,
...
+                                 const T * __restrict__ area,
+                                 const int lx,
+                                 const int m) {
+  int index[4];
+  const int idx = blockIdx.x * blockDim.x + threadIdx.x;
+  const int str = blockDim.x * gridDim.x;
+
+  for (int i = (idx + 1); i < m; i += str) {
+    const int k = (msk[i] - 1);
+    const int f = (facet[i]);
```
timfelle:
```suggestion
    const int k = msk[i] - 1;
    const int f = facet[i];
```

timfelle (approved):

Shiyu-Sandy-Du (approved):

PR #2194, opened 2025-11-26
User guide for meshing

timofeymukha:
A section on high order meshing should be added later.


njansson:
Puuurfect!

Isn't there one more file where the link should be added, I think it is the manual index page

njansson (approved):

timfelle (approved):

PR #2197, opened 2025-11-27
Make zeroing of scratch entries optional

njansson:
After looking at our current use of scratch fields, none of them (as far as I can see) relied on the returned data being zero.


njansson:
Duplicate of #2199

PR #2201, opened 2025-11-28
Reduce default registry size

timfelle:
Decrease the default size of the registry and its expansion size to optimize resource usage.


njansson:
> Tried it on LUMI with CCE 17, and for some reason it **did not** resolve the issue (T_T)

Seems to be a CCE 17 related bug, works fine with 16. Still not bad to reduce to memory footprint I would say

timofeymukha (approved):

PR #2217, opened 2025-12-03
Multiphase over integration

gloopydoop:
Finally a commit to multiphase!
In this PR I add a way of performing over integration (dealiasing) on the term

$\nabla \cdot \left( \phi (1-\phi) \frac{\nabla \phi}{|\nabla \phi|}\right)$

If one considers the test function too, this is 4 polynomials multiplied together, so it's not a 3/2 rules, it's a 4/2=2 rule.

You should be able to enable it with `"dealias_normal": true` under `scalar`.

There's also a little profiler wrapping each computations, I'm seeing (from small tests)
```

                        Total time        Avg. time       Range  +/-
 --------------------------------------------------------------------
 n_dealias            0.9233649E+02    0.7694707E-02    0.1670377E-01
 n_no_dealias         0.3231977E+01    0.2693314E-03    0.2046465E-04
 ```


tuananhdao (approved):

Shiyu-Sandy-Du:
why is it slower by 30 times? I would probably expect it's 10 times when assenble the total forcing term, and approximately the same time for the pressure and velocity equation (so in total much smaller than 30 times).

pschlatt1:
I am wondering in this regard whether implementing some type of 2/3 rule instead could be beneficial, so that one does not need to dealias many individual terms...? Or is there a drawback on doing that?

timofeymukha:
Can we merge this in, so it doesn't hang around in the PR list :)?

njansson:
> Can we merge this in, so it doesn't hang around in the PR list :)?

Sounds like a good plan

njansson:
> Can we merge this in, so it doesn't hang around in the PR list :)?

Bump this thread, should we get it in?

timofeymukha:
@gloopydoop merges tomorrow.

PR #2226, opened 2025-12-06
Add AGENTS.md

timofeymukha:
AGENTS.md has become a standard for providing instructions to AI, see agents.md on the web.

I experimented today with Codex, and wrote this file, which only contains information on how to help users setup a simulation case. I am generally very happy with the little testing I have done. 

- It easily setup a correct case file for a flat plate simulation with blasius inlet and IC. When I say correct, I mean the case file was valid and had the stuff I specified baked in. I was not very detailed, so of course the actual setup is crap (random timestep size and so on).
- Generated the mesh with genmeshbox, given the number of elements in the boundary layer and min and max spacings inside it (so it ran Python under the hood to produce the csv that is fed to the mesher)
- Finally, I prompted: "I want my simulation to compute a kinetic energy field every time step and save it to the hard drive every 10 seconds." It managed to figure out it needs a user file and that it can use the field_writer simcomp. The user file was essentially correct, maybe an extra if statement one could skip, but compiles and does what I asked for.

So, I think this is a start! I don't know how much we can eventually rely on AI to write code but I think
- Help with case setup.
- Simple to moderate things in the user file.
- Doing annoying stuff like adding all the necessary things to the build system for a new test or whatever.
Is totally within reach, and adding AGENT.md files at key locations may boost this dramatically.

image

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,8 +8,9 @@ of turbulent flows.
 ## Helping users setup new Neko simulation cases
 Neko is a simulation software, and a common scenario is that the user will ask
 you to help setup a simulation. Alternatively, they may want tips about specific
-settings or and explanation about a setup they find in an existing case. The instructions under this header are specifically for these scenarios. We
-will refer to a concrete simulation to as "the case". 
+settings or and explanation about a setup they find in an existing case. The
+instructions under this header are specifically for these scenarios. We will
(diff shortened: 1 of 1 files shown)


njansson (approved):

timfelle (approved):

timofeymukha:
@njansson Can you force-merge it in please?

PR #2227, opened 2025-12-07
Fix OpenCL memory leak: add missing clReleaseKernel calls

tuananhdao:
## Context

- The OpenCL backend had a memory leak where `cl_kernel` objects created with `clCreateKernel()` were never released

## Steps to reproduce

- I observed this on a macbook pro m4
  - `--with-opencl --enable-real=sp`
  - gfortran - GNU Fortran (Homebrew GCC 14.2.0_1) 14.2.0
  - gcc - Apple Clang 17.0.0
- Run the `examples/tgv` case with longer end time and observe the host memory (RAM) increases over time

## Changes

- Added `clReleaseKernel(kernel)` after every `clEnqueueNDRangeKernel()` call

### DIFF
--- a/src/math/bcknd/device/opencl/math.c
+++ b/src/math/bcknd/device/opencl/math.c
@@ -175,7 +175,7 @@ void opencl_cfill_mask(void* a, void* c, int* size, void* mask, int* mask_size,
                                   &global_item_size, &local_item_size,
                                   0, NULL, NULL));
   CL_CHECK(clReleaseKernel(kernel));
-  }
+}
 
 /** Fortran wrapper for rzero
  * Zero a real vector
@@ -1103,7 +1103,6 @@ real opencl_glsc3(void *a, void *b, void *c, int *n,
   CL_CHECK(clEnqueueNDRangeKernel(cmd_queue, kernel, 1, NULL,
                                   &global_item_size, &local_item_size,
                                   0, NULL, &kern_wait));
-  CL_CHECK(clReleaseKernel(kernel));
 
   CL_CHECK(clEnqueueReadBuffer(cmd_queue, bufred_d, CL_TRUE, 0,
                                nb * sizeof(real), bufred, 1,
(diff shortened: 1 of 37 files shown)


src/math/bcknd/device/opencl/math.c
```diff
@@ -1070,6 +1103,7 @@ real opencl_glsc3(void *a, void *b, void *c, int *n,
   CL_CHECK(clEnqueueNDRangeKernel(cmd_queue, kernel, 1, NULL,
                                   &global_item_size, &local_item_size,
                                   0, NULL, &kern_wait));
+  CL_CHECK(clReleaseKernel(kernel));
```
njansson: This is not 100% safe since we need to read data produced by the kernel in the following call. Best is to move the release to the end of the routine

src/math/bcknd/device/opencl/math.c
```diff
@@ -1132,6 +1166,7 @@ void opencl_glsc3_many(real *h, void * w, void *v, void *mult, int *j, int *n,
   CL_CHECK(clEnqueueNDRangeKernel(cmd_queue, kernel, 2, NULL,
                                   global_item_size, local_item_size,
                                   0, NULL, &kern_wait));
+  CL_CHECK(clReleaseKernel(kernel));
```
njansson: same here,

This is not 100% safe since we need to read data produced by the kernel in the following call. Best is to move the release to the end of the routine

njansson (changes requested):
Great catch!

However, if we dependencies on the kernel e.g. read operations, we need to move the call to `clReleaseKernel` to after those operations, e.g in `glsc3, cfl` etc

njansson:
I've fixed the outstanding issues with kernel release before read operations. Thus ready to be merged

njansson (approved):

timofeymukha (approved):

PR #2228, opened 2025-12-08
Add scalars to registry, add registered data to case file

timofeymukha:
The background here is that I want to enable a smoother way of defining values for case setup entries that are shared between various parts of the case setup. Concrete examples: there are several things that depend on `g`, the gravity vector, and on some `U0`, a freestream velocity. Currently, you have to set `g` or `U0 `in each of those case objects. If you want to change the value, you have to also change it everywhere and make sure that things are in sync. As cases get more complex, this may lead to user errors.

My solution is to introduce a new entry to the case file, which looks like this:
```json
        "constants":
        [
                {
                    "name": "c1",
                    "value": 1.0
                },
                {
                    "name": "int1",
                    "value": 1
                },
                {
                    "name": "g2",
                    "value": [1, 0, 0.0]
                }
          ]
```
The naming here are very much up to discussion. In particular, scalars get two separate meanings in the case file here, because it is also used for scalar solvers... See also below.

Given this, the user can use these registered entries in other parts of the case setup, e.g.
```json
            "source_terms": [
                {
                    "type": "boussinesq",
                    "scalar_field": "temperature",
                    //"g": [ 0, 0, 1 ], can still set directly, of course
                    "g": "g2", // Point to g2 in the constants
                    "reference_value": 0,
                    "beta": 1
                }
            ]
```
So, by referencing the same entry in the registered_data, we can have a single point of definition for a parameter across case components.

### Under the hood
The constants parsed by the constructor of `case_t` early. What it does is put these entries into `neko_const_registry`, which is aa separate registry just for this purpose. For arrays, I use the vector_t registry entries, but for scalars I now expanded the registry to also accept scalars. Here, again, there is the issue of using the word "scalar". Although it fits very naturally in the registry scalar -> vector -> matrix, so I think here it is less of an issue than in the case file. Regardless, suggestions welcome.

So, an added bonus of this system is that you can add things to the registry from the case file to later use in the user file.

I add a submodule to json_utils is introduced to parse case file entries and retrieve them either directly or via the registry. I use submodule so that json_utils itself remains clean of dependencies (`registry_t` itself uses `json_utils`, so it is important). For integers, we still use the real scalars in the registry, but check that no rounding has occurred. 

### Misc
- Assignment routine to vector_t from an array of reals.
- A routine to pretty-print the registry contents.


src/registries/registry_entry.f90
```diff
@@ -44,26 +45,35 @@ module registry_entry
...
+     !> Type of the registry entry; must be supproted.
      character(len=:), private, allocatable :: type
+     !> Whether the entry is allocated
      logical, private :: allocated = .false.
 
+     !> Storage. Only one of these will be allocated at a time.
      type(field_t), private, pointer :: field_ptr => null()
      type(vector_t), private, pointer :: vector_ptr => null()
      type(matrix_t), private, pointer :: matrix_ptr => null()
+     real(kind=rp), private :: scalar = 0.0_rp
```
timfelle: The `!>` attach the doxygen documentation to the following line. Please use `!` to signify that this is just a comment for the coming couple of lines, and not documentation of the `field_ptr` variable.

Also lets sort them based on dimensions.
```suggestion
     ! Storage. Only one of these will be allocated at a time.
     real(kind=rp), private :: scalar = 0.0_rp
     type(vector_t), private, pointer :: vector_ptr => null()
     type(matrix_t), private, pointer :: matrix_ptr => null()
     type(field_t), private, pointer :: field_ptr => null()
```

src/registries/registry_entry.f90
```diff
@@ -31,9 +31,10 @@
 ! POSSIBILITY OF SUCH DAMAGE.
 !
 !> Defines a registry entry for storing and requesting temporary objects
-!! This is used in the scratch registries to store temporary fields, vectors,
+!! This is used in the registries to store temporary fields, vectors,
 !! and matrices.
```
timfelle:
```suggestion
!! This is used in the registries to store a scalar, vector, matrix or field.
```

src/registries/registry_entry.f90
```diff
@@ -44,26 +45,35 @@ module registry_entry
...
      logical, private :: allocated = .false.
 
+     !> Storage. Only one of these will be allocated at a time.
      type(field_t), private, pointer :: field_ptr => null()
      type(vector_t), private, pointer :: vector_ptr => null()
      type(matrix_t), private, pointer :: matrix_ptr => null()
+     real(kind=rp), private :: scalar = 0.0_rp
 
    contains
+     !> Constructors
```
timfelle: This again only applied to init_field
```suggestion
     ! Constructors
```

njansson:
Haven't had time to go through the details yet. But in general, from now when we change interfaces, we should remember to bump X in the dev branch version number 1.99.X (to indicate a change)

src/registries/registry_entry.f90
```diff
@@ -44,26 +45,35 @@ module registry_entry
...
      procedure, pass(this) :: init_field => init_register_field
      procedure, pass(this) :: init_vector => init_register_vector
      procedure, pass(this) :: init_matrix => init_register_matrix
+     procedure, pass(this) :: init_scalar => init_register_scalar
```
timfelle:
```suggestion
     procedure, pass(this) :: init_scalar => init_register_scalar
     procedure, pass(this) :: init_vector => init_register_vector
     procedure, pass(this) :: init_matrix => init_register_matrix
     procedure, pass(this) :: init_field => init_register_field
```

src/registries/registry_entry.f90
```diff
@@ -44,26 +45,35 @@ module registry_entry
...
+     !> Destructor
      procedure, pass(this) :: free => free_register
 
+     !> Getters that return a pointer to the object in the entry.
      procedure, pass(this) :: get_name
      procedure, pass(this) :: get_type
      procedure, pass(this) :: get_field
      procedure, pass(this) :: get_vector
      procedure, pass(this) :: get_matrix
+     procedure, pass(this) :: get_scalar
```
timfelle:
```suggestion
     procedure, pass(this) :: get_scalar
     procedure, pass(this) :: get_vector
     procedure, pass(this) :: get_matrix
     procedure, pass(this) :: get_field
```

src/registries/registry_entry.f90
```diff
@@ -214,4 +247,15 @@ function get_matrix(this) result(matrix_ptr)
...
+  !> Get the scalar pointer of the registry entry
+  function get_scalar(this) result(scalar_ptr)
+    class(registry_entry_t), target, intent(in) :: this
+    real(kind=rp), pointer :: scalar_ptr
+    if (this%get_type() .ne. 'scalar') then
+       call neko_error("registry_entry::get_field: " &
+            // "Registry entry is not of type 'scalar'.")
+    end if
+    scalar_ptr => this%scalar
+  end function get_scalar
```
timfelle: This should not return a pointer.
We explicitly do not want the user to be able to modify the object. It is by necessity that we use pointers for the complex objects.
```suggestion
  !> Get the scalar value of the registry entry
  function get_scalar(this) result(scalar)
    class(registry_entry_t), intent(in) :: this
    real(kind=rp) :: scalar
    if (this%get_type() .ne. 'scalar') then
       call neko_error("registry_entry::get_field: " &
            // "Registry entry is not of type 'scalar'.")
    end if
    scalar = this%scalar
  end function get_scalar
```

tests/unit/registry/test_registry.pf
```diff
@@ -134,6 +134,7 @@ contains
     type(field_t), pointer :: field
     type(vector_t), pointer :: vector
     type(matrix_t), pointer :: matrix
+    real(kind=rp), pointer :: scalar
```
timfelle, line 137: It should not be a pointer
```suggestion
    real(kind=rp) :: scalar
```

tests/unit/registry/test_registry.pf
```diff
@@ -145,24 +146,29 @@ contains
...
 
     @assertEqual(.true., registry%entry_exists('test_matrix'))
     @assertEqual(.true., registry%entry_exists('test_vector'))
     @assertEqual(.true., registry%entry_exists('test_field'))
+    @assertEqual(.true., registry%entry_exists('test_scalar'))
 
     field => registry%get_field('test_field')
     vector => registry%get_vector('test_vector')
     matrix => registry%get_matrix('test_matrix')
+    scalar => registry%get_scalar('test_scalar')
```
timfelle:
```suggestion
    scalar = registry%get_scalar('test_scalar')
```

tests/unit/registry/test_registry.pf
```diff
@@ -228,6 +241,9 @@ contains
     vector => registry%get_vector('test_alias2')
     @assertEqual(42, vector%size())
 
+    scalar => registry%get_scalar('test_alias3')
```
timfelle:
```suggestion
    scalar = registry%get_scalar('test_alias3')
```

tests/unit/registry/test_registry.pf
```diff
@@ -194,6 +202,7 @@ contains
     type(registry_t) :: registry
     type(field_t), pointer :: field, alias_ptr
     type(vector_t), pointer :: vector
+    real(kind=rp), pointer :: scalar
```
timfelle:
```suggestion
    real(kind=rp) :: scalar
```

src/common/json_utils.f90
```diff
@@ -387,6 +405,8 @@ subroutine json_get_or_default_string(json, name, value, default)
     end if
   end subroutine json_get_or_default_string
 
+
+
```
timfelle:
```suggestion
```

src/common/case_file_utils.f90
```diff
@@ -0,0 +1,119 @@
...
+    if (found) return
+
+    found = neko_registry%scalar_exists(name)
+    if (found) then
+       scalar_ptr => neko_registry%get_scalar(name)
+       val = scalar_ptr
+    else
+       val = default
+    end if
+  end subroutine json_get_real_from_registry_or_entry_or_default
```
timfelle:
```suggestion
  module subroutine json_get_real_from_registry_or_entry_or_default(json, name,&
       val, default)
    type(json_file), intent(inout) :: json
    character(len=*), intent(in) :: name
    real(kind=rp), intent(out) :: val
    real(kind=rp), intent(in) :: default

    logical :: found

    call json%get(name, val, found)
    if (found) return

    found = neko_registry%scalar_exists(name)
    if (found) then
       val = neko_registry%get_scalar(name)
    else
       val = default
    end if
  end subroutine json_get_real_from_registry_or_entry_or_default
```

src/common/case_file_utils.f90
```diff
@@ -0,0 +1,119 @@
...
+    character(len=*), intent(in) :: name
+    real(kind=rp), allocatable, intent(out) :: val(:)
+    type(vector_t), pointer :: vec_ptr
+
+    logical :: found
+
+    call json%get(name, val, found)
+    if (found) return
+
+    vec_ptr => neko_registry%get_vector(name)
```
timfelle: Remember to sync the vector. You have no guarantee that the vector in the registry is actually synced to the host. Actually the expectation is that it is not.
```suggestion
    vec_ptr => neko_registry%get_vector(name)
    call vec_ptr%sopy_from(DEVICE_TO_HOST, sync = .true.)
```

timfelle (commented):
Great idea to extend it with scalars. I guess all we need is some rank 3 object (`tensor` perhaps) then we support everything up to rank 4.

timfelle:
How does this interact with the aliasing stuff. Is that also part of the same section in the json?

src/common/case_file_utils.f90
```diff
@@ -0,0 +1,119 @@
...
+  use registry, only : neko_registry
+  use vector, only : vector_t
+  implicit none
+
+contains
+
+!> Retrieves a real either from the json or from the corresponding scalar
+!! in the `neko_registry`.
+!! @details First tries to retrieve the array from the JSON, looking for a real
+!! entry unde the given name. If not found, the name is looked up in the
```
vbaconnet:
```suggestion
!! entry under the given name. If not found, the name is looked up in the
```

src/common/case_file_utils.f90
```diff
@@ -0,0 +1,119 @@
...
+    if (found) return
+
+    scalar_ptr => neko_registry%get_scalar(name)
+    val = scalar_ptr
+  end subroutine json_get_real_from_registry_or_entry
+
+!> Retrieves a real either from the json or from the corresponding scalar
+!! in the `neko_registry`, otherwise sets the default value.
+!! @details First tries to retrieve the array from the JSON, looking for a real
+!! entry unde the given name. If not found, the name is looked up in the
```
vbaconnet:
```suggestion
!! entry under the given name. If not found, the name is looked up in the
```

src/common/case_file_utils.f90
```diff
@@ -0,0 +1,119 @@
...
+       val = scalar_ptr
+    else
+       val = default
+    end if
+  end subroutine json_get_real_from_registry_or_entry_or_default
+
+!> Retrieves a real array either from the json or from the corresponding vector
+!! in the `neko_registry`.
+!! @details First tries to retrieve the array from the JSON, looking for a real
+!! vector entry unde the given name. If not found, the name is looked up in the
```
vbaconnet:
```suggestion
!! vector entry under the given name. If not found, the name is looked up in the
```

timfelle:
Also remember to add a scalar specific test for the registry. Similar to field vector and matrix registries.

tests/unit/registry/test_registry.pf
timofeymukha: If we use the pointer, we can change the value in the registry by changing `scalar`, right? Which would be consistent with the other types. Not that it really matters for the test..

timfelle: Do we want to be able to change the value of the scalar? I guess it might be useful in some situations. I only though of global constants, but i guess there might be other usecases.

src/bc/inflow.f90
```diff
@@ -71,7 +71,7 @@ subroutine inflow_init(this, coef, json)
     real(kind=rp), allocatable :: x(:)
 
     call this%init_base(coef)
-    call json_get(json, 'value', x)
+    call json_get_from_registry_or_entry(json, 'value', x)
```
vbaconnet: so this is searching for `"value"` in the registry, if it cannot be found in the boundary condition JSON object? Isn't it a bit too generic of a name to have in the registry? Could it clash with the shear stress?

src/bc/shear_stress.f90
```diff
@@ -180,7 +180,7 @@ subroutine shear_stress_init(this, coef, json)
     type(json_file), intent(inout) ::json
     real(kind=rp), allocatable :: value(:)
 
-    call json_get(json, 'value', value)
+    call json_get_from_registry_or_entry(json, 'value', value)
```
vbaconnet: see my previous comment for `inflow.f90`, is the value keyword intended to be the same as for the inflow BC?

src/case.f90
```diff
@@ -492,8 +529,8 @@ subroutine case_init_common(this)
             path = this%output_directory, fmt = trim(string_val))
        call json_get_or_default(this%params, 'case.checkpoint_control', &
             string_val, "simulationtime")
-       call json_get_or_default(this%params, 'case.checkpoint_value', &
-            real_val, 1e10_rp)
+       call json_get_from_registry_or_entry(this%params, &
+            'case.checkpoint_value', real_val, 1e10_rp)
```
vbaconnet, line 545: Why the change from `json_or_default`? I have nothing against adding that parameter to the registry, just seems a bit arbitrary... I'm sure there's a use case for it that I missed :)

vbaconnet (commented):
I think in general this is super nice and will be really nice for global variable management and a very dynamic simulation workflow! Just a bit unsure about the inflow/shear stress `"value"` thingy.

Sorry if some of my comments have been addressed before already, it's very possible that I missed some discussions..

src/bc/inflow.f90
timofeymukha: Oh, yeah, this is actually a very good point. I have underestimated the complexity here.

We need to support the keyword in the json and the name in the registry to be different of course. I got tunnelvisioned by "g" :-)).

src/case.f90
timofeymukha: This still preserves the same default value. But here it is the same situation as in your other comments, of course. This will look for "case.checkpoint_value" in the registry if not found in the JSON :-)

timfelle: Is there a reason for splitting this out in the first place. And not just look at the entry type in `json_get` and `json_get_or_default`?

timofeymukha: Then we would need the json_utils module to depend on the registry, which is to be avoided because we use json in registry for the aliases..

timfelle: Not if you moved it to a submodule, just like the new subroutines you just made.

src/bc/inflow.f90
timofeymukha: @vbaconnet The new version of the routines adresses this hopefully. We first lookup a real under "value", then lookup a string under "value", and that string is then treated as the name to lookup in the registry.

vbaconnet: > I got tunnelvisioned by "g" :-)).

Yea for sure, I mean it great for global variables like that so that makes sense!

> then lookup a string under "value"

So another string can be passed down t the subroutine as an "alias" in the registry?

src/case.f90
vbaconnet: > This will look for "case.checkpoint_value" in the registry if not found in the JSON :-)

but is it intentional that `checkpoint_value` should be looked up in the registry? that's what i meant with that I don't see the direct use case for this. Is it for the nekotop backwards adjoint stuff?

timfelle: No no. You place a value in the registry `constant_1: 2.0`.
And then you specify `case.checkpoint_value: "constant_1"`

timfelle: So you can use `constant_1` in multiple places of the case file, instead if writing 2.0 everywhere.

vbaconnet: Trying to wrap my head around this, sorry...

Based on your example: Won't the subroutine `json_get_real_from_registry_or_entry_or_default` lookup `case.checkpoint_value` in the JSON and get `"constant_1"`, and then just return? 

```.f90
    call json%get(name, val, found)
    if (found) return
```

If the JSON looks like `"checkpoint_value": "constant_1"` then `found` will be `.true.`, right? And `"constant_1"` will be loaded into `val`? If so, won't that prevent the lookup in the registry that follows?

```fortran
    found = neko_registry%scalar_exists(name)
    if (found) then
       scalar_ptr => neko_registry%get_scalar(name)
       val = scalar_ptr
    else
       val = default
    end if
```

And if `"case.checkpoint_value"` does not exist in the JSON, then that same string (stored in `name`) will be looked up in the registry. In any case it is always `name` that is looked up in the registry, so I don't see where the string `"constant_1"` would play a role here.

timfelle: So I think the subtlety here lies in the type of `val`. If you ask the `json%get` for a real under `name` then it will return true if a real value was found. If the entry exists, but the value is not a real value, then it will return false.

And you are right that there is a bug in the lookup of the registry. It should indeed not look for name, but rather get the value at name and look for that value. Good spotting.

vbaconnet: > If the entry exists, but the value is not a real value, then it will return false.

Ah that makes more sense.. thanks

timofeymukha: @vbaconnet So, the logic is

```
    ! Try to find a real for the keyword name in the json
    ! e.g. name="checkpoint_value" and we search for "checkpoint_value" : 1"
    call json%get(name, val, found)
    if (found) return ! If we found a real, we are done, val contains the value

    ! Try again to find the same keyword, but with a value, which is a string. It must exist, or we fail (no default here).
    ! "checkpoint_value" : "output_freq"
    call json_get(json, name, reg_name)
    
   ! reg_name now contains the name of the scalar to lookup in the registry.
    val = neko_const_registry%get_scalar(reg_name)
```

vbaconnet: 🤯 amazing!

src/registries/registry_entry.f90
```diff
@@ -44,26 +44,35 @@ module registry_entry
...
+     type(field_t), private, pointer :: field_ptr => null()
 
    contains
-     procedure, pass(this) :: init_field => init_register_field
+     ! Constructors
+     procedure, pass(this) :: init_scalar => init_register_scalar
      procedure, pass(this) :: init_vector => init_register_vector
      procedure, pass(this) :: init_matrix => init_register_matrix
+     procedure, pass(this) :: init_field => init_register_field
+     !> Destructor
```
timfelle, line 68: I think we can simplify things quite a bunch by doing a generic init
```suggestion
     generic :: init => init_scalar, init_vector, init_matrix, init_field
     !> Destructor
```

src/registries/registry_entry.f90
```diff
@@ -44,26 +44,35 @@ module registry_entry
...
+     !> Name of the registry entry
      character(len=:), private, allocatable :: name
+     !> Type of the registry entry; must be supproted.
      character(len=:), private, allocatable :: type
+     !> Whether the entry is allocated
      logical, private :: allocated = .false.
 
-     type(field_t), private, pointer :: field_ptr => null()
+     ! Storage. Only one of these will be allocated at a time.
+     real(kind=rp), private :: scalar = 0.0_rp
```
timfelle: Do we want to support integers as well? I think a lot of case usage would be integers as well.

src/registries/registry_entry.f90
timofeymukha: I am a bit scared that down the road we will add more types and the type constructor signatures might clash at some point.

src/registries/registry_entry.f90
timofeymukha: Good question, I kind of assumed that scalars will effectively act as any number for my purpose here, but maybe one should split... `real_scalar` and `integer_scalar`?

timfelle: I would say it is quite unsafe to convert between integers and reals more than you absolutely have to. Sure we can call them that.

src/registries/registry_entry.f90
timfelle: Sure, I could see an integer vector or something be interesting at some point where I guess you would also just give a size.

Also the vector and scalar_integer would indeed clash as well.

src/registries/registry_entry.f90
timofeymukha: What I added now is checking that conversion did not produce any rounding, keeping for now just the real scalars in the registry. With the recent pr, json_utils also check that conversion is smooth.

timofeymukha:
Alright, I am removing "don't merged". Some key updates since the draft

- Renamed the new json utils, added handling of sp / dp. Routines combined into generic interfaces, the same way as the core json utils. A detail is that we have to convert from rp to dp or sp when we extract the scalar from the registry, which always stores in rp. But I guess that is fine.
- Handle real -> integer conversion the same way as the core json_utils.
- Renamed `registered_data` to `constants`
- Introduced `neko_const_registry`, where `case_t` put all the parsed data. So, `neko_registry` is not touched by this anymore.
- Added documentation
- Now using the new json routines for all real, integer, and array json entries. Admittedly, for a lot of those, one would never use the new functionality. But I think it is just easier for both us and the users, that the possibility is there for all the parameters, without having to track, which support it and which not.
- Updated hemi to use the new functionality: it fits nicely with the double freestream_velocity entries in the Blasius bc and ic.
- Added more unit tests.

doc/pages/user-guide/case-file.md
```diff
@@ -89,9 +91,77 @@ but also defines several parameters that pertain to the simulation as a whole.
...
+The `constants` object allows the user to define parameters that are global to
+the case file, and can be referred to when setting the values of other
+parameters. Two types of parameters can be defined: scalars and arrays. Each is
+represented as an array in the `constants` object. In turn, each object inside
+these arrays has a `name` and `value` entry. Here is an example:
+
+```json
+"constants":
+{
+  "scalars":
```
timfelle: Can it not just be a combined list instead of split between scalars and arrays?

src/registries/registry_entry.f90
timfelle: But that means a register still cannot contain an integer. Why do you not add it on the registry level instead of the json_utils level. It could be useful to store integers programatically as well as through the json system.

doc/pages/user-guide/case-file.md
timofeymukha: Actually yeah, it seems like it can. And then the integer thing becomes easier also... Let me play a bit

timofeymukha: The trouble is that json_fortran cannot distinguish between 1.0 and 1. So we if we do not provide the type by the name of the sub-entry, like integer_scalalrs: [ ], then there is no way of telling whether to add a number like 1.0 / 1 to the registry as an integer or as a real.

timofeymukha: But I like the idea of removing "scalars" and "arrays", and this we can definitely have. But I think intergers will have to be registered as reals in the registry :(.

timfelle: No. They have to be reals in the json file. The registry could easily be written to support integers.
I can give it a try and do a PR against yours if you want it.

timofeymukha: Give me a few mins :-)

doc/pages/user-guide/case-file.md
```diff
@@ -38,8 +39,9 @@ multiple scalar fields, the `name` property of each scalar field is used to
 identify the scalar field in the user file, defaulted to `s_1, s_2, ...`.
 
 The `version` keyword is reserved to track changes in the format of the file.
-The subsections below we list all the configuration options for each of the high-level objects.
-Some parameters will have default values, and are therefore optional.
+The subsections below we list all the configuration options for each of the
```
```suggestion
The subsections below list all the configuration options for each of the
```

doc/pages/user-guide/case-file.md
```diff
@@ -80,7 +82,7 @@ but also defines several parameters that pertain to the simulation as a whole.
 | `output_partitions`   | Whether to write a `partitions.vtk` file with domain partitioning.                                    | `true` or `false`                               | `false`       |
 | `output_checkpoints`  | Whether to output checkpoints, i.e. restart files.                                                    | `true` or `false`                               | `false`       |
 | `checkpoint_control`  | Defines the interpretation of `checkpoint_value` to define the frequency of writing checkpoint files. | `nsamples`, `simulationtime`, `tsteps`, `never` | -             |
-| `checkpoint_value`    | The frequency of sampling in terms of `checkpoint_control`.                                           | Positive real or integer                        | -             |
+| `checkpoint_value`    | The frequency of sampling in terms of `checkpoint_control`.                                           | Positive real or integer,                       | -             |
```
```suggestion
| `checkpoint_value`    | The frequency of sampling in terms of `checkpoint_control`.                                           | Positive real or integer                        | -             |
```

doc/pages/user-guide/case-file.md
timofeymukha: Alright, add integer_scalar to the registry. But currently unused. The gist of the issue is the part in case.f90 when we parse the constants array and add things to the registry

```fortran
          ! Try to find an array
          call json_subdict%get('value', real_vals, found)
          if (found) then
             call neko_const_registry%add_vector(size(real_vals), &
                  trim(string_val))
             vec => neko_const_registry%get_vector(trim(string_val))
             vec%x = real_vals
             ! Try to find a single value
          else

             !! How to distinguish integers here? Using integer_val will check that the value is a "valid" integer,
             !! i.e. 0 after the decimal point. But and intry with value like 1.0 is valid as both real and integer, so what type of
             !! scalar do we register?
             call json_get(json_subdict, 'value', real_val)
             call neko_const_registry%add_real_scalar(real_val, trim(string_val))
          end if
```

timfelle: A solution would be to add a `type` keyword in the json which is optional, but can be used to declare the type explicitly. If that keyword is not found, then we interpret the values as reals.

examples/rayleigh_benard_cylinder/rayleigh.case
```diff
@@ -12,6 +12,31 @@
...
+            "scalars":
+            [
+                {
+                    "name": "c1",
+                    "value": 1
+                },
+                {
+                    "name": "c2",
+                    "value": 2
+                }
```
vbaconnet: Are these actually used in the rayleigh example? Or is it just to test when printing the contents of the registry? If this is still your a toy case for this PR you can disregard my comment :)

examples/rayleigh_benard_cylinder/rayleigh.f90
```diff
@@ -81,6 +81,9 @@ subroutine initial_conditions(scheme_name, fields)
     real(kind=rp) :: rand, r,z
     type(field_t), pointer :: s
 
+
+    call neko_registry%print_contents()
```
vbaconnet: Any particular reason for this, or just to be verbose for the user?

src/common/case_file_utils.f90
```diff
@@ -0,0 +1,348 @@
...
+    if (found) return
+
+    ! Try to find a string. It must exist
+    call json_get(json, name, reg_name)
+
+    ! Retrieve the value from the registry
+    val = real(neko_const_registry%get_scalar(reg_name), kind=dp)
+  end subroutine json_get_or_lookup_double
+
+!> Retrieves an intiger either from the json or from the corresponding scalar
```
vbaconnet:
```suggestion
!> Retrieves an integer either from the json or from the corresponding scalar
```

src/common/case_file_utils.f90
```diff
@@ -0,0 +1,348 @@
...
+    end if
+  end subroutine json_get_or_lookup_or_default_double
+
+!> Retrieves an integer either from the json or from the corresponding scalar
+!! in the `neko_registry`, otherwise sets the default value.
+!! @details First tries to retrieve the value from the JSON, looking for an int
+!! entry under the given name. If not found, it looks for the same entry, but
+!! with a string value. If this fails, the default value is assigned. Otherwise,
+!! the retrieved string is the name looked up in the `neko_const_registry` as a
+!! scalar, and the data is copied from there, using explicit convresionto an
```
vbaconnet:
```suggestion
!! scalar, and the data is copied from there, using explicit conversion to an
```

src/common/case_file_utils.f90
```diff
@@ -0,0 +1,348 @@
...
+
+    ! Try to find a string. It must exist
+    call json_get(json, name, reg_name)
+
+    ! Retrieve the array from the registry
+    vec_ptr => neko_const_registry%get_vector(reg_name)
+    val = real(vec_ptr%x, kind=dp)
+  end subroutine json_get_or_lookup_double_array
+
+!> Retrieves ain int array either from the json or from the corresponding vector
```
vbaconnet:
```suggestion
!> Retrieves an int array either from the json or from the corresponding vector
```

src/common/case_file_utils.f90
```diff
@@ -0,0 +1,348 @@
...
+    vec_ptr => neko_const_registry%get_vector(reg_name)
+    val = int(vec_ptr%x)
+
+    do i = 1, size(val)
+       if (.not. abscmp(real(val(i), kind=rp), vec_ptr%x(i))) then
+          call neko_error("json_get_or_lookup_integer_array: " &
+               // "Value retrieved from registry entry '" // reg_name &
+               // "' is not an integer array.")
+       end if
+    end do
```
vbaconnet, line 401: Does this not work?
```suggestion
    do i = 1, size(val)
       test_integer_conversion(val(i), vec_ptr%x(i))
    end do
```

src/case.f90
```diff
@@ -168,6 +173,33 @@ subroutine case_init_common(this)
...
+
+          ! Try to find an array
+          call json_subdict%get('value', real_vals, found)
+          if (found) then
+             call neko_const_registry%add_vector(size(real_vals), &
+                  trim(string_val))
+             vec => neko_const_registry%get_vector(trim(string_val))
+             vec%x = real_vals
+             ! Try to find a single value
+          else
```
vbaconnet: Just to be picky...
```suggestion
          else
             ! Try to find a single value
```

vbaconnet (approved):
This is definitely a huge chunk of work... and a great one as well! Excited to start testing this out myself.

src/common/case_file_utils.f90
timofeymukha: No. But that routine had to go anyway now :)

timofeymukha:
@timfelle @vbaconnet update:

- Strict type checking on for json parsing.
- Adaptions required in simcomp and fluid IO / compute_value handling, see code here.
- Changes in json_utils and case_file_utils to check types via this `var_type` thing you will see. Basically we can use json_fortran to inquire the type. A lot of new error messages based on that. 
- Updated examples case files to handle strict typing.
- Integer scalar entires now possible in registry
- `constants` structure flattened: now an array of objects, no "arrays" and "scalars" subdivision.

Gotta add more tests for the json routines. Thanks for baring with me through the updates!

src/registries/registry.f90
```diff
@@ -84,20 +89,38 @@ module registry
      !> Get pointer to a stored matrix by index.
      procedure, pass(this) :: get_matrix_by_index => &
           registry_get_matrix_by_index
+     !> Get pointer to a stored real scalar by index.
+     procedure, pass(this) :: get_real_scalar_by_index => &
+          registry_get_real_scalar_by_index
+     !> Get pointer to a stored integer scalar by index.
+     procedure, pass(this) :: get_integer_scalar_by_index => &
+          registry_get_integer_scalar_by_index
```
timfelle: I am always a bit confused by these. Do we really forsee anyone using index based lookups? you can never know what the index is.

timofeymukha: I am have no problem killing these. I agree that it is borderline impossible to make use of them.

timofeymukha: Removed.

njansson (approved):

src/common/case_file_utils.f90
```diff
@@ -0,0 +1,404 @@
+! Copyright (c) 2025, The Neko Authors
```
timfelle:
```suggestion
! Copyright (c) 2026, The Neko Authors
```

src/common/case_file_utils.f90
```diff
@@ -0,0 +1,404 @@
...
+    call json_get(json, name, reg_name)
+
+    ! Retrieve the value from the registry
+    val = real(neko_const_registry%get_real_scalar(reg_name), kind=sp)
+  end subroutine json_get_or_lookup_real
+
+!> Same as `json_get_or_lookup_real`, but for double precision.
+!! @param[inout] json The json to retrieve the parameter from.
+!! @param[in] name The full path to the parameter.
+!! @param[out] value The variable to be populated with the retrieved parameter
```
timfelle: Missing indentation. Should be level with subroutine

src/common/case_file_utils.f90
timfelle: Please add indentation to function descriptions

timfelle (approved):

PR #2229, opened 2025-12-10
added a zero pressure gradient tripped turbulent boundary layer example

Mr-2c:
the mesh is a bit large, so let me know if it is appropriate as an example. also the trip.f90 is basically from @vbaconnet from a code from @MartinKarp based on the original code from @adampep, so not sure how the acknowledgment works here, but I added all their names to the trip.f90 file

### DIFF
--- a/examples/ZPG_trippedTBL/README.md
+++ b/examples/ZPG_trippedTBL/README.md
@@ -1,0 +1,30 @@
+# Flow
+This example is a zero pressure gradient (ZPG) turbulent boundary layer (TBL)
+that starts from a laminar Blasius profile which is then tripped about 10*delta99
+downstream of the inlet.
+
+The inspiration for this example comes from Schlatter & Örlü, JFM, 2010
+The mesh is a shorter version of the computational domain used in that work.
+
+Compile using:
+
+```
+makeneko trip.f90 TBL_ZPG.f90
+```
+
+The case file is setup to include all the important parameters, but is not
+ready for a production run, and neither is the mesh.
+The resolutions should be okay for a reasonable simulation but the
+streamwise extent needs to be extended if the goal is to reach Re_theta=4000
+as explianed in the TBL_ZPG.f90
+
+NOTE: the z averaging may not work with FP32 (--enable-real=sp)
+
+
+# Tripping
+Introduces a volume forcing along a line parallel to the y or z direction.
+
+`trip.f90` implements the `trip` module to be used in the user file. See instructions
+in the header of the file.
+
+Make sure to enable the user source term in the case file as well.
--- a/examples/ZPG_trippedTBL/README.md
+++ b/examples/ZPG_trippedTBL/README.md
@@ -14,11 +14,23 @@ makeneko trip.f90 TBL_ZPG.f90
 
 The case file is setup to include all the important parameters, but is not
 ready for a production run, and neither is the mesh.
-The resolutions should be okay for a reasonable simulation but the
-streamwise extent needs to be extended if the goal is to reach Re_theta=4000
-as explianed in the TBL_ZPG.f90
 
-NOTE: the z averaging may not work with FP32 (--enable-real=sp)
+The current mesh was generated using genmeshbox with the following command:
+
+```
+genmeshbox 0.0 162.5 0.0 65.0 0.0 78.0 96 64 112 .false. .false. .true.
+```
+
+The resolutions should be okay for a reasonable simulation 
+(wall-normal resolution is even too fine)
+but the streamwise extent needs to be extended if the goal is to reach Re_theta=4000
+this can be done with the following genmeshbox command:
+
+```
+genmeshbox 0.0 1950.0 0.0 65.0 0.0 78.0 1152 64 112 .false. .false. .true.
+```
+
+IMPORTANT NOTE: the z averaging in statistics may not work with FP32 (--enable-real=sp)
 
 
 # Tripping
--- a/examples/ZPG_trippedTBL/TBL_ZPG.case
+++ b/examples/ZPG_trippedTBL/TBL_ZPG.case
@@ -1,0 +1,98 @@
+{
+"version": 1.0,
+"case": {
+//    "restart_file": "previous.chkp",
+    "mesh_file": "box.nmsh",
+    "output_at_end": true,
+    "job_timelimit": "03:50:00",
+    "output_boundary": true,
+    "output_checkpoints": false,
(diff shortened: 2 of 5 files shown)


examples/ZPG_trippedTBL/README.md
```diff
@@ -0,0 +1,30 @@
+# Flow
+This example is a zero pressure gradient (ZPG) turbulent boundary layer (TBL)
```
timofeymukha, line 2: @Mr-2c just two initial things:
- Remove the .log
- Remove the mesh and instead add the genmeshbox command to genereate it to the readme. If you are using a file to prescribe the element size distribution, add those.

examples/ZPG_trippedTBL/TBL_ZPG.f90
```diff
@@ -0,0 +1,333 @@
...
+       y = ymax * &
+        (tanh(gamma*y)-tanh(gamma*y_shift)) / &
+        (tanh(gamma*(1+y_shift))-tanh(gamma*y_shift))
+       msh%points(i)%x(2) = y
+    end do
+
+  end subroutine user_mesh_scale
+
+!=============================================================================!
+! this makes a "fake" sine-shaped profile for the inflow as a proxy for the Blasius profile
```
timofeymukha, line 217: We cannot use the built-in Blasius?

examples/ZPG_trippedTBL/TBL_ZPG.f90
```diff
@@ -0,0 +1,333 @@
...
+      w%x(msk_ind,1,1,1) = 0.0_rp
+      if (y.lt.1.0_rp) then
+        u%x(msk_ind,1,1,1) = sin(y*PI/2)
+      end if
+    enddo
+
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_memcpy(u%x, u%x_d, u%size(), HOST_TO_DEVICE, sync=.false.)
+       call device_memcpy(v%x, v%x_d, v%size(), HOST_TO_DEVICE, sync=.false.)
+       call device_memcpy(w%x, w%x_d, w%size(), HOST_TO_DEVICE, sync=.false.)
```
timofeymukha: Just use 
call u%copy_from(HOST_TO_DEVICE, sync = .false.)

etc. No if needed

examples/ZPG_trippedTBL/TBL_ZPG.case
```diff
@@ -0,0 +1,98 @@
...
+            },
+            "projection_space_size": 0,
+            "absolute_tolerance": 1e-7,
+            "max_iterations": 20
+        },
+        "pressure_solver": {
+            "type": "gmres",
+            "preconditioner": {
+                    "type": "hsmg"
+            },
```
timofeymukha, line 46: Could you test how phmg performs here? It would be good to have an example with it!

examples/ZPG_trippedTBL/TBL_ZPG.f90
njansson: As long as the mesh is axis aligned, built-in "should work"

examples/ZPG_trippedTBL/README.md
njansson: Unless the mesh is huge, can't we store both the compressed nmsh and the genmeshbox command?

examples/ZPG_trippedTBL/TBL_ZPG.case
Mr-2c: I somehow want this to be plug and play, so they just take the example and run it and have a DNS of ZPG boundary layer, so I am a bit against adding not fully tested stuff, if that okay? perhaps we can update the turbulent channel flow with that?

examples/ZPG_trippedTBL/README.md
Mr-2c: > @Mr-2c just two initial things:
> 
> * Remove the .log

I somehow wanted to have the .log so the users can see how the mesh was created. but I can remove it if it is inconstant with our guidelines

timfelle: You can easily just add that to the readme as well. Something like:

To generate the mesh use `some command`
The expected output should be
```
xmin, xmax, Nel, periodic:    0.000000   162.500000   96 F
ymin, ymax, Nel, periodic:    0.000000    65.000000   64 F
zmin, zmax, Nel, periodic:    0.000000    78.000000  112 T
```

Mr-2c: great idea, will do

examples/ZPG_trippedTBL/TBL_ZPG.f90
Mr-2c: I will add a comment to emphasize this but will somehow want to make as few changes as possible now, unless you think it is important

examples/ZPG_trippedTBL/README.md
Mr-2c: > Unless the mesh is huge, can't we store both the compressed nmsh and the genmeshbox command?

I am adding the genmeshbox command to the readme file. is that sufficient?

examples/ZPG_trippedTBL/TBL_ZPG.f90
Mr-2c: done

examples/ZPG_trippedTBL/TBL_ZPG.case
```diff
@@ -0,0 +1,98 @@
...
+        "dealias": true
+    },
+    "fluid": {
+        "scheme": "pnpn",
+        "rho": 1,
+        "mu": 0.00070,
+        "initial_condition": {
+            "type": "user"
+//            "type": "field",
+//	    "file_name": "field0.f00008"
```
timfelle, line 31: Could we clean this up for the final state of the PR.

examples/ZPG_trippedTBL/README.md
Mr-2c: > Unless the mesh is huge, can't we store both the compressed nmsh and the genmeshbox command?

the full mesh is 1.8GB before compression and 253MB after compression, so I think it cannot be included. the current smaller one just let me know whether you want to keep to remove, I am fine either way...

timfelle: I think it would be best to just remove the mesh and let the user generate it themselves. Especially if you need to compress the mesh to include it.

PR #2234, opened 2025-12-15
Add missing end_section

njansson:
This fixes the logging for the compressible solver, which was of compared to the incompressible one


timofeymukha (approved):

timfelle:
I am quite surprised this works.
The `neko_log%free()` procedure should return an error stating that the log is unbalanced if any of these are missing

timfelle:
I see, the logger is never freed...

timfelle (approved):

PR #2235, opened 2025-12-17
Fix mean_output_t initilization

timofeymukha:
Asked AI why it thinks we get this weird error with user_stats.csv failing to appear randomly in the pytest CI, and it directly found this very sad bug. 

`this%start_time` was not initialized, so randomly the check of time > this%start_time would fail...

We should not have let that error slide like that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Changelog
+
+## Develop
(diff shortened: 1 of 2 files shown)


timfelle:
Lets also add values in the class definition as well. which will also make sure this does not happen.

njansson (approved):

timfelle (approved):

PR #2236, opened 2025-12-18
Fix HDF5 bug

timfelle:
Address a bug in the HDF5 handling related to field assignment by ensuring the name is only assigned when it is empty.


njansson (approved):
🍻

nick-morse (approved):

PR #2240, opened 2025-12-22
Reduce data movement and launch latency in compressible scheme

njansson:
This PR refactors the compressible euler's step function. The transition from `field_math` to backend specific kernels were necessary to allow for better vectorisation (CPU) and reduced kernel launch latency (accelerators). Data movement has also been reduce for all backends.


tuananhdao:
Thank you! This will remove bunch of kernel launches.

tuananhdao (approved):

PR #2244, opened 2026-01-02
Feature/ale

arash-mp:
## Summary
This PR implements the Arbitrary Lagrangian-Eulerian (ALE) formulation to support moving mesh simulations in Neko. It introduces a new "ale_manager" to handle grid velocities and mesh deformation, modifies "fluid_pnpn" to include grid velocity terms, and updates the IO and checkpoints to support ALE simulations.

### Core & Solver
* **New Module:** Added "src/ale/" containing "ale_manager.f90" (handles mesh motion logic) and core definitions.
* **Solver Update:** Modified "fluid_pnpn.f90" and "rhs_maker" to incorporate ALE advection terms (fluid velocity relative to mesh velocity).
* **BC type:** Implements a new "moving_boundary" boundary condition (**field_moving.f90**).

### IO & Checkpointing
* Updated "checkpoint.f90" and "chkp_file.f90" to save/restart simulations with moving meshes.

### Examples
* Added two new examples in "examples/":
    * "ocyl_cylinder3D": A cylinder hinged from the top and oscillate like a pendulum.
    * "ocyl_ellipse3D": A moving ellipse that simultaneously rotates around its axis.

## ToDo:
* GPU implementation (Currently CPU only).
* OIFS time-integration support.

## References:
Implementation is based on:
1- Fischer, P., Schmitt, M. and Tomboulides, A., 2017. Recent developments in spectral element simulations of moving-domain problems. Recent progress and modern challenges in applied mathematics, modeling and computational science, pp.213-244.
2- Ho, L.W., 1989. A Legendre spectral element method for simulation of incompressible unsteady viscous free-surface flows (Doctoral dissertation, Massachusetts Institute of Technology).

ocyl_ellipse_example


src/fluid/fluid_pnpn_bc_fctry.f90
```diff
@@ -160,6 +164,19 @@ module subroutine velocity_bc_factory(object, scheme, json, coef, user)
        allocate(inflow_t::object)
     case ("no_slip")
        allocate(zero_dirichlet_t::object)
+    case ("moving_boundary")
+      if (.not. neko_registry%field_exists('wm_x')) then
+          call neko_error("'moving_boundary' found in the BC list&
+                          &, but ALE is not active! Set 'ale_active = true' in &
+                          &case file.")
+      end if
```
timofeymukha: Just an initial comment. This moving_boundary is just a no-slip condition underneath, right? Your new bc just does a masked copy of the mesh velocity. What I don't fully understand is how the setup of the motion in the ALE is coupled to the selection of the boundary that moves. If it is the boundary that moves, is it not more logical to prescribe the motion in the JSON dictionary for the bc? What if you have more than one moving_boundary in the current setup, does it work?

arash-mp: The "moving_boundary" is similar to a Dirichlet boundary condition, but it is also used to mark boundaries that are allowed to move.

In the current implementation, a single mesh-velocity field is defined over the entire domain, assigning a velocity to every grid point. THen, the filed is multiplied by a scalar field called phi. So, we need to identify the moving_boundary regions for constructing the correct phi field.

The field phi takes values between 0 and 1. It is equal to 1 on boundaries marked as moving_boundary and 0 on all other non-periodic boundaries. When the mesh-velocity field is multiplied by phi, only the boundaries marked as moving are assigned the prescribed velocity, while all other boundaries remain fixed. Also, this method presereve the mesh quality near the wall, so mesh deformation happes mostly away from the wall.

In ale_manager, the subroutine solve_base_displacement computes the phi field by solving a Laplace equation once during initialization (in practive, we can recompute it also every N iterations, if needed). This results in a smooth spatial transition of phi from the moving boundaries into the interior of the domain.

- We can have multiple moving boundaries at the same time. The Laplace solution provides a smooth blending of motion throughout the domain. However, in the current implementation, all moving objects share the same prescribed velocity. With some hack (like defining the base mesh velocity based on coordinates to allow different object mvoe with different velocities) we can overcome this, but ofc not the best solution. 
I will think about better ways to do this also.
Also, tbh, for most practical cases I think one would have 1 moving object, but it's good to make it more flexible for several moving objects for sure.

arash-mp: Also, at moving_boundary,  we need to set the fluid velocity vector equal to mesh velcoity (to have no-slip BC). that's what masked_copy does here. It's like enforcing the kinematics boundary condition.

arash-mp: Also one way (maybe naiive, should think more about it) for generalization into several moving objects is using super position, such that we find 1 phi field for each moving body, such that phi=1 only on that body and 0 on other moving bodies, then we sum them up. and then I can set seperate motion for any number of bodies in JSON.

timofeymukha: Thanks for the clarification, Arash, now I understand much better! I think it is totally fine to leave the mesh motion working as you say for this initial PR. For the boundary conditions, it feels like a better way would be not introducing a new condition, but rather having a `moving : true / false` keyword added to ordinary boundary conditions and letting them handle that situation. In the beginning we can just add this to the no_slip condition. But later on it is probably nice if other conditions can also handle a moving boudnary, and adding a `movng_*` for each sounds not very practical. I am not 100% sure, but it seems this can be done rather easily, or? The only thing you need is to fetch the grid velocity inside the bc. Think about it, and I will also take a bit of a look.

arash-mp: I see the point. I think it's straight forward to do this. Will give it a shot today.

timofeymukha: What we do now for no-slip is
```
allocate(zero_dirichlet_t::object)
```
in the bc factory. I don't think we should add anything to zero_dirichlet_t. That type is anyway mostly meant as a utility for masks etc. I think what you can do is rename the new bc you added here to `no_slip`, and allocate that one instead of the zero_dirichlet. Then, you pretty much do exactly the same as what you already do in the moving_boundary currently, but just add treatment of the ordinary no-slip when there is no mesh movement.

njansson:
Great stuff, I think it would be good to try to get this merged as soon as possible before it starts to lag behind `develop` too much

adperezm:
@arash-mp, is this one ready for review? Also check the failing tests.

arash-mp:
> @arash-mp, is this one ready for review? Also check the failing tests.

@adperezm  Yes it's ready.
GPU part still is not done, and the failing tests I should check and ask how to fix them:D
But I think it's good to start reviewing it as it may take a lot of time!

src/ale/ale_rigid_kinematics.f90
```diff
@@ -0,0 +1,336 @@
...
+              kinematics%vel_ang(i) = w_scalar * amp_scalar * &
+                   cos(w_scalar * time%t)
+           end if
+        end do
+     end select
+
+  end subroutine compute_body_kinematics_built_in
+
+  !> Updates pivot location
+  subroutine update_pivot_location(pivot, pivot_loc, pivot_vel, time, &
```
adperezm, line 291: Why is this sort of thing not part of the type pivot_t?

Shouldnt that class just have a "update" procedure. Or a step procedure, etc.

src/ale/ale_rigid_kinematics.f90
```diff
@@ -0,0 +1,336 @@
...
+     call ab_integrate_point_pos(tracker%pos, tracker%vel_lag, &
+          current_vel, time, nadv)
+
+  end subroutine advance_point_tracker
+
+  !> Compute built-in kinematics for a body. Uses inputs from JSON.
+  !> CPU-only.
+  subroutine compute_body_kinematics_built_in(kinematics, body_conf, time)
+     type(body_kinematics_t), intent(out) :: kinematics
+     type(ale_body_t), intent(in) :: body_conf
```
adperezm: I have not read in detail what happens here. 

But just looking at the types. I feels as if there should be an ale_body_t. That body follows certain kinematics, so in the end, body_kinematics_t should be part of it or something like that, no?

src/ale/ale_manager.f90
```diff
@@ -0,0 +1,1710 @@
...
+    ! Update internal history of mesh velocity
+    call this%wm_x_lag%update()
+    call this%wm_y_lag%update()
+    call this%wm_z_lag%update()
+  end subroutine advance_mesh
+
+  ! Compute mesh stiffness with per-body gain/decay from stiff_geom
+  subroutine compute_stiffness_ale(coef, params)
+    type(coef_t), intent(inout) :: coef
+    type(ale_config_t), intent(in) :: params
```
adperezm, line 1153: Why was it needed to create a new type of type that handles parameters?

src/bc/field_moving.f90
```diff
@@ -0,0 +1,172 @@
...
+    real(kind=rp), intent(inout), dimension(n) :: x, y, z
+    type(time_state_t), intent(in), optional :: time
+    logical, intent(in), optional :: strong
+    logical :: strong_
+
+    strong_ = .true.
+    if (present(strong)) strong_ = strong
+
+    if (strong_) then
+       if (.not. associated(this%wx) .or. .not. associated(this%wy) .or. &
```
adperezm: Is there another way to check if the velocity is set?

src/bc/field_moving.f90
```diff
@@ -0,0 +1,172 @@
...
+  ! Todo
+  !-----------------------------------------------------------------------------
+  
+  subroutine field_moving_apply_scalar(this, x, n, time, strong)
+    class(field_moving_t), intent(inout) :: this
+    integer, intent(in) :: n
+    real(kind=rp), intent(inout), dimension(n) :: x
+    type(time_state_t), intent(in), optional :: time
+    logical, intent(in), optional :: strong
+    call neko_error("field_moving cannot apply scalar BCs")
```
adperezm: Just to confirm Is this type of bc only applicable to velocity

If so one can try to throw an error even earlier and prevent a user to try to set this type of bc for pressure or scalar.

src/fluid/fluid_pnpn_bc_fctry.f90
```diff
@@ -164,7 +167,7 @@ module subroutine velocity_bc_factory(object, scheme, json, coef, user)
     case ("velocity_value")
        allocate(inflow_t::object)
     case ("no_slip")
-       allocate(zero_dirichlet_t::object)
+       allocate(no_slip_t::object)
```
adperezm, line 178: Do you use the field_moving bc that you also added? Since it is not in the factory, or am I missing this?

src/fluid/fluid_pnpn.f90
```diff
@@ -100,8 +101,10 @@ module fluid_pnpn
      !! respect to the previous time-step.
      type(field_t) :: dp, du, dv, dw
 
-     !
-     ! Implicit operators, i.e. the left-hand-side of the Helmholz problem.
+     !> ALE Manager 
+     type(ale_manager_t) :: ale
```
adperezm, line 105: Adding this here does not include more memory requirements right? Unless it is initialized etc.

src/fluid/fluid_pnpn.f90
```diff
@@ -718,15 +752,32 @@ subroutine fluid_pnpn_step(this, time, dt_controller)
...
               rho%x(1,1,1,1), ext_bdf%advection_coeffs, n)
 
          ! Add the RHS contributions coming from the BDF scheme.
+         ! Blag and Blaglag are history of B matrices, mainly used for ALE.
+         ! For a normal simulation (no moving mesh), Blag and Blaglag 
+         ! are just the initial B matrix, filled at initialization.
          call makebdf%compute_fluid(ulag, vlag, wlag, f_x%x, f_y%x, f_z%x, &
               u, v, w, c_Xh%B, rho%x(1,1,1,1), dt, &
-              ext_bdf%diffusion_coeffs, ext_bdf%ndiff, n)
+              ext_bdf%diffusion_coeffs, ext_bdf%ndiff, n, &
```
adperezm: I do not like this. If possible I would make the lag b optional and have a clear if statement separating the normal case and the ale one. 

I believe that we need to keep the normal workflow of pnpn as clean as possible if we can't create a new scheme.

But I guess there's multiple opinions on this type of thing. So you can also just wait for others to make comments :)

src/fluid/fluid_pnpn.f90
```diff
@@ -718,15 +752,32 @@ subroutine fluid_pnpn_step(this, time, dt_controller)
...
+              c_Xh%Blag, c_Xh%Blaglag)
+
+      end if
+
+      if (this%ale%active) then
```
adperezm, line 781: Can't all of this be done in one line if you wrap it inside a routine in the ale object?

Like have ale be pointing to these objects and update them inside. I feel. This comment would apply to other instances of (if ALE).


If you have an object that is called ale manager, then I would hide everything inside of it. Have a pre solve stage and an after solve stage.

src/fluid/fluid_pnpn.f90
```diff
@@ -859,7 +910,12 @@ subroutine fluid_pnpn_step(this, time, dt_controller)
               this%ksp_vel, this%pc_prs, this%pc_vel, this%ksp_prs%max_iter, &
               this%ksp_vel%max_iter)
       end if
-
+      
+      ! Update mesh velocities for ALE
+      ! We update them here (end of step) for the next step.
+      ! Returns if .not. ale.
+      call this%ale%update_mesh_velocity(c_Xh, time, ext_bdf%nadv)
```
adperezm: Yes, my previous comment reference something like this. Could you have an ale%update metrics or so?

src/fluid/fluid_pnpn.f90
```diff
@@ -879,11 +935,21 @@ subroutine fluid_pnpn_setup_bcs(this, user, params)
...
     logical, allocatable :: marked_zones(:)
     integer, allocatable :: zone_indices(:)
 
+    ! For ALE, we set a flag while reading the BCs
+    character(len=:), allocatable :: bc_type_str 
+    this%ale%has_moving_boundary = .false.
+    any_moving_wall = .false.
+    ale_active_local = .false.
+    call json_get_or_default(params, &
+         'case.fluid.ale.ale_active', &
```
adperezm: If ale_active is already inside the ale group. Then just remove the ale_

src/io/fluid_output.f90
```diff
@@ -182,7 +182,7 @@ subroutine fluid_output_sample(this, t)
     class(fluid_output_t), intent(inout) :: this
     real(kind=rp), intent(in) :: t
     integer :: i
-
+    logical :: is_ale = .false.
```
adperezm: Does the fluid output type need to know if Ale is being used? I guess what it needs to know is if the mesh needs to be written or?

Then ale should just set that the mesh has to be written.

src/sem/coef.f90
```diff
@@ -47,6 +47,7 @@ module coefs
   use mxm_wrapper, only : mxm
   use device
   use utils, only : index_is_on_facet, linear_index
+  use, intrinsic :: iso_fortran_env
```
adperezm, line 55: Why this? Not really needed before

src/sem/coef.f90
```diff
@@ -159,6 +163,9 @@ module coefs
      procedure, pass(this) :: free => coef_free
      procedure, pass(this) :: get_normal => coef_get_normal
      procedure, pass(this) :: get_area => coef_get_area
+     procedure, pass(this) :: update_metrics => coef_update_metrics
+     procedure, pass(this) :: enable_B_history => coef_enable_lagged_mass
+     procedure, pass(this) :: update_B_history => coef_update_lagged_mass
```
adperezm, line 174: I would somehow remove the whole lagged mass from coef. I would put it as part of ale too.

The main thing is, what is a coef object? 

It Is an object that keeps the geometrical data for a given mesh/dof.

The lagged stuff is not a member of such a thing. It is something that is needed by ale.

The correct way for me to use this object with lags. Is to in fact have lagged coef. But of course that would require A LOT of memory.

I see why you do it like this, but for generality and correct use of the objects I really suggest to move this out.

src/sem/coef.f90
```diff
@@ -268,6 +275,10 @@ subroutine coef_init_all(this, gs_h)
     allocate(this%B(this%Xh%lx, this%Xh%ly, this%Xh%lz, this%msh%nelv))
     allocate(this%Binv(this%Xh%lx, this%Xh%ly, this%Xh%lz, this%msh%nelv))
 
+    ! We do this so in a static simulation we don't allocate extra memory
+    this%Blag => this%B
+    this%Blaglag => this%B
```
adperezm, line 286: This is what I mean. In general, in a static simulation we do not even want to have a blag attribute .

src/sem/coef.f90
```diff
@@ -1251,5 +1285,82 @@ subroutine coef_generate_cyclic_bc(coef)
 
   end subroutine coef_generate_cyclic_bc
 
+  !> Recompute and update geometric factors (ALE)
+  subroutine coef_update_metrics(this)
```
adperezm: I would call this like recompute or regenerate. When reading pnpn, I did not get this is what this call did.

adperezm (commented):
I have now left a number of comments. 

I haven't checked all the files yet. Just going little by little.

There are many comments here that are honestly just opinions or preferences. So in the grand scheme of things they do not need to be right.

I will get back to it later. Feel free to ignore, but I hope some of the comments are useful.

src/fluid/fluid_pnpn_bc_fctry.f90
arash-mp: no, I removed field_moving now. It was for an older implementaiton. Now, everything is inside no_slip_t.

src/fluid/fluid_pnpn.f90
arash-mp: @adperezm 
exactly, it does not add memoery overhead if it is not initialized.
only 4-5 integers and logical variables

njansson:
Are we good to merge?

timofeymukha:
> Are we good to merge?

Not yet. Arash has still been actively developing, e.g. a lot of commits yesterday. So a review is still needed. 

@arash-mp one small comment here: don't commit the .msh and .nmsh files if there is a .geo. In the README to the examples, you can add the commands necessary to regenerate the mesh.

arash-mp:
> > Are we good to merge?
> 
> Not yet. Arash has still been actively developing, e.g. a lot of commits yesterday. So a review is still needed.
> 
> @arash-mp one small comment here: don't commit the .msh and .nmsh files if there is a .geo. In the README to the examples, you can add the commands necessary to regenerate the mesh.

@njansson @timofeymukha 
It's fully workig on CPU now, just a few small improvments which I will make today.
I will be on vacation for a week from tomrrow, but next week when I got back, I will add the documentation, and then it's done from my side.

arash-mp:
> one small comment here: don't commit the .msh and .nmsh files if there is a .geo. I

@timofeymukha yess sure, I'll remove the .nmsh files

examples/double_oscillating_cylinders/double_oscillating_cylinders.f90
```diff
@@ -0,0 +1,52 @@
+module user
+  use neko
+  implicit none
+
+contains
+
```
timofeymukha, line 6: Just using these lines to make comments on the examples. 

- Use lowercase letters in the example folder names
- 3D can be skipped, this is the default. If you have 2D that is worth marking since it is a special mode for Neko.
- Don't save chars on example names, spell words out, e.g.  oscilating_ellipse.
- match the base file names inside the example with the folder name. So oscillating_ellipse -> oscillating_ellipse.f90, etc.
- There is a pytest integration test, that compiles all the .f90 in the examples. Please new examples there, it should be quite straightforward. This ensures code updates never break the example user files.

timofeymukha: (I know some existing examples don't follow this, so sorry about that)

timofeymukha:
@arash-mp check the copilot's comments, they all seem reasonable at first glance.

src/fluid/fluid_pnpn_bc_fctry.f90
```diff
@@ -43,7 +43,10 @@
   use dong_outflow, only : dong_outflow_t
   use symmetry, only : symmetry_t
   use non_normal, only : non_normal_t
+  use no_slip, only : no_slip_t
   use field_dirichlet_vector, only : field_dirichlet_vector_t
```
timofeymukha, line 47: neko_registry not used

src/io/chkp_file.f90
```diff
@@ -460,6 +623,24 @@ subroutine chkp_file_read(this, data)
...
+          wm_y => data%wm_y
+          wm_z => data%wm_z
+          wm_x_lag => data%wm_x_lag
+          wm_y_lag => data%wm_y_lag
+          wm_z_lag => data%wm_z_lag
+          Blag => data%Blag
+          Blaglag => data%Blaglag
+          pivot_pos => data%pivot_pos
+          pivot_vel_lag => data%pivot_vel_lag
+          basis_pos => data%basis_pos
```
timofeymukha, line 642: Is read_ale assigned a value if this if does not execute? If not, is this guarded for somehow?

examples/double_oscillating_cylinders/double_oscillating_cylinders.case
```diff
@@ -0,0 +1,144 @@
...
+        "checkpoint_value": 5000,
+//        "restart_file": "rst.chkp",
+        "runtime_statistics":{
+            "enabled": true
+        },
+        "time": {
+            "end_time": 200.0,
+            "variable_timestep": false,
+            "timestep": 0.001,
+        },
```
timofeymukha, line 14: Remove trailing comma, this is not valid JSON, AFAIK

examples/Double_ocyl_cylinder/cylinder.case
```diff
@@ -0,0 +1,144 @@
...
+            {
+                "type": "force_torque",
+                "compute_control": "tsteps",
+                "compute_value": 1,
+                "zone_id": 7,
+                "name": "right_cyl",
+                "zone_name": "right_cylinder_pivot",
+                "center": [ 4.5, 0.0, 0.0 ],
+                "center_type": "pivot",
+            },
```
timofeymukha: Remove trailing comma. Please check other places too.

src/simulation_components/force_torque.f90
```diff
@@ -549,4 +668,102 @@ subroutine force_torque_compute(this, time)
...
+    class(force_torque_t), intent(inout) :: this
+    integer, intent(in) :: zone_id
+    character(len=*), intent(in) :: center_type
+    real(kind=rp), intent(in) :: center_in(3)
+
+    character(len=128) :: log_buf
+    integer :: i, j, nbodies, nindices, body_id
+    logical :: body_found
+    logical :: ale_active = .false.
+
```
timofeymukha: Generally, do not assign a value in the declaration like you do in ale_active, because this implicitly sets a SAVE flag on it (its state persists across subroutine calls)

src/ale/ale_manager.f90
```diff
@@ -0,0 +1,1764 @@
...
+
+     type(ale_config_t) :: config
+     logical :: has_moving_boundary = .false.
+     real(kind=rp) :: abstol = 1.0e-10_rp
+     integer :: ksp_max_iter = 10000
+     logical :: res_monitor = .false.
+     character(len=:), allocatable :: ksp_solver
+     character(len=:), allocatable :: precon_type
+     type(json_file) :: precon_params
+
```
timofeymukha: Is persisting the precon_params necessary? We generally avoid storing user input this way.

doc/pages/user-guide/case-file.md
```diff
@@ -1091,6 +1092,341 @@ The parameters for the sponge source term are summarized in the table below:
...
+
+* **Boundary Conditions:** Under the `"boundary_condition"` block, any moving wall must be set to `"no_slip"` with `"moving": true`.
+* **Enable ALE:** Under the `"ale"` block, the keyword `"enabled"` must be set to `true`. 
+
+@attention If the `"ale"` block is present, the `"enabled"` keyword is mandatory.
+
+* **Body Registration:** There must be a strict mapping between moving boundaries and ALE bodies:
+  * Any `"no_slip"` boundary marked as `"moving": true` **must** be registered as an ALE body under the `"bodies"` object.
+  * Conversely, any boundary registered as an ALE body **must** be defined as a `"no_slip"` boundary with `"moving": true`.
+  * If any of these rules are violated, Neko will not start the simulation loop and will print an informative error message.
```
timofeymukha, line 1137: Is the mapping strictly 1-to-1? Then I would explicitly say "strict one-to-one mapping" so there is no room for guessing.

src/io/chkp_file.f90
arash-mp: opsss it was missing!! it's fixed now. thanks

src/simulation_components/force_torque.f90
```diff
@@ -124,7 +140,8 @@ subroutine force_torque_init_from_json(this, json, case)
     class(case_t), intent(inout), target :: case
     integer :: zone_id
     real(kind=rp), allocatable :: center(:)
-    character(len=:), allocatable :: zone_name, fluid_name
+    character(len=:), allocatable :: zone_name, fluid_name, center_type
+    character(len=:), allocatable :: effective_center_type
```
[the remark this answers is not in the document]
arash-mp: done

src/simulation_components/force_torque.f90
```diff
@@ -310,28 +357,48 @@ subroutine force_torque_init_common(this, name, fluid_name, zone_id, &
 
     call MPI_Allreduce(n_pts, glb_n_pts, 1, &
          MPI_INTEGER, MPI_SUM, NEKO_COMM, ierr)
+    ! Calculatig avg pos here
+    avg_r(1) = glsum(this%r1%x, n_pts)/glb_n_pts
```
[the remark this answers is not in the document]
arash-mp: done

doc/pages/user-guide/case-file.md
```diff
@@ -1091,6 +1092,341 @@ The parameters for the sponge source term are summarized in the table below:
...
+
+In the follwoing, the main blocks of `"ale"` object are explained.
```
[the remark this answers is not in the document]
arash-mp: done

src/simulation_components/force_torque.f90
```diff
@@ -549,4 +668,102 @@ subroutine force_torque_compute(this, time)
...
+                   j = 1
+                   do while (j <= nindices .and. .not. body_found)
+                      if (neko_ale%config%bodies(i)%zone_indices(j) == zone_id) then
+
+                         ! Body found
+                         this%moving_center = .true.
+                         this%pivot_link => neko_ale%ale_pivot(i)
+                         this%linked_body_name = neko_ale%config%bodies(i)%name
+                         body_id = neko_ale%config%bodies(i)%id
+                         body_found = .true.
```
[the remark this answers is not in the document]
arash-mp: done

src/io/chkp_file.f90
```diff
@@ -351,6 +391,112 @@ subroutine chkp_file_write(this, data, t)
...
+          end associate
+          mpi_offset = mpi_offset + n_glb_dofs * int(MPI_REAL_PREC_SIZE, i8)
+       end do
+
+       do i = 1, wm_y_lag%size()
+          byte_offset = mpi_offset + &
+               dof_offset * int(MPI_REAL_PREC_SIZE, i8)
+          ! We should not need this extra associate block, ant it works
+          ! great without it for GNU, Intel, NEC and Cray, but throws an
+          ! ICE with NAG.
```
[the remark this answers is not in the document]
arash-mp: done

doc/pages/user-guide/case-file.md
```diff
@@ -1091,6 +1092,341 @@ The parameters for the sponge source term are summarized in the table below:
...
+  \f}
+* **Cylinder** (`"cylinder"`): 
+  Calculates the distance to the Z-axis passing through the center \f$ (c_x, c_y) \f$.
+
+  \f{eqnarray*}{
+   r = \sqrt{(x - c_x)^2 + (y - c_y)^2}.
+  \f}
+
+* **Wall Distance** (`"cheap_dist"`): 
+  \f$ r \f$ is assigned from a precomputed pusodo distance field based on the boundary `zone_indices`.
```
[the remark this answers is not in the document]
arash-mp: done

examples/ocyl_cylinder3D/README.md
```diff
@@ -0,0 +1,4 @@
+This example illustrates how to set up the Arbitrary Lagrangian-Eulerian (ALE) framework in Neko to simulate a simple oscillating cylinder.
+
+To generate the mesh, execut `generate_mesh.sh` in the mesh folder.
```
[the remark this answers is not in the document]
arash-mp: done

src/fluid/fluid_pnpn_bc_fctry.f90
arash-mp: done

src/simulation_components/force_torque.f90
arash-mp: done, also fixed at other places

src/common/checkpoint.f90
```diff
@@ -120,13 +137,22 @@ subroutine chkp_free(this)
...
     if (associated(this%dtlag)) nullify(this%dtlag)
 
+    ! ALE cleanup
+    if (associated(this%wm_x)) nullify(this%wm_x)
+    if (associated(this%wm_y)) nullify(this%wm_y)
+    if (associated(this%wm_z)) nullify(this%wm_z)
+    if (associated(this%wm_x_lag)) nullify(this%wm_x_lag)
+    if (associated(this%wm_y_lag)) nullify(this%wm_y_lag)
+    if (associated(this%wm_z_lag)) nullify(this%wm_z_lag)
+    if (associated(this%basis_vel_lag)) nullify(this%basis_vel_lag)
```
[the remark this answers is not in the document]
arash-mp: done.

src/ale/ale_manager.f90
```diff
@@ -0,0 +1,1764 @@
...
+    if (this%config%nbodies == 0) return
+
+    call neko_log%message(" ")
+    call neko_log%message("Starting base mesh motion solve ...")
+    n = coef%dof%size()
+
+    ! Save original h1/h2
+    h1_restore = coef%h1
+    h2_restore = coef%h2
+
```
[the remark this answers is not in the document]
arash-mp: done. Also fixed at other places

src/ale/ale_manager.f90
arash-mp: I changed it, also for some other variables

src/fluid/fluid_pnpn.f90
```diff
@@ -694,6 +718,18 @@ subroutine fluid_pnpn_step(this, time, dt_controller)
       call this%bcs_vel%apply_vector(f_x%x, f_y%x, f_z%x, &
            this%dm_Xh%size(), time, strong = .false.)
 
+      if (this%ale%active) then
+         if (oifs) then
+            call neko_error("ALE is not yet supported with OIFS time integration.")
+         end if
```
timofeymukha, line 725: Move this check to some much earlier place, I guess in init you already know if both are on at the same time?

arash-mp: @timofeymukha  I put it here intentionally so in future I know where the missing pieces should be added!
I can move it to init if it really should be moved there, but since there are bunch of stuff which should be added in future, and in some cases I may forget where they need to be added, I prefer to keep them where they are now (in general regarding the things which are not supported right now).

However, I can add a check in init as well (have it in both places). Is it ok like this?

src/ale/ale_manager.f90
```diff
@@ -0,0 +1,1809 @@
...
+     procedure, pass(this) :: get_tracker_pos
+     procedure, pass(this) :: compute_rotation_matrix
+     procedure, pass(this) :: prep_checkpoint => set_pivot_basis_for_checkpoint
+     procedure, pass(this) :: ghost_tracker_coord_step
+     procedure, pass(this) :: log_rot_angles
+     procedure, pass(this) :: log_pivot
+  end type ale_manager_t
+
+  type(ale_manager_t), public, pointer :: neko_ale => null()
+
```
timofeymukha, line 163: Using this place to drop a general comment, after having looked at the structure a bit more. 

I think we should probably merge this in, so that the ALE can continue developing cohesively with the other stuff going on develop. The functionality is fairly self-contained, and there is little risk it contaminates ordinary solves even if there is a bug.

What I think can generally be improved moving on is somehow lifting the ale outside of fluid. The fact that you introduced the neko_ale here kind of points into that direction as well. That pointer should perhaps be the one that owns the object, and not the fluid. I am kind of hoping it should be possible to have the ale configured at the case level. For example, the mesh motion could be done directly in simulation, outside the fluid, right? It is its own equation after all. This would also reduce the burden of adding the ALE to other schemes, both fluid and scalar. 

Anyway, a very ambitious PR with a lot done, all of these kinematic stuff, I understand like 30% of them :)).

timofeymukha (approved):

src/ale/ale_manager.f90
arash-mp: Thanks Timofey!
Regarding moving ale outise fluid we can discuss:D For now, if the current version it's ok we can merge as you said, and in future we can polish the structure more.
I'll commit a final set of minor fixes and some small bug fixes today, then I will resolve this converstation to merge.

njansson (approved):

src/ale/ale_rigid_kinematics.f90
arash-mp: @adperezm I may do this later. It's a good idea. For now I don't want to change anything anymore :D

src/bc/field_moving.f90
arash-mp: field_moving doesn't exist anymore (it's in no_slip_t), but in general, for the check that i wanted to do (to be sure if a BC is moving, then ale is already active) I had to do this (also was possible to get the keyword from the case file here).

src/io/fluid_output.f90
arash-mp: I removed it.

src/sem/coef.f90
arash-mp: Here I needed it so for sp and dp it get the bytes for DEVICE_TO_DEVICE memcpy (in coef_update_lagged_mass).
Later if I moved them from coef I will remember to remove it ;)

src/sem/coef.f90
arash-mp: done. Thanks

src/fluid/fluid_pnpn.f90
arash-mp: Done.

examples/double_oscillating_cylinders/double_oscillating_cylinders.case
arash-mp: done

doc/pages/user-guide/case-file.md
arash-mp: @timofeymukha No, it's actually many-to-1. Several no_slip ids can be mapped to 1 ALE body, but not vice versa.
I changed the doc to make it clearer

examples/Double_ocyl_cylinder/cylinder.case
arash-mp: done

arash-mp:
OK so it's finally ready to be merged.
I tested make check and pytest (sp, dp) on CPU and everything was fine.
I also prepared an integration test for ALE, but I didn't commit that here. I would prefer to include that with ALE_device PR which will be issued soon.

If there is no other things to check, I will resolve the comments and merge it in :) @timofeymukha @njansson @adperezm

adperezm:
I resolved my comments. Some things regarding structure remain for me, but I agree we should merge and polish those out as we go.

adperezm (approved):

examples/double_oscillating_cylinders/double_oscillating_cylinders.f90
arash-mp: all done, except adding the integration test. I will add that with device support soon.

arash-mp:
wait don't merge it untill I check something and will be sure (just noticed something, want to be sure).
I will tell when it's ready. @njansson

PR #2245, opened 2026-01-02
Code style improvements

timofeymukha:
This PR is sponsored by your favorite AI overlord.

On a serious note, now that an LLM agent can fix these things for you based on the flint report while you get coffee, I see no reason not to make a push for 10.0 in the repo.

Some of the edited files here are not quite bumped to 10.0 yet, but all should be improved. I stopped at some point so that the PR doesn't grow way to big.


njansson:
Nice! Do you think it is possible to "allow" for `(i,j,k,l)` instead of `(i, j, k, l)` (I recall we ended up to allow for that since it would otherwise be quite verbose

timofeymukha:
> Nice! Do you think it is possible to "allow" for `(i,j,k,l)` instead of `(i, j, k, l)` (I recall we ended up to allow for that since it would otherwise be quite verbose

I think it is allowed under some circumstances. I think you example is actually allowed. And stuff like (i,1,1,1). However, if the index variable is not single-letter, e.g. `el`, then you have to add spaces. I still think this is a good setup.

````
 # Look through the list of punctuation, but skip if inside format string '()'
   # However, we do allow a single letter variable or 2 digit number to be used
   # without spaces.
   missing-spaces-after-punctuation:
````

**Comment by timfelle** [comment]
You can enforce every rule you can describe in regex, which does have some limitations when i wanted to not lint the contents of a string. But in general you can enforce everything.

In addition to the single letter variable, we also allow 2 digit numbers if i recall correctly due to all the specializations of operators.

File: src/bc/blasius.f90 [hunk 1]
```diff
@@ -255,7 +257,8 @@ subroutine blasius_apply_vector_dev(this, x_d, y_d, z_d, time, strong, strm)
 
 
       ! Pretabulate values during first call to apply
-      if (.not. c_associated(blax_d) .and. strong_ .and. this%msk(0) .gt. 0) then
+      if (.not. c_associated(blax_d) .and. strong_ .and. this%msk(0) .gt. 0) &
+           then
```
timfelle:
```suggestion
      if (.not. c_associated(blax_d) .and. strong_ .and. m .gt. 0) then
```

src/bc/facet_normal.f90
```diff
@@ -32,10 +32,10 @@
 !
 !> Dirichlet condition applied in the facet normal direction
 module facet_normal
-  use device_facet_normal
+  use device_facet_normal, only : device_facet_normal_apply_surfvec
```
timfelle: Unused
```suggestion
```

src/fluid/flow_ic.f90
```diff
@@ -269,7 +269,7 @@ subroutine set_flow_ic_uniform(u, v, w, uinf)
     character(len=LOG_SIZE) :: log_buf
 
     call neko_log%message("Type : uniform")
-    write (log_buf, '(A, 3(ES12.6, A))') "Value: [", (uinf(i), ", ", i=1, 2), &
+    write (log_buf, '(A, 3(ES12.6, A))') "Value: [", (uinf(i), ", ", i = 1, 2), &
          uinf(3), "]"
```
timfelle: Still long
```suggestion
    write (log_buf, '(A, 3(ES12.6, A))') "Value: [", &
         (uinf(i), ", ", i = 1, 2), uinf(3), "]"
```

src/fluid/fluid_scheme_compressible_euler.f90
```diff
@@ -270,7 +270,7 @@ end subroutine fluid_scheme_compressible_euler_free
   !> @param ext_bdf Time integration controller
   !> @param dt_controller Timestep size controller
   subroutine fluid_scheme_compressible_euler_step(this, time, dt_controller)
-    use entropy_viscosity, only: entropy_viscosity_t
+    use entropy_viscosity, only : entropy_viscosity_t
```
timfelle, line 273: Why not move this to the rest of the usages?

src/fluid/fluid_scheme_compressible_euler.f90
```diff
@@ -570,7 +570,7 @@ subroutine fluid_scheme_compressible_euler_restart(this, chkp)
   end subroutine fluid_scheme_compressible_euler_restart
 
   subroutine setup_regularization(this, params)
-    use entropy_viscosity, only: entropy_viscosity_t, entropy_viscosity_set_fields
+    use entropy_viscosity, only : entropy_viscosity_t, entropy_viscosity_set_fields
```
timfelle, line 573: Why not move this to the rest of the usages?

src/scalar/scalar_pnpn.f90
```diff
@@ -210,10 +210,12 @@ subroutine scalar_pnpn_init(this, msh, coef, gs, params, numerics_params, &
       call this%s_res%init(dm_Xh, "s_res")
 
       call this%abx1%init(dm_Xh, trim(this%name)//"_abx1")
```
timfelle:
```suggestion
      call this%abx1%init(dm_Xh, trim(this%name) // "_abx1")
```

src/scalar/scalar_pnpn.f90
```diff
@@ -210,10 +210,12 @@ subroutine scalar_pnpn_init(this, msh, coef, gs, params, numerics_params, &
...
-      call neko_registry%add_field(dm_Xh, trim(this%name)//"_abx1", ignore_existing = .true.)
+      call neko_registry%add_field(dm_Xh, trim(this%name)//"_abx1", &
```
timfelle:
```suggestion
      call neko_registry%add_field(dm_Xh, trim(this%name) // "_abx1", &
```

src/scalar/scalar_pnpn.f90
```diff
@@ -210,10 +210,12 @@ subroutine scalar_pnpn_init(this, msh, coef, gs, params, numerics_params, &
...
+           ignore_existing = .true.)
 
       call this%abx2%init(dm_Xh, trim(this%name)//"_abx2")
-      call neko_registry%add_field(dm_Xh, trim(this%name)//"_abx2", ignore_existing = .true.)
+      call neko_registry%add_field(dm_Xh, trim(this%name)//"_abx2", &
```
timfelle:
```suggestion
      call neko_registry%add_field(dm_Xh, trim(this%name) // "_abx2", &
```

src/scalar/scalar_pnpn.f90
```diff
@@ -210,10 +210,12 @@ subroutine scalar_pnpn_init(this, msh, coef, gs, params, numerics_params, &
       call this%s_res%init(dm_Xh, "s_res")
 
       call this%abx1%init(dm_Xh, trim(this%name)//"_abx1")
-      call neko_registry%add_field(dm_Xh, trim(this%name)//"_abx1", ignore_existing = .true.)
+      call neko_registry%add_field(dm_Xh, trim(this%name)//"_abx1", &
+           ignore_existing = .true.)
 
       call this%abx2%init(dm_Xh, trim(this%name)//"_abx2")
```
timfelle:
```suggestion
      call this%abx2%init(dm_Xh, trim(this%name) // "_abx2")
```

timfelle (approved):

src/bc/facet_normal.f90
timofeymukha: Hmm, now I am confused as per why `device_facet_normal_apply_surfvec` is not used anywhere.

src/fluid/fluid_scheme_compressible_euler.f90
timofeymukha: You mean move it to the top? I guess @tuananhdao liked it this way and we do not strictly forbid it.

timfelle: Sure, it's just not something we use other places so for consistency i though it might be good to move.

src/bc/facet_normal.f90
timfelle: Now that I cannot tell you 😅

timofeymukha: My only guess is that it has been fused into the pnpn device kernels instead, as part of some optimization. This is the only place the routine is called on the cpu, I believe

njansson (approved):

PR #2252, opened 2026-01-05
Add step to maximize build space in workflow

njansson:


njansson:
It doesn't fix the compiler issues with nvhpc, but it resolves the `out of disk space issue`, thus we can at least track where nvhpc goes belly up

PR #2261, opened 2026-01-14
Introduce deprecation logging

timfelle:
Enhance logging functionality by introducing a deprecation logging feature, updating log levels, and providing documentation on usage. 

This allow us to mark a feature as deprecated, while optionally give an expected removal version and some extra info of how to update the code.

As an example, the use of the operator `ortho` with an implicit device array, is marked as deprecated and a new `device_ortho` function is added to replace it.

The logging is bound to new logging levels, which is documented in the `appendices.md`. With an option to throw an error on call, generating a stack trace with most compilers to make it easier to figure out where the code was called.


timofeymukha (approved):

timfelle:
One thing, should we print the warnings always to really tell users that they are using something deprecated? Of course my example have to either be completely resolved internally in Neko, or the example should be removed in that case.

My intuition is to always print them, but not introduce any with this PR. One can go through and resolve internal usage when one add new deprecations. Which could also be enforced through the CI system to make sure no new code is added to the repo which use a deprecated feature.

src/math/operators.f90
```diff
@@ -226,6 +227,9 @@ subroutine ortho(x, glb_n_points, n)
     real(kind=rp) :: c
     type(c_ptr) :: x_d
     if (NEKO_BCKND_DEVICE .eq. 1) then
+       call neko_log%deprecated('Operator: ortho, implicit device', &
+            'v2.0.0', 'Please call device_ortho instead.')
+
```
njansson: Shouldn't device ortho be moved to opr_device, and this routine all handle fallbacks for CPU and SX backends?

njansson (approved):

src/math/operators.f90
timfelle: Lets move this to a seperate PR

PR #2265, opened 2026-01-16
Fix checkpoint initialization and improve scalar lag handling

timfelle:
Enhance the initialization of scalar lags in the checkpoint and improve the handling of scalar lag terms. Address readability issues in device copy operations and resolve various challenges related to checkpoint management.

### DIFF
--- a/src/io/chkp_output.f90
+++ b/src/io/chkp_output.f90
@@ -78,7 +78,7 @@ contains
     class(chkp_output_t), intent(inout) :: this
 
     call this%free_base()
-    call this%chkp%free()
+    nullify(this%chkp)
 
   end subroutine chkp_output_free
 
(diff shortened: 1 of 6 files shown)


src/io/chkp_output.f90
```diff
@@ -78,7 +78,7 @@ subroutine chkp_output_free(this)
     class(chkp_output_t), intent(inout) :: this
 
     call this%free_base()
-    call this%chkp%free()
+    nullify(this%chkp)
```
njansson, line 81: Is something similar also needed in the hdf5 writer?

timofeymukha (approved):
Looks good, I think! We need to add a test with scalar restarts asap.

src/io/chkp_output.f90
timfelle: I do not think so. The HDF5 file interface have no local members. So it should be handled in the `generic_file_t`

njansson (approved):

PR #2271, opened 2026-01-23
Expose statistics fields to the registry

vbaconnet:
Draft PR for now so that we can discuss the implementation choices etc before committing to a full PR

# Context

I started looking at this because I wanted to access the statistics fields with probes, to monitor convergence in certain areas of the domain.

I was surprised to see that the statistics fields (`u_mean`, `uu`, etc) are not in the registry, and therefore difficult to access from `probes` some without some dangerous explicit intervention in the user file with `neko_simcomps` etc. So the obvious solution was to add those statistics fields to the registry. 

# What this PR adds/changes

`mean_field_t` previously stored the statistics field `mf` in its own `field_t` object. The change is that now `mf` is directly added to the register (with the correct name) and that field is handled on its own by the `neko_registry`.

This will affect both `fluid_stats` and `user_stats` with no other intrusive changes.

## Example use cases

This unlocks a lot of powerful functionalities from within the case file directly, such as the following setup, which initializes vorticity fields with the `curl` simcomp, averages it in time with `user_stats` and sample the field with `probes`, all without any user implementation. 

```json
	"simulation_components": 
	[		
	    {
	      // Compute vorticity and store it as w_x, w_y, w_z in the registry
	      "type": "curl",
	      "fields": ["u", "v", "w"],
	      "computed_field": "w"
	    },
	    {
      		// Classical fluid stats, only basic set of stats
      		"type": "fluid_stats",
      		"compute_control": "tsteps",
      		"compute_value": 10,
      		"output_value": 20.0,
      		"start_time": 0.0,
      		"set_of_stats":"basic"
	    },
	    {
      		// Compute mean vorticity, and creates fields mean_w_(xyz) in the registry
      		"type": "user_stats",
      		"fields": ["w_x", "w_y", "w_z"],
      		"output_file": "vort_stat"	
	    },
	    {
      		// Probe some fields, from fluid_stats and user_stats
      		"type": "probes",
      		"compute_control": "tsteps",
      		"compute_value"    : 10,
      		"fields": ["u","uu","vv","ww", "mean_w_x"],
      		"output_file":  "out.csv",
      		"points": [
	        {
	          "type": "line",
	            "start": [0.0, 0.0, 0.0],
  		    "end": [1.0, 1.0, 1.0],
  		    "amount": 10
	        }
	      ],
	    }
	],
```

# Testing etc before undrafting

- [x] Compilation on CPUs
- [x] Compilation on NVIDIA GPUs
- [x] Compilation on AMD GPUs + Cray environment
- [x] Validation testing for a basic case, like cylinder


src/registries/registry.f90
```diff
@@ -434,7 +476,10 @@ recursive function registry_get_field_by_name(this, name) result(f)
     integer :: i
 
     do i = 1, this%n_entries()
-       if (this%entries(i)%get_type() .eq. 'field' .and. &
+
+       ! index(x, 'y') searches for substrings in x matching the pattern 'y'.
+       ! .eq. 1 means that the substring 'y' was found at position 1 in x
+       if (index(this%entries(i)%get_type(), 'field') .eq. 1 .and. &
```
vbaconnet: Just an additional explanation on this, the idea is that `get_field` should work for both 'field' and 'field_ptr' types so we check for the presence of the substring 'field' at position 1...

timofeymukha:
The use case is definitely compelling, and reasonably easy access to statistics fields at run time is for sure a good thing. For UQ, for example.

An alternative way of doing this, that could be worth thinking about, is just adding the capability to directly register a fluid_stats_t :-). There are issues to consider, like how would a simcomp (e.g. probes) know to search that object and not the usual field list, etc. But this is also a more general conceptual question: whether we want to move on with making the registry more and more general. E.g. for reusing stuff like dofmaps and Krylov solvers.

That said, the capability to store field_ptrs is perhaps nice, although I am not sure I would want them to sit under the same get/add interface as ordinary fields. But the latter is a detail. A potentially serious concern is lifetime management: if an object, say fluid_stats, frees its fields, the registry with pointers to those is non the wiser, right? You just get invalid pointers. And if it somehow got the info, we currently cannot remove an entry, once it is registered. The latter could probably be implemented though. We could also theoretically say that we add the field_ptr capability, but only for user code, at own risk. If you add one, it is up to you to make sure it is valid, etc. Then, if you want to average a certain field in the statistics, you pipe it into the registry in your user file.

One could also say that if we want a field accessible from the registry, it should own it. So just add_field all the statistics. You can use the simcomps%name as a field name prefix to all the fields, to be reasonably sure you avoid clashes. 

Another scary aspect is that, as you mention, the number of fields in the registry can in principle become huge. All the name mangling can become tedious. So one could make an argument for having a registry of registries instead of one neko_registry, and then fluid_stats would add its own and keep those 44 fields there. Then, when configuring simcomps you tell it in which subregistries it should look at for the fields. The fact that we already added a bonus neko_const_registry to the code for much the same reason gives this idea some merit. That thing could instead be an item in a list of registries.

None of the above is meant as some ground truth as you see, rather a thought dump at 00:30 on a weekend :)).

src/field/mean_field.f90
```diff
@@ -94,7 +96,8 @@ subroutine mean_field_free(this)
     if (associated(this%f)) then
        nullify(this%f)
     end if
-    call this%mf%free()
+
+    nullify(this%mf)
```
timfelle:
```suggestion
    if (associated(this%mf)) then
         nullify(this%mf)
    end if
```

timfelle (approved):

vbaconnet:
The pytests have raised an interesting issue (which seems obvious now): if there are two fluid_stats simcomps in the case file, each simcomp will try to create the same fields in the registry. So the first simcomp will create `mean_u`, `uu`, etc etc, and the next one will also try to create them. 

My thinking is that if someone really wants to use two fluid stats simcomps* they should be ready to duplicate the memory accordingly (which is the behavior currently). 

Now the question is, if a field `mean_u` already exists, how should the new one be named. I propose either:
- `mean_u_1`, `mean_u_2` ... and so on for as many different fluid stats simcomps as there are, or, 
- `mean_u_{order}` where `order` is the order of the simcomp.

Now, about that asterisk:
\* The pytest is failing because there are duplicate fluid_stats simcomps which only differ in that they output 3D, 2D and 1D fields. I think we could try to allow one fluid stat simcomp to combine different outputs, if one wants to output both 2D-averaged and 1D-averaged fields. That would prevent someone from duplicating simcomps, to get different spatial averaging on the same statistics field.

timofeymukha:
> The pytests have raised an interesting issue (which seems obvious now): if there are two fluid_stats simcomps in the case file, each simcomp will try to create the same fields in the registry. So the first simcomp will create `mean_u`, `uu`, etc etc, and the next one will also try to create them.


Haha, yeah this is the name mangling need we have been discussing, but I forgot I actually had multiple stats in the test. I believe the correct thing to do here it to add a `name` component to the simcomp base time, just like we have in schemes. It's up to the user to see that they are unique (a check can be added to the simcomp executor). Then this name could be a field name prefix passed to the stats object, so things will be unique and predictable.  I can try to look into this quickly.

I would not mix up 2D and 3D. Certainly not for resolving this issue at least, this has a broader scope.

vbaconnet:
> Then this name could be a field name prefix passed to the stats object, so things will be unique and predictable. I can try to look into this quickly.

Thanks! I've updated everything now. Tests should all be working now. Please check and see if you like the way things are.

adperezm (approved):
Why is this one stuck in review? :) Is there a particular reason?

vbaconnet:
I think the tests were unhappy with some of my changes...

timofeymukha:
> Why is this one stuck in review? :) Is there a particular reason?

We need to propagate the name mangling feature I added to the underlying stats object, so that we dont get name clashes in the registry, when one has several fluid_stats compoents like in the failing test.

vbaconnet:
> We need to propagate the name mangling feature I added to the underlying stats object, so that we dont get name clashes in the registry, when one has several fluid_stats compoents like in the failing test.

That is done!

adperezm:
Jahaa! :D

Next time I will check the tests before asking :). But I guess there is quite a bit of noise in those results at the momment.

src/simulation_components/user_stats.f90
```diff
@@ -219,15 +219,24 @@ subroutine user_stats_init_common(this, name, start_time, coef, avg_dir, &
...
+    unique_name = ""
+
     this%name = name
     this%start_time = start_time
     this%time = start_time
 
+    ! If a name is specified and is not the default name, add it
+    ! as a prefix to the mean field names, followed by a "/".
+    if (trim(name) .ne. "user_stats") then
+       unique_name = name // "/"
```
timofeymukha: Interesting choice of separator. My head defaulted to _, but maybe it is good to have a special one for the prefix

timofeymukha: Or go full fortran with % :-))

vbaconnet: Slowly paving the way towards hdf5... ;)

vbaconnet: i chose that because it represents something hierarchical, that the fields are contained within the simulation component of name `name`. It's also easier to parse later if those names end up written in text files like for `probes`, you can just split the name by `"/"` directly in python.

timofeymukha:
@vbaconnet Looks like we are good to go, right?

vbaconnet:
Not yet, I want to add some documentation about how the logic works with the `"/"` separator, etc!

timfelle (commented):
I think it is quite important to test this on the Cray system. We might run into dimensionality issues similar to the scratch registry if we have too many entries in the global registry. Have you tried running an example there to see if the segfault show up?

vbaconnet:
> I think it is quite important to test this on the Cray system. We might run into dimensionality issues similar to the scratch registry if we have too many entries in the global registry. Have you tried running an example there to see if the segfault show up?

I can't even compile neko on Cray anymore :( This is the latest attempt on Dardel

```
libtool: compile:  ftn -m4 -O3 -I/cfs/klemming/pdc/projects/hpcrd/software/spack/linux-sles15-zen2/cce-18.0.1/json-fortran-8.3.0-bjoug3paqr3n6uyx4eqtlprkisnapbve/include/ -c filter/mappings.f90 -o filter/mappings.o
module field_list
       ^          
ftn-855 ftn: ERROR FIELD_LIST, File = field/field_list.f90, Line = 1, Column = 8 
  The compiler has detected errors in module "FIELD_LIST".  No module information file will be created for this module.

    call this%items(len+1)%init(f)
                                ^  
ftn-1873 ftn: ERROR FIELD_LIST_APPEND, File = field/field_list.f90, Line = 125, Column = 33 
  POINTER dummy argument "PTR" is not polymorphic so its associated actual argument must not be polymorphic.
```

vbaconnet:
Just to summarize the latest changes:
- Always prepend the name in the registry by the name of the simcomp + `"/mean_"`. A bit redundant to have the `mean_` but i think for clarity it's better to have names that are a bit too verbose..
- Sneaked in a little logging for the neko_registry as well, at the NEKO_LOG_DEBUG level

Otherwise ready to go!

njansson:
> > I think it is quite important to test this on the Cray system. We might run into dimensionality issues similar to the scratch registry if we have too many entries in the global registry. Have you tried running an example there to see if the segfault show up?
> 
> I can't even compile neko on Cray anymore :( This is the latest attempt on Dardel
> 
> ```
> libtool: compile:  ftn -m4 -O3 -I/cfs/klemming/pdc/projects/hpcrd/software/spack/linux-sles15-zen2/cce-18.0.1/json-fortran-8.3.0-bjoug3paqr3n6uyx4eqtlprkisnapbve/include/ -c filter/mappings.f90 -o filter/mappings.o
> module field_list
>        ^          
> ftn-855 ftn: ERROR FIELD_LIST, File = field/field_list.f90, Line = 1, Column = 8 
>   The compiler has detected errors in module "FIELD_LIST".  No module information file will be created for this module.
> 
>     call this%items(len+1)%init(f)
>                                 ^  
> ftn-1873 ftn: ERROR FIELD_LIST_APPEND, File = field/field_list.f90, Line = 125, Column = 33 
>   POINTER dummy argument "PTR" is not polymorphic so its associated actual argument must not be polymorphic.
> ```

Sigh, we really need a CCE based runner for testing... let me see if something can be done about that.

PS. does `explain ftn-1873` give any reasonable explanation for the issue ?

vbaconnet:
> PS. does `explain ftn-1873` give any reasonable explanation for the issue ?

Unfortunately you were a second late, we found the issue and already merged it! :) #2314

vbaconnet:
Tested on Dardel, with 44 + 13 + 3 added entries in registry from fluid/user stats. Works fine!

vbaconnet:
If we're all happy with this I'd appreciate if we could merge before this gets too difficult to keep up to date with develop

PR #2272, opened 2026-01-24
Fix initialisation of multiple scalar

Shiyu-Sandy-Du:
This issue is from the nasty polymorphism of Fortran... So here is the story:
Without the fix, it still works perfectly on my local machine, however when I used dardel it crashed at the begining, especially the bc_res%msk of my second scalar equation is somehow denoted as allocated in the initialisation of my first scalar equation. And when it comes to the initialisation of the second scalar equation, `deallocate(bc_res%msk)` gives an error while `allocated(bc_res%msk)` is still weirdly `true`. But if one change the declaration from the abstract `scalar_scheme_t` to the derived `scalar_pnpn_t`, the simulation is happy again.

And here is the solution:
Now we have a wrapper for a pointer to the abstract scalar_scheme_t and we actually only allocate the array for the pointers. And then in the do loop we do the actual allocation for `scalar_pnpn_t`

I hope it's a solution making everyone happy :)


src/scalar/scalar_scheme.f90
```diff
@@ -165,6 +165,11 @@ module scalar_scheme
      procedure(scalar_scheme_restart_intrf), pass(this), deferred :: restart
   end type scalar_scheme_t
 
+  !> Pointer to a @ref `bc_t`
+  type, public :: scalar_ptr_t
+     class(scalar_scheme_t), pointer :: scheme => null()
```
timofeymukha: For consistency with other  `_ptr_t`, please rename `scheme` to `ptr`. See, for example, `field_ptr_t`, `vector_ptr_t`, `bc_ptr_t`, etc.

timofeymukha:
Hmm, this is sad. Is there some general lesson to derive here? Should we avoid allocatable arrays of abstract types? Very sad indeed.

Shiyu-Sandy-Du:
> Hmm, this is sad. Is there some general lesson to derive here? Should we avoid allocatable arrays of abstract types? Very sad indeed.

I think you are right.

njansson:
Interesting, I'll have a look when less jet lagged. One quick question, is the large diff for fdm sx necessary? If not suggest to those in a separate pr to avoid cluttering things too much

Shiyu-Sandy-Du:
> Interesting, I'll have a look when less jet lagged. One quick question, is the large diff for fdm sx necessary? If not suggest to those in a separate pr to avoid cluttering things too much

they are from formatting (otherwise it cannot pass the test)

src/scalar/scalar_scheme.f90
```diff
@@ -165,6 +165,11 @@ module scalar_scheme
      procedure(scalar_scheme_restart_intrf), pass(this), deferred :: restart
   end type scalar_scheme_t
 
+  !> Pointer to a @ref `bc_t`
```
Shiyu-Sandy-Du:
```suggestion
  !> Pointer to a @ref `scalar_scheme_t`
```

Shiyu-Sandy-Du:
> > Interesting, I'll have a look when less jet lagged. One quick question, is the large diff for fdm sx necessary? If not suggest to those in a separate pr to avoid cluttering things too much
> 
> they are from formatting (otherwise it cannot pass the test)

Okay now it's no problem anymore

src/scalar/scalars.f90
```diff
@@ -259,7 +261,7 @@ subroutine scalars_free(this)
     ! Iterate through all scalar fields
     if (allocated(this%scalar_fields)) then
        do i = 1, size(this%scalar_fields)
-          call this%scalar_fields(i)%free()
+          call this%scalar_fields(i)%scheme%free()
```
timfelle:
```suggestion
          if (allocated(this%scalar_fields(i)%scheme)) then
               call this%scalar_fields(i)%scheme%free()
               deallocate(this%scalar_fields(i)%scheme)
          end if
          
```

timfelle: Either do this allocation stuff here, or add it to the wrapper type. I would
probably suggest to do it in the wrapper, that would make it a lot simpler to
deal with.

src/scalar/scalar_scheme.f90
```diff
@@ -636,4 +644,13 @@ subroutine scalar_scheme_set_material_properties(this, params, user)
     end if
   end subroutine scalar_scheme_set_material_properties
 
+  subroutine scalar_scheme_ptr_free(this)
+    class(scalar_scheme_ptr_t), intent(inout) :: this
+
+    if (associated(this%scheme)) then
+       call this%scheme%free()
+       nullify(this%scheme)
```
timfelle: This will leak. You have been doing allocate on the pointer, so you should deallocate here.

Shiyu-Sandy-Du: do you mean deallocate ` this%scheme`?

timfelle: Yes.

Shiyu-Sandy-Du: then it should be deallocated in scalar.f90?

timfelle: Hmm, I think you should either deallocate in free, or completely remove the free function and handle it in scalar.f90.

Shiyu-Sandy-Du: umm I don't think it could be deallocated in free since it's not declared as an allocatable in the wrapper

timfelle: Well you could allocate it, so you can also deallocate it.

Shiyu-Sandy-Du: yes the allocation is done in scalars.f90 so I think maybe I need to do the second option

njansson:
Looking at the changes, couldn't one cause of this be the lack of a `select type` when the scalar is used/accessed during init?

Shiyu-Sandy-Du:
> Looking at the changes, couldn't one cause of this be the lack of a `select type` when the scalar is used/accessed during init?

Didn't realise that. But maybe we need some flexibility since the current appraoch allows different schemes for scalars.

adperezm:
If this is a bug fix, shouldn't we push towards merging it soon? Is it that no one likes the solution?

Also, @Shiyu-Sandy-Du , check the tests when you have time :)

timofeymukha:
> If this is a bug fix, shouldn't we push towards merging it soon? Is it that no one likes the solution?
> 
> Also, @Shiyu-Sandy-Du , check the tests when you have time :)

I guess we should harmonize with https://github.com/ExtremeFLOW/neko/pull/2313 because it solves the exact same problem but for a different array, right @timfelle ?

Shiyu-Sandy-Du:
> > If this is a bug fix, shouldn't we push towards merging it soon? Is it that no one likes the solution?
> > Also, @Shiyu-Sandy-Du , check the tests when you have time :)
> 
> I guess we should harmonize with #2313 because it solves the exact same problem but for a different array, right @timfelle ?

I think so. And I think we had a discussion somewhere about the difference of using allocatable or pointer

src/scalar/scalar_scheme.f90
```diff
@@ -229,6 +220,104 @@ subroutine scalar_scheme_step_intrf(this, time, ext_bdf, dt_controller, &
...
+  ! ========================================================================== !
+
+  !> A helper type that is needed to have an array of polymorphic objects
+  type, public :: scalar_scheme_wrapper_t
+     class(scalar_scheme_t), allocatable :: scalar
+   contains
+     !> Constructor. Just allocates the pointer.
+     procedure, pass(this) :: init => scalar_scheme_wrapper_init
+     !> Destructor. Just deallocates the pointer.
+     procedure, pass(this) :: free => scalar_scheme_wrapper_free
```
timofeymukha: Maybe not use the word pointer here since it is actually an allocatable.

src/Makefile.am
```diff
@@ -316,6 +316,7 @@ neko_fortran_SOURCES = \
 	bc/bcknd/device/device_dong_outflow.F90\
 	scalar/bcknd/device/scalar_residual_device.F90\
 	scalar/scalar_scheme.f90\
+	scalar/scalar_scheme_fctry.f90\
```
timofeymukha, line 319: Don't forget the changelog!

This opens up for user-written scalar schemes!

timfelle (approved):

timofeymukha (approved):

PR #2277, opened 2026-01-28
Add names to simcomp

timofeymukha:
Adds a %name to simulation_component_t. Should be unique withing the simcom_executor array, which is checked. The functionality is essentially dormant now, but can be used for name mangling (prefixing the simcomps name), extracting a simcomp by name from the executor, etc.


src/simulation_components/derivative_simcomp.f90
```diff
@@ -181,14 +188,16 @@ subroutine derivative_init_from_controllers(this, case, order, &
 
     call this%init_base_from_components(case, order, preprocess_controller, &
          compute_controller, output_controller)
-    call this%writer%init_from_components(case, order, preprocess_controller, &
-         compute_controller, output_controller, fields, filename, precision)
-    call this%init_common(field_name, computed_field, direction)
+    call this%writer%init_from_components("field_writer", case, order, &
```
vbaconnet, line 191: The name `"field_writer"` won't clash with other field writers if I have two derivative simcomps, correct? Since that component is only internal to `derivative_simcomp_t`?

vbaconnet (approved):
LGTM :)

src/simulation_components/derivative_simcomp.f90
timofeymukha: exactly, the name has little meaning for a standalone simcomp object.

njansson (approved):

PR #2281, opened 2026-01-30
Feature/write mesh control

adperezm:
Added an option to print out the mesh every time.

I am having problems on deciding how to do this. Generally I would like that one can control what is written when the write method is called, but that is not the way we do things in Neko, so for the moment it is just like this.

I guess if one creates a file_t in the user file, it is possible to control if the mesh is written by directly modifying the attribute in the object, so at least that's something.

This is based on what @arash-mp will also add in #2244.


src/io/fld_file.f90
```diff
@@ -63,6 +63,7 @@ module fld_file
   !> Interface for NEKTON fld files
   type, public, extends(generic_file_t) :: fld_file_t
      logical :: dp_precision = .false. !< Precision of output data
+     logical :: write_mesh = .false. !< Whether to write the mesh
```
timofeymukha, line 66: I think this should also be always_write_mesh, based on the logic below.

src/io/fld_file.f90
```diff
@@ -281,7 +282,12 @@ subroutine fld_file_write(this, data, t)
     !
 
     call this%increment_counter()
-    write_mesh = (this%get_counter() .eq. this%get_start_counter())
+    ! Check if I should write the mesh. Always override at the start counters
```
timofeymukha, line 285: Can this be something like
write_mesh = this%always_write_mesh .or. ((this%get_counter() .eq. this%get_start_counter()))

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Develop
+- Add option for writing the mesh in every output.
```
timofeymukha:
```suggestion
- Added an option for writing the mesh in every output field file.
```

doc/pages/user-guide/case-file.md
```diff
@@ -1160,6 +1160,7 @@ concisely directly in the table.
 | `nut_field`                             | The name of the turbulent viscosity field.                                                        | String                                                      | -             |
 | `output_control`                        | Defines the interpretation of `output_value` to define the frequency of writing checkpoint files. | `nsamples`, `simulationtime`, `tsteps`, `never`             | -             |
 | `output_value`                          | The frequency of sampling in terms of `output_control`.                                           | Positive real or integer                                    | -             |
+| `output_always_write_mesh`              | Inidicates if the mesh should be written in every output fld file.                                | `true` or `false`                                           | `false`       |
```
timofeymukha: This one is hard not to make awkward... Maybe "output_mesh_in_all_files"?

timofeymukha (commented):
Much needed. I guess we will have to check that all fld _output types support this for consistency.

doc/pages/user-guide/case-file.md
adperezm: Sure! I modified it.

doc/pages/user-guide/case-file.md
```diff
@@ -1160,6 +1160,7 @@ concisely directly in the table.
 | `nut_field`                             | The name of the turbulent viscosity field.                                                        | String                                                      | -             |
 | `output_control`                        | Defines the interpretation of `output_value` to define the frequency of writing checkpoint files. | `nsamples`, `simulationtime`, `tsteps`, `never`             | -             |
 | `output_value`                          | The frequency of sampling in terms of `output_control`.                                           | Positive real or integer                                    | -             |
+| `output_mesh_in_all_files`              | Inidicates if the mesh should be written in every output fld file.                                | `true` or `false`                                           | `false`       |
```
```suggestion
| `output_mesh_in_all_files`              | Indicates if the mesh should be written in every output fld file.                                | `true` or `false`                                           | `false`       |
```

src/io/fld_file.f90
adperezm: In the case of the fld_file, I would like to keep it as write_mesh. 

This mostly because I see a case where I create a file in the user module and I can decide if I write the mesh or not. I feel the "always_write_mesh" thing is more of a simulation case kinda thing that of a particular file instance.

src/io/fld_file.f90
adperezm: Yes! that's cleaner. I am changing it.

vbaconnet (approved):

timofeymukha (approved):

timofeymukha:
We can optimize the if statement later :). I am doing some stuff with fld, and I don't want this PR to end up in merging problems

PR #2282, opened 2026-01-30
Add buoyancy correction of the Vreman model

linneahuusko:
Add an optional correction to the Vreman model based on the local Richardson number, intended for use in ABL simulations. 

Implements the correction factor in equation (9) of [Moeng and Sullivan 2015](https://doi.org/10.1016/B978-0-12-382225-3.00201-2), generalized to take the gravity vector g as input and apply the correction in the corresponding direction(s). 

Has been successfully tested on CPU, GPU implementation remains to be added.


src/les/bcknd/cpu/vreman_cpu.f90
```diff
@@ -152,6 +160,61 @@ subroutine vreman_compute_cpu(if_ext, t, tstep, coef, nut, delta, c)
...
+                    du_n(2) = a21%x(i,1,1,e)*n(1) + a22%x(i,1,1,e)*n(2) +&
+                            a23%x(i,1,1,e)*n(3)
+                    du_n(3) = a31%x(i,1,1,e)*n(1) + a32%x(i,1,1,e)*n(2) +&
+                            a33%x(i,1,1,e)*n(3)
+
+                    ! Component parallel to n
+                    du_parallel = du_n(1)*n(1) + du_n(2)*n(2) + du_n(3)*n(3)
+
+                    ! Perpendicular (shear) components
+                    do concurrent (j = 1:3)
```
Shiyu-Sandy-Du:
```suggestion
                    do j = 1, 3
```
it's fairly enough here to use a normal do loop

src/les/bcknd/cpu/vreman_cpu.f90
```diff
@@ -152,6 +160,61 @@ subroutine vreman_compute_cpu(if_ext, t, tstep, coef, nut, delta, c)
...
+                    end do
+
+                    ! Shear magnitude squared
+                    shear_sq = sh(1)*sh(1) + sh(2)*sh(2) + sh(3)*sh(3)
+
+                    ! Richardson number
+                    ri = buoyancy / (shear_sq + NEKO_EPS)
+
+                    if (ri .le. ri_c) then
+                         correction = (1 - ri/ri_c)**0.5
```
Shiyu-Sandy-Du:
```suggestion
                         correction = sqrt(1 - ri/ri_c)
```
It's faster and of higher precision with sqrt than **0.5

Shiyu-Sandy-Du (changes requested):
Their is some formatting issues as you can see in the test below the PR. If you run the `format.sh` script in `contrib/lnit_format`, it could be resolved:) I think it's a good PR. I will submit a different PR for the device implementation after this one is merged. Nice job!

src/les/bcknd/cpu/vreman_cpu.f90
```diff
@@ -152,6 +160,61 @@ subroutine vreman_compute_cpu(if_ext, t, tstep, coef, nut, delta, c)
                * coef%mult(i,1,1,e)
        end do
     end do
+    if (if_corr .eqv. .true.) then
+          theta => neko_registry%get_field_by_name("temperature")
+          call neko_scratch_registry%request_field(dTdx, temp_indices_buoy(1), .false.)
```
timofeymukha: no need for .eqv. .true.

src/les/vreman.f90
```diff
@@ -73,16 +77,32 @@ subroutine vreman_init(this, fluid, json)
...
 
     call json_get_or_default(json, "nut_field", nut_name, "nut")
     call json_get_or_default(json, "delta_type", delta_type, "pointwise")
     ! Based on the Smagorinsky Cs = 0.17.
     call json_get_or_default(json, "c", c, 0.07_rp)
     call json_get_or_default(json, "extrapolation", if_ext, .false.)
+    call json_get_or_default(json, "buoyancy_correction", if_corr, .false.)
+    call json_get_or_default(json, "ri_c", ri_c, 0.25_rp)
+    call json_get_or_default(json, "theta0", theta0, 293.0_rp)
+    if (if_corr .eqv. .true.) then
```
timofeymukha: remove .eqv. .true.

src/les/vreman.f90
```diff
@@ -73,16 +77,32 @@ subroutine vreman_init(this, fluid, json)
...
     character(len=LOG_SIZE) :: log_buf
 
     call json_get_or_default(json, "nut_field", nut_name, "nut")
     call json_get_or_default(json, "delta_type", delta_type, "pointwise")
     ! Based on the Smagorinsky Cs = 0.17.
     call json_get_or_default(json, "c", c, 0.07_rp)
     call json_get_or_default(json, "extrapolation", if_ext, .false.)
+    call json_get_or_default(json, "buoyancy_correction", if_corr, .false.)
+    call json_get_or_default(json, "ri_c", ri_c, 0.25_rp)
+    call json_get_or_default(json, "theta0", theta0, 293.0_rp)
```
timofeymukha: - We can use Ri_c, since we capitilize Re and Pe in the case file.
- Let's use reference_temperature or match the boussinesq term and use reference_value.

src/les/vreman.f90
```diff
@@ -106,18 +129,23 @@ end subroutine vreman_init
   !! @param delta_type The type of filter size.
   !! @param if_ext Whether trapolate the velocity.
   subroutine vreman_init_from_components(this, fluid, c, nut_name, &
-       delta_type, if_ext)
+       delta_type, if_ext, if_corr, ri_c, theta0, g)
     class(vreman_t), intent(inout) :: this
     class(fluid_scheme_base_t), intent(inout), target :: fluid
-    real(kind=rp) :: c
+    real(kind=rp) :: c, ri_c, theta0
+    real(kind=rp), allocatable :: g(:)
```
timofeymukha: I think here we can have an ordinary array of 3 elements, or?

src/les/bcknd/cpu/vreman_cpu.f90
```diff
@@ -152,6 +160,61 @@ subroutine vreman_compute_cpu(if_ext, t, tstep, coef, nut, delta, c)
                * coef%mult(i,1,1,e)
        end do
     end do
+    if (if_corr .eqv. .true.) then
+          theta => neko_registry%get_field_by_name("temperature")
```
timofeymukha: The scalar name should be a json parameter, we should not force "temeperature" here. Even in the ABL context someone might name it theta, or stick to a default "s". Set the name in the constructors, and save it as a component in the type. Then, use that name here.

src/les/bcknd/cpu/vreman_cpu.f90
Shiyu-Sandy-Du: Do we fix that as a coding rule? I am wondering whether we can add similar stuffs in format.sh to avoid unintentional occasions:)

src/les/bcknd/cpu/vreman_cpu.f90
njansson: Not so relevant for this loop, but one should not forget that do concurrent isn't only about parallelism it's also flaggin for the compiler that there isn't any dependencies, like the pragma `IVDEP`. 

So even for short loops I can help to guide the confused compiler to vectorise the code

src/les/bcknd/cpu/vreman_cpu.f90
linneahuusko: I guess that is the case for `if (if_ext .eqv. .true.) then` on line 90 in that file as well then?

src/les/vreman.f90
linneahuusko: Good points, thanks!

linneahuusko:
> Their is some formatting issues as you can see in the test below the PR. If you run the `format.sh` script in `contrib/lnit_format`, it could be resolved:) I think it's a good PR. I will submit a different PR for the device implementation after this one is merged. Nice job!

I didn't know the formatter existed, thanks!

src/les/bcknd/cpu/vreman_cpu.f90
```diff
@@ -152,6 +161,61 @@ subroutine vreman_compute_cpu(if_ext, t, tstep, coef, nut, delta, c)
...
+       call neko_scratch_registry%request_field(dTdz, temp_indices_buoy(3), .false.)
+
+       ! Calculate Richardson number
+       gmag = sqrt(vlsc2(g, g, 3))
+       if (gmag > NEKO_EPS) then
+          n = g / gmag
+       else
+          call neko_error("The gravity vector must have at least one nonzero component")
+       endif
+       call grad(dTdx%x, dTdy%x, dTdz%x, temperature%x, coef)
```
linneahuusko, line 182: Should there be a call to `coef%gs_h%op` here?

Shiyu-Sandy-Du: You could if it really leads to some instabilities

timofeymukha: yeah, this is one of those things with no clear right or wrong.

src/les/bcknd/cpu/vreman_cpu.f90
timofeymukha: Yes. The linter should flag that for you!

timofeymukha: (similarly, if (.not. conidtion) instead of if (condition .eqv. .false.)

src/les/bcknd/cpu/vreman_cpu.f90
Shiyu-Sandy-Du: Maybe we need to be a bit careful about the sign of the gravity here. See the change in PR#2303

src/les/bcknd/cpu/vreman_cpu.f90
```diff
@@ -152,6 +166,61 @@ subroutine vreman_compute_cpu(if_ext, t, tstep, coef, nut, delta, c)
...
+    if (if_corr) then
+       temperature => neko_registry%get_field_by_name(scalar_name)
+       call neko_scratch_registry%request_field(dTdx, temp_indices_buoy(1), .false.)
+       call neko_scratch_registry%request_field(dTdy, temp_indices_buoy(2), .false.)
+       call neko_scratch_registry%request_field(dTdz, temp_indices_buoy(3), .false.)
+
+       ! Calculate Richardson number
+       gmag = sqrt(vlsc2(g, g, 3))
+       if (gmag > NEKO_EPS) then
+          n = g / gmag
```
Shiyu-Sandy-Du, line 178: here as well, maybe we need to think more about the sign

src/les/bcknd/cpu/vreman_cpu.f90
linneahuusko: Good point, I have changed the sign here now.

Shiyu-Sandy-Du (approved):

src/les/bcknd/cpu/vreman_cpu.f90
```diff
@@ -152,6 +166,61 @@ subroutine vreman_compute_cpu(if_ext, t, tstep, coef, nut, delta, c)
...
+          call neko_error("The gravity vector must have at least one nonzero component")
+       endif
+       call grad(dTdx%x, dTdy%x, dTdz%x, temperature%x, coef)
+       do concurrent (e = 1:coef%msh%nelv)
+          do concurrent (i = 1:coef%Xh%lxyz)
+
+             ! Buoyancy component (numerator in Ri definition)
+             buoyancy = (-g(1) * dTdx%x(i,1,1,e) + &
+                  -g(2) * dTdy%x(i,1,1,e) + &
+                  -g(3) * dTdz%x(i,1,1,e)) / ref_temp
```
Shiyu-Sandy-Du:
```suggestion
             buoyancy = (-g(1) * dTdx%x(i,1,1,e) - &
                  g(2) * dTdy%x(i,1,1,e) - &
                  g(3) * dTdz%x(i,1,1,e)) / ref_temp
```

Shiyu-Sandy-Du:
why the cuda kernel is slightly different with hip kernel?

src/les/vreman.f90
```diff
@@ -73,16 +78,31 @@ subroutine vreman_init(this, fluid, json)
...
 
     call json_get_or_default(json, "nut_field", nut_name, "nut")
     call json_get_or_default(json, "delta_type", delta_type, "pointwise")
     ! Based on the Smagorinsky Cs = 0.17.
     call json_get_or_default(json, "c", c, 0.07_rp)
     call json_get_or_default(json, "extrapolation", if_ext, .false.)
+    call json_get_or_default(json, "buoyancy_correction", if_corr, .false.)
+    call json_get_or_default(json, "scalar_field", scalar_name, "temperature")
+    call json_get_or_default(json, "Ri_c", ri_c, 0.25_rp)
+    call json_get_or_default(json, "reference_temperature", ref_temp, 293.0_rp)
```
timofeymukha: I would suggest using json_get on scalar_field, Ri_c, and reference_temperature. At least for scalar_field and reference_temperature, these are for sure conscious choices required from the user. Our scalar name does not default to temperature, for example.  I don't know if Ri_c=0.25 is like mega-default, if it is, this value could be used.

timofeymukha (commented):
Pretty much ready to merge! Just one comment about parameter defaults.

src/les/vreman.f90
timofeymukha: @linneahuusko , I changed this, and also added some sanity checks for values. Also update copyright years, and fixed some small typos.

timofeymukha (approved):

PR #2283, opened 2026-01-31
Move scalar IC out of case.f90

timofeymukha:
This PR moves the work to set up initial conditions away from the case type and into the scalars and scalar_scheme types. It provides a welcome cleanup, and consolidates scalar-related functionality inside the right types. 

I also add an integration test that tests all the initial conditions, both with a "scalar" in the case and a "scalars" with two scalars, one of which is "temperature".

P.S. I think using the temperature field as a special case with flds is cursed. I am actually not sure if things work, if "temperature" doesn't happen to be the first scalar introduced in the case file


timofeymukha:
This is a rather peculiar set of of failed CI workflows.

njansson:
For consistency, shouldn't we do similar things with `fluid`?

timofeymukha:
> For consistency, shouldn't we do similar things with `fluid`?

Yes, I will in a new PR once we get this one in. But first gotta fix the issue with field ics here..

njansson:
> > For consistency, shouldn't we do similar things with `fluid`?
> 
> Yes, I will in a new PR once we get this one in. But first gotta fix the issue with field ics here..

Great!

njansson (approved):
🐈

timofeymukha:
I'm putting a don't merge here before we resolve the field IC issue in develop. For some reason, here I do not see if for the single scalar case, which indicates that two minuses became a plus, so to speak :-)). We need to be systematic.

timofeymukha:
This diverged too much, so closing for now.

PR #2286, opened 2026-02-02
Feature/sgs stats

Shiyu-Sandy-Du:
This PR adds the statistics for the subgrid-scale contribution for the anisotropic part of Reynolds stresses and scalar fluxes. The validation is performed by an ABL toy simulation with a TKE model for the eddy diffusivity and eddy viscosity, which is down in another un-merged branch. It does not match perfectly with the reference from nek5000, but since the simulation can not even be statistically stationary and the simulations can differ slightly in terms of time step size an other details in the implementation. But the trend still tells that we are getting the correct quantity to do the statistics. The only concern here is the scheme_t is not added into registry so one could only assign both the scalar field name and the corresponding eddy diffusivity field name for scalar sgs stats. Maybe one could add the pointer to the schemes in the registry in a separate PR.
wtheta_sgs
vw_sgs
uw_sgs


pschlatt1:
Looks good to me (the results). What about the normal components, potentially with some model for the trace?

doc/pages/user-guide/statistics-guide.md
```diff
@@ -241,4 +241,119 @@ e_{us} &= \frac{\partial u}{\partial x}  \frac{\partial s}{\partial x} + \frac{\
 e_{vs} &= \frac{\partial v}{\partial x}  \frac{\partial s}{\partial x} + \frac{\partial v}{\partial y}\frac{\partial s}{\partial y}+ \frac{\partial v}{\partial z}\frac{\partial s}{\partial z} \\\\
 e_{ws} &= \frac{\partial w}{\partial x}  \frac{\partial s}{\partial x} + \frac{\partial w}{\partial y}\frac{\partial s}{\partial y}+ \frac{\partial w}{\partial z}\frac{\partial s}{\partial z} \\\\
 \end{aligned}
-$$
\ No newline at end of file
+$$
+
+# Fluid Subgrid-Scale (SGS) Statistics
+In the fluid SGS statistics in Neko, subgrid-scale (SGS) contribution to the anisotropic part of the Reynolds stresses are computed, in a similar fashion to the fluid statistics. In total, 7 statical quantities are computed.
```
timofeymukha:
```suggestion
In the fluid SGS statistics in Neko, subgrid-scale (SGS) contribution to the anisotropic part of the Reynolds stresses are computed, in a similar fashion to the fluid statistics. In total, 7 statistical quantities are computed.
```

src/les/fluid_sgs_stats.f90
```diff
@@ -0,0 +1,243 @@
...
+
+  type, public, extends(stats_quant_t) :: fluid_sgs_stats_t
+     !> Work fields
+     type(field_t) :: stats_work
+     type(field_t) :: s11_work
+     type(field_t) :: s22_work
+     type(field_t) :: s33_work
+     type(field_t) :: s12_work
+     type(field_t) :: s13_work
+     type(field_t) :: s23_work
```
timofeymukha: It seem we can use the scratch registry for these, or? Same in scalar.

src/simulation_components/fluid_sgs_stats_simcomp.f90
```diff
@@ -0,0 +1,269 @@
...
+       else
+          fname = "fluid_sgs_stats"// &
+               trim(adjustl(prefix))//"."//suffix
+       end if
+       call this%stats_output%init_base(fname)
+    end if
+  end subroutine fluid_sgs_stats_simcomp_restart
+
+  !> fluid_sgs_stats, called depending on compute_control and compute_value
+  !! @param t The time value.
```
timofeymukha: Incorrect

src/simulation_components/fluid_sgs_stats_simcomp.f90
```diff
@@ -0,0 +1,269 @@
...
+
+       sample_start_time = MPI_WTIME()
+
+       call this%stats%update(delta_t)
+       call MPI_Barrier(NEKO_COMM, ierr)
+       this%time = t
+
+       sample_time = MPI_WTIME() - sample_start_time
+
+       call neko_log%section('Fluid stats')
```
timofeymukha: Fluid SGS stats

src/case.f90
```diff
@@ -389,32 +390,33 @@ subroutine case_init_common(this)
...
+                        this%scalars%scalar_fields(i)%scheme%gs_Xh, &
                         string_val, json_subdict, 0)
                    temperature_found = .true.
                 else
                    if (temperature_found) then
                       ! if temperature is found, other scalars start from index 1
-                      call set_scalar_ic(this%scalars%scalar_fields(i)%s, &
-                           this%scalars%scalar_fields(i)%c_Xh, &
-                           this%scalars%scalar_fields(i)%gs_Xh, &
+                      call set_scalar_ic(this%scalars%scalar_fields(i)%scheme%s, &
```
timofeymukha: Have we merged in the scheme% stuff, I forgot :-)

timofeymukha (commented):
Had a quick look. Please check if we can use the scratch_registry in some of these.

The amount of code duplication in the _output types is painful, but for now, I guess it is the only way... Something to think about for us later, though.

Shiyu-Sandy-Du:
> potentially with some model for the trace

So the test case all normal components are almost zero, for example for the first component it's <2*nut*dudx> and you could have equally distribution for negative and positive dudx for the same nut value in statistics

src/les/fluid_sgs_stats.f90
Shiyu-Sandy-Du: yes but what can we benefit from that?

src/case.f90
Shiyu-Sandy-Du: no not yet, we need to first merge that PR and then it could fit itself in

src/simulation_components/fluid_sgs_stats_simcomp.f90
Shiyu-Sandy-Du: Did I do something stupid about the restart?

src/les/fluid_sgs_stats.f90
Shiyu-Sandy-Du: Just had a talk with Adalberto. Now I have some new ideas. We could just let it be there for this PR, and I will write a new type call `Sij_t` including all 6 components and one logical to tell whether it's already computed in the current time step. This logical will turn `false` at the beginning of each fluid step. An `update` subroutine calculates all 6 components if the logical is `false` and should be called at each time the strain rate needs to be calculated. The benefit is that we can use the same field in both `sgs_stats` and `les_model` simcomps, avoiding memory waste.

Shiyu-Sandy-Du: Maybe an even more aggresive approach: instead of strain rate tensor, we compute the velocity gradient tensor and rewrite all usage of Sij in  simcomp into velocity gradients. This approach could include sigma and vreman model into consideration.

njansson (approved):

doc/pages/user-guide/statistics-guide.md
```diff
@@ -249,3 +249,118 @@ e_{vs} &= \frac{\partial v}{\partial x}  \frac{\partial s}{\partial x} + \frac{\
...
+      "compute_value": 4,
+      "output_control": "simulationtime",
+      "output_value": 20,
+      "start_time": 60.0,
+      "avg_direction":"xz",
+    }
+  ]
+~~~~~~~~~~~~~~~
+
+Otherwise, if one does not use model for the scalar eddy diffusivity but relates it with the eddy viscosity, the following would work is one specify a turbulent Prandtl number to be for instance 0.7 and the output will give the eddy diffusivity field as the edy viscosity field divided by the turbulent Prandtl number:
```
timofeymukha:
```suggestion
Otherwise, if one does not use a model for the scalar eddy diffusivity but relates it to the eddy viscosity, one needs to specify the turbulent Prandtl number, for instance, 0.7. The output will give the eddy diffusivity field as the eddy viscosity field divided by the chosen turbulent Prandtl number:
```

src/les/fluid_sgs_stats.f90
```diff
@@ -0,0 +1,247 @@
...
+  use field_list, only : field_list_t
+  use stats_quant, only : stats_quant_t
+  use registry, only : neko_registry
+  use scratch_registry, only : neko_scratch_registry
+  implicit none
+  private
+
+  type, public, extends(stats_quant_t) :: fluid_sgs_stats_t
+     !> Work fields
+     type(field_t) :: stats_work
```
timofeymukha, line 51: Since these are all temporaries, and you use scratch, there is no need to store them in the type, just declare them in the update routine.

src/les/fluid_sgs_stats.f90
```diff
@@ -0,0 +1,247 @@
...
+
+    if (present(nut_field)) then
+       this%nut => neko_registry%get_field_by_name(trim(nut_field))
+    else
+       this%nut => neko_registry%get_field_by_name('nut')
+    end if
+
+
+    ! Initialize work fields
+    call this%stats_work%init(this%u%dof, 'stats')
```
timofeymukha, line 124: Can we use the scratch for this one as well?

src/les/scalar_sgs_stats.f90
```diff
@@ -0,0 +1,239 @@
...
+     type(mean_field_t) :: alphat_mean !< <alphat>
+
+     type(mean_field_t) :: alphatdsdx !< <alphat*dsdx>
+     type(mean_field_t) :: alphatdsdy !< <alphat*dsdy>
+     type(mean_field_t) :: alphatdsdz !< <alphat*dsdz>
+
+     !> gradients
+     type(field_t), pointer :: dsdx_work
+     type(field_t), pointer :: dsdy_work
+     type(field_t), pointer :: dsdz_work
```
timofeymukha, line 69: Same here, no need to have them inside the type, right? also for stats_work here, can that be a scratch field?

src/simulation_components/fluid_sgs_stats_simcomp.f90
```diff
@@ -0,0 +1,271 @@
...
+       write(log_buf, '(A,E15.7)') 'Simulation start time:', time%start_time
+       call neko_log%warning(log_buf)
+       write(log_buf, '(A,E15.7)') 'Fluid stats start time:', this%start_time
+       call neko_log%warning(log_buf)
+       write(log_buf, '(A)') 'Assigning the statistics start time to ' &
+            // 'the simulation start time.'
+       call neko_log%warning(log_buf)
+       this%start_time = time%start_time
+       this%time = time%start_time
+    end if
```
timofeymukha, line 242: Can this be moved to the constructor?

src/simulation_components/scalar_sgs_stats_simcomp.f90
```diff
@@ -0,0 +1,360 @@
...
+       write(log_buf, '(A,E15.7)') 'Simulation start time:', time%start_time
+       call neko_log%warning(log_buf)
+       write(log_buf, '(A,E15.7)') 'scalar stats start time:', this%start_time
+       call neko_log%warning(log_buf)
+       write(log_buf, '(A)') 'Assigning the statistics start time to ' &
+            // 'the simulation start time.'
+       call neko_log%warning(log_buf)
+       this%start_time = time%start_time
+       this%time = time%start_time
+    end if
```
timofeymukha, line 328: Same here

src/simulation_components/scalar_sgs_stats_simcomp.f90
```diff
@@ -0,0 +1,360 @@
...
+       call this%stats%update(delta_t)
+       call MPI_Barrier(NEKO_COMM, ierr)
+       this%time = t
+
+       sample_time = MPI_WTIME() - sample_start_time
+
+       call neko_log%section('scalar stats')
+       write(log_buf, '(A,E15.7)') 'Sampling at time:', t
+       call neko_log%message(log_buf)
+       write(log_buf, '(A33,E15.7)') 'Simulationtime since last sample:', &
```
timofeymukha:
```suggestion
       write(log_buf, '(A33,E15.7)') 'Simulation time since last sample:', &
```

src/simulation_components/fluid_sgs_stats_simcomp.f90
```diff
@@ -0,0 +1,271 @@
...
+       call this%stats%update(delta_t)
+       call MPI_Barrier(NEKO_COMM, ierr)
+       this%time = t
+
+       sample_time = MPI_WTIME() - sample_start_time
+
+       call neko_log%section('Fluid SGS stats')
+       write(log_buf, '(A,E15.7)') 'Sampling at time:', t
+       call neko_log%message(log_buf)
+       write(log_buf, '(A33,E15.7)') 'Simulationtime since last sample:', &
```
timofeymukha:
```suggestion
       write(log_buf, '(A33,E15.7)') 'Simulation time since last sample:', &
```

timofeymukha (changes requested):
Hi Shiyu,

Some comments about temproraries. There is also a number of issues Codex found, which I think, upon inspection, all have merit. Here comes:

1. **High: `scalar_sgs_stats` crashes/behaves incorrectly in `alphat` mode**  
`scalar_sgs_stats_update` always executes `field_cmult2(this%alphat, this%nut, 1.0_rp / this%pr_turb)` even when initialized via `init_alphat`, where `this%nut` is never set. That dereferences a null pointer and also overwrites provided `alphat` data.  
Reference: `src/les/scalar_sgs_stats.f90` line 188.

2. **Medium: memory ownership leak in `scalar_sgs_stats_free` (nut-based init path)**  
`init_nut` allocates `this%alphat`, but `free` only nullifies it; it never calls `%free()` and `deallocate`. This leaks heap memory if the object is re-initialized/freed multiple times.  
References: `src/les/scalar_sgs_stats.f90` lines 148-149, 223.

3. **Medium: `stat_fields` allocations are not released in new SGS stats destructors**  
Both new stats types allocate `this%stat_fields%items(...)` but do not call `this%stat_fields%free()` in destructor, unlike `fluid_stats_t`.  
References: `src/les/fluid_sgs_stats.f90` lines 136, 210-230; `src/les/scalar_sgs_stats.f90` lines 119, 161, 210-225; comparison: `src/fluid/fluid_stats.f90` line 659.

4. **Low: user-guide docs do not match implemented JSON contract for `scalar_sgs_stats`**  
Implementation requires an `alphat` sub-dictionary with `nut_dependency` and either `alphat_field` or (`Pr_t`, `nut_field`), but the parameter table still documents top-level `alphat_field`, and default filename text is also wrong (`fluid_sgs_statsX*`).  
References: `src/simulation_components/scalar_sgs_stats_simcomp.f90` lines 121-128; `doc/pages/user-guide/statistics-guide.md` lines 307, 311.

5. **Low: review-policy/doc quality issues in new/changed Fortran files**  
- Copyright year inconsistency (new file uses `2025`; modified `coef.f90` still `2020-2024` despite 2026 edits).  
- Several new procedures have incomplete/inaccurate Doxygen params (for example scalar simcomp constructor docs mention `u,v,w` though dummies are `s,coef,...`).  
References: `src/les/scalar_sgs_stats.f90` line 1; `src/sem/coef.f90` line 1; `src/simulation_components/scalar_sgs_stats_simcomp.f90` lines 155-163, 211-220.

src/les/fluid_sgs_stats.f90
Shiyu-Sandy-Du: yes I think we talked about it at some point, and the conclusion was we address it in a separate PR since it also appears in fluid and scalar stats

src/les/fluid_sgs_stats.f90
Shiyu-Sandy-Du: yes, we address it in a separate PR together with fluid and scalar stats

src/les/scalar_sgs_stats.f90
Shiyu-Sandy-Du: yes, we address it in a separate PR together with fluid and scalar stats

timofeymukha:
@Shiyu-Sandy-Du , two more comments from the AI, as per latest changes, both correct I think.

1. **Undefined behavior in first constructor call due uninitialized `nut_dependency`**
   - In `src/les/scalar_sgs_stats.f90:56`, `nut_dependency` has no default initialization.
   - Both constructors call `this%free()` first, and `free()` branches on that flag at `src/les/scalar_sgs_stats.f90:226`.
   - If the uninitialized value is `.true.`, `free()` dereferences `this%alphat` (potentially disassociated), which can crash.
   - Suggested fix: initialize `nut_dependency = .false.` in type declaration and guard with `associated(this%alphat)` before calling `this%alphat%free()`.

2. **Owned `alphat` target is orphaned (leak) in nut-dependent mode**
   - Nut-dependent init allocates a new field target at `src/les/scalar_sgs_stats.f90:151`.
   - Destructor only calls `this%alphat%free()` and then `nullify(this%alphat)` (`src/les/scalar_sgs_stats.f90:227`, `src/les/scalar_sgs_stats.f90:232`); it never `deallocate(this%alphat)` when owned.
   - This loses the allocated target object and leaks across reinitializations.

Shiyu-Sandy-Du:
> @Shiyu-Sandy-Du , two more comments from the AI, as per latest changes, both correct I think.
> 
>     1. **Undefined behavior in first constructor call due uninitialized `nut_dependency`**
>        
>        * In `src/les/scalar_sgs_stats.f90:56`, `nut_dependency` has no default initialization.
>        * Both constructors call `this%free()` first, and `free()` branches on that flag at `src/les/scalar_sgs_stats.f90:226`.
>        * If the uninitialized value is `.true.`, `free()` dereferences `this%alphat` (potentially disassociated), which can crash.
>        * Suggested fix: initialize `nut_dependency = .false.` in type declaration and guard with `associated(this%alphat)` before calling `this%alphat%free()`.
> 
>     2. **Owned `alphat` target is orphaned (leak) in nut-dependent mode**
>        
>        * Nut-dependent init allocates a new field target at `src/les/scalar_sgs_stats.f90:151`.
>        * Destructor only calls `this%alphat%free()` and then `nullify(this%alphat)` (`src/les/scalar_sgs_stats.f90:227`, `src/les/scalar_sgs_stats.f90:232`); it never `deallocate(this%alphat)` when owned.
>        * This loses the allocated target object and leaks across reinitializations.

For its second comment, should we go through the entire neko to correct similar issues? E.g. u v w in fluid_scheme

timofeymukha:
> > @Shiyu-Sandy-Du , two more comments from the AI, as per latest changes, both correct I think.
> > ```
> > 1. **Undefined behavior in first constructor call due uninitialized `nut_dependency`**
> >    
> >    * In `src/les/scalar_sgs_stats.f90:56`, `nut_dependency` has no default initialization.
> >    * Both constructors call `this%free()` first, and `free()` branches on that flag at `src/les/scalar_sgs_stats.f90:226`.
> >    * If the uninitialized value is `.true.`, `free()` dereferences `this%alphat` (potentially disassociated), which can crash.
> >    * Suggested fix: initialize `nut_dependency = .false.` in type declaration and guard with `associated(this%alphat)` before calling `this%alphat%free()`.
> > 
> > 2. **Owned `alphat` target is orphaned (leak) in nut-dependent mode**
> >    
> >    * Nut-dependent init allocates a new field target at `src/les/scalar_sgs_stats.f90:151`.
> >    * Destructor only calls `this%alphat%free()` and then `nullify(this%alphat)` (`src/les/scalar_sgs_stats.f90:227`, `src/les/scalar_sgs_stats.f90:232`); it never `deallocate(this%alphat)` when owned.
> >    * This loses the allocated target object and leaks across reinitializations.
> > ```
> 
> For its second comment, should we go through the entire neko to correct similar issues? E.g. u v w in fluid_scheme

If we find something, then yes. But u, v, w in fluid_scheme should be fine, because those pointers just point to fields in the neko_reigstry. One never runs an `allocate` statement on those pointers in the way you do here.

src/simulation_components/fluid_sgs_stats_simcomp.f90
Shiyu-Sandy-Du: yes but I think it should be addressed together with fluid and scalar stats in another PR

timofeymukha (approved):
I think we are good to go. 

One thing to keep in mind for later is the registration of stats to the registry, that @vbaconnet added. We should have it here as well, I suppose, with the associated name mangling via `%name`

PR #2287, opened 2026-02-02
Fixing tgv after adding %name to simcomps

adampep:
There seems to be some inconsistency in simcomps right now, as "computed_field" seems to be still present in a number of them.

### DIFF
--- a/examples/tgv/tgv.case
+++ b/examples/tgv/tgv.case
@@ -49,7 +49,7 @@
             {
                 "type": "curl",
                 "fields": [ "u", "v", "w" ],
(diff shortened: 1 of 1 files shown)


timofeymukha:
You are right. In that PR I started out with the curl simcomp, and set out to mangle the names appropriately. But then I realized it needs some thinking, so in the rest of the simcomps the %name is dormant. So, my bad. But let's merge this in so that the case runs.

timofeymukha (approved):

njansson (approved):

PR #2292, opened 2026-02-03
Appendix describing the "fld" format

timofeymukha:
Dep on #2285 by accident, but here only fld-format.md is to be reviewed. The text is generated by AI, based on our IO code, with some minor editing from my side. I am not very well-versed in the format, so those well-familiar with it, please critically review this.

I know there are different versions of the format around, it would be good if this appendix tracks exactly the specification we use in Neko.

### DIFF
--- a/doc/pages/appendices/fld-format.md
+++ b/doc/pages/appendices/fld-format.md
@@ -1,0 +1,142 @@
+# Neko .fld field format {#fld-format}
+
+\tableofcontents
+
+This appendix documents the Nek5000 file format as written and read by Neko in
+``src/io/fld_file.f90``. We refer to this format as `.fld` for brevity. Neko
+writes ``.fld`` files using MPI-IO, and the layout on disk follows the code
+paths in that module.
+
+## File naming and series
+
+Neko writes a file series in Nek5000 style:
+
+- Data files: ``<base>0.f#####`` (zero-padded index)
+- Metadata file: ``<base><start_counter>.nek5000``
+
+The ``.nek5000`` file contains the filetemplate and time range for the series
+and is used by Neko when reading a series.
+
+## File overview
+
+Each ``.fld`` data file consists of:
+
+1. **Header** (132 characters)
+2. **Test pattern** (single precision, value ``6.54321``)
+3. **Element index list** (``glb_nelv`` integers)
+4. **Field blocks** (mesh, velocity, pressure, temperature, scalars)
+5. **Metadata blocks** (per-element min/max, only for 3D)
+
+The presence and ordering of the field blocks is described by ``rdcode`` in the
+header.
+
+## Header
+
+The header is a 132-character record written with the format:
+
+```
+#std <fld_data_size> <lx> <ly> <lz> <glb_nelv> <glb_nelv> <time> <step>
+     <1> <1> <rdcode[10]>
+```
+
+In code, this is:
+
+```
+'#std', i1, i2, i2, i2, i10, i10, e20.13, i9, i6, i6, 10a
+```
+
+Fields:
+
+- ``fld_data_size``: byte size of field data (``MPI_REAL_SIZE`` or
+  ``MPI_DOUBLE_PRECISION_SIZE``)
+- ``lx, ly, lz``: polynomial orders (GLL grid size per element direction)
+- ``glb_nelv``: global number of elements (written twice)
+- ``time``: simulation time
+- ``step``: output counter
+- The two ``i6`` values are currently written as ``1``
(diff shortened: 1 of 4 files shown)


doc/pages/appendices/fld-format.md
```diff
@@ -0,0 +1,142 @@
...
+
+The presence and ordering of the field blocks is described by ``rdcode`` in the
+header.
+
+## Header
+
+The header is a 132-character record written with the format:
+
+```
+#std <fld_data_size> <lx> <ly> <lz> <glb_nelv> <glb_nelv> <time> <step>
```
adperezm, line 38: I do not know if we should mention that in practice the data size is 4 or 8 for single, double respectively. Maybe that is understood, but I have seen people confused.

doc/pages/appendices/fld-format.md
```diff
@@ -0,0 +1,142 @@
...
+- ``step``: output counter
+- The two ``i6`` values are currently written as ``1``
+- ``rdcode``: 10-character field descriptor
+
+### rdcode
+
+``rdcode`` is a compact descriptor of which fields are present and in which
+order:
+
+- ``X``: mesh coordinates
```
adperezm, line 64: I do not know if we need to mention that if, for example, X is not present, then it means that the mesh is not written.

Again, it is maybe understood but just saying.

doc/pages/appendices/fld-format.md
```diff
@@ -0,0 +1,142 @@
...
+- Scalar fields: ``2 * glb_nelv`` values
+  - ``min, max`` per element
+
+For 2D output, these metadata blocks are not written.
+
+## 2D fields and Z-velocity
+
+When writing 2D fields, Neko sets ``gdim = 2`` (``lz = 1``). If a Z-velocity
+component exists, it is stored as an additional scalar field in the output.
+This is a semantic convention rather than a special file layout.
```
adperezm, line 134: We should make sure that this happens automatically. I guess martin fixed it that way.

Because in theory, you can still have a Z component of velocity in a 2D file but it will never be seen by, e.g., visit.

adperezm (approved):

adperezm:
It looks good, so feel free to ignore my comments and merge.

doc/pages/appendices/fld-format.md
timofeymukha: I think it is good to add these small things. ALso what you mentioned in the beeting, that you must always have u,v,w (for 3d), and just 1 component is not allowed.

doc/pages/appendices/fld-format.md
```diff
@@ -0,0 +1,142 @@
...
+paths in that module.
+
+## File naming and series
+
+Neko writes a file series in Nek5000 style:
+
+- Data files: ``<base>0.f#####`` (zero-padded index)
+- Metadata file: ``<base><start_counter>.nek5000``
+
+The ``.nek5000`` file contains the filetemplate and time range for the series
```
vbaconnet:
```suggestion
The ``.nek5000`` file contains the file template and time range for the series
```

doc/pages/appendices/fld-format.md
```diff
@@ -0,0 +1,142 @@
...
+and is used by Neko when reading a series.
+
+## File overview
+
+Each ``.fld`` data file consists of:
```
vbaconnet: Should we add somewhere that the files are binary?

vbaconnet (approved):
Cool stuff!

doc/pages/appendices/fld-format.md
pschlatt1: and that the typical endianness is little endian (I guess).

But I had a question: I know we dicussed this, but I probably forgot the outcome, but do we really want to call these "fld" files? The fld files were the old ASCII Nekton files. I typically call it the f files. 

I know we talked about that, and I think @adampep had a good solution at the time, but I don't remember it.

timofeymukha: Not really, the solution was to continue with fld, because it is everywhere, including the typenames in the code. But also be clear what we mean by that.

adperezm:
Are we missing something here?

adperezm:
> Are we missing something here?

Oh, I guess we are waiting for #2285 :)

timofeymukha:
> > Are we missing something here?
> 
> Oh, I guess we are waiting for #2285 :)

Right, I actually forgot about that :-))

timfelle (changes requested):
No changelog update

timfelle (approved):

PR #2298, opened 2026-02-04
Feature/full element point zones

adperezm:
Added the possibility to mark all points in an element if even one of them meet the criterium.

I plan to use this to initialize interpolators with only parts of the domain. And to eventually write out masked fields.

I decided to keep point masks instead of having "element_masks" as a conscious decision. I do not think we need those type of mask that store element ids instead of points ids.


src/mesh/point_zones/box_point_zone.f90
vbaconnet, line 104: (docstring)
```suggestion
  !! @param full_elements Whether to mark all points in the element containing
  !! points that satisfy the criterion.
  !! @param xmin Lower x-bound of the box coordinates.
```

src/mesh/point_zones/cylinder_point_zone.f90
```diff
@@ -105,17 +107,19 @@ end subroutine cylinder_point_zone_init_from_json
   !! @param p0 Coordinates of the first endpoint.
```
vbaconnet, line 109: (docstring)
```suggestion
  !! @param full_elements Whether to mark all points in the element containing
  !! points that satisfy the criterion.
  !! @param p0 Coordinates of the first endpoint.
```

src/mesh/point_zones/sphere_point_zone.f90
vbaconnet, line 100: (docstring)
```suggestion
  !! @param full_elements Whether to mark all points in the element containing
  !! points that satisfy the criterion.
  !! @param x0 Sphere center's x-coordinate.
```

src/mesh/point_zone.f90
```diff
@@ -204,11 +207,12 @@ end subroutine register_point_zone
   !! @param name Name of the point zone.
   !! @param invert Flag to indicate wether or not to invert the selection
   !! of points.
-  subroutine point_zone_init_base(this, size, name, invert)
+  subroutine point_zone_init_base(this, size, name, invert, full_elements)
```
vbaconnet, line 212: (docstring)
```suggestion
  !! @param full_elements Whether to mark all points in the element containing
  !! points that satisfy the criterion.
  subroutine point_zone_init_base(this, size, name, invert, full_elements)
```

vbaconnet (approved):
Very very cool considering the possibilities that this can open, really excited to try it out.

timofeymukha (approved):

PR #2299, opened 2026-02-04
Fix uniqueness check for simulation component names

Shiyu-Sandy-Du:


timofeymukha:
Barring the string to int comparison, did you encounter a simcomp with an empty name? How did that occur?

Shiyu-Sandy-Du:
> Barring the string to int comparison, did you encounter a simcomp with an empty name? How did that occur?

Oh wait, it's my fault on another branch (I didn't assign name to sgs_stats simcomps).

PR #2302, opened 2026-02-05
Enable unit and integration tests for intel

timofeymukha:
I will regret getting into this again :D. But would be nice to put in place.


timofeymukha:
The tests start running but hang for infinity, perfect :D

timfelle:
Current progress: 

dofmap.f90: Dofmap is not constructed correctly. Incorrect values are returned from function `dofmap_facetidx`.

We need to investigate contents of inputs to make sure `tuple4_i4_t` behaves correctly.

njansson:
> Current progress:
> 
> dofmap.f90: Dofmap is not constructed correctly. Incorrect values are returned from function `dofmap_facetidx`.
> 
> We need to investigate contents of inputs to make sure `tuple4_i4_t` behaves correctly.

This is the typical issue for Intel, and now also the issue that is shown in #2324 for llvm based CCE. I wonder if this is related to the assignment of these tuples in the hash table. 

Also in the past there were some issues with the select type for `tuple4_i4_t` and `tuple_i4_t`. even with both types present in own select clauses, some compilers (incorrectly?) matched them to the same clause since they were derived  from `tuple_t` (hence the current obfuscated construct in htable)

PR #2303, opened 2026-02-05
inverse the direction of gravity vector for Boussinesq

Shiyu-Sandy-Du:
In that way, we can always prescribe our gravity vector to the ground in an ABL simulation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Develop
+- *BREAKING* The sign of the Boussinesq source term is fixed such that the input
+  gravity vector could be prescribed correctly.
 - Added an option for writing the mesh in every output field file.
 - *BREAKING* All simcomps now have a `name` keyword in the case file. A default
   name is assigned, but all `name`s must be unique. If you have two or more
--- a/examples/rayleigh_benard_cylinder/rayleigh.case
+++ b/examples/rayleigh_benard_cylinder/rayleigh.case
@@ -25,7 +25,7 @@
                 {
                     "type": "boussinesq",
                     "scalar_field": "temperature",
-                    "g": [ 0.0, 0.0, 1.0 ],
+                    "g": [ 0.0, 0.0, -1.0 ],
                     "reference_value": 0.0,
                     "beta": 1.0
                 }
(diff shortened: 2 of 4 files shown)


timofeymukha (commented):
Darn, this is correct. If z is up, g should be negative in z for hot fluid to rise....
This change will be very difficult to notice if you have an old case using the Boussinesq , because it is not like we can throw a warning or something :(((.
I wish we caught this much earlier!

But feels like this has to be done :(. Make sure to add a breaking change notice to the changleog.


@adperezm any thoughts?

timofeymukha:
Also gotta make sure the correct sign is used in all the buoyancy-related ABL stuff being added.

Shiyu-Sandy-Du:
> Also gotta make sure the correct sign is used in all the buoyancy-related ABL stuff being added.

I commented on the buoyancy-corrected Vreman model, do we have other stuffs apart from Deardorff?

linneahuusko:
> > Also gotta make sure the correct sign is used in all the buoyancy-related ABL stuff being added.
> 
> I commented on the buoyancy-corrected Vreman model, do we have other stuffs apart from Deardorff?

I don't think so, not yet. But we need to keep this in mind when writing the wall models also. I'm really annoyed I didn't think of it before, but I guess for the ABL g is usually viewed as a scalar and the sign is included elsewhere in the equations

linneahuusko (approved):

adperezm:
> Darn, this is correct. If z is up, g should be negative in z for hot fluid to rise.... This change will be very difficult to notice if you have an old case using the Boussinesq , because it is not like we can throw a warning or something :(((. I wish we caught this much earlier!
> 
> But feels like this has to be done :(. Make sure to add a breaking change notice to the changleog.
> 
> @adperezm any thoughts?

Yeah, this looks good. I guess I would have seen it, but I was setting the source term in the old way in the userfile haha.

adperezm (approved):
Again, This was conceptually wrong but the RBC example worked because the source term was added with the right sign.

It is a good catch and we should change it.

timofeymukha (changes requested):
Add a line to the changelog please, mark as breaking.

timofeymukha (approved):

PR #2306, opened 2026-02-06
Feature/names for bcs

adperezm:
As the title says. The idea is to be able to access the mask easier, at least from the user module. One can do:

``` fortran   
    bc_list => neko_user_access%case%fluid%bcs_vel
    bc => bc_list%get_from_zone_index(3)
```

Or 

``` fortran   
    bc_list => neko_user_access%case%fluid%bcs_vel
    bc => bc_list%get_from_name("wall")
```

For the latter to work on would need to add the name in the case file:

``` json
   {
    "type": "no_slip",
    "zone_indices": [ 3 ],
    "name" : "wall"
   }
```

What I don't like is that I need to know if it is in bcs_vel or bcs_prs but it is what it is for now.

This might be redundant if one knows that the bcs are added in the order you put them in the case file, bu I find this useful at least for a very particular thing I am doing haha. I do not think it hurts to have it but let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Changelog
 
 ## Develop
-- Added possibility to assign names to boundary conditions in the case file. The `bc_list_t` now supports item retrieval by name or zone_index.
+- Added possibility to assign names to boundary conditions in the case file. The 
+  `bc_list_t` now supports item retrieval by name or zone_index.
 - Added the `full_elements` option to point_zones. Allows including all points
   in an element in the mask.
 - *BREAKING* The sign of the Boussinesq source term is fixed such that the input
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Develop
+- Added posibility to retrieve BCs from names and zone_indices.
 - Added an option for writing the mesh in every output field file.
 - *BREAKING* All simcomps now have a `name` keyword in the case file. A default
   name is assigned, but all `name`s must be unique. If you have two or more
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Develop
-- Added posibility to retrieve BCs from names and zone_indices.
+- Added possibility to assign names to boundary conditions in the case file. The `bc_list_t` now supports item retrieval by name or zone_index.
 - Added an option for writing the mesh in every output field file.
 - *BREAKING* All simcomps now have a `name` keyword in the case file. A default
   name is assigned, but all `name`s must be unique. If you have two or more
--- a/doc/pages/user-guide/case-file.md
+++ b/doc/pages/user-guide/case-file.md
@@ -386,7 +386,7 @@ the streamwise direction. Only one condition, corresponding to zone index 3 (the
 wall) is the specified in the case file.
 
 It is possible to assign specific names to the boundary conditions through the
-`name` keyword. Boundary conditions can then be retireved in the code by using
+`name` keyword. Boundary conditions can then be retrieved in the code by using
 the name or the `zone_index` where it is applied.
 
 #### Available conditions
--- a/doc/pages/user-guide/case-file.md
+++ b/doc/pages/user-guide/case-file.md
@@ -385,7 +385,7 @@ periodic pipe simulation, two periodic zones comprise the boundary conditions in
 the streamwise direction. Only one condition, corresponding to zone index 3 (the
 wall) is the specified in the case file.
 
-It is possible to name specific names to the boundary conditions through the
+It is possible to assign specific names to the boundary conditions through the
 `name` keyword. Boundary conditions can then be retireved in the code by using
 the name or the `zone_index` where it is applied.
 
--- a/doc/pages/user-guide/case-file.md
+++ b/doc/pages/user-guide/case-file.md
@@ -386,8 +386,12 @@ the streamwise direction. Only one condition, corresponding to zone index 3 (the
 wall) is the specified in the case file.
 
 It is possible to assign specific names to the boundary conditions through the
-`name` keyword. Boundary conditions can then be retrieved in the code by using
+`name` keyword. Boundary conditions can then be retireved in the code by using
 the name or the `zone_index` where it is applied.
+
+The default name of the boundary conditions is given by the `<variable>_bc_<zone_index>`
+pattern. i.e., the pressure boundary condition that applies in zone index 5 can be
+retrieved by the `pressure_bc_5` name.
 
 #### Available conditions
 The conditions to apply is specified by `type` keyword inside each of the JSON
--- a/doc/pages/user-guide/case-file.md
+++ b/doc/pages/user-guide/case-file.md
@@ -385,6 +385,10 @@ periodic pipe simulation, two periodic zones comprise the boundary conditions in
 the streamwise direction. Only one condition, corresponding to zone index 3 (the
 wall) is the specified in the case file.
 
+It is possible to name specific names to the boundary conditions through the
+`name` keyword. Boundary conditions can then be retireved in the code by using
+the name or the `zone_index` where it is applied.
+
 #### Available conditions
 The conditions to apply is specified by `type` keyword inside each of the JSON
 objects. The full list of possible conditions for the fluid is specified in the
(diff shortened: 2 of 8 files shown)


doc/pages/user-guide/case-file.md
```diff
@@ -385,6 +385,10 @@ periodic pipe simulation, two periodic zones comprise the boundary conditions in
 the streamwise direction. Only one condition, corresponding to zone index 3 (the
 wall) is the specified in the case file.
 
+It is possible to name specific names to the boundary conditions through the
```
timofeymukha:
```suggestion
It is possible to assign specific names to the boundary conditions through the
```

src/bc/bc_list.f90
```diff
@@ -63,6 +63,10 @@ module bc_list
      procedure, pass(this) :: append => bc_list_append
      !> Get the item at the given index.
      procedure, pass(this) :: get => bc_list_get
+     !> Get the item with the given name.
+     procedure, pass(this) :: get_from_name => bc_list_get_from_name
+     !> Get the item that applies to the given zone_index.
+     procedure, pass(this) :: get_from_zone_index => bc_list_get_from_zone_index
```
timofeymukha: Use `_by_` intstead of `_from_` to be the same as in the registry_t and field_list_t.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Develop
+- Added posibility to retrieve BCs from names and zone_indices.
```
timofeymukha:
```suggestion
- Added possibility to assign names to boundary conditions in the case file. The `bc_list_t` now supports item retrieval by name or zone_index.
```

src/fluid/euler_bc_fctry.f90
```diff
@@ -85,6 +85,8 @@ module subroutine density_bc_factory(object, scheme, json, coef, user)
     do i = 1, size(zone_indices)
        call object%mark_zone(coef%msh%labeled_zones(zone_indices(i)))
     end do
+    call json_get_or_default(json, "name", object%name, "density_bc")
```
timofeymukha: It would be good if the factories assigned unique names, otherwise the bc_list_t will return the first occurring one. What we can use is that the bcs necessarily partition the zone indices, so a unique name can be created from the zone_indices. Like `bc_` + concatenated zone indices? Just an idea.

src/bc/bc.f90
```diff
@@ -85,6 +85,10 @@ module bc
      !> Indicates wether the bc has been updated, for those BCs that need
      !! additional computations
      logical :: updated = .false.
+     !!> Name of the bc
+     character(len=:), allocatable :: name
+     !!> Zone indices where the bc is applied
+     integer, allocatable :: zone_indices(:)
    contains
```
timofeymukha, line 93: Highjacking this for a general comment. The functionality is very nice, I have been planning to add this myself :-). An issue we have is that bc_t does not have to be constructed from a zone at all, the mask can be built in a number of different ways. So, I planned to push bc_t to a lower step in the hierarchy, and make that type only be constructable from a zone_index list. 

However, I think we can push things forward and merge this PR, but what we need to add is making sure that name and zone_indices are always allocated and destroyed. The deallocation should be put into `bc_t%free_base`. And I think we can use `bc_t%finilize_base` to check whether name and zone_indices are allocated, and, if not, allocate to an empty string and... not sure... one-element array with a 0? The latter as a temporary indicator that the bc was not built from a zone.

adperezm: That sounds good. I will add that.

adperezm: Updated now.

src/bc/bc.f90
```diff
@@ -561,6 +573,16 @@ subroutine bc_finalize_base(this, only_facets)
             HOST_TO_DEVICE, sync = .true.)
     end if
 
+    if (.not. allocated(this%name)) then
+       allocate(character(len=1) :: this%name)
+       this%name = ""
```
timofeymukha: Is this len=1 or 0 :)? I think for strings you can just have the = and it will autoallocate the right length, or?

timofeymukha (approved):

src/bc/bc.f90
adperezm: Haha I guess I allocated of lenght 1 but did not fill it :).

I can change it if you want. I also think it allocates with = but I am afraid of everything in Fortran nowadays. Too much python lately :P

src/bc/bc.f90
```diff
@@ -85,6 +85,10 @@ module bc
      !> Indicates wether the bc has been updated, for those BCs that need
      !! additional computations
      logical :: updated = .false.
+     !!> Name of the bc
```
timfelle, line 89:
```suggestion
     !> Name of the bc
```

src/bc/bc.f90
```diff
@@ -85,6 +85,10 @@ module bc
...
+     character(len=:), allocatable :: name
+     !!> Zone indices where the bc is applied
```
timfelle, line 91:
```suggestion
     !> Zone indices where the bc is applied
```

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Develop
+- Added possibility to assign names to boundary conditions in the case file. The `bc_list_t` now supports item retrieval by name or zone_index.
```
timfelle: Consistent formatting to the rest
```suggestion
- Added possibility to assign names to boundary conditions in the case file. The 
  `bc_list_t` now supports item retrieval by name or zone_index.
```

timfelle (commented):
Just a couple of small comments for now

doc/pages/user-guide/case-file.md
```diff
@@ -385,6 +385,10 @@ periodic pipe simulation, two periodic zones comprise the boundary conditions in
 the streamwise direction. Only one condition, corresponding to zone index 3 (the
 wall) is the specified in the case file.
 
+It is possible to assign specific names to the boundary conditions through the
+`name` keyword. Boundary conditions can then be retireved in the code by using
```
vbaconnet, line 389:
```suggestion
`name` keyword. Boundary conditions can then be retrieved in the code by using
```

src/fluid/euler_bc_fctry.f90
```diff
@@ -125,6 +134,11 @@ module subroutine pressure_bc_factory(object, scheme, json, coef, user)
     do i = 1, size(zone_indices)
        call object%mark_zone(coef%msh%labeled_zones(zone_indices(i)))
     end do
+
+    write(buf,'("pressure_bc_",I0)') zone_indices(1)
```
vbaconnet, line 138: Is this the kind of thing one might want to log at the NEKO_DEBUG level?

vbaconnet (approved):

src/fluid/euler_bc_fctry.f90
vbaconnet: just thinking that hardcoded strings like that should be somehow communicated to the user in some way (could also be documentation).

timfelle: > just thinking that hardcoded strings like that should be somehow communicated to the user in some way (could also be documentation). 

I completely agree. Generally it would be good if each BC would report on itself, similar to a lot of the other logging we do.
And the name mangling pattern we decide on should be documented, possibly in the code style such that anyone writing a new bc would be able to find the pattern.

adperezm: Then I'm thinking that it is best if I, in fact, set the name to be an empty string as default.

The reason that I had not made the name unique in my initial PR is because I do not see anyone looking for a bc (by name) if they did not set the name themselves. Or I feel they shouldn't. In that case I would search for it by its zone index.

With that said. We could report when the user defined name is assigned just so things live for posterity in the log. But I disagree about the hard coded one. I will change it to an empty string in that case.

vbaconnet: Fair enough, I kind of liked how you did it but it's true that if there is no explicit naming from the user side then there is no reason for giving it a name by default. The only thing would be if we forced the bc names to be unique (like simcomps) but that is not the case here so.. dealer's choice :)

adperezm: Okay, so in the end I did as you asked :D. So it is reporting the name and I added some lines in the docs.

I decided to not put it in the DEBUG log level because I though It is nice to have in in the log, since maybe the case file can change from one run to the other.

adperezm:
If there are no further comments, lets please merge this.

PR #2309, opened 2026-02-09
Fix pkg_resources error by updating setuptools version

timfelle:
Adjust the setuptools version constraint to resolve the pkg_resources error.

> [!NOTE]
> Must be forced due to issues with python dependencies for `flint`.


adperezm (approved):

timofeymukha (approved):

timfelle:
It is only @njansson who can force push this stuff correct?

PR #2310, opened 2026-02-09
Add field-based source term

timofeymukha:
Adds a source term, which takes its values from fields in the registry. Mostly intended for scenarios where another part of the simulation is responsible for computing the source based on its own logic.

P.S. Does anyone remember what the `variable_name` argument in the constructor was supposed to be used for? I think it is not used anywhere atm.


timfelle:
> P.S. Does anyone remember what the `variable_name` argument in the constructor was supposed to be used for? I think it is not used anywhere atm.

Based on the docstring, I would expect it to be what the source term is applied to, Pressure, velocity, scalar or something.

src/source_terms/field_source_term.f90
```diff
@@ -0,0 +1,154 @@
...
+    end do
+
+  end subroutine field_source_term_init_from_components
+
+  !> Destructor.
+  subroutine field_source_term_free(this)
+    class(field_source_term_t), intent(inout) :: this
+
+    call this%free_base()
+    call this%registry_fields%free()
```
vbaconnet, line 136: The field list destructor also deallocates the field_t targets to which the pointers point to, isn't that dangerous? (see [`field_list_free()`](https://neko.cfd/docs/develop/d1/d13/field__list_8f90_source.html#l00138))

I think this is out of scope of your PR but I very much dislike this behavior of the field list destructor. The field list is populated with pointers to fields allocated externally, but for some reason its destructor takes on the responsibility of deallocating those fields. 

I guess it's a safety to avoid memory leakage but in this case you're literally freeing fields from the registry before the actual registry frees them... It's more about the principle than actually preventing bugs, although you never know.

vbaconnet (approved):
Cool :) A step closer to case file self-sufficiency? 

I just had a comment about the field_list destructor.. but since it's a bit out of scope I think we may skip it for now :)

src/source_terms/field_source_term.f90
timfelle: This is indeed a great point, and something we have been looking at for a bit now. For now, some progress is made in PR #2312, where we start to streamline wrapper and pointer types, to make it explicit if an object is just pointed to, or it is owned by the object.

timofeymukha: Yes, you are absolutely right @vbaconnet . In fact, we typically use field_list as a non-owning object to just gather pointers together. Tim has been working on this stuff, as you see, and I hope we sort it out soon and have clear conventions and guidelines!

vbaconnet: > For now, some progress is made in PR #2312, where we start to streamline wrapper and pointer types, to make it explicit if an object is just pointed to, or it is owned by the object.

Cool! I totally missed those discussions, sorry..

vbaconnet: Oh yea now it makes total sense, just had to read the PR a bit... that's great!

Shiyu-Sandy-Du (approved):

timfelle (changes requested):
No changelog update

timfelle (approved):

Issue #2301, opened 2026-02-05
Json type check inconsistency


arash-mp:
Currently,
checkpoint_control looks for a real input even if the control is "tsteps". So the following will lead to crash:
        "checkpoint_control": "tsteps",
        "checkpoint_value": 5000.0,

but at the same time, output_control looks for an integer when the control type is time_step. so the following works:
            "output_control": "tsteps",
            "output_value": 100

PR #2311, opened 2026-02-09
Fix checkpoint value parsing

timofeymukha:
Fix issue #2301.

Additionally, fixes some default values to be the same as in the documentation (no defaults).


adperezm (approved):

vbaconnet (approved):

timfelle (commented):
The checkpoint options are kinda funky now.

If a user does not specify `case.output_checkpoints` at all, then we expect they intend to save checkpoints. Therefore it is mandatory to have `case.checkpoint_control` in the case file, since we now removed the default value for that. Is that really intentional?Or would it not be more intuitive for the default of `case.output_checkpoints` to be false, since if a user wants checkpoints, they probably know how they want them to be spaced.

This currently cause the `rayleigh_benard` case to fail.

timofeymukha:
> The checkpoint options are kinda funky now.
> 
> If a user does not specify `case.output_checkpoints` at all, then we expect they intend to save checkpoints. Therefore it is mandatory to have `case.checkpoint_control` in the case file, since we now removed the default value for that. Is that really intentional? Or would it not be more intuitive for the default of `case.output_checkpoints` to be false, since if a user wants checkpoints, they probably know how they want them to be spaced.
> 
> This currently cause the `rayleigh_benard` case to fail.

Ugh, yeah, good catch! I guess both defaults are defensible; we can discuss at the meeting today!

timfelle (dismissed):
Also missing changelog update

timofeymukha:
@timfelle @adperezm @vbaconnet Should be good to go now, hopefully. Check out the description I added to the case file. Also, output_checkpoints is now mandatory, as we discussed.

doc/pages/user-guide/case-file.md
```diff
@@ -80,17 +80,46 @@ but also defines several parameters that pertain to the simulation as a whole.
...
+point, if one sets `output_at_end` to `true`, this will trigger a write as per
+usual. However, if `output_at_end` is `false`, Neko will still write a special
+checkpoint file, with the filename called `joblimit#####.chkp`. This is done so
+that the user is at least provided a restart file, and none of the computer time
+spent on the simulation is wasted. Generally, however, it is recommended to
+have `output_at_end` set to `true` in tandem with `job_timelimit`, so that what
+exactly gets written is controlled by the case file settings.
+
+
+
```
timfelle: Why so many empty lines?
```suggestion

```

src/case.f90
```diff
@@ -512,19 +512,19 @@ subroutine case_init_common(this)
             always_write_mesh = logical_val)
     end if
 
-    call json_get_or_default(this%params, 'case.fluid.output_control',&
-         string_val, 'org')
+    call json_get(this%params, 'case.fluid.output_control', string_val)
 
     if (trim(string_val) .eq. 'org') then
```
timfelle, line 517: I do not think we document that a user can set this to `org`.

src/case.f90
```diff
@@ -538,19 +538,29 @@ subroutine case_init_common(this)
     !
     ! Save checkpoints (if nothing specified, default to saving at end of sim)
     !
-    call json_get_or_default(this%params, 'case.output_checkpoints',&
-         logical_val, .true.)
+    call json_get(this%params, 'case.output_checkpoints',&
+         logical_val)
```
timfelle: No reason to break this line
```suggestion
    call json_get(this%params, 'case.output_checkpoints', logical_val)
```

src/case.f90
timofeymukha: Yeah, that was kind of intentional to be honest, it was always a bit of a legacy feature :P. Ideally, this whole setup should be moved to the fluid type, and without us having to pass the whole case json to it.

PR #2312, opened 2026-02-09
Add field wrapper and pointer types with usage guidelines

timfelle:
Introduce guidelines for implementing wrapper and pointer types. Add a Field wrapper and update the pointer management in field_list. Introduce field_array to manage the lifetime of its components effectively.

This is the first step in trying to manage how we deal with complex object lifetime in a streamlined fashion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
 ## Develop
+- Updated field types with a wrapper and ensure lifetime management of field
+  data in field lists and arrays.
+- Updated Developer Patterns documentation with new information on how to manage
+  pointers and lists of complex objects.
 - Added cache cleanup job for CI workflows upon PR closure.
 - Updated compiler check workflows to run on release branches and master.
 - Removed commented-out workflow sections.
--- a/doc/pages/developer-guide/dev_patterns.md
+++ b/doc/pages/developer-guide/dev_patterns.md
@@ -209,11 +209,59 @@ so, code repetition is fine, it will eventually cease to exist naturally.
    True code duplication is when a change in one place will also necessary a
    change in the other.
 
+## F. Complex and Polymorphic arrays.
 
+We often need to use arrays of complex or polymorphic types. Objects which
+support this, must implement a wrapper type following one of the two patterns
+below. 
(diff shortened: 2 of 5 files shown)


vbaconnet (approved):
great stuff! I assume that the intention is to also used this in the registry? In the `entry_t` type? Sorry if this is repeating some discussions that have already been had..

timfelle:
> great stuff! I assume that the intention is to also used this in the registry? In the `entry_t` type? Sorry if this is repeating some discussions that have already been had..

Don't be sorry, I am not sure the discussions were made in public anyhow 😅
Right now the `entry_t` in particular is designed to own the objects explicitly for the scratch and global registries. 

Are you missing a non-owning registry? I am not exactly sure how useful that would be.

vbaconnet:
> Are you missing a non-owning registry? I am not exactly sure how useful that would be.

For some reason I thought this was to replace the private allocatable field in the entry_t type, but I guess it’s not necessary

timfelle:
> For some reason I thought this was to replace the private allocatable field in the entry_t type, but I guess it’s not necessary

Ah, no this is to expand that particular way of doing things to more things, since it is much safer in terms of memory leaking and premature deallocation.

adperezm (approved):
Looks nice!

I have some questions to just keep in mind.

Did I misunderstand that now field_arrays will contain multiple fields that live in there, i.e., are created and destroyed there. While field lists are collections of pointers?

If I did not misunderstand, then I think the names are not that telling, but then, I would also not have any better names :). If I misunderstood, then of course ignore.

Similarly, I feel the name "field_wrapper_t" doesn't tell much. Maybe it shouldn't. But something like field manager or something else, no?

Lastly, in the case of the wrapper itself, maybe this is a stupid question, but why do we need to create the field_array with the wrapper and not the field directly? If the problem is deallocation and such, the field_array could deal with that directly no?

Again, I trust the process, and these things were probably already discussed, so feel free to disregard.

timofeymukha:
> Looks nice!
> 
> I have some questions to just keep in mind.
> 

We decided to keep the field_list_t name because changing that will break a gazzilion lines of code, including user :-)).

Regarding the last question, you cannot have an array of pointers in Fortran, you need to wrap.

PR #2313, opened 2026-02-10
Implement polymorphic wrapper for simulation components

timfelle:
Introduce a private polymorphic wrapper for simulation components to enhance memory management and encapsulation. 
Update related functions to utilize the new wrapper, ensuring proper initialization, cleanup, and allocation checks. 

Fix the TGV example to align with these changes.

Edit:
We remove the appending mechanism. It is not used, as the open factories superseded it. The object itself is now public, eliminating the getter, since protections were marginal at best.


timfelle (commented):
@timofeymukha, I am not sure we can move to a pointer based system for polymorphic objects, since the pointer system we discussed is not supported by our factories. Could you have a look and see what you think of this approach. It still encapsulates and moves management internal to the wrapper.

timofeymukha:
Haven't checked the code, but if this breaks tgv, don't forget to change the tgv cases in all the various tests, there are some .case scattered here and there :).

njansson (approved):

njansson:
Are we good to merge this?

timfelle:
> Are we good to merge this?

Well I think so, but I have not tested it on an extensive number of machines and compilers. So if you believe that is required for this then we need to do so.

src/simulation_components/simulation_component.f90
```diff
@@ -106,9 +106,20 @@ module simulation_component
   !> A helper type that is needed to have an array of polymorphic objects
   type, public :: simulation_component_wrapper_t
      class(simulation_component_t), allocatable :: simcomp
+   contains
+     !> Constructor. Just allocates the pointer.
```
vbaconnet: Docstring mismatch with line 501?
```suggestion
     !> Constructor. Initializes the object.
```

vbaconnet (approved):
I would try to compile on Cray if you have the possibility.. It usually comes up with unpredictable errors so if we can get that out of the way it would be nice. Otherwise I guess we can test on develop and push a fix

PR #2316, opened 2026-02-12
Modernize scalar management with factory pattern

timfelle:
Align with the updated way of doing polymorphic objects. PR #2313

Introduce a factory for scalar schemes to enhance the management of scalar transport equations. Update the scalar manager to utilize the new wrapper type for better encapsulation and flexibility.


src/scalar/scalar_scheme.f90
Shiyu-Sandy-Du:
```suggestion

```
We don't need it anymore, right?

timfelle: True

src/scalar/scalars.f90
Shiyu-Sandy-Du: so here should be scalar_scheme_wrapper_t

timfelle: I guess I royally screwed up the rebase 😅

timfelle: There we go. should be ready now

PR #2322, opened 2026-02-16
Bugfix/cyclic logic

IsaMammadli:
Fixed a bug in `coef_generate_cyclic_bc` and also added another protection layer with `coef_check_cyclic` logic. The logic works as following:
 1. we allocate a zero vector field 'norm' and assign values of surface normals only at the periodic boundaries.
 2. Then we apply direct stiffness summation (GS_OP_ADD).
 3. This corresponds to DSS in Cartesian coordinates and if sums to zero then everything is OK. 
 4.  If not then it applies DSS in TNB basis and if this sums to zero then cyclic rotation is required and must be switched on. This still assumes rotation around z-axis.
 5. If this doesnt also sum to zero then there will be warning  that "Cyclic is required but code must be modified" .

In general `coef_check_cyclic` is written in a way that it can decide on its own if cyclic 1) is not required 2)is required 3)is required but current "rotation around z-axis" assumption does not work. So it can be part of coef%init routine. But as we discussed, there are multiple coef_t objects and I didnt observe correct results from them except `fluid_pnpn%c_Xh` object. So it is limited again to this` fluid_pnpn%c_Xh` only.
 
Note that, we stop the program only when the cylic flag is set to true and (1)  there are no periodic boundary conditions or (2) cyclic is not required (DSS in Cartesian is sufficient)  and (3) current cyclic rotation logic does not suffice.  The remaining case is supposed to be when cyclic is true and geometry is aligned/deformed correctly for 2D rotation. With the case (2) we also solve the bug in coef_generate_cyclic_bc, but additional check is still there.

In case cyclic is false these checks will not be done, rotation angles will be R11=1 R12=0, corresponding to no rotation and since cyclic=false, rotate_cyc subroutine will be skipped.

I used `neko_eps` in few places but I might need to increase tolerances to be safer.


src/sem/coef.f90
```diff
@@ -1252,4 +1274,129 @@ subroutine coef_generate_cyclic_bc(coef)
...
+       else if (ipass .eq. 2 .and. norm_dss .eqv. .false.) then
+          call neko_error("Cylic rotation is required, but " // &
+                          "rotation logic must be modified.")
+       !if (ipass .eq. 2 .and. norm_dss .eqv. .true.)
+       !current logic is sufficient, proceed.
+       end if
+    end do
+    deallocate(normx)
+    deallocate(normy)
+    deallocate(normz)
```
timfelle, line 1397: Only the host arrays are freed. The device arrays must also be freed, otherwise
we get a pretty bad leak.

timfelle (changes requested):

src/sem/coef.f90
IsaMammadli: thanks! made the changes.

src/sem/coef.f90
```diff
@@ -1252,4 +1274,134 @@ subroutine coef_generate_cyclic_bc(coef)
...
+
+    allocate(normx(lx, lx, lx, this%msh%nelv))
+    allocate(normy(lx, lx, lx, this%msh%nelv))
+    allocate(normz(lx, lx, lx, this%msh%nelv))
+
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_map(normx, normx_d, ntot)
+       call device_map(normy, normy_d, ntot)
+       call device_map(normz, normz_d, ntot)
+    end if
```
njansson, line 1321: Maybe we should stash these arrays and pointers in the `coef` type, such that we don't have to reallocate them for each call to check cyclic (which now is done for each time-step in pnpn

timofeymukha: Isnt it only called in init?

njansson: Correct, my bad

src/sem/coef.f90
```diff
@@ -39,14 +39,16 @@ module coefs
...
-  use device_math, only : device_rone, device_invcol1, device_glsum
+  use device_math, only : device_rone, device_invcol1, &
+       device_glsum, device_absval
   use device_coef, only : device_coef_generate_geo, &
        device_coef_generate_dxydrst
   use mxm_wrapper, only : mxm
   use device
-  use utils, only : index_is_on_facet, linear_index
+  use utils, only : index_is_on_facet, linear_index, &
+       neko_error, neko_warning
```
timfelle: Neko warning is never called
```suggestion
       neko_error
```

src/sem/coef.f90
```diff
@@ -1230,19 +1241,30 @@ subroutine coef_generate_cyclic_bc(coef)
...
                    un = coef%get_normal(i, j, k, pe, pf)
                    len = sqrt(un(1) * un(1) + un(2) * un(2))
-                   d = coef%dof%y(i, j, k, pe) * un(1) - coef%dof%x(i, j, k, pe) * un(2)
-
-                   coef%cyc_msk(nc) = linear_index(i, j, k, pe, lx, ly, lz)
-                   coef%R11(nc) = un(1)/len * sign(1.0_rp, d)
-                   coef%R12(nc) = un(2)/len * sign(1.0_rp, d)
-
-                   nc = nc + 1
+                   if (len.gt.NEKO_EPS) then
```
timfelle:
```suggestion
                   if (len .gt. NEKO_EPS) then
```

src/sem/coef.f90
```diff
@@ -1230,19 +1241,30 @@ subroutine coef_generate_cyclic_bc(coef)
...
+                      d = coef%dof%y(i, j, k, pe) * un(1) &
+                           - coef%dof%x(i, j, k, pe) * un(2)
+
+                      coef%cyc_msk(nc) = linear_index(i, j, k, pe, lx, ly, lz)
+                      coef%R11(nc) = un(1) / len * sign(1.0_rp, d)
+                      coef%R12(nc) = un(2) / len * sign(1.0_rp, d)
+                      nc = nc + 1
+                   else
+                      call neko_error("x and y components of surface normals " // &
+                           "are zero. Cyclic rotations must be " // &
```
timfelle:
```suggestion
                      call neko_error("x and y components of surface " // &
                           "normals are zero. Cyclic rotations must be " // &
```

src/fluid/fluid_pnpn.f90
```diff
@@ -285,7 +285,7 @@ subroutine fluid_pnpn_init(this, msh, lx, params, user, chkp)
 
     call json_get_or_default(params, "case.fluid.cyclic", this%c_Xh%cyclic, &
          .false.)
-
+    call this%c_Xh%check_cyclic()
```
timfelle, line 288:
```suggestion
    call this%c_Xh%check_cyclic()

```

timfelle (commented):
I think this looks good. But I do not know if we have an example which uses it in order to try running it.

njansson (approved):

timfelle (approved):

PR #2327, opened 2026-02-18
Feature/vector list

adperezm:
Added vector list with the same characteristics as field_list_t. @timfelle , if you already had this somewhere, then of course ignore :).

I think it would be nice to add names to the vectors. The particular use-case I was thinking about was to further extend #2325 once it is merged.

I would like that functionality to accept a list of keys to read, and since `fld_file_data_t` uses `vector_t`, it would be nice if one could search a vector by name. This would also help to have a unique interface with hdf5, I believe.

I do not like that these searches need to always traverse the list. But I suppose that having hashtables with these types is overkill, considering that we never have crazy many entries.


src/math/vector.f90
```diff
@@ -48,6 +48,7 @@ module vector
   type, public :: vector_t
      !> Vector entries.
      real(kind=rp), allocatable :: x(:)
+     character(len=80) :: name = "" !< Name of the field
```
timfelle:
```suggestion
     character(len=80) :: name = "" !< Name of the vector
```

src/math/vector.f90
```diff
@@ -95,6 +102,12 @@ subroutine vector_init(v, n)
        call device_sync()
     end if
 
+    if (present(name)) then
+       v%name = name
+    else
+       v%name = "Vector"
+    end if
```
timfelle: How about just letting the default name stay empty instead?
```suggestion
    end if
```

src/math/vector.f90
```diff
@@ -129,6 +142,8 @@ subroutine vector_free(v)
 
     v%n = 0
 
+    v%name = ""
```
timfelle: These can go together
```suggestion
    v%name = ""
```

src/math/vector.f90
```diff
@@ -179,6 +196,8 @@ subroutine vector_assign_scalar(v, s)
        call cfill(v%x, s, v%n)
     end if
 
+    v%name = "Vector"
+
```
timfelle: I don't think you want to reset the name here?
```suggestion
```

src/math/vector.f90
```diff
@@ -193,6 +212,28 @@ subroutine vector_assign_array(v, array)
        call v%copy_from(HOST_TO_DEVICE, .true.)
     end if
 
+    v%name = "Vector"
+
```
timfelle: Also here, no reason to set the name
```suggestion
```

src/math/vector_list.f90
```diff
@@ -0,0 +1,212 @@
+module vector_list
```
timfelle, line 34: We are missing the copyright statement here

src/math/vector_list.f90
```diff
@@ -0,0 +1,212 @@
...
+    class(vector_list_t), intent(inout) :: this
+    type(vector_t), pointer :: f
+    integer, intent(in) :: i
+    f => this%items(i)%ptr
+  end function vector_list_get_by_index
+
+  !> Get an item pointer by array index
+  !! @param i The index of the item.
+  function vector_list_get_by_name(this, name) result(f)
+    class(vector_list_t), intent(inout) :: this
```
timfelle: This is a target of a pointer here

src/math/vector_list.f90
```diff
@@ -0,0 +1,212 @@
...
+          vector_list_assign_to_vector_ptr
+     procedure, pass(this) :: assign_to_vector => vector_list_assign_to_vector
+
+     !> Get device pointer for a given index.
+     procedure, pass(this) :: x_d => vector_list_x_d
+     !> Get pointer to the raw data array for a given index.
+     procedure, pass(this) :: x => vector_list_x
+     !> Get number of items in the list.
+     procedure, pass(this) :: size => vector_list_size
+     !> Get the size of dofmap for an item in the list.
```
timfelle:
```suggestion
     !> Get the size of an item in the list.
```

timfelle (changes requested):
Great work! 
There are a few things to look over.

src/math/vector.f90
adperezm: We can. I was copying what we do in `field_t` But I also think an empty name is fine. It also aligns more with what happens with the different assignments.

src/math/vector_list.f90
```diff
@@ -0,0 +1,245 @@
...
+
+  end subroutine vector_list_assign_to_ptr
+
+  !> Point item at a given index.
+  !! @param i The index of the item.
+  !! @param ptr An encapsulated vector pointer to point the item to.
+  subroutine vector_list_assign_to_vector_ptr(this, i, ptr)
+    class(vector_list_t), intent(inout) :: this
+    integer, intent(in) :: i
+    type(vector_ptr_t), target, intent(in) :: ptr
```
vbaconnet, line 218: I don't think you need the target here, do you? Or?
```suggestion
    type(vector_ptr_t), intent(in) :: ptr
```

vbaconnet: I guess it's the way it is done in field_list as well...

vbaconnet (approved):

src/math/vector_list.f90
```diff
@@ -0,0 +1,245 @@
+! Copyright (c) 2018-2025, The Neko Authors
```
timfelle:
```suggestion
! Copyright (c) 2018-2026, The Neko Authors
```

It is now 2026 😉

src/math/vector_list.f90
adperezm: Yes, I just kept the same. But I will test it and see if it fails after removing it.

vbaconnet: I mean if it compiles it compiles :) Asking because last time it caused issues but only with Cray. But it has been working for the field list so i gues there's no reason to remove it, i was just curious

adperezm:
Let me know what else I should change.

timfelle (approved):

src/math/vector_list.f90
timfelle: Also you just added the file, so it should not state 2018 as the start for this file

src/math/vector_list.f90
timfelle: @vbaconnet Is this conversation resolved?

vbaconnet: I think so!

PR #2328, opened 2026-02-20
Fix dimensions of inputs for tensor operators

timfelle:
Adjust the dimensions of input arrays in the tensor operator functions to ensure proper variable extent.

### DIFF
--- a/src/global_interpolation/legendre_rst_finder.f90
+++ b/src/global_interpolation/legendre_rst_finder.f90
@@ -236,12 +236,12 @@ contains
     real(kind=rp), intent(inout) :: resy(n_pts)
     real(kind=rp), intent(inout) :: resz(n_pts)
     integer, intent(in) :: el_list(n_pts)
-    real(kind=rp) :: r_legendre(this%Xh%lx)
-    real(kind=rp) :: s_legendre(this%Xh%lx)
-    real(kind=rp) :: t_legendre(this%Xh%lx)
-    real(kind=rp) :: dr_legendre(this%Xh%lx)
-    real(kind=rp) :: ds_legendre(this%Xh%lx)
-    real(kind=rp) :: dt_legendre(this%Xh%lx)
+    real(kind=rp) :: r_legendre(1, this%Xh%lx)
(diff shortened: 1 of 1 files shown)


njansson (approved):

vbaconnet:
Just curious, did this cause any issues or is it just for consistency?

adperezm:
I can't figure out why that input needed the (1), can someone explain? haha

The value of the different degrees of the Legendre polynomials at a given coordinate are, to me, better represented as a 1d array. So It would make more sense if I am reading this code to leave it as it was.

adperezm:
> I can't figure out why that input needed the (1), can someone explain? haha
> 
> The value of the different degrees of the Legendre polynomials at a given coordinate are, to me, better represented as a 1d array. So It would make more sense if I am reading this code to leave it as it was.

Okay, @vbaconnet now explained it is because the routine expects a transpose there. 

In this case, I would then suggest that if we want to change it, we make an explicit transpose operation into a temporal array or so. I believe that we can make the *_legendre arrays 2d (column vectors shape [lx,1]), but they should all at least have the same shape when they are declared.

timfelle:
> Just curious, did this cause any issues or is it just for consistency?

Niclas was seeing some compiler warnings about this, since the extend of the variables was not what was expected. It might be benign, but in general it is probably good to fix.

timfelle:
> Okay, @vbaconnet now explained it is because the routine expects a transpose there.
> 
> In this case, I would then suggest that if we want to change it, we make an explicit transpose operation into a temporal array or so. I believe that we can make the *_legendre arrays 2d (column vectors shape [lx,1]), but they should all at least have the same shape when they are declared.

Well there is already a specialisation for the explicit case of [lx,1], so performance is already there, but the compiler still looks for the second dimension explicitly being there. But yes in theory it would be good to have 2D variant, even though in practise I am not sure it would add any benefits except cleanup the calls and variable definitions.

adperezm (approved):
I think for readability I would still prefer if all of them are 1d however, since these are already "helper" arrays I think it's fine to change. 

If we modify this routine later on, we can change back if needed.

PR #2330, opened 2026-02-22
Speedup hash table iterators 

njansson:
Rework hash table structure and iterators to use a linked list connecting all inserted (valid) entries.

Using a linked list will significantly reduce the time it takes to iterate through a table, O(table_size) => O(inserted_entries), performed for example during mesh and gather-scatter init. Furthermore, the impact of this optimisation will be mostly visible for large problems, with many dofs per MPI rank (e.g. when using an accelerator).

### DIFF
--- a/.github/workflows/check_lint.yml
+++ b/.github/workflows/check_lint.yml
@@ -145,6 +145,11 @@ jobs:
               continue
             fi
 
+            # If the file is the `src/adt/htable.f90` file, skip it.
+            if [[ $file == "src/adt/htable.f90" ]]; then
+              continue
+            fi
+
             printf "\t- $file"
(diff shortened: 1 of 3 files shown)


timfelle (approved):
Looks great to me. 
Note: It must be force pulled since flint cannot run for this file. I still do not understand why, maybe the nested select statemens, it is the only thing i see that we dont use everywhere else.

njansson:
> Looks great to me. Note: It must be force pulled since flint cannot run for this file. I still do not understand why, maybe the nested select statemens, it is the only thing i see that we dont use everywhere else.

Should we add htable to the same skip list as stack?

timfelle:
I actually thought it was 😅

adampep (approved):

PR #2333, opened 2026-02-26
Fix unsafe nullify of pointers

njansson:
This fixes observed issues (SEGV) when deallocating a hash table.

### DIFF
--- a/src/adt/htable.f90
+++ b/src/adt/htable.f90
@@ -277,6 +277,9 @@ contains
     class(htable_t), intent(inout) :: this
     integer i
 
+    nullify(this%head)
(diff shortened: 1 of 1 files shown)


src/adt/htable.f90
```diff
@@ -277,6 +277,9 @@ subroutine htable_free(this)
     class(htable_t), intent(inout) :: this
     integer i
```
timfelle:
```suggestion
    integer :: i
```

timfelle (approved):

Shiyu-Sandy-Du (approved):

PR #2338, opened 2026-02-26
Ensure proper vectorisation for fj fortran

### LINKED ISSUE #2334: Euler residual calculations does not vectorise with Fujitsu
The loop nests inside the Runge-Kutta stages does not vectorise properly using `mpifrtpx`, despite using `do concurrent`

njansson:
This uses contiguous pointers to ensure that the residual calculations for the Euler equations are properly vectorised on Fugaku when compiling with Fujitsu's Fortran compiler. Fixes #2334


tuananhdao (approved):
Great! Thanks.

timfelle (approved):

PR #2340, opened 2026-02-27
Tiny correction to avoid diverging AMR from current develop

adampep:
Just commenting a few lines of "default" values for not mesh initialised bc.


timofeymukha (approved):
Leaves things unallocated, so not perfect of course. But let's consider it a temporary hack :-)

njansson (approved):

PR #2342, opened 2026-02-27
Feature/vtkhdf

timfelle:
This PR introduce VTKHDF as a possible output type for Neko.

Notable choices
- We currently only support direct subdivision to linear elements. These elements are constructed such that every GLL point sits at the corners of an element. This is horribly inefficient in terms of space but is robust.
- We allow the user to specify output precision for fields only. We default to working precision and will reduce the precision to working if the user requests a higher output precision. A warning is thrown, to inform the user.

Plan for the PR:

- [x] Basic implementation for storing a mesh and a field.
- [x] Simple linear hex element storage.
- [x] Temporal data support rather than 1 file per timestep.
    https://www.kitware.com/how-to-write-time-dependent-data-in-vtkhdf-files/
    https://docs.vtk.org/en/latest/vtk_file_formats/vtkhdf_file_format/vtkhdf_specifications.html#temporal-data
- [x] Add tests of the functionality. 
    - [x] Make sure HAVE_HDF5 is passed down to the tests.
    - [x] Unit tests to make sure a saved file contain the expected fields.
- [x] Test for Multiple MPI ranks and multi nodes, preferably on a distributed file system.
- [x] Test for single and double working precision.
- [x] Add user defined output precision.
- [x] Ensure restart works as intended.
    - Still limited to static meshes. Polynomial order etc cannot be changed mid run.

Planned extensions, separate PRs:

- [ ] Ensure support for the field writer simcomp.
    - This currently works if the user do not wish a separate file.
- [ ] Consider how "special" fields should be done. `u`,`v`,`w` -> `Velocity`, `p` -> `Pressure` and so on.
- [ ] High order element storage. VTK Lagrange elements
    https://scicomp.stackexchange.com/questions/42092/vtk-arbitrary-order-lagrange-elements-node-positions-ordering-on-reference-tri
    https://www.kitware.com/modeling-arbitrary-order-lagrange-finite-elements-in-the-visualization-toolkit/#:%7E:text=The%20new%20cells%20in%20VTK,may%20vary%20in%20Lagrange%20cells.
- [ ] External data for the temporal data, HDF5 feature. 
    It should be possible to split the HDF5 file to multiple files, 1 for each timestep, where the extra files just contain the field data, and the mesh and such global information is stored in the main file. This is very similar to Nek5000 files where the mesh is just stored in the first file.


Makefile.am
```diff
@@ -147,10 +147,10 @@ EXTRA_DIST = \
...
-	$(FLINT) score -d 0 src/ -r flinter_rc.yml
+# if ENABLE_FLINT
+# lint:
+# 	$(FLINT) score -d 0 src/ -r flinter_rc.yml
 
-check-local: lint
+# check-local: lint
 
-endif
+# endif
```
njansson:
```suggestion
 if ENABLE_FLINT
 lint:
 	$(FLINT) score -d 0 src/ -r flinter_rc.yml

 check-local: lint

 endif
```

timfelle: I know, I know. That's what I meant with it needs cleanup and such 😅

njansson:
Should we default to vtkhdf for TGV, or create an additional case?

timfelle:
> Should we default to vtkhdf for TGV, or create an additional case?

We could just add an extra VTKHDF case file with tgv. But due to the extra requirement of HDF5 I would probably not set it in any of the default files.

Makefile.am
```diff
@@ -147,10 +147,10 @@ EXTRA_DIST = \
...
-	$(FLINT) score -d 0 src/ -r flinter_rc.yml
+ if ENABLE_FLINT
+ lint:
+ 	$(FLINT) score -d 0 src/ -r flinter_rc.yml
 
-check-local: lint
+ check-local: lint
 
-endif
+ endif
```
njansson:
```suggestion
if ENABLE_FLINT
lint:
	$(FLINT) score -d 0 src/ -r flinter_rc.yml

check-local: lint

endif
```

njansson: My bad in prev. comment

njansson (approved):
Great stuff! 🍻

src/io/vtkhdf_file.F90
```diff
@@ -0,0 +1,1258 @@
...
+
+    call h5open_f(ierr)
+
+    call h5pcreate_f(H5P_FILE_ACCESS_F, plist_id, ierr)
+    info = MPI_INFO_NULL%mpi_val
+    call h5pset_fapl_mpio_f(plist_id, NEKO_COMM%mpi_val, info, ierr)
+
+    inquire(file = fname, exist = file_exists)
+    if (file_exists) then
+       call h5fopen_f(fname, H5F_ACC_RDWR_F, file_id, ierr, access_prp = plist_id)
```
adperezm: I guess this is very much needed since we append to a file. But I do not know if we should write a message saying that we are doing it, since this behavior is different from fld-file (where we overwrite) and someone just changing might get confused.

src/io/vtkhdf_file.F90
```diff
@@ -0,0 +1,1258 @@
...
+
+    if (present(t)) then
+       call vtkhdf_write_steps(vtkhdf_grp, t)
+    end if
+
+    ! For static mesh, only write geometry on the first call
+    if (associated(msh)) then
+       call vtkhdf_write_mesh(vtkhdf_grp, dof, msh, VTK_cell_type, &
+            this%amr_enabled, t)
+    end if
```
adperezm, line 230: After seeing this, I suppose that each new simulation will write to a new vtkhdf file. But if a user would want to keep dumping fields to the same file, then we could also add a check to write the mesh if it already does not exist in the file.

Or do you think this would be redundant? Feel free to ignore this one.

src/io/vtkhdf_file.F90
```diff
@@ -0,0 +1,1258 @@
...
+
+    do ie = 1, nelv
+       base = (ie - 1) * npts_per_cell
+
+       select case (vtk_type)
+       case (12) ! VTK_HEXAHEDRON
+          do ii = 1, lx - 1
+             do jj = 1, ly - 1
+                do kk = 1, lz - 1
+                   conn(idx + 1) = base + &
```
adperezm, line 277: I am honestly confused why this work in our files. Since every rank is looping in e from 1 to nelv. In the global sense, when vtk reads the connectivity wouldn't multiple conn(idx) point to the same points in the lists?

I suppose at some point in below in the code we add an offset, but I comment this here in case. This would be like a local connectivity in that case.

src/io/vtkhdf_file.F90
```diff
@@ -0,0 +1,1258 @@
...
+
+    ! --- Create filespaces and memspaces for 1D datasets ---
+    call h5screate_simple_f(1, vdims(1:1), filespace, ierr)
+    call h5screate_simple_f(1, dcount(1:1), memspace, ierr)
+    call h5pcreate_f(H5P_DATASET_CREATE_F, dcpl_id, ierr)
+    call h5sselect_hyperslab_f(filespace, H5S_SELECT_SET_F, &
+         doffset_1d, dcount, ierr)
+    call h5pset_chunk_f(dcpl_id, 1, chunkdims, ierr)
+
+    ! --- NumberOfPoints dataset (per-partition) ---
```
adperezm, line 415: per-partition here means that every rank writes how many points it has?

src/io/vtkhdf_file.F90
```diff
@@ -0,0 +1,1258 @@
...
+    call h5sselect_hyperslab_f(filespace, H5S_SELECT_SET_F, &
+         doffset_1d, dcount, ierr)
+    call h5dwrite_f(dset_id, H5T_NATIVE_INTEGER, connectivity, dcount(1:1), &
+         ierr, file_space_id = filespace, mem_space_id = memspace, &
+         xfer_prp = xf_id)
+    call h5sclose_f(memspace, ierr)
+    call h5dclose_f(dset_id, ierr)
+    call h5pclose_f(dcpl_id, ierr)
+    call h5sclose_f(filespace, ierr)
+    deallocate(connectivity)
```
adperezm: So we do write the local connectivity. Then I will have questions in a broader comment :)

src/io/vtkhdf_file.F90
```diff
@@ -0,0 +1,1258 @@
...
+    call h5dwrite_f(dset_id, H5T_STD_U8LE, cell_types_byte, dcount(1:1), ierr, &
+         file_space_id = filespace, mem_space_id = memspace, xfer_prp = xf_id)
+    call h5sclose_f(memspace, ierr)
+    call h5dclose_f(dset_id, ierr)
+    call h5pclose_f(dcpl_id, ierr)
+    call h5sclose_f(filespace, ierr)
+    deallocate(cell_types_byte)
+
+    if (present(t)) then
+       ! Open Steps group
```
adperezm, line 603: Wouldn't it be good to separate this step group initialization?

I suppose we would want to reuse the mesh writing even if we do not write steps. This can also just be left for later :)

adperezm (commented):
Very nice work! So far I am up to the write_mesh part but I have one question that I would like to clarify before proceeding. This might just come from me doing this in python, so it is perhaps just a different approach there.

I see that we always build just local connectivity, and I see that there is a part where each rank writes how many points it had and that. I have not checked the reader but does that mean that we are bound to keep reading with the same amount of ranks?

I suppose that here, the reader understand that this was written in parallel and that there is an "offset" in the connectivity, but I feel that the file changing depending on the number of ranks used to write it is a bit shaky, or am I understanding wrong?

A way that goes more in line with fld (obviously we want to get away from it but just for reference) is just to write a global connectivity (including offsets in there already) and just write in the "NumberOfPoints" the total. So a 1d array with one entry. 

In that case I think the offsets array might need to change to a global. But I think that one is more natural, since we always get the same file, no matter how many ranks were used to write it. 

Did I completely misunderstand what is happening here?

adperezm:
> I suppose that here, the reader understand that this was written in parallel and that there is an "offset" in the connectivity, but I feel that the file changing depending on the number of ranks used to write it is a bit shaky, or am I understanding wrong?

Okay, I now understand that this helps to see the partitioning. 

I understand the data is unchanged, which is the important thing. But this is something to be aware off if we see any problem in the future.

src/io/vtkhdf_file.F90
timfelle: Well all of these are relevant if the mesh changes in time, so AMR, but they
have to be there if the fields vary in time as well, they can just be 0 in order
to indicate that the mesh did not change. And the step group is not initialized
here, it is initialized in the `write_steps` routine.

src/io/vtkhdf_file.F90
timfelle: Yes, that is correct.

src/io/vtkhdf_file.F90
timfelle: It is because the mesh is partitioned into a seperate partition for each mpi
rank. So each partition have their own part of the mesh, with their own
connectivity list.

src/io/vtkhdf_file.F90
timfelle: In the write mesh function we check if the data is already present through the
`h5lexists_f`. So we do not write `points` again if they already exists etc.. One
thing is if you modify the mesh with different GLL points or something, then it
goes belly up, we currently only support a static mesh for now, but will look
into AMR later, so it is good to do the check inside the function instead.

timfelle:
> > I suppose that here, the reader understand that this was written in parallel and that there is an "offset" in the connectivity, but I feel that the file changing depending on the number of ranks used to write it is a bit shaky, or am I understanding wrong?
> 
> Okay, I now understand that this helps to see the partitioning. 
> 
> I understand the data is unchanged, which is the important thing. But this is something to be aware off if we see any problem in the future. 

I can investigate this a bit, but i think it only changes the offsets, I think we can make it work with changing number of ranks, but the partition information would be overwritten for the entire time series. @adperezm Would that still work for you?

Edit, nvm the point distribution would be wrong and connectivity would also not be right. Only solution is to write elements in global some globally consistent order, but I dont think we have that in the code.

adperezm:
> > > I suppose that here, the reader understand that this was written in parallel and that there is an "offset" in the connectivity, but I feel that the file changing depending on the number of ranks used to write it is a bit shaky, or am I understanding wrong?
> > 
> > 
> > Okay, I now understand that this helps to see the partitioning.
> > I understand the data is unchanged, which is the important thing. But this is something to be aware off if we see any problem in the future.
> 
> I can investigate this a bit, but i think it only changes the offsets, I think we can make it work with changing number of ranks, but the partition information would be overwritten for the entire time series. @adperezm Would that still work for you?
> 
> Edit, nvm the point distribution would be wrong and connectivity would also not be right. Only solution is to write elements in global some globally consistent order, but I dont think we have that in the code.

Hi, I am okay with things as they are now. I am happy with the PR. This was more a comment in case we see something weird in the future.

Right now, as I understood it, every rank writes a portion of the data in a contiguous way, and that is the way that works also for fld, etc. So I am happy with that. Now I just need to get used to the new things that hdf brings :) 

If something breaks for visualization, we can always check that later. 

I suggest we continue as is, and then later we see if we need to change to some global connectivity for some use case.

njansson:
Are we good to merge?

timfelle:
> Are we good to merge?

We just need @adperezm to resolve conversations if he is happy with the responses, otherwise we will continue the conversations.

timofeymukha:
@timfelle I think the two comments on deallocation from the bot are correct. The one with mpi -> MPI is cute :-). Unfortunately, I cannot triage the ones on H5 usage. Please take a look, and feel free to dismiss.

timofeymukha (approved):

PR #2344, opened 2026-02-27
Fix infinite loop in global interpolation

vbaconnet:
Increment of the iteration count was missing in the device_find_legendre_rst, causing infinite loops if the convergence critera was not reached.

thanks @adperezm for finding the bug!

### DIFF
--- a/src/global_interpolation/legendre_rst_finder.f90
+++ b/src/global_interpolation/legendre_rst_finder.f90
@@ -203,6 +203,7 @@ contains
     converged = .false.
     !Iterate until found, not heavily optimized
     do while (.not. converged)
+       iter = iter + 1
        call device_find_rst_legendre(rst, pt_x, pt_y, pt_z, &
             this%x_hat%x_d, this%y_hat%x_d, this%z_hat%x_d, &
             resx, resy, resz, &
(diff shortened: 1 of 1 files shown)


src/global_interpolation/legendre_rst_finder.f90
```diff
@@ -211,7 +212,7 @@ subroutine find_rst_legendre_device(this, rst, pt_x, pt_y, pt_z, &
        !This can be made more approriate... avoid memcpy at least
        conv_sum = device_vlsc3(conv_pts%x_d,conv_pts%x_d,conv_pts%x_d,n_pts)
        converged = conv_sum .lt. 0.5
-       print *, conv_sum
+       !print *, conv_sum
```
vbaconnet, line 215: Leave this as a comment for now, this could be good to have as a trace of what to monitor if/when we'll be touching global interpolation again

vbaconnet:
@njansson I think it would be good to cherry-pick this into the release bugfixes if it hasn't been released yet!

adperezm (approved):
Good catch!

njansson:
> @njansson I think it would be good to cherry-pick this into the release bugfixes if it hasn't been released yet!

Indeed, will do!

njansson (approved):

PR #2351, opened 2026-03-02
Feature/volflow output

pschlatt1:
Added diagnostic output to the vol flow routines, to be able to extract the pressure gradient directly. Will be later used in the channel and pipe examples.


timofeymukha:
@pschlatt1 you need to
- Merge in develop into your branch
- Run `make depend` in the `src` folder and commit the .depends file.
- Run the ./contrib/lint_format/format.sh tool and commit its changes.

PR #2357, opened 2026-03-03
Feature/json check expected size

arash-mp:
Adding optional argument "expected_size" to json_get_*_array.
If it is provided, the size of the input should match the expected size, otherwise it throws error.


adperezm (approved):

src/common/json_utils.f90
```diff
@@ -259,78 +259,149 @@ end subroutine json_get_string
...
+    if (present(expected_size)) then
+       if (actual_size /= expected_size) then
+          write(str_actual, '(I0)') actual_size
+          write(str_expected, '(I0)') expected_size
+          call neko_error("Parameter '" // trim(name) // &
+               "' has incorrect size: got " // &
+               trim(str_actual) // ", but expected " // trim(str_expected))
+       end if
+    end if  
+
```
vbaconnet, line 285: You repeat this block of code over every utility routine, should it be in its own routine?

arash-mp: @vbaconnet done!

vbaconnet (approved):
Thanks and sorry for the extra work!

src/common/json_utils.f90
```diff
@@ -607,4 +642,25 @@ subroutine json_extract_item_from_name(json, name, i, item)
...
+  !> Routine to validate array sizes against an expected size
+  !! @param[in] name The name of the parameter for error reporting
+  !! @param[in] actual_size The size found in the JSON
+  !! @param[in] expected_size (Optional) The size it is supposed to be
+  subroutine check_expected_size(name, actual_size, expected_size)
+    character(len=*), intent(in) :: name
+    integer, intent(in) :: actual_size
+    integer, optional, intent(in) :: expected_size
+    character(len=32) :: str_actual, str_expected
+
```
timofeymukha, line 661: It makes more logical snse to not have expected_size as optional here, since then the routine does nothing at all. And then add if (present( ) ) wrappers where you call it.

arash-mp: Done!

timofeymukha (approved):

PR #2364, opened 2026-03-07
Unit test harmonization

timofeymukha:
I am trying to maximize as much as possible the conventions used in all unit tests. 

- Wrapped all .pf code in a module.
- This caused some module name issues, e.g. `module tuple`, which coincides with the name of the tested module. So, renamed .pf filies to start with test_ to avoid collisions. This was already a convention used in most recent tests.
- Added a Makefile.am target to refresh the Makefile file for the tests, without running a general `configure`. 
   To execute it: `make refresh-unit-tests-makefiles`. The command was generated by an LLM, so please check it. One had to be a bit carefull because a simple glob failed due to some Makefiles not actually being present in the configure file. So we have stale test directories basically?
- Made .inc files look for Makefile updates in Makefile.in files of unit tests. This basically makes sure that a test recompiles fully if you update the makefile via `configure`.
- Added  @NEKO_PKG_FCFLAGS@  to `FFLAGS` of all tests, to not have to think about it.


njansson:
Great to harmonise things. However, if we now do this big (necessary) cleanup, wouldn't it be best if we go all the way and use proper `use pfunit` in the modules, and add setup and teardown routines for all tests?

timfelle (approved):
I kinda agree with Niclas on the setup and teardowns. But maybe it would be fine in a separate PR

timofeymukha:
> Great to harmonise things. However, if we now do this big (necessary) cleanup, wouldn't it be best if we go all the way and use proper `use pfunit` in the modules, and add setup and teardown routines for all tests?

That sounds like a good idea, I'll put this in the pipeline! Can we merge this as a first step, though? It would openup a range of smaller PRs for review, and I can keep the ball rolling.

njansson (approved):

PR #2366, opened 2026-03-10
Bugfix/cyclic logic

IsaMammadli:
Nick and I have done some further tests on cyclic checks and seems tolerances (for both glsum and glmax) vary from mesh to mesh. In order to avoid interruptions and unnecessary warnings, the tests are removed. I would later add some documentations and maybe an example (a code piece that does similar checks in the user file) to be instructive.


IsaMammadli:
I just observed smth with GPU runs, will resubmit shortly.

adperezm (approved):

nick-morse (approved):

PR #2370, opened 2026-03-10
Add serial and parallel unit test templates

timofeymukha:
Adds two templates for unit tests: serial and parallel. The idea is that any test will start out as a copy of one of these.


timofeymukha:
@njansson @timfelle I stripped down this PR to just the templates. Added some setUp and tearDown. Please comment, particularly on the parallel one.

njansson (approved):
Puurfect 🐈‍⬛

vbaconnet (approved):
looks fine to me!

PR #2371, opened 2026-03-10
Improve adt test and file structure

timofeymukha:
Dep #2370 

Improves tests for types under src/adt. Some types were not tested at all, some additional tests added for others. Sparse comments on what is tested added.

In the type modules themselves, `public` is stripped from type definitions instead a list of `public : type_name` is arranged at the top, declaring the `data` type inside as a comment. This makes it much easier to find the concrete type necessary for data at hand.


njansson:
Style comments aside (I'm in favour of having public at the type declaration), I do recall that certain compilers can't declare things `public` before they are declared, thus this long list has to be moved to before `contains`, which kind of beats the purpose here.

timofeymukha:
> Style comments aside (I'm in favour of having public at the type declaration), I do recall that certain compilers can't declare things `public` before they are declared, thus this long list has to be moved to before `contains`, which kind of beats the purpose here.

OK, I refactored to having tables in the module level docstrings. i would really like to hava quick lookup of the types somewhere

njansson (approved):

timfelle (approved):

PR #2372, opened 2026-03-10
Add script to add a unit test or a .pf to existing suite

timofeymukha:
Dep #2370

This PR adds a Python script that can
- Create a new pFUnit test suite
- Add a new .pf to an existing suite

The Python script is wrapped by two shell scripts for the two corresponding functionalities, for convenience.

Full disclosure, the script is AI-generated. But I have now gone through the entire thing and understand how it works. I have also tested the functionality--- it seems to work like a charm. 

Some notes:
- The new suite or .pf starts off as a copy from the template cases we add in #2370.
- The files are then patched to have a new name.
- For the test suite case, the build system is then patched. All the edits are first done in memory after loading the original files in their state prior to the new additions. As a final guard, once the new files are prepared in-memory, it is not written to disk directly, but rather via a temporary directory. So if anything fails during the process, the original files are left completely untouched. This is honestly a bit overkill, I think, and I couldn't emulate a "failure" and verify that the safeguard works. But generally things work, so I guess we can leave it there.


timofeymukha:
I gotta go through the python more thoroughly, putting dont merege for now and convertin to draft.

timofeymukha:
Reworked and ready for review as soon as #2370 is in.

njansson (approved):

timfelle (approved):

PR #2373, opened 2026-03-11
Memcpy pressure in set_flow_ic_common

vbaconnet:
I thought it was annoying that if I use user initial condition and set a pressure field, I have to memcpy the pressure but not the velocities. Also the [documentation of user initial condition](https://neko.cfd/docs/develop/d6/def/user-file.html#user-file_user-ic) is a bit misleading: 

> There is no need to add the transfer to GPU memory in this user routine, it will be done under the hood afterwards.

But apart from this, why aren't we doing the `gs%op` on the pressure? Clearly it is done in the compressible initial condition..

### DIFF
--- a/src/fluid/flow_ic.f90
+++ b/src/fluid/flow_ic.f90
@@ -435,11 +435,10 @@ contains
 
     ! If we are on GPU we need to move (u,v,w) back to the host
     ! since set_flow_ic_common copies it again to the device.
-    ! NOTE: p can stay on the device since it is not copied by
-    ! set_flow_ic_common
-    call u%copy_from(device_to_host, .false.)
-    call v%copy_from(device_to_host, .false.)
-    call w%copy_from(device_to_host, .true.)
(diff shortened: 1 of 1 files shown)


timofeymukha (approved):

timofeymukha:
Do pressure ICs play a role in PnPn? I am used to it just being solved for, without any time advancement as such, that is how most p-U splittings work, but PnPn is always full of suprises. Anyway, we should have the copy for consistency regardless.

src/fluid/flow_ic.f90
```diff
@@ -435,11 +435,10 @@ subroutine set_flow_ic_fld(u, v, w, p, file_name, &
 
     ! If we are on GPU we need to move (u,v,w) back to the host
```
vbaconnet:
```suggestion
    ! If we are on GPU we need to move (u,v,w) and p back to the host
```

njansson (approved):

PR #2377, opened 2026-03-13
Append _m to modules

timofeymukha:
Happy Friday:-D.

I know we haven't greenlighted this yet, but I asked AI to do it so we can have an actual look at how it looks.

The neko module remains as neko to not break user code.

### DIFF
--- a/contrib/calc_lift_from_field/calc_lift_from_field.f90
+++ b/contrib/calc_lift_from_field/calc_lift_from_field.f90
@@ -6,7 +6,7 @@
 !! Martin Karp 17/01-24
 program calc_lift_from_field
   use neko
-  use matrix
+  use matrix_m
   implicit none
 
   character(len=NEKO_FNAME_LEN) :: inputchar, mesh_fname, field_fname, hom_dir, output_fname
--- a/contrib/map_to_equidistant_1d/map_to_equidistant_1d.f90
+++ b/contrib/map_to_equidistant_1d/map_to_equidistant_1d.f90
@@ -5,8 +5,8 @@
 !> Shiyu Du 16/01-24 note: the program is currently tested on CPUs, GPU is not tested yet.
 program map_to_equidistant_1d
   use neko
-  use fast3d
-  use tensor
+  use fast3d_m
+  use tensor_m
   implicit none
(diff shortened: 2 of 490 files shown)


timofeymukha:
Tagging people just to take a look

timofeymukha:
Wow, it passed CI on the first shot. All hail our AI overloards!

vbaconnet:
What was the reason for doing this? For consistency with "_t"?

timofeymukha:
> What was the reason for doing this? For consistency with "_t"?

The main reason is name shadowing if you declare a variable name that coincides with the name of the module. E.g. you have ` use module dofmap , only dofmap_t`, and then inside some routine you have a dummy argument called `dofmap`. It seems most compilers let this fly, but it is not good practice. At the same time, it is often very natural to use a variable name that is the type name sans the _t, which is very commonly the module name.

vbaconnet:
> The main reason is name shadowing if you declare a variable name that coincides with the name of the module. 

I see.. Would be good (once the PR is ready) to try to compile this on Cray as well to make sure nothing breaks!

timofeymukha:
> > The main reason is name shadowing if you declare a variable name that coincides with the name of the module.
> 
> I see.. Would be good (once the PR is ready) to try to compile this on Cray as well to make sure nothing breaks!

For sure. But hopefully this actually makes compilers very happy. But you never know :-))

adperezm:
> Cray

Yeah, cray is the only one that has complained about Aliasing to me haha. So I would hope this is a good thing. But what I do not know is if we agreed that the _m would be the identifier. I guess in Fortran this is the only way? Since it is not case-sensitive?

timofeymukha:
> > Cray
> 
> Yeah, cray is the only one that has complained about Aliasing to me haha. So I would hope this is a good thing. But what I do not know is if we agreed that the _m would be the identifier. I guess in Fortran this is the only way? Since it is not case-sensitive?

We didn't really agree on anything, so this PR was like 20% Friday trolling. But the _m was discussed for a long time, and a few weeks back @njansson said he was warming up to the idea, so I wanted us to be able to see how it looks.

PR #2380, opened 2026-03-13
Vtkhdf/lagrange vtk

### LINKED ISSUE #2379: High order element storage. VTK Lagrange elements
https://scicomp.stackexchange.com/questions/42092/vtk-arbitrary-order-lagrange-elements-node-positions-ordering-on-reference-tri
https://www.kitware.com/modeling-arbitrary-order-lagrange-finite-elements-in-the-visualization-toolkit/#:%7E:text=The%20new%20cells%20in%20VTK,may%20vary%20in%20Lagrange%20cells.

timfelle:
The goal here is to implement support for the VTK "Arbitrary Lagrange" cell types.
Visualization software such as Paraview, have spotty support for it, but this can give us a lever to be in front of the curve, should things change.
Specifically for paraview, the files can be loaded correctly but colour interpolation is still linear between GLL points, rather than correctly based on the polynomial basis.

Depends on #2342. Fixes #2379

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Develop
 
+- Added `Arbitrary Lagrange Cell` support for the VTKHDF output. This moves the
+  existing subdivision-based output to an optional setting under
+  `output_subdivide`. The default output will now be in the VTKHDF format with
+  `Arbitrary Lagrange Cell`s, which correctly represent our higher-order
+  elements. 
 - Added an AI policy to the contribution guidelines.
 - Added simple support for VTKHDF. For now it can be used for fluid outputs.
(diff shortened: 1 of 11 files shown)


njansson (approved):
🍻 🦄

nick-morse:
Really nice, preserving the spectral element structure is a big advantage, even if ParaView mostly does linear interpolation. One thing to keep in mind would be whether the limited higher-order subdivision in ParaView at the moment and future VTK support assume a Lagrange reference hexahedron or a GLL one, since right now the data is saved as a Lagrange element but at the GLL points. Either way, it's okay for now, as ParaView mostly does linear interpolation between the points.

nick-morse (approved):

tuananhdao (approved):

PR #2381, opened 2026-03-13
Add AI policy to contribution guidelines

timofeymukha:
I added a mix of things from various projects mentioned in the blogpost that @adperezm shared.


adperezm (approved):
Nice! I think this is good to have.

Are we adding the need for an "AI was used at some point in this PR" disclaimer? I guess these days, there will be some form of LLM help, so maybe it should just be assumed, or?

timfelle (approved):

timofeymukha:
> Nice! I think this is good to have.
> 
> Are we adding the need for an "AI was used at some point in this PR" disclaimer? I guess these days, there will be some form of LLM help, so maybe it should just be assumed, or?

Yeah, I don't think a disclaimer is needed. You just gotta be able to defend the code you are pushing in.

PR #2383, opened 2026-03-16
add known issues in statistics guide

### LINKED ISSUE #2278: Fluid/scalar statistics output filled with zeros at end_time
When running a Neko simulation with the simulations components `fluid_stats`or `scalar_stats` enabled, it sometimes occurs that the resulting `.csv` file has the last timestep repeated twice, with the second one having zeros everywhere.

Minimal example of the output (imagine `end_time = 200`, `output_value = 100` and 1D domain with 3 nodes at 0,5,10 m):
```
time; z; temperature
  0;   0; 302.1
  0;   5; 302.8
  0;  10; 303.0
100;   0; 301.0
100;   5; 302.9
100;  10; 303.4
200;   0; 301.4      <--- final timestep
200;   5; 302.1
200;  10; 303.5
200;   0;   0.0      <--- final timestep is repeated, with
200;   5;   0.0      <--- these lines filled with zeros 
200;  10;   0.0      <---
```

The issue is present both with `fluid_stats` and `scalar_stats`, and I think it's a problem when `end_time`is an exact multiple of `output_value` (e.g. `end_time = 1000.0`, `output_value=100.0`).
I guess it's some kind of 'off-by-one' problem that comes up in the main statistics loop.

There are of course multiple "dirty" workarounds (e.g. `csv` file manipulation, choosing a slightly different `output_value`...), but I still think it'd be worth to fix it in the source code.

I attach a screenshot of a more complex `scalar_stats0.csv` output as example (`end_time = 14000.0`,`output_value=100.0`):

Image

### LINKED ISSUE #2378: In fluid/scalar statistics time 0 is always filled with zeros
Hello,

Just a heads up. 

When running the scalar or fluid statistics simulation component with `"start_time" = 0.0` in the `.case` file, the output file results filled with zeros for time 0.0.

This means that 
1- the initial conditions do not appear in the stats file
2- because of that, if you take an average of your CSV file starting from time 0, the average will include these zeros and the resulting profiles might be wrong.

This is an issue somehow similar to #2278 in the sense that it can sneakily mess up your results when working with the statistics (e.g. plotting profiles).

------

How to reproduce: 
- put statistics in your case file:
```
"simulation_components": [
            {
                "type": "fluid_stats",
                "compute_control": "simulationtime",
                "compute_value": 10.0,
                "output_value": 100.0,
                "start_time": 0.0,    //  <----------------
                "avg_direction": "xy",
                "set_of_stats": "basic"
            }
]
```

- inspect the resulting `fluid_stats0.csv` file visually:

Image

loretet:
Added a warning to the "Statistics guide" section of the documentation to include two known issues, #2278  and #2378.

Both issues are not proper bugs and can be easily worked around, but I think the user should still be made aware of their existence (otherwise both issues can silently mess up the user analysis, more or less significantly)


doc/pages/user-guide/statistics-guide.md
```diff
@@ -367,3 +367,20 @@ Otherwise, if one does not use a model for the scalar eddy diffusivity but relat
...
+This means that one should discard the first $N$ rows of the `.csv` file, where $N$ is the number of elements in the non-averaged direction. \
+This also means that the initial conditions _do not_ appear in the `stats0` file. \
+This issue has been discussed more in detail in [#2378](https://github.com/ExtremeFLOW/neko/issues/2378). A possible workaround is to simply ignore the first $N$ rows of the file when checking the results, or to use a `"start_time"` slightly higher than `0.0`.
+
+2. when `"end_time"` is an exact multiple of `"output_value"` (e.g. `"end_time": 1000.0`, `"output_value": 100.0`), the resulting `.csv` file has the last timestep repeated twice, with the second one having zeros everywhere (both for fluid and scalar statistics). \
+This means the last $N$ rows of the file should be discarded. The issue only appears when `"output_at_end"` is turned on. \
+This issue has been discussed more in detail in [#2278](https://github.com/ExtremeFLOW/neko/issues/2278). The obvious workaround is to turn off `"output_at_end"`; alternatively, one can set `"end_time"` as a non-multiple of `"output_value"` (e.g. `"end_time": 1001.0`, `"output_value": 100.0`).
```
adperezm: I am not sure this is the obvious workaround. It is more common to need to run simulations over multiple days, so turning this off might not be a good idea.

doc/pages/user-guide/statistics-guide.md
```diff
@@ -367,3 +367,20 @@ Otherwise, if one does not use a model for the scalar eddy diffusivity but relat
 | 2 | \f$ \langle \alpha_t \frac{\partial s}{\partial x} \rangle \f$ | X-Velocity|
 | 3 | \f$ \langle \alpha_t \frac{\partial s}{\partial y} \rangle \f$ | Y-Velocity|
 | 4 | \f$ \langle \alpha_t \frac{\partial s}{\partial z} \rangle \f$ | Z-Velocity |
+
+
+# Known issues {#known-issues}
+There are at least two known issues that is worth keeping in mind when working with the statistics component in Neko: 
+1. if one computes statistics from time 0 (`"start_time": 0.0`), the first entry in the `stats0` file will be zero-filled (both for scalar and fluid statistics). \
+This means that one should discard the first $N$ rows of the `.csv` file, where $N$ is the number of elements in the non-averaged direction. \
+This also means that the initial conditions _do not_ appear in the `stats0` file. \
```
adperezm: I do not think that initial conditions need to appear in statistics files. I would expect anything in the statistics file to be averaged in some way. In the case of start_time = 0, The reason you get zeros is because there is because delta__averaging_t = 0, and therefore everything is weighted by zero. So it is an average in a sense.

adperezm (commented):
Thanks for putting this up!

Isn't it best to simply recommend to always use weighted averages for processing any of the statistics from Neko? Is there any case where that would fail?

I would say that when post-processing the data, it is important to inspect the results. However, I do not see the harm in giving this information.

On the other hand, I am still not sure that these things are actually read.

doc/pages/user-guide/statistics-guide.md
```diff
@@ -367,3 +367,20 @@ Otherwise, if one does not use a model for the scalar eddy diffusivity but relat
 | 2 | \f$ \langle \alpha_t \frac{\partial s}{\partial x} \rangle \f$ | X-Velocity|
 | 3 | \f$ \langle \alpha_t \frac{\partial s}{\partial y} \rangle \f$ | Y-Velocity|
 | 4 | \f$ \langle \alpha_t \frac{\partial s}{\partial z} \rangle \f$ | Z-Velocity |
+
+
+# Known issues {#known-issues}
+There are at least two known issues that is worth keeping in mind when working with the statistics component in Neko:
```
vbaconnet: Whether these are _issues_ is, in my opinion, subjective. 
```suggestion
# Note to users {#note-to-users}

```

doc/pages/user-guide/statistics-guide.md
```diff
@@ -367,3 +367,20 @@ Otherwise, if one does not use a model for the scalar eddy diffusivity but relat
 | 2 | \f$ \langle \alpha_t \frac{\partial s}{\partial x} \rangle \f$ | X-Velocity|
 | 3 | \f$ \langle \alpha_t \frac{\partial s}{\partial y} \rangle \f$ | Y-Velocity|
 | 4 | \f$ \langle \alpha_t \frac{\partial s}{\partial z} \rangle \f$ | Z-Velocity |
+
+
+# Known issues {#known-issues}
+There are at least two known issues that is worth keeping in mind when working with the statistics component in Neko: 
+1. if one computes statistics from time 0 (`"start_time": 0.0`), the first entry in the `stats0` file will be zero-filled (both for scalar and fluid statistics). \
```
vbaconnet: A zero-filled stats file is always dumped at `start_time`, not only when it is set to `0.0`. Please change accordingly.

doc/pages/user-guide/statistics-guide.md
```diff
@@ -367,3 +367,20 @@ Otherwise, if one does not use a model for the scalar eddy diffusivity but relat
...
+This means that one should discard the first $N$ rows of the `.csv` file, where $N$ is the number of elements in the non-averaged direction. \
```
vbaconnet: this might be too specific, please keep in mind that not everyone uses 1D-averaged `csv` files.

doc/pages/user-guide/statistics-guide.md
```diff
@@ -367,3 +367,20 @@ Otherwise, if one does not use a model for the scalar eddy diffusivity but relat
...
+This also means that the initial conditions _do not_ appear in the `stats0` file. \
+This issue has been discussed more in detail in [#2378](https://github.com/ExtremeFLOW/neko/issues/2378). A possible workaround is to simply ignore the first $N$ rows of the file when checking the results, or to use a `"start_time"` slightly higher than `0.0`.
+
+2. when `"end_time"` is an exact multiple of `"output_value"` (e.g. `"end_time": 1000.0`, `"output_value": 100.0`), the resulting `.csv` file has the last timestep repeated twice, with the second one having zeros everywhere (both for fluid and scalar statistics). \
+This means the last $N$ rows of the file should be discarded. The issue only appears when `"output_at_end"` is turned on. \
+This issue has been discussed more in detail in [#2278](https://github.com/ExtremeFLOW/neko/issues/2278). The obvious workaround is to turn off `"output_at_end"`; alternatively, one can set `"end_time"` as a non-multiple of `"output_value"` (e.g. `"end_time": 1001.0`, `"output_value": 100.0`).
+
+Both these issues are relatively easy to avoid or work around, but can silently mess up the postprocessing of a simulation for the unaware user.
```
vbaconnet: I don't think this is necessary.
```suggestion

```

doc/pages/user-guide/statistics-guide.md
```diff
@@ -367,3 +367,20 @@ Otherwise, if one does not use a model for the scalar eddy diffusivity but relat
...
 | 3 | \f$ \langle \alpha_t \frac{\partial s}{\partial y} \rangle \f$ | Y-Velocity|
 | 4 | \f$ \langle \alpha_t \frac{\partial s}{\partial z} \rangle \f$ | Z-Velocity |
+
+
+# Known issues {#known-issues}
+There are at least two known issues that is worth keeping in mind when working with the statistics component in Neko: 
+1. if one computes statistics from time 0 (`"start_time": 0.0`), the first entry in the `stats0` file will be zero-filled (both for scalar and fluid statistics). \
+This means that one should discard the first $N$ rows of the `.csv` file, where $N$ is the number of elements in the non-averaged direction. \
+This also means that the initial conditions _do not_ appear in the `stats0` file. \
+This issue has been discussed more in detail in [#2378](https://github.com/ExtremeFLOW/neko/issues/2378). A possible workaround is to simply ignore the first $N$ rows of the file when checking the results, or to use a `"start_time"` slightly higher than `0.0`.
```
vbaconnet: - Whether this is an issue that needs a workaround can be discussed. I would simply refer the user to the discussion on this issue.
- See my previous comment about csv files and the `start_time`

doc/pages/user-guide/statistics-guide.md
vbaconnet: This is the intended behavior of `output_at_end` and has already been documented (see [the case file documentation](https://neko.cfd/docs/develop/dd/d33/case-file.html)).

Image

vbaconnet (changes requested):
Thanks for opening the PR! 

I agree with Adalberto's comments. 

The two "issues" you have raised only appear if you use an arithmetic mean for the final averaging. Therefore I would suggest that we:

- recommend the user use weighted averages exclusively, as is done in the provided post-processing script `average_fields_in_time`
- inform the user that a zero-filled stats file will always be dumped at (or slightly after) `start_time` so they know what to expect.
- Maybe refer to the documentation for the case file to remind them about the behavior of `output_at_end` (also not an issue if using weighted averaging).

doc/pages/user-guide/statistics-guide.md
loretet: Actually I don't get a zero-filled stats file first entry if I start from something different than 0.0, I just tested it

doc/pages/user-guide/statistics-guide.md
loretet: I do think it needs a workaround, but what I proposed is probably not a proper workaround, so I'll just take out that part

doc/pages/user-guide/statistics-guide.md
loretet: True, I did not think of that. I will just take out the workaround suggestions: I think the users can get there by themselves or open a discussion

loretet: I have added a reference to the case file documentation, but I'd still keep this here as well as it's strictly related to the `stats0` files

doc/pages/user-guide/statistics-guide.md
vbaconnet: You are right! My bad. 

That is so strange, I ran the cylinder example as a test case and then I got it to write directly at the start time. 

I did some more tests and it turns out that if you carefully align the output control with the start time you can get it to write at start_time. But in general that is not true. 

So this would write at 1.0, but that is because of the output_value of course :)
```json
"output_value": 1.0,
"start_time": 1.0
```

So then it all comes down to that our output controllers just trigger in general at the first timestep.

Then I will resolve this conversation.

vbaconnet:
Hi, I pushed some stuff just to reformat a bit with the 80 column limit.

doc/pages/user-guide/statistics-guide.md
loretet: Should we then change it to something like this maybe?
```diff
- 1. if one computes statistics from time 0 (`"start_time": 0.0`), the first entry in the `stats0` file will be zero-filled (both for scalar and fluid statistics).
+ 1. often, the first entry in the `stats0` file will be zero-filled (both for scalar and fluid statistics). An exception to this is when `"output_value"` is aligned to `"start_time"`.
```

doc/pages/user-guide/statistics-guide.md
```diff
@@ -367,3 +445,26 @@ Otherwise, if one does not use a model for the scalar eddy diffusivity but relat
...
+2. When `"end_time"` is an exact multiple of `"output_value"` (e.g.
+   `"end_time": 1000.0`, `"output_value": 100.0`), the resulting `.csv` file
+   has the last timestep repeated twice, with the second one having zeros
+   everywhere (both for fluid and scalar statistics).
+   This is the intended behaviour of `output_at_end` - please see
+   [the case file documentation](https://neko.cfd/docs/develop/dd/d33/case-file.html).
+   This issue has been discussed more in detail in
+   [#2278](https://github.com/ExtremeFLOW/neko/issues/2278).
+
+Note also that the initial conditions _do not_ appear in the `stats0` file.
```
vbaconnet: If we already say that the first stats file is filled with zeroes, is this still necessary?
```suggestion

```

doc/pages/user-guide/statistics-guide.md
```diff
@@ -367,3 +445,26 @@ Otherwise, if one does not use a model for the scalar eddy diffusivity but relat
...
+
+
+# Note to users {#note-to-users}
+
+The general recommendation is to always use weighted averages to analyse the
+statistics. When using an arithmetic average, it is important to keep in mind
+the following two characteristics of the `stats0` files:
+
+1. If one computes statistics from time 0 (`"start_time": 0.0`), the first
+   entry in the `stats0` file will be zero-filled (both for scalar and fluid
```
vbaconnet: To make it more general, the name "stats0" is specific to the `csv` format
```suggestion
   sample will be zero-filled (both for scalar and fluid
```

vbaconnet:
Thanks for making all the chnages! I added two comments but they are quite minor, up to you if you want to change or not

vbaconnet (approved):

njansson (approved):

PR #2385, opened 2026-03-16
MOST wall model for ABL simulations

loretet:
This PR implements the Monin-Obukhov Similarity Theory (MOST) wall model on CPU and GPU.

## Description
MOST is the most common wall model used in atmospheric science. Here, the friction velocity $u_\*$ is computed based on some empirical similarity functions which depend on the thermal stratification of the boundary layer (in this version we use the same functions as the [PALM](https://palm.muk.uni-hannover.de/trac) model). The WSS $\tau$ is then computed from $u_\*$ in the usual way and applied as a boundary condition. \
More info can be found in Stull (1988) or Rotach and Holtslag (2025).

The MOST implementation is very similar to the `rough_log_law` wall model, but it is made more complex by a necessary branching in the algorithm: the similarity functions and stress expression have different shapes based on whether a Dirichlet or Neumann boundary condition is used for temperature, and also based on the local stratification of the flow (which has to be computed for every timestep and point). This makes the GPU implementation not very efficient, I think (especially if it is a noob like me to code it :D)

## Implementation
The MOST wall model requires more specification than `rough_log_law`. An example of implementation is as follows:
```
 "boundary_conditions": [
       {
            "type": "wall_model",
            "model": "most",
            "kappa": 0.4,     // Von Karman constant
            "z0": 0.1,     // the momentum roughness length
            "z0h": 0.1,      // the heat roughness length (if negative, Zilitinkevich formula is used)
            "type_of_temp_bc": "neumann",   // supports either Neumann or Dirichlet as b.c.
            "bottom_bc_flux_or_temp": 0.05,     // the temperature or flux value imposed at the boundary
            "zone_indices": [
                       5
             ],
            "h_index":1     // the model samples the flow at the first index above the boundary
        },
....
]
```

## Limitations/issues
At the moment, in a typical ABL simulation it is necessary to specify multiple times in the `case` file the same variables (e.g. `g` or `q`). The redundancy of parameters in the `case` file can be reduced by using the `constants` group (#2228). \
Another limitation is that - for now - it is only possible to use this wall model on one boundary at the time (it is unlikely that anybody will ever need it on two different boundaries, but it's still a possible future work).

## Results
Here is a comparison made between the CPU and GPU implementation of the wall model in Neko. Both the rough log-law and the MOST model appear. The simulation had a geostrophic wind flow of 10 m/s and a bottom heat flux of 0.05Kms-1, used buoyancy-corrected Vreman as SGS model and has 24 elements (polynomial degree 7).\
Coding-wise, the CPU and GPU implementations seem to be exactly the same for me, and I think the slight difference in the results is acceptable although maybe a bit surprising (tested on Dardel). Here, $v$ is the _spanwise_ velocity.

uv

t

Note: Nek5000 doesn't appear in the plots. This is because Nek5000 and Neko have always given slightly different results in some ABL flows, and we are still investigating this issue. However, the discrepancy within the codes should _not_ be due to the wall model, hence the PR.


loretet:
Quick update after March 17 developers' meeting: 
- The difference between CPU and GPU is concerning for Niclas. Philipp suggested investigating the influence of the time average by doing batch averages and Siavash suggested to also look at very close time average after restarting from the same file. I will do that and publish some updates
- Tim suggested to open a separate PR/issue/discussion for the difference between CPU and GPU implementation of the `rough_log_law` model, which shows differences even if it is a much simpler model. Will do that as well

src/wall_models/bcknd/cpu/most_cpu.f90
```diff
@@ -0,0 +1,440 @@
...
+    character(len=*), intent(in) :: bc_type
+    real(kind=rp), dimension(n_nodes), intent(inout) :: tau_x, tau_y, tau_z
+    integer :: ts_idx(3)
+    integer :: i, count
+    integer, parameter :: max_count = 50
+    real(kind=rp) :: ui, vi, wi, ti, ts, q, hi
+    real(kind=rp) :: magu, utau, normu, z0h
+    real(kind=rp) :: L_ob, L_upper, L_lower, L_old
+    real(kind=rp) :: Ri_b, f, dfdl, fd_h, L_new, L_sign
+    real(kind=rp), parameter :: g = 9.80665_rp
```
linneahuusko: The gravity should not be hardcoded here since we allow it as a parameter in the case file for other components (e.g., Boussinesq, Deardorff). I guess you are assuming the direction of gravity to be wall normal, which makes sense in our cases, but how can we generalize it? Making it truly general I think would be a lot of work, but I think we should at least let the user provide a g_vector and then put an error if it has anything other than a component in the z direction.

src/wall_models/most.f90
```diff
@@ -0,0 +1,221 @@
...
+    type(coef_t), intent(in) :: coef
+    integer, intent(in) :: msk(:)
+    integer, intent(in) :: facet(:)
+    integer, intent(in) :: h_index
+    type(json_file), intent(inout) :: json
+    real(kind=rp) :: kappa, z0, z0h_in
+    character(len=:), allocatable :: bc_type
+    real(kind=rp) :: bc_value
+
+    call json_get_or_default(json, "kappa", kappa, 0.41_rp)
```
linneahuusko: Why 0.41? I would say 0.4 is more standard in ABL studies and textbooks.

src/wall_models/bcknd/cpu/most_cpu.f90
```diff
@@ -0,0 +1,440 @@
...
+          z0h = z0h_in
+       end if
+
+       ! Get q, Ri_b, f_ptr, dfdl_ptr based on bc_type
+       ! Maybe redundant, but needed to initialise Rib
+       call select_bc_operators(bc_type,bc_value,q,ts,ti,kappa,utau,z0h,hi)
+       call compute_Ri_b(bc_type, g, hi, ti, ts, magu, kappa, q, Ri_b)
+
+       ! Get q, Ri, f, dfdl based on bc_type
+       call compute_Ri_b(bc_type, g, hi, ti, ts, magu, kappa, q, Ri_b)
```
linneahuusko: This line is a duplicate of line 208

src/wall_models/bcknd/cpu/most_cpu.f90
```diff
@@ -0,0 +1,440 @@
...
+             if (abs(dfdl) < 1.0e-12_rp) call neko_error("Division by zero in dfdl")
+             L_new = L_ob - f/dfdl
+             ! Avoid regime crossing during Newton iter (otherwise crash)
+             if (L_new*L_sign <= 0.0_rp) then
+                ! "damp update" (stay on same side)
+                L_new = 0.5_rp * L_ob
+             end if
+             ! Bound L_ob
+             L_ob = sign(max(abs(L_new), 1.0e-6_rp), L_sign)
+             L_ob = sign(min(abs(L_ob), 1.0e6_rp), L_sign)
```
linneahuusko: Isn't this redundant since we have the check for convergence below? (I'm not 100% sure I understand the purpose of this)

src/wall_models/bcknd/cpu/most_cpu.f90
```diff
@@ -0,0 +1,440 @@
...
+    write(log_buf, '(A,E15.7)') 'ts: ', ts
+    call neko_log%message(trim(log_buf))
+    write(log_buf, '(A,E15.7)') 'ti: ', ti
+    call neko_log%message(trim(log_buf))
+    write(log_buf, '(A,E15.7)') 'q: ', q
+    call neko_log%message(trim(log_buf))
+    write(log_buf, '(A,E15.7)') 'hi: ', hi
+    call neko_log%message(trim(log_buf))
+    call neko_log%end_section()
+
```
linneahuusko: Here I think the goal should be to get the mean, min and max of these variables, and also write them to a separate file

src/wall_models/bcknd/cpu/most_cpu.f90
```diff
@@ -0,0 +1,440 @@
...
+  end function slaw_m_stable
+
+  function slaw_h_stable(z,L_ob,z0h) result(slaw)
+    real(kind=rp), intent(in) :: z,L_ob,z0h
+    real(kind=rp) :: slaw
+
+    slaw = log(z/z0h)-corr_h_stable(z,L_ob)+corr_h_stable(z0h,L_ob)
+  end function slaw_h_stable
+
+  function corr_m_stable(z,L_ob) result(corr)
```
linneahuusko: Do we want to put references for these formulations? Since there are different approaches to this floating around in the literature

src/wall_models/bcknd/device/cuda/rough_log_law_kernel.h
```diff
@@ -1,11 +1,45 @@
...
+
 #ifndef ROUGH_LOG_LAW_KERNEL_H
 #define ROUGH_LOG_LAW_KERNEL_H
 
 #include <cmath>
 #include <algorithm>
 
 /**
- * CUDA kernel for the rough log-law wall model.
+ * HIP kernel for the rough log-law wall model.
```
linneahuusko: This shouldn't be HIP, right? (Since the file is in the `cuda` directory)

src/wall_models/bcknd/device/hip/rough_log_law_kernel.h
linneahuusko:
```suggestion
 * HIP kernel for the rough log-law wall model.
```

src/wall_models/most.f90
```diff
@@ -0,0 +1,221 @@
...
+    call json_get_or_default(json, "z0", z0, 0.1_rp)
+    call json_get_or_default(json, "z0h", z0h_in, -10.0_rp) ! if z0h not specified, assign negative value (tmp)
```
linneahuusko: Maybe clarify in the comment here that a proper value will be computed later (I'm not sure I like this solution but I can't think of anything better right now)

src/wall_models/most.f90
```diff
@@ -0,0 +1,221 @@
...
+    type(field_t), pointer :: v
+    type(field_t), pointer :: w
+    type(field_t), pointer :: temp
+    integer :: i
+    real(kind=rp) :: ui, vi, wi, magu, utau, normu
+
+    u => neko_registry%get_field("u")
+    v => neko_registry%get_field("v")
+    w => neko_registry%get_field("w")
+    temp => neko_registry%get_field("temperature")
```
linneahuusko: Maybe we should allow for other names of the scalar field than `"temperature"`? Like in the buoyancy corrected Vreman, where we put `"scalar_field"` in the json file (that does add another parameter to the json but it might be good to not assume field names)

src/wall_models/wall_model_fctry.f90
```diff
@@ -30,18 +30,22 @@
 ! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 ! POSSIBILITY OF SUCH DAMAGE.
 !
+
+! new comments put by Lorenzo on Nov 26 2025
 submodule (wall_model) wall_model_fctry
-  use vreman, only : vreman_t
+  use vreman, only : vreman_t ! add new models
```
linneahuusko: This is not an addition you made I guess but why is Vreman here?

linneahuusko (commented):
Nice work! I didn't look much at the device files since I don't really know what to look for there, but the rest looks good. I left some minor comments and thoughts about things we should consider how to do.

src/wall_models/bcknd/device/hip/most_kernel.h
```diff
@@ -0,0 +1,401 @@
...
+
+        // Zilitinkevich 1995 correlation for thermal roughness
+        T z0h;
+        if (z0h_in < 0.0) {
+            const T nu = 1.46e-5; // Kinematic viscosity of air [m^2/s]
+            const T Re_z0 = (utau * z0) / nu;
+            z0h = z0 * exp(-0.1 * sqrt(Re_z0));
+        } else {
+            z0h = z0h_in;
+        }
```
timofeymukha, line 380: Just a first comment, haven't gotten to look at the entire code yet.

I really don't like the hardcoded nu here, because it is completely independent of the material properties in the simulation's case file. I think one either picks up the mu / rho from the case file (The wall model should have access to these, I think), or we just make z0h mandatory, and people have to use whatever correlation they want, but outside Neko.

timofeymukha:
@loretet I'm going to order a copilot review now. It will probably spit out a lot of comments, and for now you can just focus on those related to the documentation. Like, it will hopefully pick up that the copyright year is wrong in some files, and similar stuff. For "tougher" comments, read them if you want, or you can wait for me to triage them first.

src/wall_models/bcknd/device/hip/most_kernel.h
loretet: Yes I agree, consider it some kind of placeholder for now: I want to include rho in the wall model computation and I will include mu as well (rho HAS to be there so it's a natural step to add mu as well). Thanks for spotting it!

src/wall_models/bcknd/cpu/most_cpu.f90
loretet: Yes you are right, we can/should definitely generalise it. I have to include rho (and mu) as well in the wall model inputs, I will add g as well (then the user can specify them all in a `constants` block in the case file)

src/wall_models/bcknd/device/most_device.F90
```diff
@@ -0,0 +1,95 @@
...
+    type(c_ptr), intent(in) :: ind_r_d, ind_s_d, ind_t_d, ind_e_d
+    type(c_ptr), intent(in) :: n_x_d, n_y_d, n_z_d, h_d
+    type(c_ptr), intent(inout) :: tau_x_d, tau_y_d, tau_z_d
+    real(kind=rp), intent(in) :: kappa, z0, z0h_in, bc_value
+    character(len=*), intent(in) :: bc_type ! passed as a normal Fortran string
+
+    ! bc_type must be C-compatible string
+    character(kind=c_char, len=len_trim(bc_type)+1), target :: bc_type_c
+    ! add null terminator for C/C++ compatibility
+    bc_type_c = trim(bc_type) // c_null_char
```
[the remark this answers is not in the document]
timofeymukha: @loretet Haven't looked in details at the contents of the suggestion here, but how about you just encode the bc type with an integer or even a boolean when passing things to the device kernels? Just to avoid messing with cross-language string conversion

src/wall_models/bcknd/cpu/most_cpu.f90
loretet: I agree, I didn't want to include this part at first but I left it in the end so that the user can just have a very rough idea if there are weird values or NaNs. A better diagnostics is necessary for sure

src/wall_models/bcknd/device/cuda/most_kernel.h
```diff
@@ -0,0 +1,401 @@
...
+        ui -= normu * nx;
+        vi -= normu * ny;
+        wi -= normu * nz;
+
+        // The magnitude used for MOST is the magnitude of the tangential vector
+        T magu = sqrt(ui*ui + vi*vi + wi*wi);
+        magu = fmax(magu, (T)1e-6); // Avoid division by zero in extreme cases
+
+        // Initial utau estimate
+        T utau = kappa * magu / log(hi/z0);
```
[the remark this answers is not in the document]
timofeymukha: This is a good point, but the check should not be in the kernel loop. I think nothing prevents us to check in the constructor, right? That h is larger than z0 and z0h for each node?

src/wall_models/bcknd/device/hip/most.hip
```diff
@@ -0,0 +1,71 @@
...
+                                          *n_nodes, *lx, *kappa, *z0, *z0h_in, *bc_value);
+        } else if (strncmp(bc_type, "dirichlet", 9) == 0) {
+          hipLaunchKernelGGL(HIP_KERNEL_NAME(most_compute<real,1>),
+                                          nblcks, nthrds, 0, (hipStream_t) glb_cmd_queue,
+                                          (real *) u_d, (real *) v_d, (real *) w_d, (real *) temp_d,
+                                          (real *) h_d, (real *) n_x_d, (real *) n_y_d, (real *) n_z_d,
+                                          (int *) ind_r_d, (int *) ind_s_d, (int *) ind_t_d, (int *) ind_e_d,
+                                          (real *) tau_x_d, (real *) tau_y_d, (real *) tau_z_d,
+                                          *n_nodes, *lx, *kappa, *z0, *z0h_in, *bc_value);
+                                        }
```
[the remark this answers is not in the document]
timofeymukha: This can be disregraded, at kernel level the input should be guaranteed to be valid. Just make sure we have a check somewhere in the code.

src/wall_models/most.f90
```diff
@@ -0,0 +1,221 @@
...
+    integer, intent(in) :: msk(:)
+    integer, intent(in) :: facet(:)
+    integer, intent(in) :: h_index
+    type(json_file), intent(inout) :: json
+    real(kind=rp) :: kappa, z0, z0h_in
+    character(len=:), allocatable :: bc_type
+    real(kind=rp) :: bc_value
+
+    call json_get_or_default(json, "kappa", kappa, 0.41_rp)
+    call json_get_or_default(json, "z0", z0, 0.1_rp)
```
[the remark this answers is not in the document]
timofeymukha: Correct

src/wall_models/most.f90
```diff
@@ -0,0 +1,221 @@
...
+    this%bc_value = bc_value
+  end subroutine most_init_from_components
+
+  !> Destructor for the most_t (base) class.
+  subroutine most_free(this)
+    class(most_t), intent(inout) :: this
+
+    call this%free_base()
+
+  end subroutine most_free
```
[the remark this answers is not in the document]
timofeymukha: Correct

src/wall_models/most.f90
```diff
@@ -0,0 +1,221 @@
...
+    type(coef_t), intent(in) :: coef
+    type(json_file), intent(inout) :: json
+
+    call this%partial_init_base(coef, json)
+    call json_get_or_default(json, "kappa", this%kappa, 0.41_rp)
+    call json_get_or_default(json, "z0", this%z0, 0.1_rp)
+    call json_get_or_default(json, "z0h", this%z0h_in, -10.0_rp)
+    call json_get(json, "type_of_temp_bc", this%bc_type)
+    call json_get(json, "bottom_bc_flux_or_temp", this%bc_value)
+
```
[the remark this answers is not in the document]
timofeymukha: Correct comment, but unrelated suggestion?

src/wall_models/wall_model_fctry.f90
```diff
@@ -30,18 +30,22 @@
...
   use spalding, only : spalding_t
-  use rough_log_law, only : rough_log_law_t
+  use rough_log_law, only : rough_log_law_t ! add new models
+  use most, only : most_t
```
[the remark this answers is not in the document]
timofeymukha: I agree. We can also remove the vreman_t from here.

src/wall_models/most.f90
```diff
@@ -0,0 +1,221 @@
...
+  subroutine most_compute(this, t, tstep)
+    class(most_t), intent(inout) :: this
+    real(kind=rp), intent(in) :: t
+    integer, intent(in) :: tstep
+    type(field_t), pointer :: u
+    type(field_t), pointer :: v
+    type(field_t), pointer :: w
+    type(field_t), pointer :: temp
+    integer :: i
+    real(kind=rp) :: ui, vi, wi, magu, utau, normu
```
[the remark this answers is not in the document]
timofeymukha: Correct if really unsused

src/wall_models/bcknd/device/most_device.F90
```diff
@@ -0,0 +1,95 @@
...
+          tau_x_d, tau_y_d, tau_z_d, n_nodes, lx, &
+          kappa, z0, z0h_in, bc_type, bc_value, tstep) &
+          bind(c, name = 'cuda_most_compute')
+       use, intrinsic :: iso_c_binding, only : c_ptr, c_int
+       use num_types, only : c_rp
+       implicit none
+       type(c_ptr), value :: u_d, v_d, w_d, temp_d
+       type(c_ptr), value :: ind_r_d, ind_s_d, ind_t_d, ind_e_d
+       type(c_ptr), value :: n_x_d, n_y_d, n_z_d, h_d
+       type(c_ptr) :: bc_type
```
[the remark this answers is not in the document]
timofeymukha: See also below. Let's avoid all of this by not passing a string to device kernels.

src/wall_models/bcknd/device/hip/most.hip
```diff
@@ -0,0 +1,71 @@
...
+ INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#include <hip/hip_runtime.h>
```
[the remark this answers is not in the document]
timofeymukha: Correct, but will be resolved if we don't use strings for bc_type

src/wall_models/bcknd/cpu/most_cpu.f90
```diff
@@ -0,0 +1,440 @@
...
+contains
+
+  !> Selects different expressions for the similarity functions in  MOST
+  !> based on the type of bottom boundary condition for temperature.
+  subroutine select_bc_operators(bc_type,bc_value,q,ts,ti,kappa,utau,z0h,hi)
+    character(len=*), intent(in) :: bc_type
+    real(kind=rp), intent(in) :: hi, ti, kappa, utau, z0h, bc_value
+    real(kind=rp), intent(inout) :: q,ts
+    select case (bc_type)
+      case ("neumann")
```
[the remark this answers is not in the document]
timofeymukha: Looks legit. We need to talk about how to format that log, but if we keep ts, the comment looks valid.

src/wall_models/bcknd/cpu/most_cpu.f90
loretet: I think it's better to keep both. This is inside the loop and acts as some kind of computational guardrail which avoids crashing (especially near neutral regime when 1/L is almost zero), while the check after the loop is more of a "physics" check that evaluates whether the L value obtained from the loop makes physical sense or not.

I think in practical terms the difference between the two checks is very small but still I'd keep them both

src/wall_models/bcknd/cpu/most_cpu.f90
loretet: That might be nice, where did we get these from? 😆

src/wall_models/most.f90
loretet: yeah I hate this solution but I think it's better than not giving the possibility at all...

src/wall_models/most.f90
```diff
@@ -0,0 +1,221 @@
+! Copyright (c) 2024, The Neko Authors
```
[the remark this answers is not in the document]
timofeymukha: So, copilot clearly doesn't know it is 2026 :D.

src/wall_models/bcknd/cpu/most_cpu.f90
```diff
@@ -0,0 +1,435 @@
...
+  end function slaw_h_stable
+
+  function corr_m_stable(z,L_ob) result(corr)
+    real(kind=rp), intent(in) :: z,L_ob
+    real(kind=rp) :: corr
+    real(kind=rp) :: a, b, c, d
+    real(kind=rp) :: zeta
+    zeta = z/L_ob
+    a = 1.0_rp
+    b = 0.6666666_rp
```
timofeymukha: Can we use `2.0_rp / 3.0_rp? Seems like an unnecessary loss of precision. I think there are a few other places like this.

src/wall_models/bcknd/device/most_device.F90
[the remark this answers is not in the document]
loretet: solved in commit 336bbe1

src/wall_models/bcknd/cpu/most_cpu.f90
loretet: True, corrected in a354f23b

src/wall_models/bcknd/cpu/most_cpu.f90
loretet: solved in commit 2d5a8bd

src/wall_models/bcknd/device/cuda/most_kernel.h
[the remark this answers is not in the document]
loretet: I put it in most.f90 and rough_log_law.f90, commit f391fd

src/wall_models/most.f90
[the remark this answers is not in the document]
loretet: Can I solve this in this way?

```diff
  subroutine most_init(this, scheme_name, coef, msk, facet, &
       h_index, json)
    class(most_t), intent(inout) :: this
    character(len=*), intent(in) :: scheme_name
    type(coef_t), intent(in) :: coef
    integer, intent(in) :: msk(:)
    integer, intent(in) :: facet(:)
    integer, intent(in) :: h_index
    type(json_file), intent(inout) :: json
    real(kind=rp) :: kappa, z0, z0h_in
-    character(len=:), allocatable :: bc_type
    real(kind=rp) :: bc_value
    real(kind=rp), pointer :: bc_value

+    call json_get(json, "type_of_temp_bc", this%bc_type)
    call json_get_or_default(json, "kappa", kappa, 0.4_rp)
    call json_get_or_default(json, "z0", z0, 0.1_rp)
    call json_get_or_default(json, "z0h", z0h_in, -10.0_rp)
    ! if z0h not specified, assign negative
    ! and compute automatically with Zilitinkevich
    call json_get(json, "type_of_temp_bc", bc_type)

    call neko_const_registry%add_real_scalar(this%bc_value, "bc_value")

    bc_value => neko_const_registry%get_real_scalar("bc_value")

    this%bc_value = bc_type

    call this%init_from_components(scheme_name, coef, msk, facet, h_index, &
+         kappa, z0, z0h_in, this%bc_type, bc_value)
  end subroutine most_init

```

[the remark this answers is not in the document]
timofeymukha: Yes. Otherwise, it is basically just adding deallocate(bc_type) in the end of the routine.

src/wall_models/most.f90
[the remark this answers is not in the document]
loretet: ...or should I keep everything as is and just deallocate it here?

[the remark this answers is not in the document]
timofeymukha: The two comments were about different variables. You have a bc_type in your type, and this must be deallocated here, if it has been allocated. so you add a standard if (allocated(this%bc_type)) deallocate(this%bc_type)

The comment above was about a local variable you introduced, also called bc_type. That one should be deallocated in the end of that subroutine. Or, you can just skip having that variable and use this%bc_type there directly.

src/wall_models/most.f90
loretet: done in commit b8955ce (following vreman.f90)

src/wall_models/most.f90
```diff
@@ -100,14 +100,22 @@ subroutine most_init(this, scheme_name, coef, msk, facet, &
     real(kind=rp) :: kappa, z0, z0h_in
     character(len=:), allocatable :: bc_type
     real(kind=rp) :: bc_value
+    real(kind=rp), pointer :: bc_value
```
loretet: Is this double declaration needed?

src/wall_models/most.f90
```diff
@@ -100,14 +100,22 @@ subroutine most_init(this, scheme_name, coef, msk, facet, &
...
     call json_get(json, "type_of_temp_bc", bc_type)
-    call json_get(json, "bottom_bc_flux_or_temp", bc_value)
+    ! call json_get(json, "bottom_bc_flux_or_temp", bc_value)
+
+    call neko_const_registry%add_real_scalar(this%bc_value, "bc_value")
+
+    bc_value => neko_const_registry%get_real_scalar("bc_value")
+
+    this%bc_value = bc_type
+
```
loretet: Not sure the logic here is clear to me (and I guess line 117 has a typo, but I fixed it already)

loretet: I am going to take this out and finish the PR without this functionality. I will create another branch where we can experiment a bit more before pushing it

loretet:
We have now (finally) finished with this PR. It is ready for the final review and merge.

Only thing that's missing is having the mean of some variables in the diagnostics, but we'll do this with a later PR. 

The physics and inputs are all there and fully working on both CPU and GPU

linneahuusko:
Since Lorenzo wrote the description of this PR we have added diagnostics of key variables to the log. An example is shown as it appears in the log file for a single timestep, and as the variables can be plotted over time.

```
  ------------------------------------------------------------------------------
 Step =           4 t =   2.750849608518615E-01                     [   0.00% ]
 ------------------------------------------------------------------------------

   --------Preprocessing---------
   CFL:  0.5017  dt: 6.8771240E-02

   ------------Fluid-------------

    ----Wall model diagnostics----
    --- sum ---
    Ri_b:   0.2611577E-03
    L_ob:   0.0000000E+00
    utau:   0.3207647E+05
    magu:   0.2046056E+06
    ti:   0.6784000E+07
    ts:   0.6784000E+07
    q:  -0.6886292E-01
```

image

The diagnostics are intended to include the mean, min, and max values but currently only have the sum, which can be divided by `(Nx*8)*(Ny*8)`, where Nx, Ny are the number of elements in the x and y directions, to get an approximation of the mean. The full functionality will be added once the required functions are in place in other parts of the code.

src/wall_models/bcknd/device/hip/most.hip
```diff
@@ -0,0 +1,83 @@
...
+     contributors may be used to endorse or promote products derived
+     from this software without specific prior written permission.
+
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+     LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;∂
```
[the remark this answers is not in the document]
timofeymukha: That's a funny one!

timofeymukha (approved):
I purge the rho_val and mu_val stuff completely and we just pass stub values for now. To be fixed in a future PR for all wall models, as agreed.

loretet:
> I purge the rho_val and mu_val stuff completely and we just pass stub values for now. To be fixed in a future PR for all wall models, as agreed.

Sorry, I misunderstood. I thought we could just leave them as they were since they defaulted ot 1.0 and 1e-10 already. My bad! But thanks for doing it.

I opened an issue to track the rho and mu thing: #2476

njansson (approved):

PR #2387, opened 2026-03-17
rewrite multigrid cycle to no longer be recursive

### LINKED ISSUE #2374: solver crashes when phmg and runtime_statistics are used together
When "runtime_statistics" : {"enabled" : true} , and at the same time phmg precond is used, the solver crashes at first time step, and gives:
 *** ERROR: Profile region renamed ***.

Two profilers are having the same ids when phmg is used to get this behaviour.

pohm01:
Fixes indexing mistake when freeing tamg smoother.

Addresses #2374 by removing the profiler calls in phmg.


njansson:
This also resolve the issue that recent CCE (19 and later) take ages (4h+) to compile phmg

timfelle:
There is no need to remove the profiler regions if they are useful. All that need to be done is either find an unused ID or let the system supply an ID by removing the Id from the call site.

pohm01:
Ah, I had removed the profiler regions because I thought it better matched the usage elsewhere in the code (most of the other linear solvers don't have profiler regions). But now I see that hsmg also uses profiler regions, so I've added back the profiler regions to phmg and added regions to tamg.

njansson:
> Ah, I had removed the profiler regions because I thought it better matched the usage elsewhere in the code (most of the other linear solvers don't have profiler regions). But now I see that hsmg also uses profiler regions, so I've added back the profiler regions to phmg and added regions to tamg.

Can't we keep the all regions, or will that make the runtime profiler sad?

pohm01:
what do you mean by the all regions?

njansson:
> what do you mean by the all regions?

isn't the smoother region missing now ?

pohm01:
Figured if I want one for the smoother, it should go inside the smoother itself

njansson:
> Figured if I want one for the smoother, it should go inside the smoother itself

True, make sense!

njansson (approved):
🍻 🦫 🎉

timfelle (approved):

PR #2391, opened 2026-03-18
Use device_unmap in destructors

### LINKED ISSUE #2350: Missing deassociate calls in destructors
Many destructors handling device pointers miss a call to deassociate and only call device_free.

timofeymukha:
Propagates the use of device_unmap through the codebase.


njansson:
Wouldn't it be a bit more safe if we still do the test with `c_associated`, to avoid that we try to `unmap` and `free` a device pointer?  (or if we want to avoid these if, we could add the check in `unmap`, not sure what is best but somehow it's nice not to hide too many important details)

timfelle:
> Wouldn't it be a bit more safe if we still do the test with `c_associated`, to avoid that we try to `unmap` and `free` a device pointer? (or if we want to avoid these if, we could add the check in `unmap`, not sure what is best but somehow it's nice not to hide too many important details)

From what I can see, all the updates here mirror the initializers. So I think we should not add the extra `c_associated` guards. These types are written such that if devices are enabled we map the arrays. By using unmap, you will be throwing an error if someone messed with the internals and deallocated or nullified something. I might even suggest we move the unmap outside the allocated check. Of course in that case you need to unmap, then do the allocation check.

njansson:
> > Wouldn't it be a bit more safe if we still do the test with `c_associated`, to avoid that we try to `unmap` and `free` a device pointer? (or if we want to avoid these if, we could add the check in `unmap`, not sure what is best but somehow it's nice not to hide too many important details)
> 
> From what I can see, all the updates here mirror the initializers. So I think we should not add the extra `c_associated` guards. These types are written such that if devices are enabled we map the arrays. By using unmap, you will be throwing an error if someone messed with the internals and deallocated or nullified something. I might even suggest we move the unmap outside the allocated check. Of course in that case you need to unmap, then do the allocation check.

Fully agree that we do mirror init and in principle it will be fine. But without checking the content of the pointer, we don't know if something bad has happened during the run. E.g. some consumer has deallocated an array etc (such an error will be a pain to find (now we just silently ignores it, which is probably also not the best thing))

timfelle:
> > > Wouldn't it be a bit more safe if we still do the test with `c_associated`, to avoid that we try to `unmap` and `free` a device pointer? (or if we want to avoid these if, we could add the check in `unmap`, not sure what is best but somehow it's nice not to hide too many important details)
> > 
> > 
> > From what I can see, all the updates here mirror the initializers. So I think we should not add the extra `c_associated` guards. These types are written such that if devices are enabled we map the arrays. By using unmap, you will be throwing an error if someone messed with the internals and deallocated or nullified something. I might even suggest we move the unmap outside the allocated check. Of course in that case you need to unmap, then do the allocation check.
> 
> Fully agree that we do mirror init and in principle it will be fine. But without checking the content of the pointer, we don't know if something bad has happened during the run. E.g. some consumer has deallocated an array etc (such an error will be a pain to find (now we just silently ignores it, which is probably also not the best thing))

Which is why I argue that we move it outside the `if (allocated` block.
The unmap does nothing, if the host array is not mapped, and the device pointer is not associated. Otherwise it will throw an error if something is wrong with the mapping, no longer associated to the expected pointer etc.

timofeymukha:
My master plan to settle everything in yesterday's small PR has clearly failed :-). 

It's true that if someone deallocates the host array randomly, the device pointer will leak. But passing an unallocated array to a dummy argument which is not itself allocatable is illegal, or? I guess we have to have an else if clause where we check the c_associate on x_d and free that if associated. I would not put that stuff in unmap, it should just reverse a valid map.

So in the end we will have about the same verbosity level as for just adding a dissassociate directly and not using unmap, but unmap gives us extra validity checks.

To be honest, in 90% of these types, one should just use vector_t and matrix_t instead of storing x and x_d. Then, it will just be a matter of calling vector_t%free, and we can be very verbose and careful in that destructor.

njansson:
> My master plan to settle everything in yesterday's small PR has clearly failed :-).
> 
> It's true that if someone deallocates the host array randomly, the device pointer will leak. But passing an unallocated array to a dummy argument which is not itself allocatable is illegal, or? I guess we have to have an else if clause where we check the c_associate on x_d and free that if associated. I would not put that stuff in unmap, it should just reverse a valid map.
> 
> So in the end we will have about the same verbosity level as for just adding a dissassociate directly and not using unmap, but unmap gives us extra validity checks.
> 
> To be honest, in 90% of these types, one should just use vector_t and matrix_t instead of storing x and x_d. Then, it will just be a matter of calling vector_t%free, and we can be very verbose and careful in that destructor.



> My master plan to settle everything in yesterday's small PR has clearly failed :-).
> 
> It's true that if someone deallocates the host array randomly, the device pointer will leak. But passing an unallocated array to a dummy argument which is not itself allocatable is illegal, or? 

True, my understanding of the standard is that passing an unallocated to a not (marked) allocatable dummy argument should be illegal (but a language lawyer might be needed decide on that)

>I guess we have to have an else if clause where we check the c_associate on x_d and free that if associated. I would not put that stuff in unmap, it should just reverse a valid map.
> 

Agree, check and then call unmap instead of a foolproof unmap, like a free in C, it would accept anything ;)

> So in the end we will have about the same verbosity level as for just adding a dissassociate directly and not using unmap, but unmap gives us extra validity checks.
> 
> To be honest, in 90% of these types, one should just use vector_t and matrix_t instead of storing x and x_d. Then, it will just be a matter of calling vector_t%free, and we can be very verbose and careful in that destructor.

In an ideal world yes... but with the current state of compilers, maybe not. These structures prevents optimisation (sigh), unless do concurrent or we sprinkle pragmas everywhere (yeah I know it's shouldn't be needed but it is sadly)

timfelle (approved):
Lets get this one rolling. The discussions here were more about the general direction rather than the changes here. All we did here was move to map -> unmap pairing rather than map -> free+disassociate.

timfelle (commented):
@njansson, happy?

njansson (approved):

PR #2404, opened 2026-03-22
Add omp pragmas in key kernels

### LINKED ISSUE #2336: Ensure performant multithreading of key kernels
Auto parallelisation produces less optimal multithreaded version of key kernels like `ax_helm`.  Hints, pragmas etc is needed to steer the parallelisation to create threads over the region of most work, e.g. elements.

njansson:
This PR add openmp statement in key kernels, which would not parallelise efficiently using auto parallelisation. fixes #2336

### DIFF
--- a/src/math/bcknd/cpu/ax_helm_cpu.f90
+++ b/src/math/bcknd/cpu/ax_helm_cpu.f90
@@ -1,4 +1,4 @@
-! Copyright (c) 2021-2024, The Neko Authors
+! Copyright (c) 2021-2026, The Neko Authors
 ! All rights reserved.
 !
 ! Redistribution and use in source and binary forms, with or without
@@ -36,7 +36,6 @@ module ax_helm_cpu
   use coefs, only : coef_t
   use space, only : space_t
   use mesh, only : mesh_t
-  use math, only : addcol4
   implicit none
   private
 
@@ -63,9 +62,10 @@ contains
     type(coef_t), intent(in) :: coef
     real(kind=rp), intent(inout) :: w(Xh%lx, Xh%ly, Xh%lz, msh%nelv)
     real(kind=rp), intent(in) :: u(Xh%lx, Xh%ly, Xh%lz, msh%nelv)
+    integer :: i
 
-
-    select case (Xh%lx)
+    !$omp parallel private(i)
+    select case(Xh%lx)
     case (14)
        call ax_helm_lx14(w, u, Xh%dx, Xh%dy, Xh%dz, Xh%dxt, Xh%dyt, Xh%dzt, &
             coef%h1, coef%G11, coef%G22, coef%G33, coef%G12, coef%G13, &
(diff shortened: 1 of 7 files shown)


src/math/bcknd/cpu/ax_helm_cpu.f90
```diff
@@ -63,9 +62,10 @@ subroutine ax_helm_compute(w, u, coef, msh, Xh)
     type(coef_t), intent(in) :: coef
     real(kind=rp), intent(inout) :: w(Xh%lx, Xh%ly, Xh%lz, msh%nelv)
     real(kind=rp), intent(in) :: u(Xh%lx, Xh%ly, Xh%lz, msh%nelv)
+    integer :: i
 
-
-    select case (Xh%lx)
+    !$omp parallel private(i)
```
timfelle: All the other places you place the private statement on the `do` rather than the
`parallel` How about doing it consistently?

src/math/bcknd/cpu/ax_helm_full_cpu.f90
```diff
@@ -1909,6 +1923,7 @@ subroutine ax_helm_stress_lx10(au, av, aw, u, v, w, Dx, Dy, Dz, &
     real(kind=rp) :: s11, s22, s33, s12, s13, s23
     real(kind=rp) :: u1, u2, u3, v1, v2, v3, w1, w2, w3
 
+    !$omp do
```
timfelle, line 1926: No private variables?

src/math/bcknd/cpu/ax_helm_cpu.f90
njansson: Fair enough, I'll remove them since they aren't needed (due to the placement in a subroutine)

src/math/bcknd/cpu/ax_helm_full_cpu.f90
njansson: It's in a subroutine, so per the standard, local variables will be automatically thread private

src/math/bcknd/cpu/ax_helm_cpu.f90
```diff
@@ -124,8 +124,15 @@ subroutine ax_helm_compute(w, u, coef, msh, Xh)
...
-
+    if (coef%ifh2) then
+       !$omp do
+       do i = 1, coef%dof%size()
+          w(i,1,1,1) = w(i,1,1,1) + &
+               coef%h2(i,1,1,1) * coef%B(i,1,1,1) * u(i,1,1,1)
+       end do
+       !$omp end do
+    end if
+    !$omp end parallel
```
timfelle, line 135: Could addcol4 not do this internally?

timfelle (approved):

src/math/bcknd/cpu/ax_helm_cpu.f90
njansson: For normal axhelm it should be "fine", it would only be fork-join-barrier-fork-join-barrier. But for the stress formulation it would be too many fork-join so I tried to make it consistent.

The bigger question is how one want's to deal with `math.f90`, if one just do `!$omp do` there, we would only loose out on possible loop fusions in various part of the code. But then we have to make sure that it's always called from within a parallel region (a bigger mess to untangle)

timfelle: Yea I can see the issue. Is that something `do concurrent` potentially could help resolve? Almost all math operations are point wise and do not use the concurrent system.

njansson: Seems like we need to have full `f2018` support for `do concurrent`, or rather support for `local/shared` to ensure that compilers generate good threading. Fujitsu is an exception here, while CCE is rather sad without

njansson: I'll have a look at this in a different pr (related to getting good performance out of pnpn)

pohm01 (approved):

PR #2411, opened 2026-03-24
Feature/no_slip_t

arash-mp:
I added no_slip_t for no_slip BC. It extends zero_dirichlet_t, but allows more flexibility. For example, as needed for ALE, setting a given velocity to wall instead of zero.

It only overrides `init`, `apply_vector`, `apply_vector_dev`, `free` from zero_dirichlet_t.
If moving: true is not set explicitly, the behaviour should be exactly the same as zero_dirichlet, and no additional changes in the existing codes or case files are needed.


adperezm (approved):

timfelle (approved):

adperezm:
So what about this? This was merged on ALE no? Do we close it? @arash-mp

arash-mp:
> So what about this? This was merged on ALE no? Do we close it? @arash-mp

Yes,
I will clsoe it now.

arash-mp:
It is already merged with ALE.

PR #2412, opened 2026-03-24
Expose global interp parameters to the user via JSON subdict

vbaconnet:
This PR does three main things:
1. Expose the `padding` parameter to `import_fields` and downstream routines in `fld_file_data` (previously only had `tolerance`, but padding is equally important).
2. Expose `tolerance` and `padding` to the probes simcomp.
3. Allow global interpolation to be initialized from a JSON subdirectory containing the relevant parameters. This functionality is applied to probes and fluid/scalar IC. This is a *breaking change* for fluid/scalar initial condition. 
```json
"initial_condition": {
  "type": "field",
  "file_name": "field0.f00001",
  "interpolate": true,
   "interpolation": {
      "tolerance": 1e-8, <- previously this was given in the "initial_condition" object. 
      "padding": 1e-3
    }
}
```

Compiled and tested on Cray with the `advecting_cone` example.

---

More minor additions:
- `global_interpolation`: 
  - Added two public variables for default values of tolerance and padding. These are based on the defaults Martin left in the code.
  - Made tolerance and padding double precision variables. Martin used a bit of a mix as you can see from the diff, but here for robustness they are changed to always be in double precision.
- `json_utils`:
  - Added a routine `json_get_subdict_or_empty` which does what its name says. Is extends the capability of the existing `json_get_subdict` to not throw an error if the requested subdict is not found. This makes things clearer and avoids checks/if statements when initializing `probes` and the initial condition. It lives under the `json_get_or_default` interface currently
- Documentation:
  - Updated the index page to match the contents of the `user-guide` pages.
  - @timofeymukha I took the liberty to try something in the programming examples page with foldable menus to hide the code snippets. I think these things help a lot with readability, I like them quite a lot. @Mr-2c had some comments about the doc being too long! Let me know what you think, otherwise we can revert back to the original.

### Note
I also started experimenting with `block` statements, I find them to make the code a bit clearer by focusing the scope of variables on parts of the code instead of having a long list of variables declared at the beginning. But that's personal preference :) Curious to hear what people think


src/fluid/flow_ic.f90
```diff
@@ -401,24 +407,22 @@ end subroutine set_flow_ic_point_zone
...
   !! values onto the current mesh.
-  !! @param tolerance If interpolation is enabled, tolerance for finding the
-  !! points in the mesh.
-  !! @param sample_mesh_idx If interpolation is enabled, index of the field
-  !! file where the mesh coordinates are located.
+  !! @param file_name If interpolation is enabled the name of the "fld" file
+  !! containing the mesh coordinates (if they are not in `file_name`).
+  !! @param global_interp_subdict If interpolation is enabled, subdict
+  !! containing the interpolation parameters to use.
   subroutine set_flow_ic_fld(u, v, w, p, file_name, &
```
adperezm, line 418: I understand the reason for this one, but the risk of doing this sort of thing is that the actual inputs to the function are less readable, since they are defined inside the dictionary.

I think we can keep this, since this gives us good flexibility, but for future implementations, we must be conscious to not over-rely on it.

Or at least I remember seeing codes (particularly in python) where inputs were given packed in a dictionary and it was always messy to understand what was needed.

src/fluid/flow_ic.f90
```diff
@@ -429,7 +433,8 @@ subroutine set_flow_ic_fld(u, v, w, p, file_name, &
 
     call import_fields(file_name, mesh_file_name, &
```
adperezm: For example here, already we start to hide things. I would not say that it is the best idea to nest the dictionary to many routines too deep. It is good to have it on the user-facing side, but I do like having access to the particular input.

Again, just comments, I think we can keep this as you have modified it.

adperezm (approved):
I added some comments but I think they are not crucial and might also be coming from misunderstanding.

Overall I like this! because it is also the first step in removing those env variables related to interpolation :D

src/fluid/flow_ic.f90
vbaconnet: Yes I agree that is really frustrating and I've had the same experience with Python.

For parts of the code like initial condition & sponge that are meant to be used purely from a case file perspective, I think there is an effort to be put into documentation, to show _exactly_ what the component will be reading from the JSON and what the user can play around with. 

For routines like `import_fields` the actual individual parameters are not removed, you can still provide them individually.

src/fluid/flow_ic.f90
vbaconnet: Yea here I use the subdict because it is all "case-file" based functionalities so it is just a matter of passing those parameters down to global interpolation. But for `import_fields`, `fld_file_data%import_fields` and `global_interpolation`, you can always specify individual inputs if you want to. 

This is also in case the configuration parameters for the global interpolation starts to grow, then you don't need to adjust all the interfaces and add arguments, you'd just modify the global interpolation constructors

src/fluid/flow_ic.f90
adperezm: Ohhh then that its perfect. So ignore :)

src/global_interpolation/global_interpolation.f90
```diff
@@ -33,41 +33,45 @@
...
+  use mask, only : mask_t
   use structs, only : array_ptr_t
-  use, intrinsic :: iso_c_binding, only: c_ptr, C_NULL_PTR, c_associated
+  use, intrinsic :: iso_c_binding, only : c_ptr, C_NULL_PTR, c_associated
   implicit none
   private
 
   integer, public, parameter :: GLOB_MAP_SIZE = 4096
+  real(kind=dp), public, parameter :: GLOBAL_INTERP_TOL = NEKO_EPS*1e3_dp
+  real(kind=dp), public, parameter :: GLOBAL_INTERP_PAD = 1e-2_dp
```
timofeymukha, line 75: Could we have either GLOB or GLOBAL for the same set? You know my preferences I guess, I prefer more explicit :).

src/simulation_components/probes.F90
```diff
@@ -180,44 +182,82 @@ subroutine probes_init_from_json(this, json, case)
...
 
     call probes_show(this)
-    call this%init_from_components(case%fluid%dm_Xh, output_file, name)
+
+    block
+      type(json_file) :: interp_subdict
+      call json_get_or_default(json, "interpolation", interp_subdict)
+      call this%global_interp%init(case%fluid%dm_Xh, &
+           params_subdict = interp_subdict)
+
```
timofeymukha, line 231: I guess this is a good place to show what I would perhaps change. Is it really necessary to have the JSON passed down to objects that are not themselves initialized through the user file? To me it looks like you could just have the global_interp constructor accept the tolerance and padding, without any optionals. The module now provides default values. So if the calling code doesn't know what to do, it can grab those default from the module and use them. Like here you could handle the JSON, and if it isn't there, just use the defaults, right? I would perhaps go through the routine you added where we either pass values or JSON (so 3 optionals), and see if we can tighten it up.

src/common/json_utils.f90
```diff
@@ -450,6 +450,32 @@ subroutine json_get_subdict(json, key, output)
 
   end subroutine json_get_subdict
 
+  !> Extract a sub-object from a json object and returns
+  !! an empty object if the key is missing.
+  subroutine json_get_subdict_or_empty(json, key, output)
+    type(json_file), intent(inout) :: json
```
timofeymukha, line 457: Good to have, but I would pull it out of the get_or_default umbrella. When reading your code I was first confused about what is going on, i.e. what is the default here. The name of the routine makes it very clear.

timofeymukha (commented):
Thanks Victor. Nice addition, the only thing I didn't get convinced about is the need for this point:

> Added the possibility to initialize `global_initialization` from a JSON subdict.

But please push back if it is actually really needed, I mighta missed it.

src/global_interpolation/global_interpolation.f90
```diff
@@ -119,9 +123,9 @@ module global_interpolation
      !> Turns true if points are redistributed to their respective owners
      logical :: all_points_local = .false.
      !> Tolerance for Newton solve to find the correct rst coordinates.
-     real(kind=rp) :: tol = NEKO_EPS*1e3_rp
+     real(kind=dp) :: tolerance = GLOBAL_INTERP_TOL
```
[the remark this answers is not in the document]
timofeymukha: Is this why the CI died maybe?

timofeymukha:
The copilot comments warrant a look. Particularly the stuff with optional forwarding, we gotta really be careful with that: that was the cause of the recent regression with projections code. This is partially why I reacted negatively to the 3 optionals in the routine signatures, these things are just too easy to miss up.

src/global_interpolation/global_interpolation.f90
vbaconnet: sorry i don't understand what you mean... :D

vbaconnet:
> the only thing I didn't get convinced about is the need for this point:
> > Added the possibility to initialize `global_initialization` from a JSON subdict.

Sorry for the longer comment, I want to make sure I can explain my thinking process around this choice :)

---

That's a good question and I should have been more clear about why I decided to add this. 

The main features that were needed for this PR were:
1. Expose global interpolation parameters (tolerance, padding) to `probes`
2. Expose the `padding` parameter to all components currently using global interpolation:
    - "case file" components: fluid/stats IC and sponge
    - "user file" components: `import_fields`, `fld_file_data%import_fields`, `fld_file_data%generate_interpolator`.

The reason for introducing an `"interpolation"` subdict for components that use global interpolation was to homogenize the case file interface, to separate between (i) what keywords/parameters are configuring the actual component, and (ii) what keywords/parameters are configuring _specifically_ the global interpolation. I think that makes it clearer and removes potential confusions on the semantics of the `tolerance` and `padding` keywords in the context of the component that is being configured.

Now for the part about initializing `global_interpolation` with a JSON subdict. fluid/scalar IC, probes and sponge are components that delegate work to interpolation tools, but that is not their main purpose. They will, however, ask for the exact same JSON subdict for configuring the global interpolation. If you see the interpolation as work that is being "delegated", then to me it makes sense that the JSON subdict is also just passed down as a hot potato to be handled by the global interpolation. If they want to grab something from that JSON on the way they can (see for example in `fld_file_data%generate_interpolator`), but ultimately the `global_interpolation_t` object takes care of that subdict for its initialization.

Granted, at this stage there are only two parameters which are just reals, so it feels a bit unnecessary. These are my arguments for why passing down JSON to global interpolation is needed:
1. It protects the "uniform" way of passing down parameters so that the JSON interface is always consistent. Both from the user side and the developer side. Let's say someone unaware would add a new simcomp that needs interpolation, and in the constructor queries the case file for interpolation parameters with keys `"interpolation_params.tol"` and `"interpolation_params.pad"`? Then you'd have to document that, you'd have a different way of passing arguments for different components, although it's the exact same thing that is being done in the end. Instead, from the user side, we guarantee/enforce that: "whenever a component uses interpolation, configuration parameters should always be passed in this format, with these exact keys". It makes documentation much easier and users are happy with a consistent user interface.
2. If if additions are made to global interpolation which require more/new configuration parameters, you wouldn't have to adjust the procedures/constructors of all components that use global interpolation. If global interpolation needs more/new parameters, they can just be added to the subdict in the case file, and will be taken care of once they arrive at the global interpolation constructor. For components that are more "user file"-based like `import_fields`, etc, you can see that those can still be configured by individual parameters `..., tolerance = tol, padding = pad, ...`, so there is no JSON dependency enforced. You'd still need to add dummy arguments to those, ultimately it limits the amount of work needed when things need to change.

Curious to hear your thoughts! I'm perfectly fine with keeping global interpolation free of any JSON stuff as well. It's not critical for the actual features that are needed right now :)

src/common/json_utils.f90
vbaconnet: Fair!

src/simulation_components/probes.F90
vbaconnet: See my comment below, hopefully it answers some of the questions on this review comment?

vbaconnet: > I would perhaps go through the routine you added where we either pass values or JSON (so 3 optionals), and see if we can tighten it up.

Not sure what you mean by tighten it up? Do you mean removing the subdict?

src/io/import_field_utils.f90
```diff
@@ -84,14 +85,16 @@ module import_field_utils
   !! to device when necessary, i.e. only the required fields are copied to
   !! device.
   subroutine import_fields(fname, mesh_fname, u, v, w, p, t, s_target_list, &
-       s_index_list, interpolate, tolerance)
+       s_index_list, interpolate, tolerance, padding, global_interp_subdict)
     character(len=*), intent(in) :: fname
```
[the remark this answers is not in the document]
vbaconnet: Working on it.

src/io/import_field_utils.f90
```diff
@@ -219,7 +217,8 @@ subroutine import_fields(fname, mesh_fname, u, v, w, p, t, s_target_list, &
 
     ! Call the import of fields
     call fld_data%import_fields(u, v, w, p, t, s_target_list, s_index_list, &
-         interpolate_, tolerance = tolerance)
+         interpolate_, tolerance = tolerance, padding = padding, &
+         global_interp_subdict = global_interp_subdict)
```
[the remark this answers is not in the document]
vbaconnet: Wrong. tolerance, padding and global_interp_subdict are also optional in `fld_data%import_fields`, so not an issue.

src/global_interpolation/global_interpolation.f90
```diff
@@ -169,44 +184,98 @@ module global_interpolation
...
+    call this%init_dof(dof, comm = comm, tol = tol, pad = pad, mask = mask)
+
+  end subroutine global_interpolation_init_json_dof
+
   !> Initialize the global interpolation object on a dofmap.
   !! @param dof Dofmap on which the interpolation is to be carried out.
   !! @param tol Tolerance for Newton iterations.
   !! @param pad Padding of the bounding boxes.
   !! @mask  Mask that indicates which portions of the domain to include
+  !! initialization instead of tol and pad.
```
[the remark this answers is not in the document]
vbaconnet: Working on it.

src/global_interpolation/global_interpolation.f90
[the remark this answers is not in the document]
vbaconnet: Yes :) I will update the tests with the new variable name if that is okay.

src/global_interpolation/global_interpolation.f90
vbaconnet: oh sorry i see now!

src/simulation_components/probes.F90
```diff
@@ -180,44 +182,82 @@ subroutine probes_init_from_json(this, json, case)
...
+    character(len=:), allocatable, intent(inout) :: output_file
+    character(len=*), intent(in) :: name
+    real(kind=dp), intent(in), optional :: tolerance, padding
+
+    character(len=1024) :: header_line
+    real(kind=rp), allocatable :: global_output_coords(:,:)
+    integer :: i, ierr
+    type(matrix_t) :: mat_coords
+
+    call this%global_interp%init(dof, tol = tolerance, pad = padding)
```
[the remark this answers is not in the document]
vbaconnet: Wrong, tolerance and padding are also optional in all the init subroutines, so not an issue.

src/io/fld_file_data.f90
```diff
@@ -515,6 +535,18 @@ subroutine fld_file_data_generate_interpolator(this, global_interp, to_dof, &
...
     allocate(z_coords(to_Xh%lx, to_Xh%ly, to_Xh%lz, to_msh%nelv))
 
+    ! JSON subdict takes precedence over dummy argument, but if neither is
+    ! provided throw an error
+    if (present(global_interp_subdict)) then
+       call json_get_or_default(global_interp_subdict, "tolerance", tol_, &
+            tol_)
+    else if (present(tolerance)) then
+       tol_ = tolerance
+    else
```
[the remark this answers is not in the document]
vbaconnet: Working on it.

src/io/fld_file_data.f90
```diff
@@ -186,8 +196,11 @@ subroutine fld_file_data_import_fields(this, u, v, w, p, t, &
        end if
 
        ! Generates an interpolator object and performs the point search
+       ! NOTE: optional arguments are managed in the subroutine, but if both
+       ! are provided the JSON subdict takes precedence
        call this%generate_interpolator(global_interp, dof, msh, &
-            tolerance = tolerance)
+            tolerance = tolerance, padding = padding, &
+            global_interp_subdict = global_interp_subdict)
```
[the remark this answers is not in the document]
vbaconnet: Wrong, tolerance, padding and globla_interp_subdict are also optional in `this%generate_interpolator`, so not an issue.

src/io/fld_file_data.f90
```diff
@@ -542,9 +574,17 @@ subroutine fld_file_data_generate_interpolator(this, global_interp, to_dof, &
...
-         this%nelv, fld_Xh, tol = tolerance)
+    ! fld data.
+    if (present(global_interp_subdict)) then
+       call global_interp%init(this%x%x, this%y%x, this%z%x, &
+            this%gdim, this%nelv, fld_Xh, &
+            global_interp_subdict)
+    else
+       call global_interp%init(this%x%x, this%y%x, this%z%x, &
+            this%gdim, this%nelv, fld_Xh, &
+            tol = tolerance, pad = padding)
```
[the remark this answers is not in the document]
vbaconnet: Wrong, tolerance and padding are also optional in all the init subroutines, so not an issue.

src/global_interpolation/global_interpolation.f90
```diff
@@ -169,44 +184,98 @@ module global_interpolation
...
+
+    real(kind=dp) :: tol, pad
+
+    call json_get_or_lookup_or_default(params_subdict, 'tolerance', &
+         tol, GLOBAL_INTERP_TOL)
+    call json_get_or_lookup_or_default(params_subdict, 'padding', &
+         pad, GLOBAL_INTERP_PAD)
+
+    call this%init_xyz(x, y, z, gdim, nelv, Xh, comm = comm, tol = tol, &
+         pad = pad)
```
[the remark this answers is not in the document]
vbaconnet: comm is also optional in init_xyz, so not an issue.

src/global_interpolation/global_interpolation.f90
```diff
@@ -169,44 +184,98 @@ module global_interpolation
...
+    type(mask_t), intent(in), optional :: mask
+
+    real(kind=dp) :: tol, pad
+
+    call json_get_or_lookup_or_default(params_subdict, 'tolerance', &
+         tol, GLOBAL_INTERP_TOL)
+    call json_get_or_lookup_or_default(params_subdict, 'padding', &
+         pad, GLOBAL_INTERP_PAD)
+
+    call this%init_dof(dof, comm = comm, tol = tol, pad = pad, mask = mask)
```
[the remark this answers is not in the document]
vbaconnet: comm is also optional in init_dof, so not an issue.

src/global_interpolation/global_interpolation.f90
```diff
@@ -215,19 +284,22 @@ subroutine global_interpolation_init_dof(this, dof, comm, tol, pad, mask)
     ! to get the right dimension (see global_interpolation_init_xyz).
     if (.not. present(mask)) then
        call this%init_xyz(dof%x(:,1,1,1), dof%y(:,1,1,1), dof%z(:,1,1,1), &
-            dof%msh%gdim, dof%msh%nelv, dof%Xh, comm,tol = tolerance, pad=padding)
+            dof%msh%gdim, dof%msh%nelv, dof%Xh, comm = comm, &
+            tol = tol, pad = pad)
```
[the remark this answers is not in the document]
vbaconnet: tol and pad are also optional in init_xyz, so not an issue.

src/global_interpolation/global_interpolation.f90
```diff
@@ -215,19 +284,22 @@ subroutine global_interpolation_init_dof(this, dof, comm, tol, pad, mask)
...
+          call neko_error("Mask size must be a multiple of the number of" // &
+               " elements in the mesh.")
        end if
        ! Initialize with the masked coordinates
-       call this%init_xyz(dof%x(mask%get(),1,1,1), dof%y(mask%get(),1,1,1), dof%z(mask%get(),1,1,1), &
-            dof%msh%gdim, temp_nelv, dof%Xh, comm,tol = tolerance, pad=padding)
+       call this%init_xyz(dof%x(mask%get(),1,1,1), dof%y(mask%get(),1,1,1), &
+            dof%z(mask%get(),1,1,1), dof%msh%gdim, temp_nelv, dof%Xh, &
+            comm = comm, tol = tol, pad = pad)
     end if
```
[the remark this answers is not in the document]
vbaconnet: tol and pad are also optional in init_xyz. Not an issue.

vbaconnet:
> The copilot comments warrant a look. Particularly the stuff with optional forwarding, we gotta really be careful with that: that was the cause of the recent regression with projections code. 

The issue with the projections code was that it was forwarding optional arguments to procedures where those arguments were non-optional.

What I am doing here is forwarding optional arguments into procedures where those arguments are _also_ optional. This delegates the detection of the presence of such argument to the downstream procedure. This is allowed under the [Fortran 2008 standard](https://j3-fortran.org/doc/year/10/10-007r1.pdf), section 12.5.2.12:

> 4. Except as noted in the list above, [an optional dummy argument that is *not* present] may be supplied as an actual argument corresponding to an optional dummy argument, which is then also considered not to be present.

(I opened that document for the first time this morning, please someone correct me if this is wrong.)

However I agree that passing everything without distinction of presence can be a bit harder to read. I think it depends on the context of the subroutine.

src/simulation_components/probes.F90
timofeymukha: @vbaconnet I will answer here so we can have the discussion threaded. Thanks for the detailed reply. If you view global interpolation as a user-facing type, having a JSON-based constructor is defensible. In fact I think I conflated two things in my comment there. The part that i didn't like is signatures like

fld_file_data_import_fields(s_target_list, s_index_list, interpolate, tolerance, padding, &
       global_interp_subdict)

with the latter 3 arguments being optional. As per your second point, this number will just grow if more parameters for interpolation are needed. My suggestion is to have two routines, one for json and one for direct parameter input, with no optionals. The former calls the later one after parsing the JSON. Basically, just like you have for the constructors of the interpolation type. You can bind the two routines by a generic fld_file_data_import_fields, so one can call both via the same name. Same approach for any other routines that have the same optionals, if any. Does that work?

vbaconnet: I see, I get your point about having multiple optional arguments. I think what you suggest is quite reasonable!

If we do it that way then there wouldn't be a need to pass a subdict to global interpolation, right? Since you'd end up calling the "direct parameter input" one anyways? 

Actually since all simcomps except `probes` use `import_fields` we can just modify `import_fields` as you suggest. Then you'd only need to modify that high-level subroutine but not touch the other ones in `fld_file_data`.

adperezm: I would still have the json options in global interpolation. While it might not be a user-facing object now, we might end up having it at some point. 

At least for me, most of the post-processing in my cases at some point end up calling an interpolation step. I would appreciate being able to pass all the interpolation settings in a packed way. If it is already here, I would not remove it.

edit: I agree that the initialization should be init(xyz, tol, padding, etc) or init(xyz, json_packed). But it should not be init(xyz, tol, padding, json_packed) with all being optional.

timofeymukha: Just having an extra constructor is totally fine I think. So, I leave it up to you :). I guess we all agree on killing the optionals now!

vbaconnet: Sorry I get why we'd want to separate into different constructors, but why killing the optionals? tol and pad could still be optional in the "individual parameters" constructors, no?

njansson (approved):

vbaconnet:
This is ready for merge :) If someone could activate auto-merge

PR #2413, opened 2026-03-24
Interface for HDF5 - HDF5 probes

adperezm:
## hdf5

As per discussed in a previous PR, this one simply exposes some wrappers that allow to write matrices, vectors, fields. It only allows to read matrices and vectors, since initializing fields requires dofs. It also allows to write and read integer and rp attributes.

An example on how things work is shown below. Since I limited it to those types, one needs to make some copies but it would be possible to simply write arrays, as done in VTKHDF. However I leave it like this for now.
```f90
! Set up the name
write(group_name, '(A,I0)') "probes/Step_", out_int
call this%fout%open("w")
call this%fout%set_active_group("probes")
! Write Nsteps in group
call this%fout%write_attribute("NSteps", out_int)
! Write out the data
call this%fout%set_active_group(trim(group_name))
do i = 1, this%n_fields
   call copy(this%vec_out%x, this%out_values(:,i), this%vec_out%size())
   this%vec_out%name = trim(this%which_fields(i))
   call this%fout%write_dataset(this%vec_out)
end do
! Write the time as an attribute
time_ = time%t
call this%fout%write_attribute("time", time_)
call this%fout%close()
```

## probes

On the probes side, I made it so we can write and read probes from hdf5, but I limited reading from rank0, since the probe object is very much designed like that. The readers do work in parallel. Generally the default behavior is that they distribute the data in the typical linear-load-balanced way.

Here, I gave 2 options for outputs. I like to have each "sample" in a different group, so the outputs might look like this:

```bash
/                        Group
/probes                  Group
/probes/Step_1           Group
/probes/Step_1/p         Dataset {200/Inf}
/probes/Step_1/s         Dataset {200/Inf}
/probes/Step_1/u         Dataset {200/Inf}
/probes/Step_1/v         Dataset {200/Inf}
/probes/Step_1/w         Dataset {200/Inf}
/probes/Step_2           Group
/probes/Step_2/p         Dataset {200/Inf}
/probes/Step_2/s         Dataset {200/Inf}
/probes/Step_2/u         Dataset {200/Inf}
/probes/Step_2/v         Dataset {200/Inf}
/probes/Step_2/w         Dataset {200/Inf}
```

In this case, each step has the time attribute as well.

However, I grew worried about performance in this way, so I did as @timfelle suggested and instead appended data:
```bash
/                        Group
/probes                  Group
/probes/coordinates      Dataset {200/Inf, 3}
/probes/p                Dataset {1800/Inf}
/probes/s                Dataset {1800/Inf}
/probes/time             Dataset {9/Inf}
/probes/u                Dataset {1800/Inf}
/probes/v                Dataset {1800/Inf}
/probes/w                Dataset {1800/Inf}
```

There are some docs on how to choose which to use. I do not use probes that much, so I prefer to leave the options, since I am a bit out of touch with the best way to use it.

I think that all the work that @timfelle did on the hdf5 side for vtkhdf should eventually be made public to use by the hdf5 file, as I said before, but I understand that might be too much. We can see how we feel later on.

I have tested but if possible, I would like someone else to also test it before we merge this. Then again, I do not think this would break any existing functionality.


timofeymukha:
I guess Tim and Niclas are in a way better position to review the actual code, I'll just say that
- I really like the added functionality.
- We gotta think about the file_t situation. Like, we cannot collect all the routines for all possible file formats in one type :))). At some point we just have to say: if you want to do this specialized stuff, create a variable with the needed concrete type, or take care of the `select type` stuff yourself.

adperezm:
> I guess Tim and Niclas are in a way better position to review the actual code, I'll just say that
> 
> * I really like the added functionality.
> * We gotta think about the file_t situation. Like, we cannot collect all the routines for all possible file formats in one type :))). At some point we just have to say: if you want to do this specialized stuff, create a variable with the needed concrete type, or take care of the `select type` stuff yourself.

Yes, I agree. The thing is that file_t, as it is, is very well posed for very structured outputs as we currently have. But if we want hdf5 because it gives flexibility, then I would say that constraining through the `write` and `read` methods only is a bit of a counter intuitive thing to do.

If I wanted to keep the same pipeline, then I would have needed to create a 'probe_data' type, keep all the hdf5 stuff private and let file%write decide what to do based on the probe_data type. But then if someone else needed almost the same output but with a different attribute, then they have to create a new data type and a new writer handler.

then again, of course performance is still a concern. So these methods I add are still probably suboptimal, since there might be operations that are unnecessarily repeated, etc. One would need to weight these things.

src/io/hdf5_file.F90
```diff
@@ -587,15 +663,966 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+    ! ===================
+    call h5lexists_f(this%active_group_id, trim(data_name), dset_exists, ierr)
+    if (dset_exists) then
+       ! Openr the data set
+       call h5dopen_f(this%active_group_id, trim(data_name), dset_id, ierr)
+       ! Get the current rank of the of the dataset
+       call h5dget_space_f(dset_id, filespace, ierr)
+       call h5sget_simple_extent_ndims_f(filespace, temprank, ierr)
+       if (temprank .ne. 2) then
+          call neko_error("Dataset " // trim(data_name) // " is not a rank 2 matrix in file " // trim(this%get_fname()))
```
timfelle:
```suggestion
          call neko_error("Dataset " // trim(data_name) // &
               " is not a rank 2 matrix in file " // trim(this%get_fname()))
```

src/io/hdf5_file.F90
```diff
@@ -587,15 +663,966 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+       end if
+       ! Get the current shape and close the filespace
+       call h5sget_simple_extent_dims_f(filespace, tempddims, tempmaxddims, ierr)
+       call h5sclose_f(filespace, ierr)
+    else
+       call neko_error("Dataset " // trim(data_name) // " does not exist in current group " // trim(this%get_fname()))
```
timfelle:
```suggestion
       call neko_error("Dataset " // trim(data_name) &
            // " does not exist in current group " // trim(this%get_fname()))
```

src/io/hdf5_file.F90
```diff
@@ -587,15 +663,966 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+       if (pe_rank .eq. 0) then
+          counts = total_count
+       else
+          counts = 0
+       end if
+       offset = 0
+    end if
+    call MPI_Scan(counts, offset, 1, MPI_INTEGER, &
+         MPI_SUM, NEKO_COMM, ierr)
+    offset = offset - counts ! Not using exclusive scan
```
timfelle, line 1476: Why not use `MPI_Exscan` instead?

src/io/hdf5_file.F90
```diff
@@ -587,15 +663,966 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+            filespace, dset_id, ierr, dcpl_id = dcpl_id)
+       call h5sclose_f(filespace, ierr)
+       call h5pclose_f(dcpl_id, ierr)
+    end if
+
+    ! ===========================
+    ! Set up writing the data set
+    ! ===========================
+    dcount = [int(stride_ax_1, hsize_t), int(stride_ax_2, hsize_t), int(stride_ax_3, hsize_t), int(counts, hsize_t)] ! local size of the tensor
+    doffset = [0_hsize_t, 0_hsize_t, 0_hsize_t, int(offset, hsize_t) + append_offset] ! offset for this rank in the global tensor
```
timfelle: Long lines

src/io/hdf5_file.F90
```diff
@@ -587,15 +663,966 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+    ddims = [int(strides, hsize_t), int(total_count, hsize_t)] ! global size of the matrix
+    chunkdims = [int(strides, hsize_t), max(int(max_count, hsize_t), 1_hsize_t)]
+    ddims_max = [int(strides, hsize_t), H5S_UNLIMITED_F]
+    call h5lexists_f(this%active_group_id, trim(mat%name), dset_exists, ierr)
+    if (dset_exists) then
+       if (this%overwrite) then
+          ! retrieve the dset id for the existing data set
+          if (pe_rank .eq. 0) then
+             write(*,*) "Dataset ", trim(mat%name), " already exists in file ", trim(this%get_fname()), " and will be overwritten."
+             write(*,*) "This only works if the global shape is the same"
```
timfelle: Message for the log maybe?

src/io/hdf5_file.F90
```diff
@@ -587,15 +663,966 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+    else
+       call h5fcreate_f(fname, H5F_ACC_TRUNC_F, &
+            this%file_id, ierr, access_prp = this%plist_id)
+    end if
+
+    ! Set the active group to the root of the file
+    call this%set_active_group()
+
+    if (pe_rank .eq.0) then
+       write(*,*) "Opened HDF5 file: ", trim(fname), " with counter: ", counter
```
timfelle: Debug message maybe?

src/io/hdf5_file.F90
```diff
@@ -587,15 +663,966 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+    end if
+
+  end subroutine hdf5_file_open
+
+  !> Close the file
+  subroutine hdf5_file_close(this)
+    class(hdf5_file_t), intent(inout) :: this
+    integer :: ierr
+
+    if (this%active_group_id .ne. -1_hid_t .and. this%active_group_id .ne. this%file_id) then
```
timfelle: Long line

src/io/hdf5_file.F90
```diff
@@ -587,15 +663,966 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+    integer(hid_t) :: filespace, attr_id
+    integer(hsize_t), dimension(1) :: dcount
+
+    ! Get the precision
+    precision_hdf = h5kind_to_type(rp, H5_REAL_KIND)
+
+    ! ====================
+    ! Create the attribute
+    ! ====================
+    dcount = [int(1, hsize_t)]
```
timfelle, line 1645:
```suggestion
    dcount = 1_hsize_t
```

src/io/hdf5_file.F90
```diff
@@ -53,16 +59,86 @@ module hdf5_file
...
+    character(len=1024) :: base_fname
+    character(len=1024) :: fname
+    character(len=1024) :: path, name, suffix
+
+    fname = trim(this%get_base_fname())
+    call filename_split(fname, path, name, suffix)
+
+    ! Append a counter
+    !write(base_fname, '(A,A,"_",I0,A)') &
+    !     trim(path), trim(name), this%get_start_counter(), trim(suffix)
```
timfelle, line 120: Careful, this might break HDF5 based checkpoints.

src/simulation_components/probes.F90
```diff
@@ -536,8 +544,41 @@ subroutine probes_init_from_components(this, dof, output_file, name)
...
+          call this%fout%set_overwrite(.false.)
+       else
+          call this%fout%set_overwrite(.true.)
+       end if
+       call mat_coords%init(3, this%n_local_probes, "coordinates")
+       call copy(mat_coords%x, this%xyz, 3*this%n_local_probes)
+
+       !> Set up output
+       call this%fout%open("w")
+       call this%fout%set_active_group([character(len=1000) :: "probes"])
```
timfelle: Should you not just do "probes", why does it have to be an array cast to length
1000 characters?

src/simulation_components/probes.F90
```diff
@@ -731,6 +841,20 @@ subroutine read_probe_locations(this, xyz, n_local_probes, n_global_probes, &
     type is (csv_file_t)
        call read_xyz_from_csv(xyz, n_local_probes, n_global_probes, ft)
        this%seq_io = .true.
+    type is (hdf5_file_t)
+       call file_in%open("r")
+       call file_in%set_active_group() ! Empty sets it to the root group "/"
```
timfelle: Why do you need `set_active_group` here. Does the open operation not open into root?

src/io/hdf5_file.F90
```diff
@@ -587,15 +663,966 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+    call h5sselect_hyperslab_f(filespace, H5S_SELECT_SET_F, doffset, dcount, ierr)
+    ! Create the corresponding memory space (buffer) for my local data
+    call h5screate_simple_f(dset_rank, dcount, memspace, ierr)
+
+    ! =======================
+    ! Cast and write the data
+    ! =======================
+    if (this%precision == sp) then
+       allocate(write_buffer_sp(mat%get_nrows(), mat%get_ncols()))
+       if (mat%size() > 0) write_buffer_sp = real(mat%x, kind=sp)
```
timfelle, line 1122: This check should be at the beginning of the write subroutine. At this point you
have already created the dataset and all that.

src/simulation_components/probes.F90
```diff
@@ -536,8 +544,41 @@ subroutine probes_init_from_components(this, dof, output_file, name)
           !! Write the data to the file
           call this%fout%write(mat_coords)
        end if
+    class is (hdf5_file_t)
+
+       !> This is always on cpus.
+       if (this%append_out) then
+          call this%fout%set_overwrite(.false.)
```
timfelle: The `select type` statement give you direct access to the specific filetype
stored so you should access that pointer instead of the original object.
```suggestion
          call ft%set_overwrite(.false.)
```

timfelle: This will also eliminate the requirement for changing `file_t`

timfelle (commented):
I like the fact that you are starting this work. I am not exactly sure how much sense I find in extending the file_t. I think anyone intending to write something with as specialized a notation as you do here would already declare their type to be `hdf5_file_t` directly rather than the generic one. My vote would be to not implement things as specific as these to `file_t`

src/io/hdf5_file.F90
adperezm: Oh shit, you are right. I will check that ASAP

adperezm: I will not override the original get_name. That is how things were working before. Now I will use this one internally to be used only when the more new route of opening an HDF5 is used.

src/io/hdf5_file.F90
adperezm: I will comment it out for now.

src/io/hdf5_file.F90
adperezm: I am commenting all these messages for now.

src/io/hdf5_file.F90
adperezm: Yeah but the data set was created with the correct precision. All I do now is cast.

Data set is created even if mat has size zero since some ranks might be empty. But the data set needs to be created by all.

src/io/hdf5_file.F90
adperezm: I want to keep it as vector to keep consistency with the other rank writes. In case later someone else want to modify, I believe this will be a bit of a better template.

src/io/hdf5_file.F90
adperezm: I am most used to using Scan from the FLD files, so I just kept it.

src/simulation_components/probes.F90
adperezm: Yes, this is a good point. I will change this and update.

src/simulation_components/probes.F90
adperezm: When doing it this way, fortran was truncating my strings in weird places. It seems that explicitly setting the lenght helped not having the names in the file be truncated.

src/simulation_components/probes.F90
adperezm: It does open into the root. I do some "redundant" operations here to allow someone else modify it without checking the other part of the code. I hoped some probe user would check modify what they did not like here.

I do not want to remove it yet.

src/io/hdf5_file.F90
```diff
@@ -587,15 +662,977 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+    call h5open_f(ierr)
+    call h5pcreate_f(H5P_FILE_ACCESS_F, this%plist_id, ierr)
+    call h5pset_fapl_mpio_f(this%plist_id, mpi_comm, mpi_info, ierr)
+
+    ! Open the file
+    inquire(file = fname, exist = file_exists)
+    if (file_exists) then
+       call h5fopen_f(fname, H5F_ACC_RDWR_F, this%file_id, ierr, &
+            access_prp = this%plist_id)
+    else
```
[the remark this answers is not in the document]
adperezm: This is fine. The mode flag will take care of things here.

adperezm:
Updated now

src/io/file.f90
timfelle, line 1: Should probably be reverted since the only change is removing an empty line

timfelle (commented):
Generally gret work. Your latest changes really cleaned things up. One thing to make sure works is the hdf5 checkpoints. I can try to remember lookimg at it tomorrow.

src/io/hdf5_file.F90
timfelle: No I was talking about the if size > 0. You have now created a dataset which
might have size 0. You should probably guard against a 0 size matrix in the
beginning of the subroutine and either throw an error or just return.

src/io/hdf5_file.F90
timfelle: The Exscan is not really different. All you do is remove the second line.
```fortran
    call MPI_Exscan(counts, offset, 1, MPI_INTEGER, &
         MPI_SUM, NEKO_COMM, ierr)
```

src/io/hdf5_file.F90
```diff
@@ -587,15 +660,977 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+       if (pe_rank .eq. 0) then
+          counts = total_count
+       else
+          counts = 0
+       end if
+       offset = 0
+    end if
+    call MPI_Scan(counts, offset, 1, MPI_INTEGER, &
+         MPI_SUM, NEKO_COMM, ierr)
+    offset = offset - counts ! Not using exclusive scan
```
timfelle:
```suggestion
    call MPI_Exscan(counts, offset, 1, MPI_INTEGER, &
         MPI_SUM, NEKO_COMM, ierr)
```

src/io/file.f90
```diff
@@ -338,5 +338,4 @@ subroutine file_set_subdivide(this, subdivide)
        end if
     end select
   end subroutine file_set_subdivide
-
 end module file
```
timfelle, line 342:
```suggestion
  end subroutine file_set_subdivide

end module file
```

src/simulation_components/probes.F90
```diff
@@ -35,11 +35,12 @@
...
-  use logger, only : neko_log, LOG_SIZE, NEKO_LOG_DEBUG
-  use utils, only : neko_error, nonlinear_index
-  use field_list, only : field_list_t
+  use num_types, only: rp, dp
+  use matrix, only: matrix_t
+  use vector, only: vector_t
+  use logger, only: neko_log, LOG_SIZE, NEKO_LOG_DEBUG
+  use utils, only: neko_error, nonlinear_index
+  use field_list, only: field_list_t
   use time_state, only : time_state_t
```
timfelle: It was decided at a dev meeting that we should have the space here...
```suggestion
  use num_types, only : rp, dp
  use matrix, only : matrix_t
  use vector, only : vector_t
  use logger, only : neko_log, LOG_SIZE, NEKO_LOG_DEBUG
  use utils, only : neko_error, nonlinear_index
  use field_list, only : field_list_t
  use time_state, only : time_state_t
```

timfelle (approved):

src/simulation_components/probes.F90
timfelle: Right, since it is an array you run into issues with determining what the sizes are. Why not just use the path strings that HDF5 is built on anyway. 
`/probes/coordinates`: direct from root to coordinates.
`probes`: look for a local subgroup called `probes`.

src/simulation_components/probes.F90
timfelle: Right, i guess it should be updated now, since your layout is `/probes/coordinates`? to keep it consistent with the output.

adperezm: Yes, it changed. But I did use the empty `set_active_group()` in the part where I read the probes. I will actually remove it now :), since I think it is unlikely that it will be changed by anyone else at this stage.

adperezm:
> Generally gret work. Your latest changes really cleaned things up. One thing to make sure works is the hdf5 checkpoints. I can try to remember lookimg at it tomorrow.

Thank you!

I also updated now to have more of your suggestions in, particularly the paths "path/to/group"

I left alone the previous hdf5 code now. But I can also check that I do not break anything.

timfelle (commented):
> > Generally gret work. Your latest changes really cleaned things up. One thing to make sure works is the hdf5 checkpoints. I can try to remember lookimg at it tomorrow.
> 
> Thank you!
> 
> I also updated now to have more of your suggestions in, particularly the paths "path/to/group"
> 
> I left alone the previous hdf5 code now. But I can also check that I do not break anything. 

From my tests, you should be all good regarding the checkpoints. They are as broken as they have always been :sweat_smile:

timfelle:
I think we are at a good point. Could you update the PR description so that it matches the latest changes, and then we can request Niclas and the others to take a look at it as well so we can get it rolling.

doc/pages/user-guide/simcomps.md
```diff
@@ -321,6 +325,10 @@ time_1, p_1_field_0, p_1_field_1, ..., p_1_field_N_f-1
 time_N_p, p_N_p_field_0, p_N_p_field_1, ..., p_N_p_field_N_f-1
 ~~~~~~~~~~~~~~~
 
+The `append_output` keyword only works for `hdf5` files. 
+It sets the behaviour of the written probes. 
+If `true` they are written in one group and each sample appends its data.
+
```
timfelle, line 332: If we are happy with the structure, I think it would be good to provide an
example HDF5 header. That would make it a lot simpler for people to understand
the HDF5 structure.

adperezm: Yes! I added it now. Let me know if you would like to change anything there.

src/io/hdf5_file.F90
adperezm: Okay! It should be added now

adperezm:
@timofeymukha , @njansson , @vbaconnet , It would be nice if we merge this. It does not seem to affect any functionality at the moment. So I hope you can take a quick look.

doc/pages/user-guide/simcomps.md
```diff
@@ -321,6 +325,43 @@ time_1, p_1_field_0, p_1_field_1, ..., p_1_field_N_f-1
...
+As an example, the file structure for a simulation where we sample "u", "v",
+and "w" is the following when `append_output=false`:
+~~~~~~~~~~~~~~~{.bash}
+/                        Group
+/probes                  Group
+/probes/Step_1           Group
+/probes/Step_1/u         Dataset {200/Inf}
```
vbaconnet, line 339: What does `/Inf` mean`?

doc/pages/user-guide/simcomps.md
```diff
@@ -321,6 +325,43 @@ time_1, p_1_field_0, p_1_field_1, ..., p_1_field_N_f-1
...
+/probes/Step_2/v         Dataset {200/Inf}
+/probes/Step_2/w         Dataset {200/Inf}
+/probes/coordinates      Dataset {200/Inf, 3}
+~~~~~~~~~~~~~~~
+
+Note that every sampled field is saved under the `probes/Step_i` group. Where
+`i` is the sample number. In this case, the sampled time is saved as an attribute
+of the `Step_i` group and accessed by the `time` keyword.
+
+The default behaviour is `append_output=true`. This mode behaves as the `.csv`
```
vbaconnet, line 353: Just curious if the default=.true. was just to keep the same behavior as csv for backwards compatibility or if there was another reason?

src/io/hdf5_file.F90
```diff
@@ -587,15 +660,1001 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+       call h5fcreate_f(fname, H5F_ACC_TRUNC_F, &
+            this%file_id, ierr, access_prp = this%plist_id)
+    end if
+
+    ! Set the active group to the root of the file
+    call this%set_active_group()
+
+    !if (pe_rank .eq.0) then
+    !   write(*,*) "Opened HDF5 file: ", trim(fname), " with counter: ", counter
+    !end if
```
vbaconnet:
```suggestion
```
unless you want to add this at the NEKO_DEBUG_LVL?

src/io/hdf5_file.F90
```diff
@@ -587,15 +660,1001 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+    call h5pclose_f(this%plist_id, ierr)
+    this%plist_id = -1_hid_t
+    call h5fclose_f(this%file_id, ierr)
+    this%file_id = -1_hid_t
+    call h5close_f(ierr)
+
+    !if (pe_rank .eq.0) then
+    !   write(*,*) "Closed HDF5 file: ", trim(this%get_fname())
+    !end if
+
```
vbaconnet:
```suggestion
```

src/io/hdf5_file.F90
```diff
@@ -587,15 +660,1001 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+  end subroutine hdf5_file_close
+
+
+  !> Set the active group for HDF5 files from an input string
+  !! @param this The HDF5 file object
+  !! @param An string indicating the path to the group to create or open.
```
vbaconnet:
```suggestion
  !! @param A string indicating the path to the group to create or open.
```

src/io/hdf5_file.F90
```diff
@@ -587,15 +660,1001 @@ subroutine hdf5_file_determine_real(H5T_NEKO_REAL)
...
+          call h5dset_extent_f(dset_id, ddims, ierr)
+       end if
+    else
+       ! create file space of this shape
+       call h5screate_simple_f(dset_rank, ddims, filespace, ierr, ddims_max)
+       ! Create chunk property list (needed to be able to append)
+       call h5pcreate_f(H5P_DATASET_CREATE_F, dcpl_id, ierr)
+       call h5pset_chunk_f(dcpl_id, dset_rank, chunkdims, ierr)
+       ! create the data set with the given shape
+       call h5dcreate_f(this%active_group_id, trim(vec%name), precision_hdf, &
```
vbaconnet, line 968: Sorry maybe I missed this, what happens if `vec%name` is empty again?

src/simulation_components/probes.F90
```diff
@@ -94,6 +98,8 @@ module probes
      !> Output variables
      type(file_t) :: fout
      type(matrix_t) :: mat_out
+     type(vector_t) :: vec_out
+     logical :: append_out = .false.
```
vbaconnet:
```suggestion
     logical :: append_out = .true.
```

vbaconnet (approved):
Really cool! Excited to try it out.

I guess now with hdf5 at some point we will be able to move away from the "sequential" probes that do everything on rank 0?

doc/pages/user-guide/simcomps.md
timfelle: Means it is "infinite", ie. have no limit on potential size. This allows us to just append new data everytime we do a new iteration. Int the particular case you point out it could actually have been omitted, since there each step is it's own group.

doc/pages/user-guide/simcomps.md
timfelle: It limits the number of groups in the HDF5 file, which probably makes it simpler to work with when you want to sample a few thousands of steps.

src/io/hdf5_file.F90
timfelle: This should be debug information yes.

src/io/hdf5_file.F90
timfelle: Same here.

timofeymukha:
As a note (perhaps to self), we need to expand our programming examples with the vtkhdf5 output and also using this new interface.

njansson (approved):

src/io/hdf5_file.F90
vbaconnet: Resolving this for the sake of merging the PR!

vbaconnet:
@timofeymukha I removed you as reviewer so this can merge once the checks pass! I see that copilot already did a sweep here as well but if you'd like to have a look of course you're more than welcome :)

timofeymukha:
Please proceed with the merge!

PR #2422, opened 2026-03-26
Boundary operation simcomp

timofeymukha:
This PR adds a new simulation component, boundary_operation, for computing boundary-restricted reductions over labeled mesh zones, and adds a supporting face_mask_gather_copy_0 routine to math on all backends.

The new simcomp reads a set of zone indices and a list of requested operations from the case file, evaluates those operations on the selected boundary region, stores the results on the simcomp, and can optionally log and/or write CSV output.

Supported operations:
- integral
- average
- min
- max

Example step log with two simcomps configured, with a different set of operations and `name`s zone_1_v and zone_2_v.

```
------------------------------------------------------------------------------
 Step =           1 t =   1.000000000000000E-03                     [ 100.00% ]
 ------------------------------------------------------------------------------

   --------Preprocessing---------
   CFL:  0.1268  dt: 1.0000000E-03

   ------------Fluid-------------
   KSP solver   | Field              Iters:   Start residual:   Final residual:
   1            | Pressure               41   0.779926897E+00   0.819055283E-04
   1            | X-Velocity              3   0.130136177E+00   0.410566471E-07
   1            | Y-Velocity              3   0.871130145E-01   0.353632576E-07
   1            | Z-Velocity              3   0.791857770E-01   0.162699233E-07
   Fluid step time (s):     0.1663182E+01

   --------Postprocessing--------

    -----------zone_1_v-----------
                 |          Average:              Min:              Max:
    1            |   5.711424416E-10  -4.549512600E-04   4.549691770E-04

    -----------zone_2_v-----------
                 |         Integral:          Average:
    1            |   0.000000000E+00   0.000000000E+00
```

Currently only tested by my on the CPU, so the new device code for coef is untested! However, I think the PR is ready for review.


adperezm:
I have not read it, but is it not possible to use the mask_t type for this thing? 

Specially if there are also new kernels. I understand that we do not want to touch old stuff, but I would really encourage to use it as it might prevent errors, in my opinion.

src/simulation_components/boundary_operation.f90
```diff
@@ -0,0 +1,504 @@
...
+    this%bc%zone_indices = this%zone_indices
+    do i = 1, size(this%zone_indices)
+       call this%bc%mark_zone(this%bc%msh%labeled_zones(this%zone_indices(i)))
+    end do
+    call this%bc%finalize()
+
+    n_pts = this%bc%msk(0)
+    if (n_pts .gt. 0) then
+       call this%field_msk%init(n_pts)
+       call this%coef%get_areas_by_mask(this%area_msk, this%bc%msk, this%bc%facet)
```
adperezm: I believe that for this application, there should be a mask that is consistent with the area array. And then you retrieve those with a typical masked gather operation, instead of creating new functions and kernels. Or did I misunderstand what happens here?

timofeymukha:
> I have not read it, but is it not possible to use the mask_t type for this thing?
> 
> Specially if there are also new kernels. I understand that we do not want to touch old stuff, but I would really encourage to use it as it might prevent errors, in my opinion.

It uses a bc_t underneath, so I think here it is better to match the old mask behaviour. But I am hoping that we will convert the bc code to use proper masks in the coming months. Note that area_msk and field_msk are not masks, they are values at masked nodes... I think I should rename that.

src/simulation_components/boundary_operation.f90
adperezm: Although I do see that the area array is a bit weird.

vbaconnet:
I think it would be really cool if we could have this to compute on either a full field, or a "masked" field as Adalberto is saying. That way you could extend that to point zones as well.

I haven't read everything either so I'll take a look eventually but just some initial thoughts since I was thinking of implementing a user-defined simcomp to compute this stuff (for example, total kinetic energy, or whatever) :)

timofeymukha:
> I think it would be really cool if we could have this to compute on either a full field, or a "masked" field as Adalberto is saying. That way you could extend that to point zones as well.
> 
> I haven't read everything either so I'll take a look eventually but just some initial thoughts since I was thinking of implementing a user-defined simcomp to compute this stuff (for example, total kinetic energy, or whatever) :)

My idea was to keep that specific to boundaries, and then have a volume_operation simcomp to work on the entire volume / point_zone. My feeling is that boundaries are an important enough special case.

src/simulation_components/boundary_operation.f90
timofeymukha: Yeah, the issue is that this%area(this%Xh%lx, this%Xh%ly, 6, this%msh%nelv) has this specific form, so a purely nodal mask will not work, you also need the face index for each node in the mask. So I think we cannot use the existing masked_gather routines. Or am I wrong?

timofeymukha: We could add a generic routine like this to math

```fortran
  !> Gather values from a face-local SEM field to a reduced contiguous vector.
  !! @param a Destination array of size `n_mask`.
  !! @param b Source face-local array indexed as `b(:, :, facet, element)`.
  !! @param mask Mask array of length `n_mask + 1`, where `mask(0) = n_mask`.
  !! @param facet Facet ids associated with the masked points.
  !! @param lx Number of points in the first SEM direction.
  !! @param ly Number of points in the second SEM direction.
  !! @param lz Number of points in the third SEM direction.
  !! @param n_mask Size of the mask and destination arrays.
  subroutine face_masked_gather_copy_0(a, b, mask, facet, lx, ly, lz, n_mask)
    integer, intent(in) :: lx, ly, lz, n_mask
    real(kind=rp), dimension(n_mask), intent(inout) :: a
    real(kind=rp), dimension(:, :, :, :), intent(in) :: b
    integer, dimension(0:n_mask), intent(in) :: mask
    integer, dimension(0:n_mask), intent(in) :: facet
    integer :: l
    integer :: idx(4)

    do l = 1, n_mask
       idx = nonlinear_index(mask(l), lx, ly, lz)

       select case (facet(l))
       case (1, 2)
          a(l) = b(idx(2), idx(3), facet(l), idx(4))
       case (3, 4)
          a(l) = b(idx(1), idx(3), facet(l), idx(4))
       case (5, 6)
          a(l) = b(idx(1), idx(2), facet(l), idx(4))
       end select
    end do

  end subroutine face_masked_gather_copy_0
```

I am a bit unsure if we have a lot of data living like b(:, :, facet, element) except for area, but maybe this is a better approach?

adperezm: Yeah, I like this too. What I meant was that one can keep the logic of this outside of the gpu by building a new area-based mask that can call the regular operations. Something like what I show below (in python)

If ,from a non-linear index and the face index, one can get the 1d location in the array. then the masked operations work. So what I would have done here is to create a mask with a logic like this in the cpu or wherever I want, and then simply use it with the operations that are needed.

Then again, I do not really know if it is slow to do this. You basically do things like this on the GPU, so maybe that makes it much faster. In the end, I also do not have such a strong opinion :P on this. I think it is a nice functionality to have, so please keep as you see fit.

```python

# Get the area of the lower boundary based on the area array
el = np.unique(np.where(msh.z == 0)[0])
print("The area is: ", np.sum(coef.area[el, 5, :, :]))

def facet_based_linear_index(i, j, k, l, face, lx, ly, lz):
    """
    """
    facets = 6

    # Facet 1 in neko
    if face == 0:
        facet = 0
        return j + ly * ((k - 0) + lz * ((facet - 0) + facets * ((l - 0))))
    # Facet 2 in neko
    elif face == 1:
        facet = 1
        return j + ly * ((k - 0) + lz * ((facet - 0) + facets * ((l - 0))))
    # Facet 3 in neko
    elif face == 2:
        facet = 2
        return i + lx * ((k - 0) + lz * ((facet - 0) + facets * ((l - 0))))
    # Facet 4 in neko
    elif face == 3:
        facet = 3
        return i + lx * ((k - 0) + lz * ((facet - 0) + facets * ((l - 0))))
    # Facet 5 in neko
    elif face == 4:
        facet = 4
        return i + lx * ((j - 0) + ly * ((facet - 0) + facets * ((l - 0))))
    # Facet 6 in neko
    elif face == 5:
        facet = 5
        return i + lx * ((j - 0) + ly * ((facet - 0) + facets * ((l - 0))))
    else:
        raise ValueError('Invalid facet index')

mask = []
for e in el:
    for k in range(msh.lz):
        for j in range(msh.ly):
            for i in range(msh.lx):
                if k == 0: # Corresponding to the face of my boundary, which I must know
                    mask.append(facet_based_linear_index(i, j, k, e, 5-1, msh.lx, msh.ly, msh.lz)) # substract 1 due to 0-base
mask = np.array(mask)

flat_area = coef.area.flatten()
print("the area is: ", flat_area[mask].sum()) # By this point, you can call a masked operation and get the right results
```

timofeymukha: OK, I understand. Indeed, the whole idea here was to avoid the loop over the boundary on the CPU. Otherwise, I could very easily just compute the areas on the CPU and then transfer over. Maybe I should have just done that :)). But I liked the idea of a generic math routine, so I am adding that!

adperezm: No but you still create the masks in the cpu when initializing the boundary condition no?

So what I am saying is that you create the area mask as soon as your dof mask is created. then you use it. 

Right now you are doing these indexing operations every time step that you call the operation itself, so I would disagree that this is the same of just calculating on the cpus every time.

Edit: I guess my last statement might be incorrect, since I suppose you only gather the area once.

src/simulation_components/boundary_operation.f90
```diff
@@ -0,0 +1,507 @@
...
+
+    this%integral = 0.0_rp
+    this%average = 0.0_rp
+    this%minimum = huge(0.0_rp)
+    this%maximum = -huge(0.0_rp)
+    area = 0.0_rp
+
+    if (n_pts .gt. 0) then
+       call vector_masked_gather_copy_0(this%surface_values, &
+            this%field%x(1,1,1,1), this%bc%msk, this%field%size(), n_pts)
```
vbaconnet: Out of curiosity, did it not work with just `%x`? 
```suggestion
            this%field%x, this%bc%msk, this%field%size(), n_pts)
```

adperezm: The gather copy expects a 1d array of dim n_pts (field is (:,:,:,:)). I think in fact the hack i've seen is (:,1,1,1). I did not know (1,1,1,1) worked. Maybe check in cray.

vbaconnet: Oh true...

timofeymukha: I actually think we can just pass %x, because `b` is a 4D array.

```fortran
subroutine vector_face_masked_gather_copy_0(a, b, mask, facet, lx, ly, lz, &
       n_mask)
    integer, intent(in) :: lx, ly, lz, n_mask
    type(vector_t), intent(inout) :: a
    real(kind=rp), dimension(:, :, :, :), intent(in) :: b
```

timofeymukha: No wait, I confused two things :)). Adalberto is right, the expected array is 1d.

```fortran
  subroutine vector_masked_gather_copy_0(a, b, mask, n, n_mask)
    integer, intent(in) :: n, n_mask
    type(vector_t), intent(inout) :: a
    real(kind=rp), dimension(n), intent(in) :: b
```
But I think we can still pass %x, like here for example

```fortran
  subroutine field_masked_gather_copy_0(a, b, mask, n, n_mask)
    integer, intent(in) :: n, n_mask
    real(kind=rp), dimension(n_mask), intent(inout) :: a
    type(field_t) :: b
    integer, dimension(0:n_mask) :: mask
    type(c_ptr) :: mask_d, a_d

       mask_d = device_get_ptr(mask)
       a_d = device_get_ptr(a)
       call device_masked_gather_copy_0(a_d, b%x_d, mask_d, n, n_mask)
    else
       call masked_gather_copy_0(a, b%x, mask, n, n_mask)
    end if

  end subroutine field_masked_gather_copy_0
```

timofeymukha:
I have 0 ideas what happens to pFUnit on the 2 older compilers...

adperezm:
> I have 0 ideas what happens to pFUnit on the 2 older compilers...

Could it just be that the line is too long? I have been getting that one at times from my laptop. It said xxx:132. And that 132 might be a magic number haha

src/simulation_components/boundary_operation.f90
```diff
@@ -0,0 +1,507 @@
...
+    this%maximum = -huge(0.0_rp)
+    area = 0.0_rp
+
+    if (n_pts .gt. 0) then
+       call vector_masked_gather_copy_0(this%surface_values, &
+            this%field%x, this%bc%msk, this%field%size(), n_pts)
+
+       if (this%compute_integral .or. this%compute_average) then
+          this%integral = vector_glsc2(this%surface_values, this%areas, n_pts)
+       end if
```
timofeymukha: Can't have MPI-calling routiens behind the n_pts check. Gotta fix.

src/simulation_components/boundary_operation.f90
timofeymukha: You are right, the init of the bc mask is basically a loop over the boundary anyway. So worrying about it was overkill, I should have just filled the areas on the CPU and copy over ^^.

But the good thing is that the face_masked_gather_copy is now a separate routine under math, and I guess we can keep it, as it may be useful elsewhere. And then I might as well use it in the simcomp.

timofeymukha:
@njansson if you have time, you could take a look at the kernels

njansson:
> @njansson if you have time, you could take a look at the kernels

They look fine. But haven't tested them.

src/math/bcknd/device/cuda/math_kernel.h
```diff
@@ -87,6 +87,67 @@ __global__ void masked_gather_copy_aligned_kernel(T * __restrict__ a,
...
+__device__ __forceinline__
+void face_gather_nonlinear_index(const int idx, const int lx, const int ly,
+                                 const int lz, int *index) {
+  const int idx2 = idx - 1;
+  index[3] = idx2 / (lx * ly * lz);
+  index[2] = (idx2 - (lx * ly * lz) * index[3]) / (lx * ly);
+  index[1] = (idx2 - (lx * ly * lz) * index[3] - (lx * ly) * index[2]) / lx;
+  index[0] = (idx2 - (lx * ly * lz) * index[3] - (lx * ly) * index[2]) -
+    lx * index[1];
+  index[0]++;
```
adperezm, line 99: So this -1 and +1 is due to the masking offset we have right?

src/math/bcknd/device/cuda/math_kernel.h
```diff
@@ -87,6 +87,67 @@ __global__ void masked_gather_copy_aligned_kernel(T * __restrict__ a,
   }
 }
 
+__device__ __forceinline__
+void face_gather_nonlinear_index(const int idx, const int lx, const int ly,
```
adperezm: I guess generally we keep the outputs as the first argument, no?

src/math/bcknd/device/hip/math_kernel.h
```diff
@@ -71,24 +71,85 @@ __global__ void masked_gather_copy_kernel(T * __restrict__ a,
...
-template< typename T >
-__global__ void masked_gather_copy_aligned_kernel(T * __restrict__ a,
-                                   T * __restrict__ b,
-                                   int * __restrict__ mask,
-                                   const int n,
+template< typename T >
+__global__ void masked_gather_copy_aligned_kernel(T * __restrict__ a,
+                                   T * __restrict__ b,
+                                   int * __restrict__ mask,
+                                   const int n,
```
adperezm, line 78: These things here are just github weird behavior right? These aligned kernels were not modified.

src/math/bcknd/device/hip/math_kernel.h
```diff
@@ -71,24 +71,85 @@ __global__ void masked_gather_copy_kernel(T * __restrict__ a,
...
+
+  for (int m = idx; m < n_mask; m += str) {
+    const int f = facet[m + 1];
+    face_gather_nonlinear_index(mask[m + 1], lx, ly, lz, index);
+
+    switch (f) {
+    case 1:
+    case 2:
+      a[m] = b[face_gather_idx(index[1], index[2], f, index[3], n1, n2, 6)];
+      break;
```
adperezm, line 137: Why the break?

src/sem/bcknd/device/cuda/coef_kernel.h
```diff
@@ -34,6 +34,20 @@
  POSSIBILITY OF SUCH DAMAGE.
 */
 
+__device__ inline void coef_nonlinear_index(const int idx, const int lx,
+                                            int *index) {
+  const int idx2 = idx - 1;
+  index[3] = idx2 / (lx * lx * lx) + 1;
+  index[2] = (idx2 - (lx * lx * lx) * (index[3] - 1)) / (lx * lx) + 1;
+  index[1] = (idx2 - (lx * lx * lx) * (index[3] - 1)
```
adperezm: Since you added a face masked gather, is this one still used?

adperezm (approved):
I left some comments but otherwise I am happy with this functionality. Nice!

src/sem/bcknd/device/cuda/coef_kernel.h
timofeymukha: Yes, good catch! the sem directory should be untouched

src/math/bcknd/device/hip/math_kernel.h
timofeymukha: Yeah, the diff disappears when I remove the whitespace in the settings here. Should be fine I think.

src/math/bcknd/device/hip/math_kernel.h
timofeymukha: It is C++ weirdness.

When the switch (f) runs:

- The program jumps directly to the matching case label.
- From that point onward, execution proceeds sequentially downward.
It continues until:
1. a break is encountered  (normal case), or
2. the switch block ends (fall-through all the way)

So it won't just stop on its own. We use that for case 1, 3, and 5, actually. For example, for case 1, it seemingly does nothing, but actually it will fall down to case 2 and then stop at the break.

src/math/bcknd/device/cuda/math_kernel.h
timofeymukha: No, this is fortran vs C indexing. idx is the Fortran index, idx2 is the C (0-based). Then the index arrays is computed in c indices, and finally converted back to Fortran. So this does not depend on the offset in the mask array as such.

The masking offset pops up in the loop on line 130, in particular in the +1 in `const int f = facet[m + 1];`

src/math/bcknd/device/cuda/math_kernel.h
timofeymukha: Fixed

timofeymukha:
@vbaconnet can you take another look at this one?

njansson (approved):

vbaconnet:
> @vbaconnet can you take another look at this one?

Sorry, not sure how I missed this! I will look now!

src/simulation_components/boundary_operation.f90
```diff
@@ -0,0 +1,505 @@
...
+               "[integral, average, min, max]")
+       end select
+    end do
+
+    call this%bc%init_from_components(this%coef, 0.0_rp)
+    this%bc%zone_indices = this%zone_indices
+    do i = 1, size(this%zone_indices)
+       call this%bc%mark_zone(this%bc%msh%labeled_zones(this%zone_indices(i)))
+    end do
+    call this%bc%finalize()
```
vbaconnet, line 229: It would be super cool also to be able to provide BC names in addition to zone indices. You could retrieve the zone_indices of each bc by looping through `case%fluid%bc_prs/vel` and do a `get_by_name` to grab the corresponding BC and use its internal `zone_indices` to populate the simcomp zone_indices. 

I don't know if it adds too much extra work? But up to you, it is very much a QoL thing. So I'll approve it anyways since it already looks great!

src/simulation_components/boundary_operation.f90
```diff
@@ -0,0 +1,505 @@
...
+    this%average = 0.0_rp
+    this%minimum = huge(0.0_rp)
+    this%maximum = -huge(0.0_rp)
+    area = 0.0_rp
+
+    call vector_masked_gather_copy_0(this%surface_values, &
+         this%field%x, this%bc%msk, this%field%size(), n_pts)
+
+    if (this%compute_integral .or. this%compute_average) then
+       this%integral = vector_glsc2(this%surface_values, this%areas, n_pts)
```
vbaconnet, line 413: A silly question: does this give the same result as multiplying with the mass matrix? Or is it not correct, since you need to do a surface integral, and that's why you use `area`?

vbaconnet (approved):
LGTM :) nice to have some tests with it as well!

I wanted to ask why `boundary_operation` and not `boundary_operationS` but at this stage of the PR it may be risky to open this topic :D

src/simulation_components/boundary_operation.f90
timofeymukha: I am not sure, I just copied what we do in the Neumann bcs :-). But coef builds areas independently from the mass matrix, so I assume there is a difference.

adperezm: I do not think it does. When this PR was opened I tested in PySEMTools and did not get the same.

For averages, it would be the same, but for integrals, just summing the mass over a face does not give the area, apparently haha.

src/simulation_components/boundary_operation.f90
timofeymukha: Victor, I think the idea is cool, but I am gonna merge without if for now. I am a little unsure if tapping into fluid is universally the right thing to do. For example, if one has scalars then one also has bc names there...

vbaconnet: Fair!

PR #2423, opened 2026-03-26
Append scalar name to scalar_stats simcomp name and field prefix

nick-morse:
Appends the scalar field name to the default name for each `scalar_stats` simcomp as `scalar_stats_{scalar name}`, so that the simcomp names would be `scalar_stats_s`, `scalar_stats_temp` etc., with the `scalar_stats` entries in the field registry having the same style prefix.

This allows users to have statistics for different scalars without needing to specify a different simcomp name for each, and also avoids confusing field registry error messages happening now when the simcomp names are the same. 

Also fixes the default `scalar_stats` field registry prefix to match the `fluid_stats` logic.

### DIFF
--- a/doc/pages/user-guide/statistics-guide.md
+++ b/doc/pages/user-guide/statistics-guide.md
@@ -207,7 +207,8 @@ using the `fluid_stats` simulation component (see
 ## Using statistics
 Similar to fluid statistics, scalar statistics are enabled in the case file
 as a simcomp with an additional argument `field` for the name of the scalar
-field to be averaged:
+field to be averaged, which will be appended to the default prefix of the
+`scalar_stats` field registry entries as `scalar_stats_{field}/*`:
 
 | Name              | Description                                                                                                           | Admissible values    | Default value                            |
 | ----------------- | --------------------------------------------------------------------------------------------------------------------- | -------------------- | ---------------------------------------- |
@@ -222,9 +223,10 @@ field to be averaged:
 
 \*The name of the written statistics file will by default be
 `scalar_statsX0.f0000X,..., scalar_statsX0.f0000Y` where X is the number of
-the first outputted statistic of the current run. Note that if you want to
-compute statistics for multiple scalars, you will need to specify an
-independent `output_filename` for each.
+the first outputted statistic of the current run. If `field` is specified,
+it will be appended to the default filename as `scalar_stats_{field}X0.f0000X`.
+Otherwise, you can specify an independent `output_filename` for each scalar you
+compute statistics for.
 
 In addition, one can specify the usual controls for the output, in the same
 manner as for fluid statistics. For example, if one wants to compute only the
--- a/src/scalar/scalar_stats.f90
+++ b/src/scalar/scalar_stats.f90
@@ -192,7 +192,7 @@ contains
 
     ! If a name is specified and is not the default name, add it
     ! as a prefix to the mean field names, followed by a "/".
-    if (present(name) .and. trim(name) .ne. "scalar_stats") then
+    if (present(name)) then
        unique_name = name // "/"
     else
        unique_name = "scalar_stats/"
--- a/src/scalar/scalar_stats.f90
+++ b/src/scalar/scalar_stats.f90
@@ -190,8 +190,6 @@ contains
        this%n_stats = 42
     end if
 
-    ! If a name is specified and is not the default name, add it
-    ! as a prefix to the mean field names, followed by a "/".
     if (present(name)) then
        unique_name = name // "/"
     else
--- a/src/scalar/scalar_stats.f90
+++ b/src/scalar/scalar_stats.f90
@@ -192,8 +192,10 @@ contains
 
     ! If a name is specified and is not the default name, add it
     ! as a prefix to the mean field names, followed by a "/".
-    if (present(name) .and. trim(name) .ne. "fluid_stats") then
+    if (present(name) .and. trim(name) .ne. "scalar_stats") then
        unique_name = name // "/"
+    else
+       unique_name = "scalar_stats/"
     end if
 
     ! Initialize work fields
--- a/src/simulation_components/scalar_stats_simcomp.f90
+++ b/src/simulation_components/scalar_stats_simcomp.f90
@@ -136,6 +136,9 @@ contains
        call json_get(json, "output_filename", filename)
        call scalar_stats_simcomp_init_from_components(this, name, s, u, v, w, &
             p, coef, start_time, hom_dir, stat_set, filename)
+    else if (sname_provided) then
+       call scalar_stats_simcomp_init_from_components(this, name, s, u, v, w, &
+            p, coef, start_time, hom_dir, stat_set, "scalar_stats_" // trim(sname) // "0")
     else
        call scalar_stats_simcomp_init_from_components(this, name, s, u, v, w, &
             p, coef, start_time, hom_dir, stat_set)
@@ -188,7 +191,7 @@ contains
        this%default_fname = .false.
        stats_fname = fname
     else
-       stats_fname = trim(this%name) // "0"
+       stats_fname = "scalar_stats0"
        this%default_fname = .true.
     end if
 
--- a/src/simulation_components/scalar_stats_simcomp.f90
+++ b/src/simulation_components/scalar_stats_simcomp.f90
@@ -103,11 +103,19 @@ contains
     real(kind=rp) :: start_time
     type(field_t), pointer :: s, u, v, w, p
     type(coef_t), pointer :: coef
+    logical :: sname_provided
+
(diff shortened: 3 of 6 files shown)


src/scalar/scalar_stats.f90
```diff
@@ -192,8 +192,10 @@ subroutine scalar_stats_init(this, coef, s, u, v, w, p, set, name)
 
     ! If a name is specified and is not the default name, add it
     ! as a prefix to the mean field names, followed by a "/".
-    if (present(name) .and. trim(name) .ne. "fluid_stats") then
+    if (present(name) .and. trim(name) .ne. "scalar_stats") then
```
vbaconnet: I think you could remove the `trim(name) .ne. "scalar_stats"` test since it is also not present for the fluid stats.
```suggestion
    if (present(name)) then
```

src/simulation_components/scalar_stats_simcomp.f90
```diff
@@ -104,16 +104,17 @@ subroutine scalar_stats_simcomp_init_from_json(this, json, case)
     type(field_t), pointer :: s, u, v, w, p
     type(coef_t), pointer :: coef
 
-    call json_get_or_default(json, "name", name, "scalar_stats")
+    call json_get_or_default(json, 'field', &
+         sname, 's')
+    call json_get_or_default(json, "name", &
+         name, "scalar_stats_" // sname)
```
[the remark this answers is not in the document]
vbaconnet: Good point, but don't apply the AI's suggestion please :)

vbaconnet: Just as a precaution
```suggestion
         name, "scalar_stats_" // trim(sname))
```

src/scalar/scalar_stats.f90
```diff
@@ -192,8 +192,10 @@ subroutine scalar_stats_init(this, coef, s, u, v, w, p, set, name)
 
     ! If a name is specified and is not the default name, add it
     ! as a prefix to the mean field names, followed by a "/".
```
vbaconnet:
```suggestion

```

vbaconnet (approved):
Few comments on my end :) Tried to make your life easy so you could commit things from github if you're happy with my suggestions.

The AI's comment about documentation is maybe worth some consideration?

src/scalar/scalar_stats.f90
nick-morse: Oops, thanks, I was referencing an old commit

timofeymukha (approved):
I guess this is a good QoL improvement. Just update the documentation and the changelog please.

nick-morse:
I made the default base output file name also follow `scalar_stats_{field}`, if `field` is provided. Otherwise, it retains the original `scalar_stats` base file name. I think this is also a nice QoL improvement to avoid having to provide output filenames for each scalar to avoid accidentally overwriting other files. I also updated the changelog/docs :)

nick-morse:
Maybe it's better to not take the simcomp name for the stats file, I'll look at it

vbaconnet:
I think the way you did it is fine, no? at this point it’s unlikely people will give a name to this simcomp if it’s already adding the suffix based on the specified field to create a unique name. It kind of removes the need for providing a name to have “unique” simcomps. Therefore if someone wants to name the simcomp it’s a more “intentional” choice and the file can reflect that :)

nick-morse:
> I think the way you did it is fine, no? at this point it’s unlikely people will give a name to this simcomp if it’s already adding the suffix based on the specified field to create a unique name. It kind of removes the need for providing a name to have “unique” simcomps. Therefore if someone wants to name the simcomp it’s a more “intentional” choice and the file can reflect that :)

Yes, it would usually not come up, but I'd rather not change the output `scalar_stats` filename prefix unless `output_filename` is explicitly provided in the case file

src/scalar/scalar_stats.f90
```diff
@@ -190,10 +190,10 @@ subroutine scalar_stats_init(this, coef, s, u, v, w, p, set, name)
        this%n_stats = 42
     end if
 
-    ! If a name is specified and is not the default name, add it
-    ! as a prefix to the mean field names, followed by a "/".
-    if (present(name) .and. trim(name) .ne. "fluid_stats") then
+    if (present(name)) then
        unique_name = name // "/"
+    else
+       unique_name = "scalar_stats/"
```
timfelle, line 196: Careful with using `/` in names. If you want to use HDF5 with this you cannot have `/` in the name, so it might be good to avoid.

timfelle: Of course one could look for the / in the name when writing the hdf5 file and that way get the proper grouping

nick-morse: Right now this is how it is done for the `fluid_stats`, perhaps that could be changed to play nicer with hdf5 as you said

vbaconnet: Wonder what happens if you create a dataset named “foo/bar”, if it actually creates the group “foo” and places “bar” inside it? Maybe @adperezm knows?

My idea with the “/“ was to create an artificial hierarchy in the registry which would transfer nicely to hdf5.

vbaconnet: > Careful with using `/` in names. If you want to use HDF5 with this you cannot have `/` in the name, so it might be good to avoid.

Oh maybe that was the answer to my question. What does "you cannot" mean here?

timfelle: > Wonder what happens if you create a dataset named “foo/bar”, if it actually creates the group “foo” and places “bar” inside it? Maybe @adperezm knows?
> 
> My idea with the “/“ was to create an artificial hierarchy in the registry which would transfer nicely to hdf5.

Ah, I am wrong. It is only a problem for the VTKHDF standard as the extra group would confuse the readers. It is correct that hdf5 would create the group and place the dataset inside.

timfelle (approved):

src/scalar/scalar_stats.f90
adperezm: But here vtkhdf is important, since we would want to visualize these files. We should make it happy.

Would paraview not like it if we create nested groups inside pointData? I never tried. I can check later today if we are not sure

timfelle: For VTKHDF you cannot. The format explicitly specifies that each point data array must sit under `/VTKHDF/PointData/ArrayName`, where `ArrayName` cannot contain `/` or `.`.

https://docs.vtk.org/en/latest/vtk_file_formats/vtkhdf_file_format/vtkhdf_specifications.html#limitations

timfelle: If you introduce extra subgroups under `PointData`, they will in best case be ignored as only datasets are read. I don't think it will actually break the file, but the data will not show up in Paraview or other readers.

adperezm: Okay, then I think your original comment has merit. We should rethink how we name these things. I feel that it is more likely that vtkhdf would be used rather than just storing files you can't see with plain hdf.

vbaconnet: Should we piggyback on this PR to change the "/" into something else then?

nick-morse: Maybe also checking whether prohibited characters are present when adding any field?

doc/pages/user-guide/statistics-guide.md
```diff
@@ -207,7 +207,8 @@ using the `fluid_stats` simulation component (see
 ## Using statistics
 Similar to fluid statistics, scalar statistics are enabled in the case file
 as a simcomp with an additional argument `field` for the name of the scalar
-field to be averaged:
+field to be averaged, which will be appended to the default prefix of the
+`scalar_stats` field registry entries as `scalar_stats_{field}/*`:
 
 | Name              | Description                                                                                                           | Admissible values    | Default value                            |
```
linneahuusko, line 213: `name` is mentioned in simcomps.md but it is not in this table, should it be?

linneahuusko: Okay looking at the other files I believe it should be. I'm not sure I understand what happens if neither `field` nor `name` is provided, will the files be overwritten if there are multiple scalars?

linneahuusko (approved):

doc/pages/user-guide/statistics-guide.md
nick-morse: TLDR; if you do statistics of multiple scalars, just specify the `field` you want for each. 

Details: If `name` is provided, it is the name of the simulation component and will just change the prefix of the field in the field registry. `field` controls the scalar field you are doing statistics of, if it is not specified it takes the default scalar name `s`, and if it is specified, it will be appended to the simcomp name and its field registry prefix. If neither is specified, the simcomp name and filename prefix is `scalar_stats` for statistics of the default field `s`

PR #2435, opened 2026-04-02
Overset interface as a boundary condition

adperezm:
Hi,

This PR introduces a new boundary condition that can be used to connect Neko simulations through an overlapping interface.

This is a first implementation, and it does not add the sub-stepping that is generally required, but I would like to get it in. I do not want to be issuing very large pull requests. This one says I modified 21 files, but that is also due to multiple repeated codes among kernels and formatting. So please do not be discouraged to review. **The main operations happen in two files.** This should also be very self contained.

I had to make small modifications to some "unrelated" files to make cray in Lumi happy. It has been tested on CPUs and AMD-Cray. Not on Nvidia, as I do not have multiple of those.

For this, we need to run Neko on MPMD mode, so I added some docs.

The implementation as it is does not require any user-file input. It works from the case file and it should work fine with multiple over set meshes in the same run. I think I will need to add some use interface for cases in which the mesh moves out of the initial boundaries, but that will come later.

Below is a little colorful fluid dynamics:
overlapped

image

image


doc/pages/user-guide/case-file.md
```diff
@@ -441,6 +441,7 @@ table below.
 | blasius_profile     | A Blasius velocity profile.                                                                                                                            |
 | user_velocity       | The `field_dirichlet_vector_t` user-defined Dirichlet condition for velocity.                                                                          |
 | user_pressure       | The `field_dirichlet_t` user-defined Dirichlet condition for pressure.                                                                                 |
+| overset_interface   | A Dirichelt condition that prescribes values from another neko simulation running concurrently.                                                        |
```
```suggestion
| overset_interface   | A Dirichlet condition that prescribes values from another neko simulation running concurrently.                                                        |
```

vbaconnet (approved):
I feel like this is the culmination of so much work from so many different people, and seeing it coming together is very very cool :) 

Took me a while to understand the power of initializing the global interpolator with the global NEKO_COMM! Really love this, so excited to try it out.

src/bc/overset_interface.f90
```diff
@@ -0,0 +1,402 @@
...
+    call this%z_interface_dof%copy_from(DEVICE_TO_HOST, sync = .true.)
+
+  end subroutine gather_interface_dofs_
+
+  !> Set up the global interpolator.
+  subroutine setup_interpolator_(this)
+    class(overset_interface_t), intent(inout) :: this
+
+    call this%interface_interpolator%init(this%dof, NEKO_GLOBAL_COMM, &
+         mask = this%domain_element_mask)
```
vbaconnet: Sorry, not to be that guy........ do we want `tol` and `pad` to be passed down as an option here? If the interfaces have to overlap I don't think there would be interpolation issues but just in case?

adperezm: We do, actually. It is better. I will come to you for suggestions on that.

adperezm: I added this now. Let me know your thoughts.

vbaconnet: great!

doc/pages/user-guide/case-file.md
```diff
@@ -575,6 +576,27 @@ A more detailed description of each boundary condition is provided below.
     "zone_indices": [1, 2]
   }
   ```
+* `overset_interface`, a dirichlet boundary condition that retrieves values
```
timofeymukha:
```suggestion
* `overset_interface`, a Dirichlet boundary condition that retrieves values
```

doc/pages/user-guide/case-file.md
```diff
@@ -575,6 +576,27 @@ A more detailed description of each boundary condition is provided below.
...
+  interface are included in one boundary. This avoids repeated calls to
+  interpolation routines.
+
+  As a note, both meshes must be overlapping with at least one element.
+
+  Since this requires another concurrent simulation, you must execute neko in
+  [multiple-program-multiple-data (MPMD)](#user-file_tips_mpmd) mode. Note that both simulations are otherwise
+  independent, therefore the rest of the user case can be modified as seen fit.
+
+  *Time-step* must be the same between the simulations. variable time-step is not
```
timofeymukha:
```suggestion
  The *time-step* must be the same between the simulations. A variable time-step is not
```

doc/pages/user-guide/user-file.md
```diff
@@ -658,6 +658,69 @@ use `sync = .true.` as a starting point.
 Finally, observe that we use the flag `NEKO_BCKND_DEVICE` to check if we are
 indeed running on GPUs. In that case, `NEKO_BCKND_DEVICE` would be equal to 1.
 
+### Running on multiple program multiple data (MPMD) mode {#user-file_tips_mpmd}
```
timofeymukha:
```suggestion
### Running in multiple program multiple data (MPMD) mode {#user-file_tips_mpmd}
```

doc/pages/user-guide/user-file.md
```diff
@@ -658,6 +658,69 @@ use `sync = .true.` as a starting point.
...
+
+Note that the operator `:` is the one that indicates that both processes
+share the same global communicator **this is not the same as running with &**.
+The latter generally runs two independent sessions that do not necessarily talk
+to each other. Check your MPI implementation's way of handling MPMD. 
+If the other program is not neko, then you just need to make sure
+to split the global communicator with a color that is not the same as the id 
+that you gave to neko.
+
+In supercomputers, if `slurm` is used, generally the jobs need to be launched
```
timofeymukha:
```suggestion
On supercomputers, if `slurm` is used, generally the jobs need to be launched
```

doc/pages/user-guide/user-file.md
```diff
@@ -658,6 +658,69 @@ use `sync = .true.` as a starting point.
...
+with the `--multi-prog` flag accompanied by a `mpmd.conf` file that indicates 
+which rank runs which program. An example for running on a GPU cluster, where
+gpus must be selected looks like this:
```
timofeymukha:
```suggestion
GPUs must be selected looks like this:
```

src/device/device.F90
```diff
@@ -795,6 +795,7 @@ subroutine device_map_r1(x, x_d, n)
     end select
 
     call device_alloc(x_d, s)
+    if (s .eq. 0) return
     call device_associate(x, x_d)
```
timofeymukha, line 846: @njansson @timfelle Just marking this as the big general thing to discuss here, so that we don't merge this in without consideration. There was some discussion on Zulip, but it also fits into the destructor checks we were working on.

src/bc/overset_interface_vector.f90
```diff
@@ -0,0 +1,498 @@
+! Copyright (c) 2020-2026, The Neko Authors
```
timofeymukha:
```suggestion
! Copyright (c) 2026, The Neko Authors
```

src/bc/overset_interface_vector.f90
```diff
@@ -0,0 +1,498 @@
...
+  use json_utils, only : json_get_subdict_or_empty
+  use field_list, only : field_list_t
+  use, intrinsic :: iso_c_binding, only : c_ptr, c_size_t
+  use time_state, only : time_state_t
+  implicit none
+  private
+
+  !> Extension of the user defined dirichlet condition `field_dirichlet`
+  ! for the application on a vector field.
+  type, public, extends(bc_t) :: overset_interface_vector_t
```
timofeymukha: You forgot to change the docstring here

src/bc/overset_interface.f90
```diff
@@ -0,0 +1,410 @@
...
+  end subroutine overset_interface_update
+
+  !===================
+  ! Helper subroutines
+  !===================
+
+  !> Build masks.
+  subroutine build_masks_(this)
+    class(overset_interface_t), intent(inout) :: this
+    type(mask_t) :: temp_mask
```
timofeymukha, line 350: If the helper routines are not type-bound, do not use `this`, because that argument name basically tells the reader that we are working on an object that is `pass`ed.

src/bc/overset_interface.f90
```diff
@@ -0,0 +1,410 @@
...
+    character(len=:), allocatable :: field_name
+
+    call json_get(json, "field_name", field_name)
+    call this%init_from_components(coef, field_name)
+    if (allocated(field_name)) deallocate(field_name)
+
+    !> Store the interpolation settings
+    call json_get_subdict_or_empty(json, "interpolation", &
+         this%interpolation_settings)
+
```
timofeymukha: What if I directly init from components, do I have a chance to configure the interpolation through that constructor? The ideal split is to never assign type components in the json constructor, but only use it to parse all the JSON and then forward to a constructor from components.

src/bc/overset_interface_vector.f90
```diff
@@ -0,0 +1,498 @@
...
+    call vector_masked_scatter_copy(this%bc_w%field_bc%x(:,1,1,1), this%w_interface, &
+         this%interface_dof_mask, this%bc_w%dof%size())
+
+
+  end subroutine overset_interface_update
+
+  !===================
+  ! Helper subroutines
+  !===================
+
```
timofeymukha, line 423: So are these type-specific after all? Then why not make them TBPs?

timofeymukha (commented):
Adalberto, I did a shallow review so far so preliminary comments. I still haven't really looked into how the interpolation works. I look forward to understanding how it manages to grab a field from another sim :-)). So might add some comments later on!

P.S. Is there something prohibiting us to already now add it to the scalar? Is there anything that needs modification besides for the scalar bc factory?

adperezm:
> Adalberto, I did a shallow review so far so preliminary comments. I still haven't really looked into how the interpolation works. I look forward to understanding how it manages to grab a field from another sim :-)). So might add some comments later on!
> 
> P.S. Is there something prohibiting us to already now add it to the scalar? Is there anything that needs modification besides for the scalar bc factory?

Nice! thank you! I will go through the comments as soon as possible. I agree with all of them.

At some point we can discuss more about that interpolation step. Victor also wanted to understand it more :)

src/bc/overset_interface_vector.f90
adperezm: I will make them so. All of this code was inside finalize at first and it was bit messy to read so took it out into these helpers. Since this was called as `call this%finalize()` I did not want to have other calls to `this` inside. But I guess in the end that is what I do haha.

src/bc/overset_interface.f90
adperezm: Modified (up soon)

src/bc/overset_interface.f90
adperezm: @timofeymukha , I modified now things such that it would comply with your suggestions.

I resolved comments that were trivial and left this one and the one regarding the device map.

adperezm:
> Adalberto, I did a shallow review so far so preliminary comments. I still haven't really looked into how the interpolation works. I look forward to understanding how it manages to grab a field from another sim :-)). So might add some comments later on!
> 
> P.S. Is there something prohibiting us to already now add it to the scalar? Is there anything that needs modification besides for the scalar bc factory?

Hi, Now the corrections to device are in. And I also added things to the scalar factory. So I think we are in a good spot to finalize this one.

adperezm:
It seems the HIP environment is not happy. Is it us or github?

adperezm:
@timfelle, @njansson, in the end I added an optional size input to device_associate in this PR. It turns out that in the interpolation routines we do rely on calling device_map with a host buffer of size != 0 but then we map to device with a runtime defined n = 0.

I will check if in the future I can remove that, but for now I do not want to get too deep into modifying the interpolation objects.

timofeymukha:
@adperezm I had an AI explain to me how the interpolation works, it's really cool! I had it compose the following addition to the overset bc type dosctring (a bit modified manually by myself).

```f90
  !! @details
  !! This boundary condition is intended for overset or Chimera-style coupling,
  !! via MPMD runs where two separate Neko simulations share the same
  !! global MPI job. In that setup, each simulation has its own local
  !! `NEKO_COMM`, selected through `NEKO_COMM_ID`, while overset interpolation
  !! is performed collectively on `NEKO_GLOBAL_COMM`, which spans all ranks from
  !! both simulations.
  !!
  !! The boundary condition does not access donor values directly from another
  !! simulation's registry or memory. Instead, it uses `global_interpolation_t`
  !! as a distributed point-evaluation service. During `finalize()`, the
  !! boundary DOFs are gathered as interpolation target points and
  !! `find_points()` is called on `NEKO_GLOBAL_COMM` so both simulations are
  !! searched for donors.
  !!
  !! Donor selection is controlled by geometry and masking, not by
  !! `NEKO_COMM_ID`. The interpolator is initialized with `domain_element_mask`,
  !! which excludes all elements that touch the overset interface on the local
  !! simulation. As a result, the receiving simulation does not search for donor
  !! values in its own interface-adjacent elements; those points must instead be
  !! found in the remaining searchable part of the global domain, i.e. in
  !! the overlapping donor simulation.
``` 

This basically answers two questions: 1) How does one sim get data from the other one 2) If one searches globally for the interface points, why are they not found twice (on both sims) -> the answer is the masking machinery. I think it will be useful to have this clarified at a conceptual level in the dosctring. But feel free to edit or discard!

adperezm:
> @adperezm I had an AI explain to me how the interpolation works, it's really cool! I had it compose the following addition to the overset bc type dosctring (a bit modified manually by myself).
> 
> ```fortran-modern
>   !! @details
>   !! This boundary condition is intended for overset or Chimera-style coupling,
>   !! via MPMD runs where two separate Neko simulations share the same
>   !! global MPI job. In that setup, each simulation has its own local
>   !! `NEKO_COMM`, selected through `NEKO_COMM_ID`, while overset interpolation
>   !! is performed collectively on `NEKO_GLOBAL_COMM`, which spans all ranks from
>   !! both simulations.
>   !!
>   !! The boundary condition does not access donor values directly from another
>   !! simulation's registry or memory. Instead, it uses `global_interpolation_t`
>   !! as a distributed point-evaluation service. During `finalize()`, the
>   !! boundary DOFs are gathered as interpolation target points and
>   !! `find_points()` is called on `NEKO_GLOBAL_COMM` so both simulations are
>   !! searched for donors.
>   !!
>   !! Donor selection is controlled by geometry and masking, not by
>   !! `NEKO_COMM_ID`. The interpolator is initialized with `domain_element_mask`,
>   !! which excludes all elements that touch the overset interface on the local
>   !! simulation. As a result, the receiving simulation does not search for donor
>   !! values in its own interface-adjacent elements; those points must instead be
>   !! found in the remaining searchable part of the global domain, i.e. in
>   !! the overlapping donor simulation.
> ```
> 
> This basically answers two questions: 1) How does one sim get data from the other one 2) If one searches globally for the interface points, why are they not found twice (on both sims) -> the answer is the masking machinery. I think it will be useful to have this clarified at a conceptual level in the dosctring. But feel free to edit or discard!



> @adperezm I had an AI explain to me how the interpolation works, it's really cool! I had it compose the following addition to the overset bc type dosctring (a bit modified manually by myself).
> 
> ```fortran-modern
>   !! @details
>   !! This boundary condition is intended for overset or Chimera-style coupling,
>   !! via MPMD runs where two separate Neko simulations share the same
>   !! global MPI job. In that setup, each simulation has its own local
>   !! `NEKO_COMM`, selected through `NEKO_COMM_ID`, while overset interpolation
>   !! is performed collectively on `NEKO_GLOBAL_COMM`, which spans all ranks from
>   !! both simulations.
>   !!
>   !! The boundary condition does not access donor values directly from another
>   !! simulation's registry or memory. Instead, it uses `global_interpolation_t`
>   !! as a distributed point-evaluation service. During `finalize()`, the
>   !! boundary DOFs are gathered as interpolation target points and
>   !! `find_points()` is called on `NEKO_GLOBAL_COMM` so both simulations are
>   !! searched for donors.
>   !!
>   !! Donor selection is controlled by geometry and masking, not by
>   !! `NEKO_COMM_ID`. The interpolator is initialized with `domain_element_mask`,
>   !! which excludes all elements that touch the overset interface on the local
>   !! simulation. As a result, the receiving simulation does not search for donor
>   !! values in its own interface-adjacent elements; those points must instead be
>   !! found in the remaining searchable part of the global domain, i.e. in
>   !! the overlapping donor simulation.
> ```
> 
> This basically answers two questions: 1) How does one sim get data from the other one 2) If one searches globally for the interface points, why are they not found twice (on both sims) -> the answer is the masking machinery. I think it will be useful to have this clarified at a conceptual level in the dosctring. But feel free to edit or discard!

Haha I like it, although I will probably change some terms. Where would you like me to put this one? at the beginning of the module or the type?

timofeymukha:
> Haha I like it, although I will probably change some terms. Where would you like me to put this one? at the beginning of the module or the type?

Doesn't matter too much I guess, but maybe the type?

timofeymukha (approved):

PR #2441, opened 2026-04-09
Data/field streamer simcomp

vbaconnet:
Adds a simcomp to stream fields from the registry using ADIOS2.

This has been successfully tested on GPUs and specifically on LUMI. 

Summary of changes:
- Added the `data_streamer` simcomp + documentation
- Added an in-situ setup to `example/cylinder`
- Modified the existing in-situ POD in `example/turb_pipe`


src/io/data_streamer.F90
```diff
@@ -88,7 +88,7 @@ subroutine data_streamer_init(this, coef)
     nelb = coef%msh%offset_el
     gdim = coef%msh%gdim
 
-#ifdef HAVE_ADIOS2
+#ifdef HAVE_ADIOS2_FORTRAN
```
adperezm: There is a thing here. The streamer uses the c++ api. But the BP FILE uses the fortran API. so the 2 preprocessing variables are different and correct. You do not need adios2 compiled with fortran to stream. So this needs to go back to the previous one. 

I can change it back. But just for information

vbaconnet: Oh I thought that since there is a call to fortran routines such as `fortran_adios2_initialize` below it needed to be fortran. But yea please change it to the correct one if you can/have time!

adperezm: I will. I will get back to this one when the timeout option gets merged. I think we can make the turb_ pipe example use this  simcomp.

timofeymukha:
Ignorant question, what does "streaming" mean in this context?

adperezm:
> Ignorant question, what does "streaming" mean in this context?

Here it means passing data from one app to another through MPMD

adperezm:
> EDIT: This has been successfully tested on GPUs and specifically on LUMI. I am also adding an in-situ variant of the cylinder example. Still missing some updates on the READMEs and documentation but overall the feature is there and is working as expected!

Very cool!

examples/cylinder/insitu_task.py
```diff
@@ -0,0 +1,237 @@
...
+    msh.ly = ds.ly
+    msh.lz = ds.lz
+    msh.gdim = ds.gdim 
+    msh.nelv = ds.nelv
+    msh.glb_nelv = ds.glb_nelv
+    return msh
+
+def log(msg):
+    if comm.Get_rank() == 0:
+        print("[INFO]", msg)
```
adperezm: I guess you wanted a different formatting for your case. But you can always do the same with pysemtools.monitoring.logger.

I like that one since it also gives time information that can be checked later in logs. But obviously not super needed here.

examples/cylinder/insitu_task.py
```diff
@@ -0,0 +1,237 @@
...
+#=========================================
+# Define some helper functions
+#=========================================
+
+def init_mesh_from_adios2(x, y, z, ds: DataStreamer):
+    """
+    Initialize a Mesh object from coordinates and 
+    """
+    msh = Mesh(comm, x = x, y = y, z = z)
+    msh.lx = ds.lx
```
adperezm: Why all the extra lx things? Mesh should initialize those automatically if the sizes of x, y, and z are right.

vbaconnet: Oh you know what I didn’t actually check. I was just worried it wouldn’t initialize things properly :) let me remove it then

examples/cylinder/insitu_task.py
vbaconnet: Oh that’s great, I was a bit lazy and didn’t look for the logger thing.

src/io/nek_adios2.cpp
```diff
@@ -51,7 +51,7 @@ extern "C" void adios2_initialize_(
     f2py_field = io_asynchronous.DefineVariable<double>("f2py_field", {gn}, {start}, {n});
     
     // If asyncrhonous execution, open the global array
-    std::cout << "create global array" << std::endl;
+    if( rank == 0 ) std::cout << "(ADIOS2) Creating global arrays" << std::endl;
```
vbaconnet: I don't think this log message is necessary, or?

adperezm: There might be situation where the code hangs here because it is waiting for the other app. In my experience, it has always been nice to know that it is waiting here. But I do not have strong opinions on it

adperezm: Oh but I think we can log something similar on the fortran side instead. And remove these ones. Since this is not formatted nicely.

vbaconnet: I added some logs around the wrappers in the data_streamer so it could be enabled at the NEKO_LOG_DEBUG. Is there a way to have some "deeper" logs in the `cpp` file that would be enabled with the NEKO_LOG_DEBUG level?

vbaconnet: We could add an `#ifdef NEKO_ADIOS2_DEBUG` statement...

adperezm: Nah, I think we can remove it. If it is waiting and the debug messages point here, then one can then debug manually things. I think it is okay to remove it

examples/cylinder/cylinder_insitu.gif
vbaconnet, line 1: This was just a bonus, but if it's too heavy we can remove it :)

examples/turb_pipe/insitu_turb_pipe.f90
```diff
@@ -55,44 +19,43 @@ subroutine compute(time)
...
-    ! Stream the data
-    call dstream%stream(u%x)
-    call dstream%stream(v%x)
-    call dstream%stream(w%x)
+    call coef%gs_h%op(w, GS_OP_ADD)
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_col2(w%x_d, coef%mult_d, n)
+    else
+       call col2(u%x, coef%mult, n)
+    end if
```
vbaconnet: @adperezm did we say this was needed, or not? The turb pipe was too big to run/test so I didn't check if that averaging over interfaces was needed or not. If it is needed I will move it to `preprocess` since `compute()` is called after the simcomps%compute so it will never get picked up.

src/io/bp_file.F90
vbaconnet, line 1: By the way, this file is not used by the data streamer. I initially used --with-adios2-fortran and saw a bunch of compilation errors so I decided to fix them but it is just piggybacking on this PR, and I have not tested or validated anything, only compilation errors.

examples/turb_pipe/insitu_turb_pipe.f90
adperezm: I think you can remove it.

src/io/bp_file.F90
adperezm: Excellent! I think it is good, since these files are not usually touched.

examples/cylinder/cylinder_insitu.gif
njansson: We should keep it !

njansson (approved):

examples/cylinder/README.md
```diff
@@ -9,3 +9,60 @@ Included files:
...
+cat > mpmd.conf << 'EOF'
+0 ./select_gpu ./neko cylinder_insitu.case
+1 python3 insitu_task.py
+2 python3 insitu_task.py
+3 python3 insitu_task.py
+4 python3 insitu_task.py
+EOF
+```
+
+This example has been succesfully executed on supercomputers such as lumi, leveraging the typically idling CPUs on a GPU compute node.
```
adperezm, line 68: I also added some docs in the overset pr. At some point we can merge these docs into one. But this is good in this one.

adperezm (approved):
Very nice!  Good job with this one.

vbaconnet:
this is ready for merge! If we're happy with it we can get it in :)

timofeymukha:
Merging in. But for the future, we should try to not be lazy and add all the "from components" constructors for new simcomps. It is a bit tedious, but I think it is worth that we keep it up!

vbaconnet:
> Merging in. But for the future, we should try to not be lazy and add all the "from components" constructors for new simcomps. It is a bit tedious, but I think it is worth that we keep it up!

Fair enough, I have been skipping those lately...

PR #2444, opened 2026-04-10
Masked field writer simcomp

vbaconnet:
Just having fun with Adalberto's newly added features :)

Basically exposes the masked fld output via a simcomp.


adperezm (approved):
Very nice! Thank you for taking this on. In theory I should have added this as well :) 

This writes out without copying the filed by using a mask in the writer routine

For other formats and to lower interpolation order etc, one can use the `submesh` feature. If one uses those, then I think hdfvtk might work, since new objects of the new shape are created and can be passed directly.

timofeymukha (approved):

timofeymukha:
Nice! In principle, we can later propagate this to simcomps that have the field writer nested!

vbaconnet:
> For other formats and to lower interpolation order etc, one can use the `submesh` feature. If one uses those, then I think hdfvtk might work, since new objects of the new shape are created and can be passed directly.

For sure. I haven't found a nice way to do it but anyone is welcome to piggyback off of this PR to get that kind of stuff in.

> Nice! In principle, we can later propagate this to simcomps that have the field writer nested!

Yes, with Adalberto's stuff I'm hoping that we could have a `field_subsampler` simcomp which creates, and populates masked fields to the registry which can be then streamed to Python..

timofeymukha:
btw are we using    `"output_format": "nek5000"` and not `fld`? Actuallynek5000 is probably better since fld is that old ascii format in reality. But we should just be consistent.

vbaconnet:
> btw are we using `"output_format": "nek5000"` and not `fld`? Actuallynek5000 is probably better since fld is that old ascii format in reality. But we should just be consistent.

Idk, since it is an input to the field_writer I used that, but it surprised me as well.

PR #2445, opened 2026-04-10
Fix gradient order in TKE model

loretet:
Just a small PR to fix a bug in the TKE SGS model.

In the `device` file for `hip`, the order of the gradients input in the kernel was wrong and this leads to an error when the model is run on GPU with hip.


timofeymukha (approved):

loretet:
Ops sorry, I canceled the Copilot request by mistake

timofeymukha:
> Ops sorry, I canceled the Copilot request by mistake

I just want to check if it will create fake complaints :D

linneahuusko (approved):

PR #2447, opened 2026-04-10
Implement Richardson wall model for ABL simulations

loretet:
Implements the Richardson wall model [1] into Neko.

The Richardson wall model is basically a "simplified" version of MOST, where the Richardson number is used directly in the computation without the need for a for-loop computing the Obukhov length. More info in the reference and #2385.

It behaves very similar to MOST, as expected.

output

Not ready to merge yet as we are finishing testing on GPU.

--------------

[1]: 
Mauritsen, T., G. Svensson, S. S. Zilitinkevich, I. Esau, L. Enger, and B. Grisogono, 2007: A Total Turbulent Energy Closure Model for Neutrally and Stably Stratified Atmospheric Boundary Layers. J. Atmos. Sci., 64, 4113–4126, https://doi.org/10.1175/2007JAS2294.1.


loretet:
I guess we can start opening this PR for reviews/comments. 

This WM is super similar to MOST and many of the modifications done here are the same as the MOST PR #2385. It's still not ready to be merged, but only few things are left

njansson (approved):

linneahuusko (approved):

PR #2459, opened 2026-04-14
Fix aliasing issue in phmg routines

njansson:
This PR fixes alias issues in the phmg routines when sub3 is used. The arguments of sub3 is `inout`, `in`, `in` thus we are not only breaking the standard by aliasing the arguments, we are also aliasing the intents ... (T_T)

### DIFF
--- a/src/multigrid/phmg.f90
+++ b/src/multigrid/phmg.f90
@@ -54,11 +54,11 @@ module phmg
   use interpolation, only : interpolator_t
   use json_module, only : json_file
   use json_utils, only : json_get_or_default, json_get
-  use math, only : copy, col2, add2, add2s2
+  use math, only : copy, col2, add2, add2s2, add2s1
   use device, only : device_get_ptr, device_stream_wait_event, glb_cmd_queue, &
        glb_cmd_event
-  use device_math, only : device_rzero, device_copy, device_add2, device_sub3,&
-       device_add2s2, device_invcol2, device_glsc2, device_col2
+  use device_math, only : device_rzero, device_copy, device_add2, &
+       device_add2s2, device_invcol2, device_glsc2, device_col2, device_add2s1
   use neko_config, only: NEKO_BCKND_DEVICE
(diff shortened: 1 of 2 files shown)


pohm01:
@njansson can you fix the usage of `device_sub3` while you are at it? It does the same aliasing (kernels want `(T*, const T*, const T*)`)

njansson:
> @njansson can you fix the usage of `device_sub3` while you are at it? It does the same aliasing (kernels want `(T*, const T*, const T*)`)

can't we just swap to `device_sub2` ?

src/multigrid/phmg.f90
```diff
@@ -509,7 +509,9 @@ subroutine phmg_jacobi_smoother(z, r, w, mg, msh, Ax, n, lvl)
           call Ax%compute(w%x, z%x, mg%coef, msh, mg%Xh)
           call mg%gs_h%op(w%x, n, GS_OP_ADD)
           call mg%bclst%apply_scalar(w%x, n)
-          call sub3(w%x, r%x, w%x, n)
+          do concurrent (j = 1:n)
+             w%x(i,1,1,1) = r%x(i,1,1,1) - w%x(i,1,1,1)
```
njansson:
```suggestion
             w%x(j,1,1,1) = r%x(j,1,1,1) - w%x(j,1,1,1)
```

pohm01:
`sub2` and `device_sub2` do `a=a-b` but here we need `a=b-a` (the negative of sub2)

njansson:
> `sub2` and `device_sub2` do `a=a-b` but here we need `a=b-a` (the negative of sub2)

`add2s1` should work den `a = c*a + b`

njansson:
> > `sub2` and `device_sub2` do `a=a-b` but here we need `a=b-a` (the negative of sub2)
> 
> `add2s1` should work den `a = c*a + b`

@pohm01 done!

timfelle (approved):

pohm01 (approved):
lgtm

PR #2464, opened 2026-04-15
Add note about restarts when using the sponge source term

linneahuusko:
If the fringe function for the sponge is defined in `fluid_user_ic` it does not work to do a restart from a checkpoint file, since `fluid_user_ic` is not called during the restart. I added a note about this in the documentation, suggesting to instead use `user_init_modules` in that case.

### DIFF
--- a/doc/pages/user-guide/case-file.md
+++ b/doc/pages/user-guide/case-file.md
@@ -951,7 +951,7 @@ the prefix `"sponge_bf_"`, meaning that `u` will be in `sponge_bf_u`, etc.
 This prefix can be changed by setting the parameter `bf_registry_prefix`.
 
 <details>
-<summary><b><u>Example using `user_init_modules`</u></b></summary>
+<summary><b><u>Example using `initialize`</u></b></summary>
 
 ```fortran
 module user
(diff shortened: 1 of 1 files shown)


vbaconnet:
oh that is a very good point. Thank you for catching this!

Then perhaps it is good to not tell users to use the initial condition at all? Just thinking that setting up the fringe is kind of different from setting up the initial condition so they should probably not be in the same subroutine anyways. Just a thought :) 

Especially since the fortran code snippet uses user%initialize...

doc/pages/user-guide/case-file.md
```diff
@@ -937,7 +937,7 @@ The reference velocity field, or `baseflow` can be set from three methods:
 Finally, the fringe function field must be filled by the user. This must be
 done through the user file by adding the fringe field to the
 `neko_registry` in either `user_init_modules` or `fluid_user_ic` (more
```
```suggestion
`neko_registry` in either `initialize` or `initial_conditions` (more
```

doc/pages/user-guide/case-file.md
```diff
@@ -937,7 +937,7 @@ The reference velocity field, or `baseflow` can be set from three methods:
...
-specifically, before the first call to compute the sponge source term).
+specifically, before the first call to compute the sponge source term). Note that `fluid_user_ic` is not called when doing a restart from a checkpoint file, so if restarts will be done the sponge should be implemented in `user_init_modules`.
```
```suggestion
specifically, before the first call to compute the sponge source term). Note that `initial_conditions` is not called when doing a restart from a checkpoint file, so if restarts will be done the sponge should be implemented in `user_init_modules`.
```

vbaconnet (approved):
In the end I think it's fine as is :) i just updated the names of the user routines since they were changed in the meantime!

loretet:
Thanks for specifying this in the PR: I had followed exactly that part of the doc when I got it wrong :D 
I think how it looks now (after Victor's modifications to the routine names) is perfect

loretet (approved):

PR #2470, opened 2026-04-17
Translation source term

adperezm:
Adds a term to emulate mesh translation in the domain with constant velocity.

This, as many other terms, is encapsulated in ALE. I thought to add it to give flexibility on users on how they want to set up their simulations. Also because I am testing some stuff and needed something very simple :)

Credits: Martin Karp used this forcing in the past. I adapted it from his user files.


adperezm:
There are some weird things with the signs here that I am confused about, so it would be great if @arash-mp can help. Maybe we discuss in person.

src/source_terms/translation_source_term.f90
```diff
@@ -0,0 +1,207 @@
...
+    work1 = this%domain_vel(1)
+    work2 = this%domain_vel(2)
+    work3 = this%domain_vel(3)
+
+    ! umesh_j * du/dx_j = div (umesh u) - u * div umesh but div umesh = 0 so we just calculate div (umesh u)
+    call field_col3(work4, u, work1, u%size())
+    call field_col3(work5, u, work2, u%size())
+    call field_col3(work6, u, work3, u%size())
+    call div(work7%x, work4%x, work5%x, work6%x, this%coef)
+    call field_add2(fx, work7, work7%size())
```
arash-mp, line 183: @adperezm just a quick note at first sight, here I think you are missing the mass matrix. am I right?
check out compute_ale_advection_no_dealias .

src/source_terms/translation_source_term.f90
```diff
@@ -0,0 +1,207 @@
...
+! INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+! BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+! LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+!
+!> Implements the `coriolis_source_term_t` type.
+!! Maintainer: Timofey Mukha
```
vbaconnet:
```suggestion
!> Implements the `translation_source_term_t` type.
```

src/source_terms/translation_source_term.f90
arash-mp: the same in y and z

src/source_terms/translation_source_term.f90
```diff
@@ -0,0 +1,207 @@
+! Copyright (c) 2024, The Neko Authors
```
arash-mp: 2026

arash-mp:
> There are some weird things with the signs here that I am confused about, so it would be great if @arash-mp can help. Maybe we discuss in person.

Sure.
The signs are ok in the code, but we can discuss it.
maybe it would be good also to give it a try with ALE (to validate the result). For that, I put an example in documentaitons how to do essentially the same thing (translation of the domain).
https://neko.cfd/docs/develop/d6/def/user-file.html#user-file_ale-base-shapes

I can also setit up quickly if you share your case and mesh file.

doc/pages/user-guide/case-file.md
```diff
@@ -710,6 +710,13 @@ define both `coriolis` and `centrifugal` source terms in a consistent way.
    applied in a user-specified region of the domain.
 9. `field`, uses fields in the `neko_registry` as values of the source term. The
    fields are selected with the `field_names` keyword.
+10. `translation` Adds a forcing term corresponding to a domain translating with
+    constant velocity. The term comes from the ALE formulation and is written as
+    \f$ (\mathbf{w}\cdot\nabla)\mathbf{u} \f$,
+    or, for spatially constant mesh velocity \f$ \nabla\cdot\mathbf{w} = 0 \f$,
```
timofeymukha, line 873: I have to admit that I spent quite a bit of time figuring out why a uniformly translating domain changes the governing equations. It is a little confusing because such a transformation is Galilean and should not change the equations. But this is basically not fully committing to a new frame but just measuring advection with respect to it, or?

I realize this must be some trivial edge case of ALE as you write, I am just not that used to thinking in those terms.

I feel like somehow improving the doc, but I am not sure how :-). Could we maybe mention a situation where this source term would be useful?

adperezm:
> > There are some weird things with the signs here that I am confused about, so it would be great if @arash-mp can help. Maybe we discuss in person.
> 
> Sure. The signs are ok in the code, but we can discuss it. maybe it would be good also to give it a try with ALE (to validate the result). For that, I put an example in documentaitons how to do essentially the same thing (translation of the domain). https://neko.cfd/docs/develop/d6/def/user-file.html#user-file_ale-base-shapes
> 
> I can also setit up quickly if you share your case and mesh file.

Yeah, I intend to test that later, but I feel there are more moving parts there out of my control to check things if something goes wrong, so that's why this sort of forcing was a simple route.

doc/pages/user-guide/case-file.md
adperezm: You are not alone in that confusion/thinking :).

Maybe I can refer to ALE docs and they explain that better? I am not sure either haha.

src/source_terms/translation_source_term.f90
adperezm: Hej, no, when source terms are added in this case, that is not needed. The mass is multiplied later.

adperezm:
> > There are some weird things with the signs here that I am confused about, so it would be great if @arash-mp can help. Maybe we discuss in person.
> 
> Sure. The signs are ok in the code, but we can discuss it. maybe it would be good also to give it a try with ALE (to validate the result). For that, I put an example in documentaitons how to do essentially the same thing (translation of the domain). https://neko.cfd/docs/develop/d6/def/user-file.html#user-file_ale-base-shapes
> 
> I can also setit up quickly if you share your case and mesh file.

Sounds good. I will check that later but first I need something simple to verify that things are okay on my side and not mix too many things.

One thing, is there a translation mode in ALE from the case file? I only saw other types of motion. This one allows me to keep the user file cleaner for now.

arash-mp:
> > > There are some weird things with the signs here that I am confused about, so it would be great if @arash-mp can help. Maybe we discuss in person.
> > 
> > 
> > Sure. The signs are ok in the code, but we can discuss it. maybe it would be good also to give it a try with ALE (to validate the result). For that, I put an example in documentaitons how to do essentially the same thing (translation of the domain). https://neko.cfd/docs/develop/d6/def/user-file.html#user-file_ale-base-shapes
> > I can also setit up quickly if you share your case and mesh file.
> 
> Sounds good. I will check that later but first I need something simple to verify that things are okay on my side and not mix too many things.
> 
> One thing, is there a translation mode in ALE from the case file? I only saw other types of motion. This one allows me to keep the user file cleaner for now.

No, Actually i wanted to add it when we talked a while ago, but I forgot:D 
it's quite simple to add, I will do it maybe this weekend

arash-mp:
to check, we can have a cylinder with inlet_vel = 1 (no forcing) 
and using this forcing, we can set inlet_vel = 0, but the domain_velocity should be -1, 0, 0. 
They should be equivalent, right?

doc/pages/user-guide/case-file.md
adperezm: > But this is basically not fully committing to a new frame but just measuring advection with respect to it, or?

And I think that is exactly the case. The thing is that I understand the term but I have not fully read how we reach it.

But this is as if advection is done by the relative velocity, rather than the absolute one. There are many more things that I think happen, but they become zero for this case of constant velocity in time.

src/source_terms/translation_source_term.f90
arash-mp: yee ok

njansson (approved):

doc/pages/user-guide/case-file.md
adperezm: I added more comments on the docs. But one can see later if they should be expanded. Eventually we might remove this in favor of just keeping ALE, but for now I want to have it in.

adperezm:
This one should be ready for merge, as comments have been adressed.

timofeymukha (approved):

PR #2472, opened 2026-04-17
field_subsampler simcomp

vbaconnet:
Introduces a simulation component `field_subsampler` that creates fields, either masked by a point_zone or with a different polynomial order.

The simcomp is used quite easily by specifying `source_fields`, taken from the registry, and either:
- a `point_zone`,
- a `polynomial_order`, or
- both

The simcomp will throw an error if neither a point zone nor a poly. order different from the one in the simulation are provided.

The result is:
- new fields added to the registry which can be grabbed by, for example, a `data_streamer`. The naming convention for the new fields are `<name_of_simcomp> + "/" + <name_of_source_field>`. **I am using the same delimiter as "/" for consistency with fluid stats, etc... But I do feel that we are creating some technical debt with having this delimiter (If we ever want to change it back to `"_"` due to vtkhdf)**
- A field writer which can output those new fields and potentially replace the original fluid output if one wants to run with one polynomial order and output at a lower one.

```.json
 {
   "type": "field_subsampler",
   "source_fields": ["u", "omega_z", "fluid_stats/mean_u"], // adds "field_subsampler_u", "field_subsampler_omega_z"... etc
   "point_zone": "my_point_zone",
   "polynomial_order": 3,
   "compute_control": "tsteps",
   "compute_value": 10
 }
 ```

## Demo

Just a quick demo to show how it works, with the cylinder example, polynomial order 3 and the following point_zone:
```json
{
    "name": "mybox",
    "geometry": "box",
    "x_bounds": [-2.0, 5.0],
    "y_bounds": [-3.0, 3.0],
    "z_bounds": [-99.0, 99.0],
    "full_elements": true // must be enabled for this simcomp.
}
```
<details>
  <summary>Show demo</summary>

The original mesh in black, overlayed by the sub-mesh masked with the point zone and polynomial order 3 in red.
image

Example showing the original flow field (top), masked field with point zone and same polynomial order (bottom left) and polynomial order 3 (bottom right)
image

</details>

## Notes on the field_subsampler implementation

Just some extra comments that I find relevant to explain the design choices in my implementation and to (potentially) discuss.

<details>
  <summary>Show notes</summary>

### Compute interface

I chose here to have `compute_` be assigned at runtime based on the inputs. The reason for that is to avoid too many if statements in the `compute` to properly handle the cases where we have space-to-space interpolation and/or `point_zone` masking enabled or not.

There are essentially three different `computes`, depending on if we have:
- a `point_zone` only: `field_subsampler_compute_pz`,
- a `polynomial_order` only: `field_subsampler_compute_Xh`
- both: `field_subsampler_compute_pz_Xh`. The particulary of this one is that we need to use a "buffer" field, with the same `lx` as `source_field`, for the masked_copy (the mask indices are built with the `lx` of the source field) 

### Field lists to avoid neko_registry calls

I don't know if this is actually an issue, but in order to avoid calling `neko_registry%get_field` too many times inside `this%compute_`, I initialize two `field_lists` that point to the fields that will be needed.

*Declaration*
https://github.com/ExtremeFLOW/neko/blob/d73afd0d9907638f4d5345b102d952191a3ad92e/src/simulation_components/field_subsampler.f90#L82-L85

*Assignment*
https://github.com/ExtremeFLOW/neko/blob/d73afd0d9907638f4d5345b102d952191a3ad92e/src/simulation_components/field_subsampler.f90#L411-L421

*Usage*
https://github.com/ExtremeFLOW/neko/blob/d73afd0d9907638f4d5345b102d952191a3ad92e/src/simulation_components/field_subsampler.f90#L513-L517

It just feels strange to me to do the `neko_registry%get_field` searches in the `compute`. I guess it won't make a huge difference in performance but the principle just bothers me a bit. Perhaps @njansson has thought on this :) 

</details>


vbaconnet:
@timofeymukha I think copilot could be helpful here for the memory management part of things. I've already asked copilot to review some things internally but I don't think it was very complete.

vbaconnet:
Thanks @timofeymukha! Those comments were actually very helpful. I haven't tagged anyone for review but if anyone wants to look at it, this is pretty much a finished product :)

adperezm (approved):
Very nice!

adperezm:
This one is ready right? We just need someone to rerun the tests.

njansson (approved):
🍻

PR #2478, opened 2026-04-21
Fix OIFS on the CPU

timofeymukha:
This seems to fix OIFS on the CPU (at least one can run hemi).

On the GPU it is still broken, and the issue does seem to be the u_gl vector. But I couldn't pinpoint the issue, so for now at least we can merge this I guess.


timofeymukha:
Is the OIFS developer still around?

njansson (approved):

vbaconnet (approved):

PR #2482, opened 2026-04-21
Decrease blocks/threads from 1024 to 256

loretet:
Decreased blocks/threads from 1024 to 256 for CUDA wall models. Works better for testing and hopefully production runs.
Done under the guide of Niclas the wise (I don't know much about this)

### DIFF
--- a/src/wall_models/bcknd/device/cuda/rough_log_law.cu
+++ b/src/wall_models/bcknd/device/cuda/rough_log_law.cu
@@ -46,8 +46,8 @@ extern "C" {
           void *tau_x_d, void *tau_y_d, void *tau_z_d,
           int *n_nodes, int *lx, real *kappa, real *rho, real *B, real *z0, int *tstep) {
     
-    const dim3 nthrds(1024, 1, 1);
-    const dim3 nblcks(((*n_nodes) + 1024 - 1) / 1024, 1, 1);
+    const dim3 nthrds(256, 1, 1);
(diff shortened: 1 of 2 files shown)


src/wall_models/bcknd/device/cuda/rough_log_law.cu
```diff
@@ -46,8 +46,8 @@ extern "C" {
           void *tau_x_d, void *tau_y_d, void *tau_z_d,
           int *n_nodes, int *lx, real *kappa, real *rho, real *B, real *z0, int *tstep) {
     
-    const dim3 nthrds(1024, 1, 1);
-    const dim3 nblcks(((*n_nodes) + 1024 - 1) / 1024, 1, 1);
+    const dim3 nthrds(256, 1, 1);
```
linneahuusko, line 49:
```suggestion
    const dim3 nthrds(1024, 1, 1);
```

src/wall_models/bcknd/device/cuda/rough_log_law.cu
```diff
@@ -46,8 +46,8 @@ extern "C" {
...
+    const dim3 nblcks(((*n_nodes) + 256 - 1) / 256, 1, 1);
```
linneahuusko, line 50:
```suggestion
    const dim3 nblcks(((*n_nodes) + 1024 - 1) / 1024 1, 1);
```

PR #2483, opened 2026-04-22
Add a global variable name length 

### LINKED ISSUE #2471: Set a standard size for field names
I started thinking about this after seeing that the `field_writer` init currently looks for field names up to 20 characters, although in field_t they are declared as being 80 characters long:

https://github.com/ExtremeFLOW/neko/blob/7d2449fc2d3ae3bf21a1345b078f36c4cf6b743e/src/field/field.f90#L55

https://github.com/ExtremeFLOW/neko/blob/7d2449fc2d3ae3bf21a1345b078f36c4cf6b743e/src/simulation_components/field_writer.f90#L101

With recent simcomp added to the code, we are starting to add more and more fields in the registry with unique names, which are usually based on names of simcomps and can be quite long. Case in point, `"fluid_stats/u_mean"` is already 18 characters... So it's a matter of time before we run into issues.

In the same vein as `NEKO_FNAME_LEN` and `LOG_SIZE`, I propose that we set a `NEKO_FIELD_NAME_LEN` in `field_t` and use that consistently. Or why not even set a standard size `NEKO_VARNAME_LEN` (or whatever we want to call it) for vector, matrix, field names.

vbaconnet:
Adds the `NEKO_VARNAME_LEN` variable to `common/utils`, which is meant to be used for `name`-like attributes such as `field_t`, `vector_t`, but also to retrieve field names from the case file like we do in many simcomps.

Fixes #2471 

~Waiting for #2413 to be merged so that this can also be added to `matrix%name`, but opening this for feedback.~


vbaconnet:
just fyi this is ready for review/merge but i know we have a lot of prs right now so I'm not tagging anyone specifically :)

timfelle (approved):

timofeymukha (approved):
Looks innocent enough :).

PR #2485, opened 2026-04-22
Update ABL example + minor fixes

loretet:
Quick PR to update the current ABL flow example.

- added sponge layer
- switch WM to MOST
- added a couple of references

-----

I also did two quick and small modifications to 
- the `rough_log_law` wall model: it now automatically assigns a default value to `B`and `kappa` if the user doesn't specify them (consistent with MOST)
- the `sponge` documentation: the way `wbf` was assigned in the coding example wasn't very safe (it actually lead to an `ALLOCATE` error when trying to run)

Both of these are quick modifications that I don't think were worth opening a separate PR for. I thought about them while modifying the case files for the example.

### DIFF
--- a/doc/pages/user-guide/case-file.md
+++ b/doc/pages/user-guide/case-file.md
@@ -1134,7 +1134,7 @@ contains
 
     end do
 
-    wbf%x(:,1,1,1) = 0.0_rp    
+    wbf%x = 0.0_rp    
     
     if (NEKO_BCKND_DEVICE .eq. 1) then
        call device_memcpy(ubf%x, ubf%x_d, ubf%size(), &
--- a/doc/pages/user-guide/case-file.md
+++ b/doc/pages/user-guide/case-file.md
@@ -1134,7 +1134,8 @@ contains
 
     end do
 
-    wbf = 0.0_rp
+    wbf%x(:,1,1,1) = 0.0_rp    
+    
     if (NEKO_BCKND_DEVICE .eq. 1) then
        call device_memcpy(ubf%x, ubf%x_d, ubf%size(), &
             HOST_TO_DEVICE, .false.)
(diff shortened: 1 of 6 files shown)


examples/shear_convection_abl/shear_convection_abl.f90
```diff
@@ -57,6 +58,46 @@ subroutine user_ic(scheme_name, fields)
        u => fields%get("u")
        v => fields%get("v")
        w => fields%get("w")
+
+       ! Implement top-mounted sponge l(z) = l*S( (z-z0)/delta )
```
timofeymukha: Didn't we conclude that having the sponge in initialize is the way to go?

doc/pages/user-guide/case-file.md
```diff
@@ -1156,7 +1134,8 @@ contains
 
     end do
 
-    wbf = 0.0_rp
+    wbf%x(:,1,1,1) = 0.0_rp
```
timofeymukha: Doesn't matter too much, but I think either using rzero or assigning directly to wbf%x is just simpler.

examples/shear_convection_abl/shear_convection_abl.f90
vbaconnet, line 13: Why not using `constant` for the sponge baseflow instead?

examples/shear_convection_abl/shear_convection_abl.f90
vbaconnet: Yes I am surprised to see it here as well.

vbaconnet (approved):

examples/shear_convection_abl/shear_convection_abl.f90
loretet: My bad, I must have left it there from the more general formulation! Fixed it now

examples/shear_convection_abl/shear_convection_abl.f90
loretet: That's cause I'm dumb, but I fixed it now!

timofeymukha (approved):

PR #2488, opened 2026-04-22
Add high-passs filter-based filtering

timofeymukha:
- Adds the HPFRT source term for simulation stabilization. 
```json5
   {
      "type": "hpfrt",
      "filter_weight": 5.0,
      "filter_modes": 2
   }
```

- Changes a the constructor structure in fitler_t and descendents for a clean split between json parsing and actual init.
- Adds instructions for copilot to check for this in review.
- I also sneak in a .temp to gitignore. I find it useful to have an ignored dir for stashing stuff like references for AI and just temporary dev things.


Shiyu-Sandy-Du:
Just a note here, we should consider collect thing like entropy viscosity, GJP and this one under the same regularization umbrella, but it could be done in a separate PR

src/source_terms/hpfrt_source_term.f90
```diff
@@ -0,0 +1,217 @@
...
+
+    this%filter_modes = filter_modes
+    this%chi = -abs(filter_weight)
+
+    allocate(transfer(nx))
+    transfer = 1.0_rp
+    k0 = nx - filter_modes
+    do i = k0 + 1, nx
+       amp = real((i - k0) * (i - k0), kind=rp) / &
+            real(filter_modes * filter_modes, kind=rp)
```
Shiyu-Sandy-Du, line 155: Why is it a sqaured kernel? can the power be somehow user-defined otherwise default to square?

timofeymukha: That is how Nek5000 does it

```fortran

      subroutine hpf_trns_fcn(diag,kut)

c      implicit none

      include 'SIZE'
      include 'PARALLEL'

      real diag(lx1*lx1)
      integer nx,k0,kut,kk,k

      real amp

c     Set up transfer function
c
      nx = lx1
      call ident   (diag,nx)
c
      kut=kut                                     ! kut=additional modes
      k0 = nx-kut
      do k=k0+1,nx
        kk = k+nx*(k-1)
        amp = ((k-k0)*(k-k0)+0.)/(kut*kut+0.)     ! Normalized amplitude. quadratic growth
        diag(kk) = 1.-amp
      enddo
```

It is also in the Nek5000 docs: "Nek5000 will then parabolically decrease the effect of the filter for each subsequent lower mode". My overall impression from using this in Nek500 is that the existing two parameters are already typically chosen in a very adhoc manner, so controlling the decay would not add much utility.

adperezm (approved):
I never realized that the filter in the RHS was a sort of penalization term... cool to see! I only ever did it explicitly to the fields haha (not very nice).

Shiyu-Sandy-Du (approved):

njansson:
@timofeymukha can you have a look at the conflicts, would be nice to get this is 🐈‍⬛ 🐈

src/filter/PDE_filter.f90
```diff
@@ -115,32 +115,59 @@ module PDE_filter
...
-    ! user parameters
-    call json_get(json, "radius", this%r)
+    call json_get(json, "filter.radius", r)
 
-    call json_get_or_default(json, "tolerance", this%abstol_filt, &
+    call json_get_or_default(json, "filter.tolerance", abstol_filt, &
          1.0e-10_rp)
 
-    call json_get_or_default(json, "max_iter", this%ksp_max_iter, 200)
+    call json_get_or_default(json, "filter.max_iter", ksp_max_iter, 200)
```
timofeymukha: Wait, did I mess this up in the merge, @Shiyu-Sandy-Du ?

timofeymukha: Should there be a `filter.` or not?

Shiyu-Sandy-Du: It was with `filter.`, but removed in  PR #2477. Maybe @timfelle has an idea

timfelle: It used to be `filter.` but that does not make any sense, what if you want to apply multiple filters in succession. Therefore, I removed it. So for instance the Brinkman source term, have a `filter` sub dictionary in the json, I now extract that and send the contents to the filter, rather than the source term settings.

timofeymukha: Sounds reasonable. I will fix the change

PR #2492, opened 2026-04-23
Fix memcpy calls in MOST wall model

linneahuusko:
A call to `device_memcpy` was accidentally duplicated in the last commit to the MOST wall model, causing erroneous behavior of the wall model. The misplaced calls has now been removed, and proper guards added to the correct calls.


src/wall_models/most.f90
```diff
@@ -274,23 +274,15 @@ subroutine most_finalize(this, msk, facet)
...
-         HOST_TO_DEVICE, sync = .false.)
-    call device_memcpy(this%h_z_idx, this%h_z_idx_d, this%n_nodes, &
-         HOST_TO_DEVICE, sync = .true.)
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+      call device_memcpy(this%h_x_idx, this%h_x_idx_d, this%n_nodes, &
+            HOST_TO_DEVICE, sync = .false.)
+      call device_memcpy(this%h_y_idx, this%h_y_idx_d, this%n_nodes, &
+            HOST_TO_DEVICE, sync = .false.)
+      call device_memcpy(this%h_z_idx, this%h_z_idx_d, this%n_nodes, &
+            HOST_TO_DEVICE, sync = .true.)
```
loretet: Should it be false or true?
```suggestion
            HOST_TO_DEVICE, sync = .false.)
```

loretet (approved):

njansson (approved):

PR #2493, opened 2026-04-23
Fix normlization in coupled_cg

timofeymukha:
Standard CPU CG uses 1.0_rp / sqrt(coef%volume) (src/krylov/bcknd/cpu/cg.f90:160), and fused_coupled_cg does the same
 (src/krylov/bcknd/device/fusedcg_cpld_device.F90:565).

So looks like a bug in coupled_cg, or?

### DIFF
--- a/src/krylov/bcknd/cpu/cg_coupled.f90
+++ b/src/krylov/bcknd/cpu/cg_coupled.f90
@@ -230,7 +230,7 @@ contains
     else
        max_iter = this%max_iter
     end if
-    norm_fac = 1.0_rp / coef%volume
+    norm_fac = 1.0_rp / sqrt(coef%volume)
 
     associate (p1 => this%p1, p2 => this%p2, p3 => this%p3, z1 => this%z1, &
          z2 => this%z2, z3 => this%z3, r1 => this%r1, r2 => this%r2, &
--- a/src/krylov/bcknd/device/cg_cpld_device.f90
+++ b/src/krylov/bcknd/device/cg_cpld_device.f90
@@ -334,7 +334,7 @@ contains
     else
        max_iter = this%max_iter
     end if
-    norm_fac = 1.0_rp / coef%volume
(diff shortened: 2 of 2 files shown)


njansson (approved):

njansson:
Yeah, we should at least be consistent in all methods

timofeymukha:
> Yeah, we should at least be consistent in all methods

Since fused_coupled_cg used a different normalization, it may explain CPU vs GPU discrepancies in e.g. ABL cases @linneahuusko @loretet. At least partially.

loretet:
> > Yeah, we should at least be consistent in all methods
> 
> Since fused_coupled_cg used a different normalization, it may explain CPU vs GPU discrepancies in e.g. ABL cases @linneahuusko @loretet. At least partially.

Wasn't it equally wrong in both CPU and GPU...?

timofeymukha:
> > > Yeah, we should at least be consistent in all methods
> > 
> > 
> > Since fused_coupled_cg used a different normalization, it may explain CPU vs GPU discrepancies in e.g. ABL cases @linneahuusko @loretet. At least partially.
> 
> Wasn't it equally wrong in both CPU and GPU...?

coupled_cg yes, but if you used fused_coupled_cg for the GPU (you probably should), then no.

loretet:
> > > > Yeah, we should at least be consistent in all methods
> > > 
> > > 
> > > Since fused_coupled_cg used a different normalization, it may explain CPU vs GPU discrepancies in e.g. ABL cases @linneahuusko @loretet. At least partially.
> > 
> > 
> > Wasn't it equally wrong in both CPU and GPU...?
> 
> coupled_cg yes, but if you used fused_coupled_cg for the GPU (you probably should), then no.

Then it wasn't that: both tests were done with `coupled_cg` :/ but still great that you caught that, it also reminded me to change it now

loretet (approved):

Issue #2476, opened 2026-04-20
Input rho and mu as vectors from fluid field in wall models


loretet:
This issue is a work in progress based on PR #2385 (more laterally, it deals with any wall model that takes `rho` and `mu` as inputs).

At the moment, all the wall models that use either `rho` or `mu` as inputs have hardcoded `rho = 1.0`and `mu = 1e-10`(standard in atmospheric LES codes). A previous implementation had them specifiable by the user in the `init_from_components` routine, but it was taken out under suggestion of @timofeymukha on the basis that there should not be any extra `rho` value setup for the wall model in the case file; the `rho` values should instead be taken from the *density **field** of the fluid at the wall*.
The same thing should happen for `mu`, afaik.

It was decided not to get into this for now and just open an issue to then fix the problem for all wall models at once.
Therefore, at the moment, the atmospheric wall models (`MOST`, `rough_log_law` and, soon, `richardson`) assume `rho=1.0` and ` mu=1e-10` and require modification of the source code to use different values.

The proposed workflow is (from @timofeymukha):
1- implement all wall models with these default values (i.e. "sweep it under the rug").
2- modify kernel signatures to take `mu` and `rho` as vectors.
3- modify rest of the code to take `rho` field from `wall_model_t`, pack the boundary part into a vector with `masked_gather_0`, and then send `%x` or `%x_d` to the compute kernel.

timofeymukha:
spalding_t already does the right thing for `nu`, I believe. So that can be copied over to rho and mu as necessary

loretet:
That's nice! So this will be a very easy fix. I'll just wait for MOST and Ri to be merged and then I'll do it on the three models at once :)

loretet:
OK, that should now be fixed. I'm closing the issue, I guess we can continue to discuss in #2495 if there's any need

PR #2495, opened 2026-04-24
Input mu and rho into wall models as masked fields

loretet:
Should solve issue #2476.

`rho` and `mu` used to be input as scalars in the models. Now they are input as fields, masked at the boundaries.


Shiyu-Sandy-Du (approved):

src/wall_models/bcknd/device/hip/richardson_kernel.h
```diff
@@ -124,7 +124,7 @@ __device__ T heat_flux_neutral(T ti, T ts, T h, T z0h, T utau, T kappa)
 }
 
 /*
- * HIP kernel for the Richardson wall model.
+ * CUDA kernel for the Richardson wall model.
```
Shiyu-Sandy-Du: Make sure the comments are still consistent when you copy a kernel from cuda to hip :)

loretet: True, laziness at its finest

src/wall_models/bcknd/device/cuda/spalding.cu
```diff
@@ -56,7 +56,7 @@ extern "C" {
                                     (int *) ind_r_d, (int *) ind_s_d, 
                                     (int *) ind_t_d, (int *) ind_e_d,
                                     (real *) n_x_d, (real *) n_y_d, 
-                                    (real *) n_z_d, (real* ) nu_d, (real *) h_d,
+                                    (real *) n_z_d, (real *) nu_d, (real *) h_d,
                                     (real *) tau_x_d, (real *) tau_y_d,
```
timofeymukha: I think we are missing the multiplication of the shear stress by rho in Spalding, or?

loretet: Ohi, that's true! I thought I added it when I added it for the other WMs and haven't even looked this time. My bad.

loretet:
Everything should be fine now.

timofeymukha (approved):

PR #2496, opened 2026-04-27
Init the data streamer from dofmap in the field list and not fluid

vbaconnet:
The data streamer was being initialized from the fluid coef/dofmap, which causes bugs if the fields passed do not match with said dofmap (typically when used with the subsampler). 

- Added some alternative constructors for the data streamer so that the simcomp can use the msh and space_t from its own fields.
- Added some checks in the data streamer simcomp to throw an error early is ADIOS2 is not initialized.


timofeymukha (approved):
Maybe change check to validate, because we use that in some types. But not obligatory :-).

adperezm (approved):
Cool! You really are making this an actual feature to use rather than some obscure thing!

vbaconnet:
> Maybe change check to validate, because we use that in some types. But not obligatory :-).

Both the field subsampler and sponge already have a `check` ;)

PR #2499, opened 2026-04-29
Feature/moving interface

adperezm:
This PR adds some functionalities to move domains around.

This one uses some user functions to morph the interface values and find points again.

Later PRs will use ALE too, but we go one step at a time.

Reviewing this should get easier once #2470 is in.

moving_cylinders

For now I stopped with the changelog. I will add an entry once this whole thing is finalized.


timofeymukha:
My brain refuses to parse what is happening in that GIF :D.

njansson (approved):
🎉

adperezm:
> My brain refuses to parse what is happening in that GIF :D.

haha here the cylinder on the left keeps moving upwards but it is periodic, so once it leaves the domain on the right it enters it from below :P

src/fluid/fluid_pnpn_bc_fctry.f90
```diff
@@ -110,11 +111,18 @@ module subroutine pressure_bc_factory(object, scheme, json, coef, user)
...
+       call json_get_or_default(json, "couple_pressure", temp_logical, .false.)
+       if (temp_logical) then
+          allocate(overset_interface_t::object)
+          select type (obj => object)
+          type is (overset_interface_t)
+             write(*,*) "Coupling pressure on overset interface"
+             call json%add("field_name", scheme%p%name)
+             obj%morph_interface => user%morph_interface
+          end select
+       else
```
timofeymukha, line 122: Is this a new parameter to be documented in case-file.md?

Should one use neko_log instead of a raw write? Alternatively, should that message be written inside the overset_interface_t? That will probably land it better in the log structure.

adperezm: Yes, you are right on both accounts. I'll fix it

timofeymukha (approved):
Let's get it in, I think! You are clearly actively working on all this functionality, and it is basically encapsulated in a single bc, so we should just keep the momentum going.

PR #2502, opened 2026-04-30
Schwarz-like iterations

adperezm:
This PR allows to add Schwarz iterations for cases with coupled simulations at the boundaries.

This one is just one file once we merge the moving interface PR I issued recently.

I checked, and things are behaving as expected with regards to multiple subsequent solves. The only thing that I am unsure about is what happens with wall_models as BCs. I think this should work if we consider that the RHS is already estimated to be at n+1 based on n. And then the sub-iterations just help us find the proper n+1 for the next step.

As in the docs:

### Schwarz iterations
This feature is enabled by setting the `schwarz_iterations` keyword inside
the `fluid` group to an integer larger than zero. In this case, each fluid
timestep solves for velocity and pressure multiple times.

The total number of passes is `1 + n_schwarz_iterations`. This feature is often
needed to increase the stability of solutions when using `overset_interface`
boundary conditions, i.e., when there are multiple coupled simulations running
in tandem.

@note In each sub-step, the right-hand-side and forcing terms that are intended
to operate at the beginning or end of a real timestep are frozen. However, the
boundary conditions are re-applied at each sub-step. This is the intended
behaviour for overset boundaries, but functionalities such as user Dirichlet
will also be called multiple times. Note that the `t` and `tstep` variables are
only updated across real timesteps. Therefore, if your user conditions depend
on these variables, they remain valid.


adperezm:
Lets not call a copilot review until the other PR is merged, otherwise it will just give a bunch of comments on something else.

adperezm:
I would appreciate if we could evaluate if this is ready to go in (I think it is), as I am soon adding a new functionality that depends on this one.

timofeymukha (approved):

src/fluid/fluid_pnpn.f90
```diff
@@ -94,6 +94,10 @@ module fluid_pnpn
 
   type, public, extends(fluid_scheme_incompressible_t) :: fluid_pnpn_t
 
+     !> Number of schwarz-like iterations to perform each time step.
+     ! Set to 0 for no iterations.
```
timfelle:
```suggestion
     !! Set to 0 for no iterations.
```

timfelle (approved):

src/fluid/fluid_pnpn.f90
timfelle: > ```suggestion
>      !! Set to 0 for no iterations.
> ```
> 

Otherwise it will not show up in the documentation.

Issue #2500, opened 2026-04-29
OpenMP not working on LUMI


loretet:
@linneahuusko and I are having a problem running Neko on LUMI's GPUs with OpenMP enabled.

This is the error we are getting at runtime:
```
lib-4425 : UNRECOVERABLE library error
INTERNAL ERROR-Unitialized descriptor for ALLOCATE statement argument.
``` 
and this error only appears when building Neko with OpenMP enabled. The exact same case, module list and configuration flags *do work* if OpenMP is NOT enabled. So it is for sure an issue related somehow to that.

And this is what `lib-4425`is, supposedly:

<details>
<summary> lib-4425 description </summary>

```
INTERNAL ERROR-Unitialized descriptor for ALLOCATE statement argument.

The descriptor for the array or pointer to an array in an ALLOCATE statement
is not initialized.  If an allocatable item is declared, the descriptor is
initialized.

This represents an internal error that usually does not occur.  The program
may have overwritten itself, corrupting the internal data used by the run
time library.

On some systems, if a Fortran module is not present, the variables used to
specify the size of an array may not be initialized and the descriptor would
not contain the information expected for the array.

The error class is UNRECOVERABLE (issued by the run time library).
```

</details>

---------

This is how we configure Neko:
`./configure MPIFC=ftn MPICC=cc MPICXX=CC CFLAGS="-O3" --enable-openmp FCFLAGS="-O0" --prefix=$PWD HIPCC=hipcc FC=ftn CC=cc --with-hip=${ROCM_PATH} HIP_HIPCC_FLAGS="-O3 --offload-arch=gfx90a" --enable-device-mpi --with-hdf5`

and this is what we have in our bash script:

<details>
<summary> run.sh </summary>

```
#!/bin/bash -l
#SBATCH --job-name=high_res_dp
#SBATCH --account=project_xxx
#SBATCH --time=0-02:00:00              
#SBATCH --partition=dev-g
#SBATCH --nodes=2           
#SBATCH --ntasks-per-node=8    
#SBATCH --gpus-per-node=8      
#SBATCH --cpus-per-task=7         

ml CrayEnv cce/19.0.0 craype-accel-amd-gfx90a rocm/6.3.4 cray-python

export OMP_NUM_THREADS=2
export MPICH_GPU_SUPPORT_ENABLED=1
export NEKO_GS_STRTGY=3

export JSON_INSTALL=/users/lorenzol/json-fortran/b
[...]

PR #2504, opened 2026-05-04
Fix-Temp: Disable broken OpenMP region

timfelle:
Comment out the OpenMP parallel region to prevent issues during execution.

This is a temporary fix of #2500


njansson (approved):
Since this is a cce >18 issue I would rather add this as a patch (as other cce fixes) rather than breaking things on several other platforms

loretet:
Seems like a decent workaround for now. I also agree with Niclas

loretet (approved):

PR #2506, opened 2026-05-05
Log: Always print job info.

timfelle:
Enable visibility of job information for all users, even those without a case. This change enhances accessibility to job details.

It is the following block which is now printed even for library consumers of neko
```
 -------Job Information--------
 Start time: 14:15 / 2026-05-04
 Running on: 1 MPI rank, using 16 thrds each
 CPU type  : 11th Gen Intel(R) Core(TM) i9-11950H @ 2.60GHz
 Bcknd type: Accelerator (CUDA)
 Dev. name : NVIDIA RTX A3000 Laptop GPU
 Real type : double precision
```

### DIFF
--- a/src/neko.f90
+++ b/src/neko.f90
@@ -179,7 +179,11 @@ contains
     call neko_const_registry%init()
     call neko_scratch_registry%init()
 
+    !
+    ! Job information
+    !
     call neko_log%header(NEKO_VERSION, NEKO_BUILD_INFO)
+    call neko_job_info(date, time)
 
     if (present(C)) then
 
@@ -213,11 +217,6 @@ contains
              call neko_log%message(args, NEKO_LOG_QUIET)
           end do
        end if
-
-       !
-       ! Job information
-       !
-       call neko_job_info(date, time)
 
        !
        ! Create case


njansson:
Wouldn't it be better to expose this via a `-v` (or --version for the GNU crowd) switch? It some how feels strange that it prints out session start etc

timfelle:
> Wouldn't it be better to expose this via a `-v` (or --version for the GNU crowd) switch? It some how feels strange that it prints out session start etc

My usecase for this is I start a neko instance, without a case. Because I construct it myself later. But that means I do not get the MPI configuration, device backend etc printed.

Exposing the -v stuff should be done from a driver, rather than the neko initializer, since anyone building their own software and using neko as a library would then not be able to control their own commandline inputs.

vbaconnet (approved):

njansson:
> > Wouldn't it be better to expose this via a `-v` (or --version for the GNU crowd) switch? It some how feels strange that it prints out session start etc
> 
> My usecase for this is I start a neko instance, without a case. Because I construct it myself later. But that means I do not get the MPI configuration, device backend etc printed.
> 
> Exposing the -v stuff should be done from a driver, rather than the neko initializer, since anyone building their own software and using neko as a library would then not be able to control their own commandline inputs.

Sure, but for that case, I would rather say one would go the route as in the API, and call `neko_init` followed by `neko_jobinfo`. 

With always writing, it's a bit strange to get this information also for all the tooling like `prepart` and rea2nbin`

timfelle:
> > > Wouldn't it be better to expose this via a `-v` (or --version for the GNU crowd) switch? It some how feels strange that it prints out session start etc
> > 
> > 
> > My usecase for this is I start a neko instance, without a case. Because I construct it myself later. But that means I do not get the MPI configuration, device backend etc printed.
> > Exposing the -v stuff should be done from a driver, rather than the neko initializer, since anyone building their own software and using neko as a library would then not be able to control their own commandline inputs.
> 
> Sure, but for that case, I would rather say one would go the route as in the API, and call `neko_init` followed by `neko_jobinfo`.
> 
> With always writing, it's a bit strange to get this information also for all the tooling like `prepart` and rea2nbin`

Sure that is also resonable. Lets do it that way instead then.

PR #2507, opened 2026-05-05
Use case timestep as initial time for variable timestep

timfelle:
The fluid initial condition is not always a great first step for detemining the timestep based on CFL. This allows a user to manually specify an initial timestep.


vbaconnet (approved):
cool!

adperezm (approved):
Nice! I thought we had this, certainly important.

PR #2508, opened 2026-05-05
Add case file JSON schema and Python validator utility

### LINKED ISSUE #1203: JSON Schema for case files
Utilize the JSON Schema system to define our case file structure.
This should allow us to supply a specification of our json files with descriptions of each entry and specify required entries in a standardized way that others could utilize.

See more:
[https://json-schema.org/learn/getting-started-step-by-step](https://json-schema.org/learn/getting-started-step-by-step)

timofeymukha:
Adds a JSON schema for our case files. Written by AI of course, but with many iterations after the initial draft, and I manually checked more or less all entries where I was sure of the real config. All example cases pass now. 

I would like Arash to double-check ALE stuff, and  Niclas the preconditioner stuff.

Also adds a Python script to run a case (or many) through the schema. 

So I think this is pretty cool, but it will require diligence on our side to update this whenever we change JSON. It is not so difficult, and probably fully AI-automatable. But still, it is one more thing to track in each PR.


vbaconnet:
just mentioning the related issue #1203

vbaconnet:
This is really nice, thanks for putting in the work! Can/should the validator script be used in the github workflow to validate the examples?

timofeymukha:
> This is really nice, thanks for putting in the work! Can/should the validator script be used in the github workflow to validate the examples?

Thanks! That would be trivial to add. We had a discussion about this at the meeting. Those present were generally positive, but we do need more input from the developer team since it will increase upkeep work. So that includes you @vbaconnet , @adperezm , @Shiyu-Sandy-Du, @njansson and others!

doc/schemas/common.schema.json
```diff
@@ -0,0 +1,357 @@
...
+        },
+        "levels": {
+          "$ref": "#/$defs/positiveIntegerOrRef"
+        },
+        "cheby_degree": {
+          "$ref": "#/$defs/positiveIntegerOrRef"
+        }
+      },
+      "additionalProperties": false
+    },
```
njansson, line 212: cheby_degree I'm pretty sure it isn't used for hsmg, or maybe with the tamg coarse grid solver?

Also would be good if @pohm01 could check the phmg params.

njansson:
Maybe I'm blind, but I can't find the scheme for the linear solvers, only preconditioners

doc/schemas/common.schema.json
```diff
@@ -0,0 +1,357 @@
...
+          "$ref": "#/$defs/phmgPreconditioner"
+        }
+      ]
+    },
+    "linearSolver": {
+      "type": "object",
+      "required": [
+        "type"
+      ],
+      "properties": {
```
timofeymukha, line 321: @njansson Here is the linear solver one.

doc/schemas/common.schema.json
pohm01: `cheby_degree` is used with the tamg coarse grid solver (`levels` are also only used for tamg).

doc/schemas/common.schema.json
```diff
@@ -0,0 +1,357 @@
...
+          "const": "phmg"
+        },
+        "pcoarsening_schedule": {
+          "type": "array",
+          "items": {
+            "type": "integer",
+            "minimum": 1
+          },
+          "minItems": 1
+        },
```
pohm01, line 282: The `pcoarsening_schedule` does work with an empty array, in which case the phmg solver directly applies the coarse grid solver to the input. Unlikely to be practical for applications, but useful for debugging the coarse grid solver.

timofeymukha:
I think we are good to go

njansson (approved):
Great stuff! ☕

timfelle (approved):

PR #2509, opened 2026-05-06
Improve folder creation safety and functionality

timfelle:
Enhance folder creation by implementing a safer `mkdir` subroutine, addressing a bug that prevented output folders from being created if the path ended with a `/`, and ensuring consistent usage of the new function across relevant modules.

### DIFF
--- a/src/common/utils.f90
+++ b/src/common/utils.f90
@@ -34,6 +34,7 @@
 !! @details Various utility functions
 module utils
   use, intrinsic :: iso_fortran_env, only : error_unit, output_unit
+  use iso_c_binding
   implicit none
   private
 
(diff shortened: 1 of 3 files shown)


njansson (approved):

timofeymukha (approved):

src/common/utils.f90
```diff
@@ -148,6 +157,42 @@ subroutine filename_chsuffix(fname, new_fname, new_suffix)
...
+    slash_pos = scan(path, '/', back = .true.)
+    if (slash_pos .gt. 0) then
+       call mkdir(trim(path(1:slash_pos-1)), dir_mode)
+    end if
+
+    path_len = len_trim(path)
+    allocate(c_path(path_len + 1))
+    do i = 1, path_len
+       c_path(i) = path(i:i)
+    end do
```
timofeymukha, line 189: Out of curiosity, why i:i and not just i.

timfelle: Well to be honest, i did not try with just i 😅

PR #2513, opened 2026-05-07
Add Cai and Sagaut Model-II wall model

timofeymukha:
This adds an explicit algebraic wall model for smooth walls. Physically, it is essentially the same as Spalding's but with a somewhat better buffer layer overlap.

In this image, Spalding is dashed, and Mode-II is with white triangular symbols (overlaps with circles and some others)
image

The relevant paper is https://doi.org/10.1063/5.0048563

The main feature of the model is that it is explicit. You get the shear stress directly from the sampled quantities with no solver iterations. To make that work, it uses a clever trick for the log-law using the Lambert W function. 

Also fixes an indexing bug in wall_model_compute_mag_field! Please double-check that one.


vbaconnet (approved):
LGTM :) and I checked the indexing for the mask, it looks correct as well!

Shiyu-Sandy-Du (approved):

njansson:
@timofeymukha  Shouldn't we just merge then ;) ?

PR #2514, opened 2026-05-07
Reduce memory usage when reading nmsh zones and curves

njansson:


Mr-2c (approved):
I understand too little for my review of the code to really count, but it fixed the problem so I approve anyway 😺

timfelle (approved):

PR #2515, opened 2026-05-07
Modularize the entropy viscosity regularization

shiyu-du-dsy:
This PR separate entropy viscosity from the Euler solver by doing the following steps:

1. Add a new simcomp call `avm_simcomp_t` for the model for the artificial viscosity method (AVM), which handle the operation of the new type `avm_t` . Now `entropy_viscosity_t` is the only one derived type of `avm_t`. This object is responsible for compute the entropy viscosity to be used at each time step.
2. Add a new class `viscous_regularization_t` to apply the artificial viscosity to the equation. So far there is only one derived type `artificial_viscosity_t`, but lately I will add the spectral vanishing viscosity (SVV) as another child.
3. To apply the viscous regularization, one needs to have both AVM simcomp and viscous regularization as the json entries. Just see the updated cases.

To do list:

- [x] Documentation
- [x] Updated examples
- [x] Changelog

And regarding other regularization method such as GJP and HPF, since they do not act through the viscous term, we could keep them in the source term. (in the pnpn solver, viscous term is treated implicitly, therefore needs extra treatment)


src/simulation_components/avm_simcomp.f90
```diff
@@ -0,0 +1,139 @@
...
+  !> Constructor from json.
+  subroutine avm_simcomp_init_from_json(this, json, case)
+    class(avm_simcomp_t), intent(inout), target :: this
+    type(json_file), intent(inout) :: json
+    class(case_t), intent(inout), target :: case
+    character(len=:), allocatable :: name
+    character(len=:), allocatable :: model_name
+    character(len=:), allocatable :: reg_coeff
+    character(len=NEKO_VARNAME_LEN) :: fields(1)
+
```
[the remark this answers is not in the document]
Shiyu-Sandy-Du: @copilot apply changes based on [this feedback](https://github.com/ExtremeFLOW/neko/pull/2515#discussion_r3208734664)

src/fluid/fluid_scheme_compressible_ns.f90
```diff
@@ -571,46 +492,35 @@ subroutine fluid_scheme_compressible_euler_restart(this, chkp)
...
-         reg_json, this%c_Xh, this%dm_Xh, this%effective_visc)
-
-    select type (reg => this%regularization)
-    type is (entropy_viscosity_t)
-       call entropy_viscosity_set_fields(reg, this%S, this%u, this%v, this%w, &
-            this%h, this%max_wave_speed, this%msh, this%Xh, this%gs_Xh)
-    end select
+    logical :: found
+    character(len=:), allocatable :: viscous_regularization_type
+
```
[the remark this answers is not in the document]
Shiyu-Sandy-Du: @copilot apply changes based on [this feedback](https://github.com/ExtremeFLOW/neko/pull/2515#discussion_r3208734706)

src/simulation_components/avm_simcomp.f90
```diff
@@ -0,0 +1,139 @@
...
+    call this%free()
+
+    ! Add fields keyword to the json so that the field_writer picks it up.
+    ! Will also add fields to the registry if missing.
+    call json_get_or_default(json, "name", name, "avm_model")
+    call json_get_or_default(json, "reg_coeff", reg_coeff, "reg_coeff")
+    fields(1) = reg_coeff
+
+    call json%add("fields", fields)
```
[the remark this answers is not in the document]
Shiyu-Sandy-Du: @copilot apply changes based on [this feedback](https://github.com/ExtremeFLOW/neko/pull/2515#discussion_r3208734727)

timofeymukha:
Avm_model become artificial viscosity method model. I kinda feel like u should choose either model or method and just write ot out, for example, artificial_viscosity_method

src/viscous_regularization/avm_model/avm_model.f90
```diff
@@ -0,0 +1,212 @@
...
+
+
+contains
+  !> Constructor for the avm_model_t (base) class.
+  !! @param case The case_t object.
+  !! @param reg_coeff_name The name of the artificial viscosity field.
+  subroutine avm_model_init_base(this, case, reg_coeff_name)
+    class(avm_model_t), intent(inout) :: this
+    class(case_t), intent(inout), target :: case
+    character(len=*), intent(in) :: reg_coeff_name
```
timofeymukha: Please consider carefully whether you really need case_t here. By the looks of it, coef and dofmap suffice.

Shiyu-Sandy-Du: yes it's fixed now

shiyu-du-dsy:
I will rebase it after the compressible Navier-Stokes PR is merged :)

tuananhdao:
> I will rebase it after the compressible Navier-Stokes PR is merged :)

Thank you for that :) I hope it will not be a hassle.

shiyu-du-dsy:
Now it's fixed with the latest compressible ns push

doc/pages/user-guide/case-file.md
```diff
@@ -2041,3 +2024,28 @@ currently supports 50 regions, with id 1..25 being reserved for internal use.
...
+Note that some regularization techniques do not have a diffusive mathematical 
+form, for example gradient jump penalty and  high-pass filter relaxation term,
+and they are included in the source terms instead. The viscous regularization
+could be set up by the following options
+
+* `type`, viscous regularization type.
+  - `artificial_viscosity`, the standard second order diffusion term,
+    $\frac{\partial}{\partial x}\left(\mu_\mathrm{artificial}\frac{\partial u}{\partial x}\right)$.
+* `reg_coeff_name`, name of the $\mu_\mathrm{artificial}$ field, usually 
+computed by a simulation components.
```
timofeymukha: grammar

src/viscous_regularization/avm/avm_fctry.f90
```diff
@@ -0,0 +1,129 @@
...
+!
+submodule (artificial_viscosity_model) artificial_viscosity_model_fctry
+  use case, only : case_t
+  use entropy_viscosity, only : entropy_viscosity_t
+  use utils, only : neko_type_error, neko_type_registration_error
+  implicit none
+
+  ! List of all possible types created by the factory routine
+  character(len=20) :: AVM_KNOWN_TYPES(1) = [character(len=20) :: &
+       "entropy_viscosity"]
```
timofeymukha, line 41: You are preparing to drop a new method here soon, right?

src/math/bcknd/cpu/ax_helm_vector_cpu.f90
timofeymukha, line 3147: wow do we have an axhelm bug?

examples/euler_2d_smooth/euler_2d_smooth.case
```diff
@@ -13,20 +13,26 @@
...
-            "polynomial_order": 3,
-            "c_avisc_low": 0.5,
-            "c_avisc_entropy": 1.0
+            "polynomial_order": 3
         },
         "fluid": {
             "scheme": "compressible",
             "gamma": 1.4,
+            "viscous_regularization":{
+                "type": "artificial_viscosity",
```
timofeymukha, line 22: So there can be a viscous regularization that is not via an artificial viscosity? Just wondering if we need all three: the object, the coef name, and a type.

timofeymukha (commented):
I think is all fine basically, just some comments.

The axhelm edit makes me scared...

doc/pages/user-guide/case-file.md
timofeymukha: Particularly the "a simulation components". Either plural or the "a".

src/viscous_regularization/avm/avm_fctry.f90
Shiyu-Sandy-Du: yes

src/math/bcknd/cpu/ax_helm_vector_cpu.f90
Shiyu-Sandy-Du: we used to have it, but it's now fixed in develop as well (by the same commit

examples/euler_2d_smooth/euler_2d_smooth.case
Shiyu-Sandy-Du: we could have svv later by filtering the gradient term and then take the divergence

njansson:
> I think is all fine basically, just some comments.
> 
> The axhelm edit makes me scared...

It was me who forced pushed the buggy vector version a week or so ago

examples/euler_2d_smooth/euler_2d_smooth.case
Shiyu-Sandy-Du: I somehow don't like this nomenclature since SVV is also an artificiall viscosity in some sense... but it's better to follow the convention

Shiyu-Sandy-Du:
@copilot resolve the merge conflicts in this pull request

timofeymukha:
@shiyu-du-dsy  It seems Codex found some important stuff!

1. [P1] Entropy history is updated in the wrong order.
     entropy_viscosity_update_lag computes the new entropy before shifting the lag series. Consequently, at step 4 both S and S_lag(1) contain step 3, while step 0 is missing. The
     BDF3 residual is therefore incorrect. Shift the lag series first, then recompute S.

  2. [P1] Existing case files are silently accepted but lose stabilization.
     The solver no longer reads case.numerics.c_avisc_low or c_avisc_entropy, and it disables regularization when the new fluid object is absent (setup logic). However, the numerics
     schema still accepts both obsolete parameters. Existing validated cases will therefore run without artificial viscosity instead of failing or preserving old behavior. Either
     provide backward compatibility or reject the old configuration and mark the change as breaking. _Tim: let's reject_

  3. [P1] Restart does not reconstruct the three entropy history levels.
     avm_simcomp_restart calls compute once. At a restarted tstep > 3, only one lag entry is populated; the next preprocessing pass evaluates BDF3 using zero/uninitialised older
     entries. The lag fields need checkpoint support or a safe restart initialization such as filling all levels with the restarted entropy.

  4. [P2] Updated examples output an unrelated, empty S field.
     The model initializes its private entropy as entropy_viscosity_S without registering it (lines 138–140), while all five examples still request "S" (Sod example). field_writer
     silently registers a new zero-valued S, so the output looks valid but is wrong. Register the model entropy under the documented name or remove/update these output entries.

  5. [P2] The base destructor does not release all owned state.
     viscous_regularization_t owns reg_coeff, reg_coeff_name, coef, and dof, but free_base only nullifies the last two. It should also nullify reg_coeff and explicitly deallocate
     reg_coeff_name. Additionally, the local allocatable reg_coeff_name in entropy_viscosity_init is not manually deallocated as required by the repository guidelines.

  6. [P2] A non-compressible case leaves the model partially initialized and then dereferences null pointers.
     The select type has no default/error branch, yet compute_entropy is called immediately afterward. If this globally available simcomp is added to a PnPn case, p and rho remain
     null. Initialization should reject unsupported fluid schemes explicitly. 

  7. [P2] Documentation does not describe a working configuration.
     The compressible-flow example includes only viscous_regularization, omitting the AVM simcomp that computes its coefficient, so copied configurations get zero viscosity. The
     later statement that this works for scalar equations is also unsupported by this PR. The new section additionally lacks the mandatory header tag, and several new modules/types/
     procedures lack required Doxygen documentation. Modified 2025-licensed source files should include 2026 in their copyright range.

shiyu-du-dsy:
Note that in the current version, the restart will re-compute the entropy viscosity based on the latest state in 1st order. For a real clean restart, the p_lag and rho_lag should be included in the check point file as well. However noting that restart is not in effect in the compressible solver, I would like to keep the issue to a fixing PR for the restart of the compressible solver.

timofeymukha (approved):

tuananhdao (approved):
LGTM!

PR #2516, opened 2026-05-07
Some changes to the integration tests

vbaconnet:
Just some suggestions for the integration tests, these are just suggested improvements but I don't feel strongly about them :)

- Allows the user to specify a genmeshbox executable through the `GENMESHBOX_EXEC` env. variable. I've been experimenting with the integration tests with GPUs on LUMI but my contrib scripts are compiled in a separate installation folder, so I thought perhaps that could come in handy for other people.
- Instead of assuming the existence of the "neko", "makeneko", "genmeshbox" executables if they aren't specified as envirionment variables, we check their presence in `PATH` and throws an error if they aren't found.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -50,6 +50,7 @@ def pytest_addoption(parser):
         "--real_precision",
         action="store",
         default="dp",
+        choices=["dp", "sp"],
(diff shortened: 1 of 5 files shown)


timfelle (approved):
I do not work much with the python test interface, but what I can see here looks great.

timofeymukha (approved):

PR #2521, opened 2026-05-11
Fix configuration logic and bug in sx detection

timfelle:
Improve language switching logic in the configuration to enhance error reporting and clean up failed confidence tests. Fix a bug in the sx detection macro to ensure proper functionality.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4,11 +4,13 @@ AM_INIT_AUTOMAKE([foreign subdir-objects])
 AM_MAINTAINER_MODE
 AC_CONFIG_MACRO_DIR([m4])
 AC_CANONICAL_HOST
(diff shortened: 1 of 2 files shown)


njansson:
Nice catch with the sx backend. Out of curiosity, why did you need to switch to c for libtool init?

timfelle:
> Nice catch with the sx backend. Out of curiosity, why did you need to switch to c for libtool init?

If you look throungh the config.log file it contains a bunch of errors where the test programs are compiled with the wrong compiler language. Atleast on my local system. So C files being compiled with mpifort and such

njansson (approved):

timofeymukha (approved):

Issue #1273, opened 2024-05-07
Norms and residuals


MartinKarp:
The intent of this issue is to have somewhere to discuss what an appropriate norm and stopping criterion for our Krylov solvers would be. Upon request, I compile here a short summary of what systems we are solving and what is currently in the code.

For the residuals for the stopping criterion in our linear solvers, we do not solve for the full system at any point in time, but for the change in time, $\Delta x = x^i-x^{i-1}$ where $x$ is either the pressure or velocity and $i$ corresponds to the value at time step $i$.

For some linear system $Ax=b$ (Poisson equation for the pressure and Helmholtz for the velocity), we transform the system to the form
$$Ax^i = b^i => A\Delta x = b^i -A x^{i-1} $$
before it enters our Krylov solvers (usually GMRES for the pressure and CG for the velocity). 

This means the magnitude of the starting rhs $b^i -A x^{i-1}$ scales with the time step $\Delta t$ and the magnitude of $x$.

This discussion is created to assess what norm we would like to minimize when solving our system, and what criterion we should have for 
$$||A\Delta x - b^i +A x^{i-1} ||_x < tol.$$

The choice of our norm $||x||_x$ for this stopping criterion is important as we would like it to preferably be independent of the number of grid points, the length of the time step, the volume of the domain, and the normalization of the flow.

Currently, the following norms might be considered:

1. The L2 norm normalized with the volume of the domain $||x||_2/\sqrt{vol} = \sqrt{(x \cdot x)/vol}$ (this is what is used currently in the code).
2. The standard L2 norm $||x||_2$.
3. The integral (maybe the wrong word) over the domain divided by the volume $\sqrt{(x  \cdot B \cdot x)/vol}$ where $B$ is the mass matrix.

What has been noted in all these cases (except the integral) is that the norm scales with the number of grid points and for all of them also that the time step length plays an important role in choosing an appropriate tolerance. @timofeymukha had some thoughts on this and asked me to start a discussion about it here.

adperezm:
Good to bring it up.

I would say that 3rd option sounds right, at least for controlling the dependence with the number of gridpoints. My reasoning for liking that one is given below, but you do not need to read :)

I did not realize that $\Delta t$ also influenced it. But of course it does :). For that one I do not have opinions or suggestions.


#=============================== Why I like the 3rd ==============================================#

If one calls the residual: $r = A\Delta x - b^i +A x^{i-1} $

[...]

PR #2523, opened 2026-05-12
Residual selection playground

timofeymukha:
## What the PR draft does
This is a PR draft that can be used as a playground to test different residual formulations for Neko.
- The testing can be done with `cg` and `gmres` on the CPU.
- Both of these tap into a new type `ksp_residual_t`, which provides two run-time selectable procedure pointers:
  1. `compute_weight`, which provides a weight for the $L_2$ norm used to collapse the residual vector $r$ to a scalar $|r| = |b - Ax|$. Here and below I use single vertical lines to denote a norm because GH markdown doesn't render $||$ well.
  2. `compute_normalization`, which computes a scalar $S$, normalizing $||r||$. You can play with this by setting `residual_weight` and `residual_normalization` in the JSON for the solver.
- Solvers are also enhanced to accept `rel_tol` as an additional convergence criterion alongside `abs_tol`. To reduce the number of optional arguments, both `abs_tol` and `rel_tol` are made non-optional. Call sites can use default values already defined in the `ksp` module.
## Residual formulations in various codes
What surprised me somewhat is that there is no established way of defining $|r| / S$, and every code I looked at does things differently. Moreover, there tend to be multiple definitions within one code. However, it seems this template is generic enough for discussion:

$$ R = \frac{\|r\|_{W,p}}{S}, $$

where $W$ is a possible weight, $p$ is the type of norm, and $S$ is the normalization scale.
### Nek5000, NekRS, and currently Neko
For these codes the common template is

$$
R = \sqrt{\frac{r^T W r}{V}}.
$$

In other words, $p = 2$, and $S = V$, where $V$ is the volume of the computational domain.

- In Neko, there is no weighting, except for multiplicity.
- The same holds in NekRS.
- In Nek5000, the weight depends on the solver path. For pressure in $P_nP_{n-2}$ the weight is $B^{-1}$, i.e. the inverse of the mass matrix. No multiplicity is used. For $P_nP_n$ the pressure residual is computed like in Neko and NekRS. For a generic Helmholtz solve, it seems there is a parameter that controls whether you only use multiplicity or *both* multiplicity and $B^{-1}$. I don't really understand why these differences between solver paths are there; perhaps someone knows?

To me, using $p=2$ and multiplicity as weights feels very natural, and in all my experiments, I settled for exactly that as default.

Regarding the inverted mass matrix, I think the key here is that the residual vector is already volume-weighted. It is, perhaps, easiest to see for $b$, which we just explicitly multiply by $B$ in the code.

This means that $r^T r$ contains a factor of $B^2$. So, $B^{-1}$ returns the norm to a pure volume-weighted norm of some raw $\hat r$, not affected by volume scaling. This makes $R$ invariant to changes in volume, which I think is a very good property. For pseudo-2D and pseudo-1D in Neko, I would say it is a must, because right now, extruding along the dummy dimension changes the residuals.

This also means that weighting the norm with $B$ (no inverse) is not a good idea, because one effectively triple-accounts for the volume weight.

### Using the first $N$ residuals as scale
Finite-volume codes tend to use the assembled matrix entries in defining $|r|$, and sometimes also for $S$. But there is also another idea used by both Star-CCM+ and Fluent for some equations:

$$ S = \max_{j=1,\ldots,5} |r|^{(j)}, $$

where $j$ are iterations or time steps, depending on the solver mode. I suppose 5 is an arbitrary small number, and for those initial iterations you just don't apply any $S$.

This $S$ is convenient to monitor the drop from some initial state of the system. It should also be pretty resilient to scaling of the problem in various ways. However, I feel like this is a horrible $S$ for transient flows, because $S$ will be defined with respect to the large residual in the beginning and make the residual artificially small later on. However, this could me something to keep in mind for a future steady-state solver.

### Global / equation scaling
Both Fluent and OpenFOAM attempt to find a scale $S$ attached to the currently solved system of equations. Fluent uses the diagonal coefficients of the matrix to that end, which is quite natural in a finite-volume setting. OpenFOAM's approach is even more generic, and takes into account both $A$ and $b$:

$$ S = \sum_i \left(\left| (Ax)_i - (A \bar{x})_i \right| + \left| b_i - (A\bar{x})_i \right|\right), $$

where $x$ is the quantity solved for, and $\bar{x}$ is its average across the domain. Note that OpenFOAM uses the $L_1$ norm. Subtracting $\bar{x}$ is there to handle cases like the pressure equation, which is arbitrary up to a constant under some boundary conditions (no Dirichlet pressure boundaries).

It appears that Nektar++ does something similar and simply scales with $|b|$, but I am unsure about the details.

### Use inverse-volume scaling for weights
This seems like a very natural choice, whatever the normalization used, since it recovers a volume-weighted quantity. I think this can be applied independently of normalization, but it plays well with the existing $S = V$ to produce a volume-independent $R$.

### An equation-based $S$ for Neko
I propose that we adopt an equation-based scaling structurally similar to OpenFOAM's, but adapted to our choice of the norm. In particular, we can use the $L_2$ norm, with whatever choice of weighting, e.g. the inverse-volume.

This gives
$$S = \sqrt{|Ax - A \bar{x}|^{2} + | b - A\bar{x}|^2}, $$
and
$$R = |r| / S$$

Since this definition is resilient to various scalings, a given absolute tolerance setting is much more stable across cases. With what we have now, just scaling up the domain volume should prompt the user to adjust the absolute tolerance.

### Add support for both relative and absolute tolerance checks
We should check both the absolute and relative tolerances and allow either to signal convergence. This will guard against absolute tolerances that are unreasonable for a given precision, for example. For some solvers or problems, one type of tolerance may be preferable to the other.

## Important detail, we solve for $\Delta x$
In Neko, the system of equations is transformed:
$$Ax = b \rightarrow A(x - x_{ref}) = b - Ax_{ref} \rightarrow A\Delta x = \hat b.$$
What we currently pass to the KSP solvers is $\Delta x$ and $\hat b$, and we use $\Delta x = 0$ as the initial guess.

Using these quantities compromises the equation-scaling idea. Firstly, $A \Delta x$ is zero. So that makes $S = |\hat b|$. Of course, this also happens to be the value of the residual $|r_0| = |A\Delta x - \hat b|$. So this means that the initial $R$ is simply always equal to 1. Effectively, this makes the absolute tolerance a relative one.

An important related point is brought up by Marin in [Issue 1273](https://github.com/ExtremeFLOW/neko/issues/1273). The idea is that the smallest meaningful residual is bounded by $\epsilon |b|$, and *not* $|\epsilon \hat b|$, where $\epsilon$ is the machine precision. If we use proper equation-based scaling, we have the correct normalization (although we use $Ax$ as well, not only $b$). But if we use $\hat b$, the normalization is not good.

Therefore, I believe we should pass $x_{ref}$ to the solver and forward it to the normalization routine, so we can get a proper $S$.


GrahamBenHarper:
When I think of linear solver tolerances for an open source library, I think the tolerances should respect the variety of different applications that the library officially supports. That is, if somebody is running a turbulent problem on a highly deformed mesh, I would hope that Neko's settings out of the box would work for them. At the same time, if somebody is running a simple laminar problem, it would be great to not have to re-tune tolerances. When I look through the examples directory, I see tolerances in a wide range between 1e-3 and 1e-8, which feels like a pretty big range, and can be daunting for new users or less experienced people.

From a user interface and a mathematics standpoint, it would be great if there were a default setting that "just works" so users don't have to guess (and maybe less people will complain "why is Neko diverging?"), but it should also be as loose as possible so Neko "runs fast". I haven't been involved in Neko for a long time, so I don't want to impose my opinions too much, but I think some of the suggestions there at the end are ultimately what you want to aim for to effectively cover a wide variety of applications well. Inverse-volume scaling gives you the volume-independent R, which will help with distorted grids (and maybe ALE applications?). Combined with equation-based S would put you very close to computing the actual residual of the PDE solution, although I'm not sure how easy that equation-based S would be. If I'm not mistaken, I thought NEK5000 does something similar to this with the P021 and P022 parameters. See, e.g. https://lists.mcs.anl.gov/pipermail/nek5000-users/2010-November/001072.html

Anyway, I think the right idea is to design a system where one tolerance tends to work well for a variety of meshes and flow parameters, and I think inverse-volume and equation-based S move in that direction.

GrahamBenHarper:
And of course, the option to switch between different tolerance schemes (relative, absolute, weighted) would be nice for the "expert" users :wink:

timofeymukha:
I completely agree, @GrahamBenHarper !

timofeymukha:
Interesting Nek5000 link btw. The divergence-based residual for PnPn-2's pressure is still in the code in fact. But the `\int  h1^2 grad u . grad u + h2^2 u^2 dV / vol < tol = p22` for Helmholz seems to be gone from the code.

PR #2530, opened 2026-05-15
Spatial averaging simcomp

timofeymukha:
Adds a simcomp that only spatially averages a list of fields.

The simcomp itself is just a front for a new output_t type, which does the averaging at write time. This mirrors what we do for statistics. Note that we need temporaries for map_2d because it mutates fields that are input to it. This is because statistics fields are zero-ed out after write anyway, so we can get a way with that there. Here, we use the scratch registry to copy over fields before sending them to map_2d. For map_1d no copy is necessary. A test is added for the 1d output.


src/simulation_components/spatial_average.f90
```diff
@@ -0,0 +1,264 @@
...
+  type, extends(output_t) :: spatial_average_output_t
+     !> Pointers to the fields to average and write.
+     type(field_list_t) :: fields
+     !> Space averaging object for 2 homogeneous directions.
+     type(map_1d_t) :: map_1d
+     !> Space averaging object for 1 homogeneous direction.
+     type(map_2d_t) :: map_2d
+     !> The dimension of the output fields. Either 1 or 2.
+     integer :: output_dim = 0
+   contains
```
adperezm: If this is an output, could we not put it in a separate file in the IO part of the repo?

adperezm (commented):
Map 1d and 2d work for every type of mesh? If it is only fro extruded ones, then it should be mentioned somewhere. Apologies if it is.

src/simulation_components/spatial_average.f90
timofeymukha: yes, only extruded, good point on that and also the file location. The code is currently broken because of some pointer reallocation issues when the output and / or simcomp array grow. Hadn't had time to debug it, and the reason I didn't request a review yet.

timofeymukha: I'll convert to draft to not confuse people

src/io/spatial_average_output.f90
```diff
@@ -0,0 +1,166 @@
...
+    call this%fields%init(size(fields))
+    do i = 1, size(fields)
+       call this%fields%assign(i, neko_registry%get_field(trim(fields(i))))
+    end do
+
+    select case (trim(avg_direction))
+    case ('x', 'y', 'z')
+       this%output_dim = 2
+       output_filename = trim(output_filename) // '.fld'
+       call this%map_2d%init_char(coef, avg_direction, 1e-7_rp)
```
vbaconnet, line 101: I'm worried about the tolerance here... I haven't had any issues for my case in single precision but I think we should still be careful. This tolerance is used in the `relcmp` calls to check relative error between numbers, and if no tolerance is specified it uses the NEKO_EPS variable. Maybe it's fine to have it as is now, and submit a fix if we see that it fails for some cases.

src/sem/map_1d.f90
```diff
@@ -20,47 +53,87 @@ module map_1d
...
+  !! layout allow a unique stack of element levels in the requested direction.
+  !! If an element level remains unassigned, the resulting point levels are not
+  !! valid for the volume accumulation or averaging steps.
+  !! @warning The requested 1D mapping direction must not be periodic. The
+  !! propagation uses the mesh gather-scatter connectivity to order element
+  !! levels from the global coordinate minimum to the global coordinate maximum;
+  !! periodic connectivity in that same direction connects those two ends and
+  !! invalidates this ordering assumption. Periodicity in the directions being
+  !! averaged over is not the issue here, only periodicity in the remaining
+  !! profile direction.
```
vbaconnet: I find this paragraph a bit confusing. What is the issue?

vbaconnet (approved):
Cool stuff!

src/sem/map_1d.f90
timofeymukha: Yeah, spend a bit of time debugging the test I added, only to realize I was using a fully periodic mesh. This is not supported by the underlying algorithm, it seems. So if you for example average across xy, the flow cannot be periodic in z.

src/io/spatial_average_output.f90
timofeymukha: Hmm, ok, the same thing is used in all stats components. I just copy-pasted it here.

src/sem/map_1d.f90
vbaconnet: Is this because of the "dirty" GS_OP_MIN in map_1d?

src/io/spatial_average_output.f90
vbaconnet: yea, it's more of a general comment I think so we know/remember that it is hardcoded

src/sem/map_1d.f90
timofeymukha: Unclear. I am trying to investigate, maybe there is some bug in map_1d...

timofeymukha: @vbaconnet OK, map_1d is fine and I removed the warning. The issue with my test was that I used a 2x2x2 element mesh, which is apparently degenerate in the fully periodic case. There was some funny business with how dof ids got assigned, which led to gs_op_add giving an incorrect result.

src/io/spatial_average_output.f90
loretet: Dumb question, not sure I understand why 1e-7 is a problem here: 1e-7 is already the max possible precision for SP, right? Is there a clean way to not-hardcode it (eg using NEKO_EPS directly, or maybe NEKO_EPS is too low of a tolerance in DP)?

loretet (approved):
Very cool and useful, I don't have any specific comment. Looking forward to testing it

src/io/spatial_average_output.f90
timofeymukha: I didnt' think abou this much to be honest,  but 1e-7 seems too low for sp since epsilon is 1.19e-07 or so.

vbaconnet: yea it's just kind of on the edge and feels a bit risky, but it has been okay until now.

> Is there a clean way to not-hardcode it (eg using NEKO_EPS directly, or maybe NEKO_EPS is too low of a tolerance in DP)?

Yes, if we just remove the tolerance from the relcmp calls the correct NEKO_EPS (_SP or _DP) will be used instead.

PR #2536, opened 2026-05-18
Add host and device array implementations

timfelle:
Introduce host and device array types. Update the project structure to accommodate new data types in the scratch registries.

These types should allow us to request temporary objects on either GPU or CPU instead of always reserving memory on both.


njansson (approved):
🍻

adperezm (approved):
This is so good!

I think this is an amazing addition. Now we just need some math operations that allow to request explicitly where it should happen (host, or device haha :) )

One comment:
We use array in another context too, as I remember we have field_arrays and field_lists, right? Could we maybe name this in a different way?

Ignore if I am mistaken about the name. And also, I do not have better suggestions either.

PR #2537, opened 2026-05-18
Boundary profile t

arash-mp:
This PR adds the `boundary_profile` simulation component to extract and log instantaneous boundary data (pressure, wall shear stress, and pressure gradients) at GLL nodes of a specific zone directly to .csv files. Works also for ALE.

One can also add `time_average` or `spanwise_average` modes, so averaging happens on the fly. But currently, it doesn't do this.
I added a python script in `cylinder` example which does time and spanwise averaging, so one can use that instead. 

Things to be done:
- [ ] Changing the name of simcomp
- [ ] Using other simcomps for some operations
- [ ] Code style
- [ ] More output formats
- [ ] etc


adperezm:
Hi, Before checking this: isn't this the same that @timofeymukha added? Or is there something extra? 

By title alone seems very similar.

arash-mp:
> Hi, Before checking this: isn't this the same that @timofeymukha added? Or is there something extra?
> 
> By title alone seems very similar.

I don't think so.
`boundary_operation` does integral, averaging etc, and `boundary_flux` computes the flux through boundries.
This one saves the values on all GLL points on boundary, so for example one can get wall shear stress, or dpdx directly from its output, without needing to put probes etc..
Unless I missed something in those :D

adperezm:
> > Hi, Before checking this: isn't this the same that @timofeymukha added? Or is there something extra?
> 
> > 
> 
> > By title alone seems very similar.
> 
> 
> 
> I don't think so.
> 
> `boundary_operation` does integral, averaging etc, and `boundary_flux` computes the flux through boundries.
> 
> This one saves the values on all GLL points on boundary, so for example one can get wall shear stress, or dpdx directly from its output, without needing to put probes etc..
> 
> Unless I missed something in those :D

Oh I see. Yeah, this one works on all GLL. I didn't see that, my bad.

In that case are you sure csv is the right file type here? That would require things to be in one rank.

You should consider building a fld file (would need to be 2d, so perhaps not the most straight forward) or at least hdf.

I haven't seen the code, maybe you have that. If not, Hdf would give you parallel capabilities. See how it is used to write probes. It's not difficult.

This is a very nice functionality to have, but writing in rank0 is maybe not robust in large cases.

Again, I have not checked the code, so ignore if this is already there.

arash-mp:
> > > Hi, Before checking this: isn't this the same that @timofeymukha added? Or is there something extra?
> > 
> > 
> > > 
> > 
> > 
> > > By title alone seems very similar.
> > 
> > 
> > I don't think so.
> > `boundary_operation` does integral, averaging etc, and `boundary_flux` computes the flux through boundries.
> > This one saves the values on all GLL points on boundary, so for example one can get wall shear stress, or dpdx directly from its output, without needing to put probes etc..
> > Unless I missed something in those :D
> 
> Oh I see. Yeah, this one works on all GLL. I didn't see that, my bad.
> 
> In that case are you sure csv is the right file type here? That would require things to be in one rank.
> 
> You should consider building a fld file (would need to be 2d, so perhaps not the most straight forward) or at least hdf.
> 
> I haven't seen the code, maybe you have that. If not, Hdf would give you parallel capabilities. See how it is used to write probes. It's not difficult.
> 
> This is a very nice functionality to have, but writing in rank0 is maybe not robust in large cases.
> 
> Again, I have not checked the code, so ignore if this is already there.


Yes I think I should add hdf. Ok I'll take a look at probes.
thankss
But at the same time, since it only writes surface points, it should work with csv for large cases as well (I think), unless one have VERY LARGE number of elements. hdf, however, is prefered for sure.

arash-mp:
another thing that can be added is a buffer so it collect the data and writes once in a while. This can also be done.

doc/pages/user-guide/simcomps.md
```diff
@@ -258,6 +259,49 @@ Optional fields for this simcomp are:
...
+
+Each point on the boundary is assigned a unique `gll_id`. This identifier acts as a key to map the GLL points defined in the `*_mesh.csv` file to their corresponding rows of data in the `*_data.csv` file.
+
+If the simulation is using the Arbitrary Lagrangian-Eulerian (ALE) framework, the updated mesh coordinates and unit normal vectors will be written to the `*_data.csv` file as the body moves. In this case, the data file header will automatically include these fields before the fluid variables: `gll_id,x,y,z,-nrm_x,-nrm_y,-nrm_z,...`
+
+Note that the unit normal vectors, written in `*_mesh.csv` or `*_data.csv` (in case of ALE), are unit normal vectors pointing **outward** from the **wall**. So they are pointing into the fluid.
+
+Mandatory fields for this simcomp are:
+- `zone_id`: The ID of the labelled boundary zone where profiles should be extracted.
```
adperezm, line 278: Why like this?

Could you not use some simcomp that first calculates some value and adds it to the registry and then query any field in the registry? Why specific options?

I understand this probably serves some use that you have now, but this seems oddly specific.

doc/pages/user-guide/simcomps.md
```diff
@@ -258,6 +259,49 @@ Optional fields for this simcomp are:
...
+
+~~~~~~~~~~~~~~~{.json}
+{
+  "type": "boundary_profile",
+  "name": "my_boundary_profile",
+  "zone_id": 1,
+  "zone_name": "profiler_wall",
+  "start_time": 5.0,
+  "format_long": true,
+  "variables": ["p", "tau_mag", "dpdx"],
```
adperezm, line 299: In other simcoms do we use "fields"? I do not remember but be sure to be consistent.

doc/pages/user-guide/simcomps.md
```diff
@@ -258,6 +259,49 @@ Optional fields for this simcomp are:
...
+Optional fields for this simcomp are:
+- `zone_name`: A descriptive name for the zone, defaults to `"profile_zone"`.
+- `start_time`: Physical simulation time to begin outputting the profiles. Defaults to `0.0`.
+- `format_long`: If set to `true`, the exported data fields are written with extended precision (`15` decimal places) instead of the default shorter format (`6` decimal places). The coordinates are always written with extended precision. Defaults to `false`.
+
+~~~~~~~~~~~~~~~{.json}
+{
+  "type": "boundary_profile",
+  "name": "my_boundary_profile",
+  "zone_id": 1,
```
adperezm, line 295: Do you need both zone name and zone id? what do they do?

src/simulation_components/boundary_profile.f90
```diff
@@ -0,0 +1,1037 @@
...
+                this%out_dpdz = .true.
+             case ('all')
+                this%out_p = .true.
+                this%out_tx = .true.
+                this%out_ty = .true.
+                this%out_tz = .true.
+                this%out_tmag = .true.
+                this%out_dpdx = .true.
+                this%out_dpdy = .true.
+                this%out_dpdz = .true.
```
adperezm, line 192: Again, is it not possible to just use whatever one has in the field registry? There are many output writers that follow that convention

src/simulation_components/boundary_profile.f90
```diff
@@ -0,0 +1,1037 @@
...
+     type(field_t), pointer :: u => null()
+     type(field_t), pointer :: v => null()
+     type(field_t), pointer :: w => null()
+     type(field_t), pointer :: p => null()
+     type(field_t), pointer :: mu => null()
+     type(coef_t), pointer :: coef => null()
+
+     type(dirichlet_t) :: bc
+     integer :: zone_id
+     character(len=60) :: zone_name
```
vbaconnet, line 75:
```suggestion
     character(len=NEKO_VARNAME_LEN) :: zone_name
```
You will need to add the `utils` module with this as well

src/simulation_components/boundary_profile.f90
```diff
@@ -0,0 +1,1037 @@
...
+               DEVICE_TO_HOST, .false.)
+          call device_memcpy(this%r3%x, this%r3%x_d, n_pts, &
+               DEVICE_TO_HOST, .false.)
+
+          call device_memcpy(this%n1%x, this%n1%x_d, n_pts, &
+               HOST_TO_DEVICE, .false.)
+          call device_memcpy(this%n2%x, this%n2%x_d, n_pts, &
+               HOST_TO_DEVICE, .false.)
+          call device_memcpy(this%n3%x, this%n3%x_d, n_pts, &
+               HOST_TO_DEVICE, .true.)
```
vbaconnet, line 382: Just a suggestion for clarity, if you want. 

```suggestion
          call this%r1%copy_from(DEVICE_TO_HOST, .false.)
          call this%r2%copy_from(DEVICE_TO_HOST, .false.)
          call this%r3%copy_from(DEVICE_TO_HOST, .false.)
          call this%n1%copy_from(DEVICE_TO_HOST, .false.)
          call this%n2%copy_from(DEVICE_TO_HOST, .false.)
          call this%n3%copy_from(DEVICE_TO_HOST, .true.)
```

src/simulation_components/boundary_profile.f90
```diff
@@ -0,0 +1,1037 @@
...
+     integer, allocatable :: gll_id(:)
+     type(vector_t) :: x_ref, y_ref, z_ref
+     type(vector_t) :: r1, r2, r3, n1, n2, n3
+
+     type(vector_t) :: p_work, mu_work
+     type(vector_t) :: dpdx_work, dpdy_work, dpdz_work
+     type(vector_t) :: s11msk, s22msk, s33msk, s12msk, s13msk, s23msk
+     type(vector_t) :: force1, force2, force3, force4, force5, force6
+
+     character(len=256) :: data_filename
```
vbaconnet, line 99:
```suggestion
     character(len=NEKO_FNAME_LEN) :: data_filename
```

src/simulation_components/boundary_profile.f90
```diff
@@ -0,0 +1,1037 @@
...
+  !> A simulation component for boundary profiles (pressure, wall shear stress, pressure gradient)
+  type, public, extends(simulation_component_t) :: boundary_profile_t
+     type(field_t), pointer :: u => null()
+     type(field_t), pointer :: v => null()
+     type(field_t), pointer :: w => null()
+     type(field_t), pointer :: p => null()
+     type(field_t), pointer :: mu => null()
+     type(coef_t), pointer :: coef => null()
+
+     type(dirichlet_t) :: bc
```
vbaconnet, line 73: I know this is what the `boundary_flux` simcomp does, so this is more of a general question, but why store a BC and not retrieve a pointer to the corresponding BC in `case%fluid%bcs_vel` and `case%fluid%bcs_prs`? The `bc_list_t` type has a `get_by_name` and `get_by_zone_index` implemented.

src/simulation_components/boundary_profile.f90
```diff
@@ -0,0 +1,1037 @@
...
+       this%z_ref%x = this%r3%x
+
+       if (NEKO_BCKND_DEVICE .eq. 1) then
+          call device_memcpy(this%x_ref%x, this%x_ref%x_d, n_pts, &
+               HOST_TO_DEVICE, .false.)
+          call device_memcpy(this%y_ref%x, this%y_ref%x_d, n_pts, &
+               HOST_TO_DEVICE, .false.)
+          call device_memcpy(this%z_ref%x, this%z_ref%x_d, n_pts, &
+               HOST_TO_DEVICE, .true.)
+       end if
```
vbaconnet, line 403: You could probably also shorten this if you wanted to. `vector_t` has the [assignment (=) operation overloaded](https://neko.cfd/docs/develop/d3/db2/vector_8f90_source.html#l00170). Then you also don't need the allocation of `this%x,y,z_ref` above. 

```suggestion
       this%x_ref = this%r1
       this%y_ref = this%r2
       this%z_ref = this%r3
```

src/simulation_components/boundary_profile.f90
```diff
@@ -0,0 +1,1037 @@
...
+         '("boundary_profile_",A,"_data.csv")') trim(this%name)
+
+    if (pe_rank == 0) then
+       open(newunit=io_unit, &
+            file=this%case%output_directory // trim(this%data_filename), &
+            status='replace', action='write')
+
+       header_str = "gll_id"
+       if (this%ale_enabled) header_str = &
+            trim(header_str) // ",x,y,z,-nrm_x,-nrm_y,-nrm_z"
```
vbaconnet, line 514: Why the `-` sign?

src/simulation_components/boundary_profile.f90
```diff
@@ -0,0 +1,1037 @@
...
+    real(kind=rp) :: dt
+
+    if (time%t < this%start_time) return
+
+    if (this%compute_controller%check(time)) then
+       dt = time%dt
+       if (dt > 0.0_rp) then
+          call this%write_profile(time)
+       end if
+       call this%compute_controller%register_execution()
```
vbaconnet, line 614: I don't think you need to touch the compute controller, this is handled by the simulation component base class. See here: https://neko.cfd/docs/develop/d9/d55/simulation__component_8f90_source.html#l00449

vbaconnet: Or maybe I misunderstood something :)

src/simulation_components/boundary_profile.f90
```diff
@@ -0,0 +1,1037 @@
...
+    character(len=64) :: path_str
+
+    call this%free()
+    call json_get_or_default(json, "name", name, "boundary_profile")
+    call this%init_base(json, case)
+
+    call json_get_or_default(json, 'fluid_name', fluid_name, 'fluid')
+    call json_get_or_lookup(json, 'zone_id', zone_id)
+    call json_get_or_default(json, 'zone_name', zone_name, 'profile_zone')
+    call json_get_or_default(json, 'start_time', start_time, 0.0_rp)
```
loretet, line 139: Maybe we can have `json_get_or_lookup_or_default` here? I guess one might want to define a common `start_time` in the `constants` registries for different components (e.g. after a transitory)

loretet (dismissed):

adperezm (approved):
I think this is a very nice functionality, and I think we should have it in, since you probably need it and I am in favor of having more in the code.

I have some comments:

1- Couldn't the calculations be made through another sim comp and then have this as an output module that *maybe* does span-wise averaging?. For example, the simcomp by @timofeymukha does add some of the things here. Sure, that one integrates, but before integrating, it needs all the values at the boundary too, all the normals, all the area masses. So you could just output from the vectors that are created there, and reuse for other things.

2 - The last comment I would have is the same as before, consider adding another form of output that is parallel.

I approve because this is a nice functionality, but I think some changes would be beneficial.

adperezm:
> But at the same time, since it only writes surface points, it should work with csv for large cases as well (I think),

We are overutilizing the csv format, which was also a temporary measure back in the day for probes. Using it makes it so that we need to do gathering operations every time, we need to have more memory in rank0, and we use only one rank for IO.

If the posibility to do things in parallel exist, then we should do it and future-prove the functionality, instead of leaving it in a state that might fail later. Better to pay the price of an extra day of development now, than later when you are running a big production case an suddenly this is your bottleneck.

src/simulation_components/boundary_profile.f90
```diff
@@ -0,0 +1,1037 @@
...
+
+  !> Compute boundary profile data.
+  !! @param this The boundary_profile_t object.
+  !! @param time Current simulation time state.
+  subroutine boundary_profile_compute(this, time)
+    class(boundary_profile_t), intent(inout) :: this
+    type(time_state_t), intent(in) :: time
+    real(kind=rp) :: dt
+
+    if (time%t < this%start_time) return
```
vbaconnet, line 607: I think we decided this would be the convention
```suggestion
    if (time%t .lt. this%start_time) return
```

src/simulation_components/boundary_profile.f90
```diff
@@ -0,0 +1,1037 @@
...
+
+    if (this%compute_controller%check(time)) then
+       dt = time%dt
+       if (dt > 0.0_rp) then
```
vbaconnet, line 611:
```suggestion
       if (dt .gt. 0.0_rp) then
```

doc/pages/user-guide/simcomps.md
arash-mp: which part is the problem? Which specific options are you refering to here?

doc/pages/user-guide/simcomps.md
arash-mp: i will remove zone_names. 
zone_id and name are both needed.

doc/pages/user-guide/simcomps.md
adperezm: My bad. The part that is:

`variables`: An array of strings specifying which variables to compute and output. Available options are:
  - `"p"`: Instantaneous static pressure.
  - `"tau_x"`, `"tau_y"`, `"tau_z"`: Cartesian components of the wall shear stress vector.
  - `"tau_mag"`: The magnitude of the wall shear stress.
  - `"dpdx"`, `"dpdy"`, `"dpdz"`: Cartesian components of the strong pressure gradient vector.
  - `"all"`: Shortcut to include all of the above.
  -

vbaconnet (commented):
Super nice feature, thanks for doing all of this work!

My comments are mostly related to readability of the code which are aesthetic suggestions but ultimately up to you to incorporate into the code or not.

__
Just one point that I wanted to expand on a bit:

Essentially this simcomp can be seen as a combination of 
 1. "compute shear stress on a boundary" (similar in essence to `force_torque` simcomp, right?), 
 2. "compute pressure gradient" (essentially `gradient` simcomp),
 3. "output vectors to csv", which can also be seen as a masked output of a field that doesn't preserve any element structure, contrary to the `field_writer` with the `point_zone` option.

If you decoupled those into separate simcomps, you could easily plug any field from the registry into the last one (for instance, fields from the statistics which do the time averaging for you directly). 

Moreover, the "shear stress on boundary" could add vectors to the registry which would be picked up by the output simcomp, or pipelined/reused somewhere else. It's definitely more work, but just a suggestion.

src/simulation_components/boundary_profile.f90
arash-mp: not all of the stuff here are calculated in other simcomps, as far as I know, for example wall shear stress. Part of each thing here may be calculated in other places, but I don't see what is the problem of calcualting what is needed, directrly where is needed! The only part which worths using from other simcomp is force for tau (from force_torque). I will check that.
normals and etc are also not calculated in other simcomps, other simcops also call them when they need it. 
But please let me know exactly what you mean. Maybe I didn't get your point here :)

adperezm: > Part of each thing here may be calculated in other places, but I don't see what is the problem of calcualting what is needed, directrly where is needed!

I approved the code, so I do not oppose this, otherwise I would not have aproved. 

The comments are because based on what I see here, you are adding something that is very specific to what you need, which is **not** wrong. But if you want others to potentially use it, then it is beneficial to see how it interacts with other parts of the code.

There are other simcomps that calculate many things on demand. If this is an important quantity, then you could consider to add the calculation of that directly into a registry in a simcomp. The output can be another simcomp.

That could allow **other** people to build from something that you wrote even if they do not need specifically this application.

timofeymukha (commented):
Commenting before reading the code ;-).

Looking at the simcomp description, my thinking is in the same direction of Adalberto and Victor. I think the central part of the simcomp is essentially sampling / outputting boundary data, which is incredibly useful. I would avoid the word "profile" because I immediately thought of a  1D profile, but maybe that is just me. So I think the best thing would be to have a boundary_data_writer type, which this simcomp just wraps. CSV is indeed a horrible format for this in general, but I'd be prepared to accept it as the first available option.

As a very rough sketch
```f90
type boundary_data_writer_simcomp_t
 type(boundary_data_writer_t) :: sampler
end type

type boundary_data_writer_t
contains
  procedure, deferred:: sample
end type

type, extends(boundary_data_writer_t) :: csv_boundary_data_writer_t
contains
  procedure :: sample
end type
```
Then we can later plugin other better formats, potentially supporting connectivity, e.g. vtk or similar, and h5.

My gut tells me that the p, grad p and tau stuff should be just stripped off here in favour of providing a list of fields and vectors in the JSON. To get the wall stress, I'd try to see if it is easy to make the force_torque simcomp register its output.

pschlatt1:
I would also avoid the word profile. A profile is, in my view, always 1D, and typically in the cross-flow direction (even though there is a spanwise profile). But at the wall, I don't think this is a profile.

arash-mp:
Thanks for all the comments.
I made the PR as draft. Feel free to add more comments/suggestions. I'll be working on it again in 2 weeks after conference+vacataion :)

adperezm:
@arash-mp , when you come back to this, if you keep the boundary data in vectors (look at Tim's boundary operations) I can help adding the hdf writer so you can do just the csv. 

I still think having this in would be very useful.

arash-mp:
> @arash-mp , when you come back to this, if you keep the boundary data in vectors (look at Tim's boundary operations) I can help adding the hdf writer so you can do just the csv. 
> 
> 
> 
> I still think having this in would be very useful.

Perfect thanks.
I'll rewrite the whole thing after i come back 👍

PR #2539, opened 2026-05-19
Add xp to cg_coupled's residual reduction

timofeymukha:
Mirror ordinary cg on the CPU in the coupled version in terms of handling residual accumulation in extended precision.


src/krylov/bcknd/cpu/cg_coupled.f90
```diff
@@ -308,18 +310,14 @@ function cg_cpld_solve(this, Ax, x, y, z, fx, fy, fz, &
...
             y%x(i,1,1,1) = y%x(i,1,1,1) + alpha * p2(i)
             z%x(i,1,1,1) = z%x(i,1,1,1) + alpha * p3(i)
-            r1(i) = r1(i) + alphm * w1(i)
-            r2(i) = r2(i) + alphm * w2(i)
-            r3(i) = r3(i) + alphm * w3(i)
-            tmp(i) = r1(i)**2 + r2(i)**2 + r3(i)**2
          end do
 
-         rtr = glsc2(tmp, coef%mult, n)
+         call second_cg_cpld_part(rtr, r1, r2, r3, coef%mult, &
```
njansson: Have you checked any compiler listings if it is better to fuse both loops?

timofeymukha: No, just copied over what was done for cg :-)

loretet:
Is it straightforward to extend this to GPU as well? For the `cg_coupled` and `fused_cg-coupled` device files

timofeymukha:
> Is it straightforward to extend this to GPU as well? For the `cg_coupled` and `fused_cg-coupled` device files

Not *as* straightforward, but definitely doable I guess. I am starting with the easy stuff :-)

src/krylov/bcknd/cpu/cg_coupled.f90
```diff
@@ -334,4 +332,28 @@ function cg_cpld_solve(this, Ax, x, y, z, fx, fy, fz, &
...
+    real(kind=xp) :: tmp
+    integer :: i, ierr
+
+    tmp = 0.0_xp
+    !$omp parallel do reduction(+:tmp)
+    do i = 1, n
+       r1(i) = r1(i) - alpha * w1(i)
+       r2(i) = r2(i) - alpha * w2(i)
+       r3(i) = r3(i) - alpha * w3(i)
+       tmp = tmp + (r1(i) * r1(i) + r2(i) * r2(i) + r3(i) * r3(i)) * mult(i)
```
adperezm: How does the casting work here? Does it automatically promote stuff? Would it make sense to cast everything explicitly to some other variable first and then do the operation?

src/krylov/bcknd/cpu/cg_coupled.f90
```diff
@@ -334,4 +332,28 @@ function cg_cpld_solve(this, Ax, x, y, z, fx, fy, fz, &
     ksp_results%converged = this%is_converged(iter, rnorm)
   end function cg_cpld_solve
 
+  !> Updates residual vectors and computes the norm in extended precision.
+  subroutine second_cg_cpld_part(rtr, r1, r2, r3, mult, w1, w2, w3, alpha, n)
```
adperezm: Just curiosity: what is with this name? :)

adperezm (approved):

src/krylov/bcknd/cpu/cg_coupled.f90
timofeymukha:
```
❯ ag "cg.f90:250:.*(optimized:|missed:)" src/krylov/bcknd/cpu/cg.f90.172t.vect
6576:krylov/bcknd/cpu/cg.f90:250:31: missed:   not vectorized: no vectype for stmt: _274 = MEM <real(kind=4)[0:D.7465]> [(real(kind=4)[0:D.6556] *)r$data_248][_273];

ag "cg_coupled\.f90:346:.*(optimized:|missed:)" src/krylov/bcknd/cpu/cg_coupled.f90.172t.vect
3024:krylov/bcknd/cpu/cg_coupled.f90:346:36: missed:   not vectorized: no vectype for stmt: _648 = MEM <real(kind=4)[0:D.8199]> [(real(kind=4)[0:D.6711] *)r1$data_493][_647];

```

So it seems this is not good in both solvers now. is this the right way to check @njansson ?

njansson: Easiest is to check with some compiler that produce decent listings, e.g. CCE via `-hlist=am` or Fujitsu.

I'll push a new version of `cg` and `gmres` based on some performance work on a64fx. But in this case, the interesting is to see what it does generate for that block of code, e.g. vectors or unrolled loops etc (easy with CCE). The complexity pre/post this pr is the same 2 x n iterations, the question is if the compiler can do things a bit better (e.g. keeping alpha in register if they are merged)

src/krylov/bcknd/cpu/cg_coupled.f90
timofeymukha: Added explicit casts

src/krylov/bcknd/cpu/cg_coupled.f90
timofeymukha: OK, extra routine removed and xp temporaries added.

njansson (approved):

PR #2541, opened 2026-05-20
Extended precision for reduction operation in device_math

timofeymukha:
Changes gl* kernelds with additive accumulation to use extended precision for the accumulating variables / buffers.

For CUDA and HIP, the kernels are templated on `T_acc`, which is set to `real_xp` in the launching function.

For OpenCL, I had to just `typedef double real_xp;` in math_kernel.cl and use real_xp directly.


njansson:
I would suggest that instead of using `real_xp` directly, we use an additional template variable in addition to `T`

timofeymukha:
> I would suggest that instead of using `real_xp` directly, we use an additional template variable in addition to `T`

Darn, that's what AI wanted to do, and I told it not to :-)). Alrighty!

njansson:
This will unfortunately break on apple silicon, where only `single` is supported. 

How about we ifdef that typedef and retain `real_xp` as `real` if we are on macOS ?

njansson:
> This will unfortunately break on apple silicon, where only `single` is supported.
> 
> How about we ifdef that typedef and retain `real_xp` as `real` if we are on macOS ?

something like 
```c
#ifdef __APPLE__
 typedef real real_xp;
#else
typedef double real_xp;
#endif 
```

njansson:
> > This will unfortunately break on apple silicon, where only `single` is supported.
> > How about we ifdef that typedef and retain `real_xp` as `real` if we are on macOS ?
> 
> something like
> 
> ```c
> #ifdef __APPLE__
>  typedef real real_xp;
> #else
> typedef double real_xp;
> #endif 
> ```

also we do have a header where this can be set in the device folder

src/math/bcknd/device/opencl/math_kernel.cl
```diff
@@ -34,6 +34,12 @@
  POSSIBILITY OF SUCH DAMAGE.
 */
 
+#ifdef __APPLE__
+typedef real real_xp;
+#else
+typedef double real_xp;
+#endif
```
timfelle: As @njansson mentioned, this should probably be in the
`src/device/device_config.h.in`, si´ce thats where `real_xp` is first defined.

timofeymukha: The thing is that we cannot include a header in the .cl. At least the AI is trying to convince me of that for seemingly valid reasons:

  1. math_kernel.cl is not compiled as a normal source file.
     It is first converted into an embedded C string by src/scripts/cltostring.sh:1, via the
     %.cl.h: %.cl rule in src/Makefile.am:641.
  2. That script does not preprocess includes.
     It just escapes the .cl text and writes a header string:

  /usr/bin/sed ... $1 >> $1.h

  So #include "device_config.h" would remain literal text in the kernel source string.

The problem is that as usual with OpenCL I am operating blindly, so it is hard for me to confirm ^^.

njansson: Both me and the AI is a bit confused. 

For CL you need to update the cltostring.sh script to also sed real_xp to @NEKO_DEV_XREAL_TYPE@

timfelle: I have added this. Can you please have a look if it is now correct @njansson. Please resolve the conversation if so.

timofeymukha: I switched the order of substitution to real_xp first in the .sh script and removed the apple guard from the cl file.

timfelle (approved):

src/math/bcknd/device/opencl/math_kernel.cl
timofeymukha: @njansson happy times?

src/math/bcknd/device/opencl/math.c
```diff
@@ -1430,34 +1445,39 @@ real opencl_glsum(void *a, int *n, cl_command_queue cmd_queue) {
   const size_t global_item_size = 256 * nb;
   const size_t local_item_size = 256;
```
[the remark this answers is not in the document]
timfelle: @timofeymukha There is still this which warrant a look.

src/math/bcknd/device/opencl/math_kernel.cl
njansson: I haven't had time to try it, but looking at the changes, I don't see how this would work on macOS. The typedef will not solve it since the replacement script will still resolve to double

src/math/bcknd/device/opencl/math.c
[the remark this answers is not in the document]
timofeymukha: I think we have this already. The line number are a bit confusing, but both glsum and glsubnorm2 have the guard

src/math/bcknd/device/opencl/math_kernel.cl
timofeymukha: But should one not compile with ssp on mac os then, so that the real_xp is single?

timofeymukha: @njansson How should we proceed here. Should I revert the PR to GPU only?

njansson: GPU only still covers the OpenCL part. Hold a bit, and I'll give it a try

njansson: Something is wrong with OpenCL. I'm looking into it

vbaconnet:
Thank you for doing this important work! I have two comments/questions:
- Why only device_math and not host math? :D 
- Would there be instances where we would not want to truncate the reduction from `xp` to `rp`? In other words t keep the reduction result in `xp`? It depends on the use case I guess.

timofeymukha:
> Thank you for doing this important work! I have two comments/questions:
> 
> * Why only device_math and not host math? :D
> * Would there be instances where we would not want to truncate the reduction from `xp` to `rp`? In other words, to keep the reduction result in `xp`? It depends on the use case I guess.

On the host, it is already implemented! Not 100% sure about the second point, but typically I would say that if you will use the values in rp computations down the line, there is not a lot of point in retaining one scalar in xp. But there may be exceptions, I guess.

vbaconnet:
> On the host, it is already implemented! 

Indeed, my bad..

src/math/bcknd/device/opencl/math_kernel.cl
njansson: OpenCL does work, the issue was the assumption that `--enable-real=sp` should work on macOS, which it doesn't since the gpus doesn't support fp64 (it does work with `--enable-real=ssp` or the native metal backend #2595, so we can merge this and move on)

src/math/bcknd/device/cuda/math_kernel.h
```diff
@@ -969,10 +971,10 @@ __global__ void glsc3_reduce_kernel( T * bufred,
 
 
 /**
- * Device kernel for glsc3
+ * Device kernel for vlsc3
  */
 template< typename T >
-__global__ void glsc3_kernel(const T * a,
+__global__ void vlsc3_kernel(const T * a,
```
njansson, line 982: vlsc?

njansson (approved):

PR #2543, opened 2026-05-20
feat: Wave-dominated compressible Navier-Stokes

tuananhdao:
## Changes

This PR enables user to declare physical dynamic viscosity mu and thermal conductivity kappa for wave-dominated problems.

It also adds an example: Reflected Shock-Boundary Layer Interaction in a Shock Tube that illustrates the new feature.

<p>
  cns_2d_shock_tube_S02_fixed_1p5x_cropped
  cns_2d_shock_tube_S02_fixed_1p5x_bottom_right_370h
</p>

## Use how?

In the user file

```fortran
  subroutine user_setup(user)
    type(user_t), intent(inout) :: user
    user%material_properties => material_properties
  end subroutine user_setup
```

Then declare

```fortran
  subroutine material_properties(scheme_name, properties, time)
    character(len=*), intent(in) :: scheme_name
    type(field_list_t), intent(inout) :: properties
    type(time_state_t), intent(in) :: time

    type(field_t), pointer :: mu, kappa
    integer :: i

    if (scheme_name .eq. "fluid") then
       mu => properties%get_by_name("fluid_mu")
       kappa => properties%get_by_name("fluid_kappa")

       do i = 1, mu%dof%size()
          mu%x(i, 1, 1, 1) = ...
          kappa%x(i, 1, 1, 1) = ...
       end do
    end if
  end subroutine material_properties
```

## A bit more info

Implemented and tested backends (by eyes):

- [x] CPU
- [x] OpenCL
- [x] HIP
- [x] CUDA

Some kernel optimizations are added.

Note: Full support for arbitrarily large physical viscosities will be enabled by IMEX schemes in a future PR.


.codex_build/cpu_install/include/neko.h
njansson, line 1: What is the purpose of this file? It looks like a copy of the header from the api installation directory?

.codex_build/opencl_install/include/neko.h
njansson, line 1: What is the purpose of this file? It looks like a copy of the header from the api installation directory?

.codex_build/cpu_install/include/neko.h
tuananhdao: It will be cleaned up.

.codex_build/opencl_install/include/neko.h
tuananhdao: It will be cleaned up.

src/mesh/mesh.f90
```diff
@@ -985,24 +991,22 @@ subroutine mesh_generate_edge_conn(this)
...
          MPI_INTEGER, MPI_SUM, NEKO_COMM, ierr)
     edge_offset = edge_offset + 1
 
     ! Construct global numbering of locally owned edges
-    ns_id => non_shared_edges%array()
-    do i = 1, non_shared_edges%size()
-       call this%ddata%set_local_to_global_edge(ns_id(i), edge_offset)
+    do i = 1, num_edge_loc
+       call this%ddata%set_local_to_global_edge(non_shared_edges(i), edge_offset)
        edge_offset = edge_offset + 1
```
njansson: What are these changes address?

Now will have n_edges more htable lookups instead of directly accessing the edge global_id from the iterator.

njansson (commented):
Nice!

src/mesh/mesh.f90
tuananhdao: Good catch. I only wanted to change how `non_shared_edges` is stored. The extra lookup was not needed, since the iterator already gives us the edge id. I changed it back to `it%data(id)` and kept the id check.

tuananhdao: Good catch. I only wanted to improve how non_shared_edges is stored. The htable lookup was unnecessary. But I reverted changes to mesh.f90 for better separation of concern.

doc/pages/user-guide/case-file.md
```diff
@@ -313,6 +313,15 @@ The compressible solver uses variable time-stepping controlled by the CFL
 number. Set `variable_timestep` to `true` and specify `target_cfl` in the time
 control object.
 
+Physical viscosity can be specified via the `material_properties` user interface
+in the user file. Stabilization uses the existing Laplacian artificial
```
timofeymukha: Question: Why just via the user interface? Would it not be possible to specify in JSON as well like for incompressible, alongside some law like Sutherlands?

src/fluid/bcknd/device/compressible_res_device.F90
```diff
@@ -565,16 +616,25 @@ subroutine evaluate_rhs_device(rhs_rho_field, rhs_m_x, rhs_m_y, &
...
-    call device_copy(coef%h1_d, effective_visc%x_d, n)
-
-    ! Calculate artificial diffusion with variable viscosity
+    ! Density is stabilized only by artificial viscosity.
+    call device_copy(coef%h1_d, artificial_visc%x_d, n)
     call Ax%compute(visc_rho%x, rho_field%x, coef, p%msh, p%Xh)
+
+    ! Momentum uses the same artificial viscosity as monolithic mode.
+    call device_copy(coef%h1_d, artificial_visc%x_d, n)
     call Ax%compute(visc_m_x%x, m_x%x, coef, p%msh, p%Xh)
```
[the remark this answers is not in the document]
tuananhdao: fixed

doc/pages/user-guide/case-file.md
tuananhdao: Agreed. Constant mu and kappa should be supported in the json format. Commit coming.

timofeymukha: Cool. I guess at this point the scheme is not euler anymore, right? It is basically a compressible navier-stokes that can be run in euler mode :-)

tuananhdao: Yes, exactly. I still want to keep the Euler structure since it's very extendable to MHD (and several other useful models).

njansson (approved):

doc/pages/user-guide/case-file.md
njansson: How about renaming it? `Euler` might be a bit misleading in a couple of years ;)

timfelle (approved):

PR #2548, opened 2026-05-21
Fix/stats file counter

### LINKED ISSUE #2546: Inconsistent behaviour of scalar stats during restart
After a restart in a case with statistics simcomps active, the stats in the first run are written in files ending with 0.csv, and after the restart there are new files written ending with 1.csv. This is currently not working for the scalar stats; it continues writing in the -0.csv file (running with HIP, I haven't checked the other backends). This should be an easy fix and I can do it but I probably will not have time until next week so I wanted to flag it :)

Image

vbaconnet:
Fixes #2546

Uses the start_counter to build the stats file name instead of the counter to ensure there is no "-1" in the name, since it is constructed before any call to write().

Alsso changes a bit the user stats so that all the stats behave in the exact same way. Whether it is the "right" way is out of scope here :) 

@linneahuusko I would appreciate if you are able to test this on a small case of yours that uses a production-like setup with scalars etc to see things actually behave the same. I tested a bit on my side but not to a great extent!

______________________________
If people are interested to read about what has been said about the statistics output strategy, I can refer to the following issue https://github.com/ExtremeFLOW/neko/issues/1696

### DIFF
--- a/src/simulation_components/scalar_stats_simcomp.f90
+++ b/src/simulation_components/scalar_stats_simcomp.f90
@@ -228,7 +228,8 @@ contains
     if (t .gt. this%time) this%time = t
     if (this%default_fname) then
        fname = this%stats_output%file_%get_base_fname()
-       write (prefix, '(I5)') this%stats_output%file_%get_counter()
+       write (prefix, '(I5)') &
+            this%stats_output%file_%file_type%get_start_counter()
        call filename_suffix(fname, suffix)
        last_slash_pos = &
             filename_tslash_pos(fname)
--- a/src/simulation_components/fluid_sgs_stats_simcomp.f90
+++ b/src/simulation_components/fluid_sgs_stats_simcomp.f90
@@ -202,7 +202,8 @@ contains
     if (t .gt. this%time) this%time = t
     if (this%default_fname) then
        fname = this%stats_output%file_%get_base_fname()
-       write (prefix, '(I5)') this%stats_output%file_%get_counter()
+       write (prefix, '(I5)') &
+            this%stats_output%file_%file_type%get_start_counter()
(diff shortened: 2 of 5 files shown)


timfelle (approved):

src/simulation_components/scalar_stats_simcomp.f90
```diff
@@ -228,7 +228,8 @@ subroutine scalar_stats_simcomp_restart(this, time)
     if (t .gt. this%time) this%time = t
     if (this%default_fname) then
```
linneahuusko, line 229: This seems to become a problem if the scalar is called anything other than `"s"`, so in my ABL simulation it keeps writing to `scalar_stats_temperature0.csv` since I call the scalar field `"temperature"`. If I add a different scalar called `"s"` the numbering works.

vbaconnet: what is "works" for you? I think right now the numbering stuff is very `fld` format-centered, but for csv you want everything dumped in the same file right?

linneahuusko: Right, for csv it would be easier if everything is dumped in the same file, yes. What I meant by "works" here is that with scalar `"s"` the scalar statistics behave the same way as the other stats (`fluid`, `fluid_sgs`, and `scalar_sgs`). With `"temperature"` as the scalar it is inconsistent (i.e. the numbering works differently for different types of stats).

Regardless of which is preferable we should make sure it behaves the same for all types of stats :)

linneahuusko: The numbering of all stats now works consistently as long as the scalar has the default name. I will open a new PR to fix that!

linneahuusko (approved):

PR #2552, opened 2026-05-22
Update deprecation logging

timfelle:
- Log level is now set such that: 5: Log use of deprecated features.
- Version checking added. If version of deletion is reached, a warning is printed to stdout on usage to warn users that features might be removed with no further notice. 
- The runtime environment variable NEKO_DEPRECATION_ERROR changes the warnings of potential removals to errors.


njansson (approved):

src/common/log.f90
```diff
@@ -509,4 +535,56 @@ subroutine log_end_section_c() bind(c, name = "log_end_section")
...
+    removal_size = 1
+    do i = 1, len_trim(removal_str)
+       if (removal_str(i:i) .eq. '.') then
+          removal_str(i:i) = ' '
+          removal_size = removal_size + 1
+       end if
+    end do
+
+    ! allocate(current_number(current_size), source=0)
+    ! allocate(removal_number(removal_size), source=0)
```
vbaconnet: remove this?
```suggestion
```

src/common/log.f90
```diff
@@ -509,4 +535,56 @@ subroutine log_end_section_c() bind(c, name = "log_end_section")
...
+       call neko_error('Invalid version string in deprecation check: ' // &
+            'NEKO_VERSION=' // trim(NEKO_VERSION) // ', ' // &
+            'removal_version=' // trim(version_removal))
+    end if
+
+    is_deprecated = .true.
+    do i = 1, current_size
+       is_deprecated = current_number(i) .ge. removal_number(i)
+       if (.not. is_deprecated) exit
+    end do
```
vbaconnet, line 590: Will this not fail if for example current_str = "2.0.0" and removal_str = "1.9.9"?

2 >= 1 -> is_deprecated = .true. -> continues the loop
0 <= 9 -> is_deprecated = .false. -> returns False, even though 2.0.0 >= 1.9.9

vbaconnet (changes requested):
I think there may be a bug in the is_deprecated function but I could be wrong.

src/common/log.f90
timfelle: Correct.

timfelle: Correct

vbaconnet (approved):

PR #2549, opened 2026-05-21
Rework Krylov solvers and Jacobi preconditioner for CPU

njansson:
This PR introduces performance improvements and parallelisation enhancements to the Krylov solvers (CPU backend) and the Jacobi preconditioner. The main focus is on optimising key computational kernels with OpenMP parallelism and SIMD vectorisation, as well as improving the efficiency of reductions and orthogonalisation in GMRES.

(_yes this does revert some of the blocked loop to an earlier verbose version for better SIMD vectorisation_) 

see #2335

Issue #2335, opened 2026-02-26
Hybrid MPI+Multithreading


njansson:
On machines with limited memory per node (e.g. Fugaku), hybrid MPI + Multithreading solves the growing memory demands from more MPI end-points.

A intermediate steps towards a fully hybrid CPU version is to use auto parallelisation + threaded do concurrent.

PR #2553, opened 2026-05-26
Rework critical loops for compressible solver on CPU

tuananhdao:
Same spirit as https://github.com/ExtremeFLOW/neko/pull/2549 but for the compressible solver.

On a MacBook M4 with GNU Fortran 14.2.0, 1 MPI rank and 8 OpenMP threads, this PR improves `Fluid compressible` profiler region median runtime by 1.392x on euler_tgv and 1.410x on euler_2d_smooth. Smoke runs produced only round-off errors to final field files compared with `develop`.

### CPU OpenMP benchmark

Build/run setup:
- Platform: MacBook M4
- Compiler: GNU Fortran 14.2.0
- Configure: `--enable-openmp`
- Flags: `-O3 -std=f2008 -march=native`
- Run: `mpirun -n 1`, `OMP_NUM_THREADS=8`
- Cases: 50 fixed timesteps, runtime statistics enabled, field output disabled
- Metric: median of 5 runs, using the `Fluid compressible` profiler region

| Case | Branch | `Fluid compressible` median (s) | Speedup |
|---|---:|---:|---:|
| `euler_tgv` | `develop` | 9.413010 | |
| `euler_tgv` | PR | 6.764413 | 1.392x |
| `euler_2d_smooth` | `develop` | 3.445488 | |
| `euler_2d_smooth` | PR | 2.442870 | 1.410x |


njansson:
Have you compared against non-blocked versions? My guess would be that they can generate better SIMD code than the blocked one.

*Update* In addition to non-blocked versions, it is also important to use the directive sallad described in #2335, otherwise one will most likely loose the SIMD/SVE code from `do concurrent`

tuananhdao:
> Have you compared against non-blocked versions? My guess would be that they can generate better SIMD code than the blocked one.
> 
> _Update_ In addition to non-blocked versions, it is also important to use the directive sallad described in #2335, otherwise one will most likely loose the SIMD/SVE code from `do concurrent`

Good point. I compared the blocked version with a simpler unblocked `parallel do simd` version for these plain array loops. On my benchmark setup, the unblocked version was slightly faster, so I switched those simple loops to that form.

I also added the directive stack from #2335 before the OpenMP/SIMD loops.

njansson:
> > Have you compared against non-blocked versions? My guess would be that they can generate better SIMD code than the blocked one.
> > _Update_ In addition to non-blocked versions, it is also important to use the directive sallad described in #2335, otherwise one will most likely loose the SIMD/SVE code from `do concurrent`
> 
> Good point. I compared the blocked version with a simpler unblocked `parallel do simd` version for these plain array loops. On my benchmark setup, the unblocked version was slightly faster, so I switched those simple loops to that form.
> 
> I also added the directive stack from #2335 before the OpenMP/SIMD loops.

Great! 🍻 
 
Sorry for be that person, but I would suggest that in `euler_res` keep the loops in the main body instead of separate functions. Main reason is that we can then open a larger parallel region containing each sub iteration, and avoid e.g. `s` fork/join of the threads (which is not for free, especially on fugaku).

Ps. The goal is to rewrite gather-scatter (in-progress) such that it can be called from within a parallel region. So when that is fixed, it should be possible to more or less have one large region for the entire residual calculation.

tuananhdao:
> > > Have you compared against non-blocked versions? My guess would be that they can generate better SIMD code than the blocked one.
> > > _Update_ In addition to non-blocked versions, it is also important to use the directive sallad described in #2335, otherwise one will most likely loose the SIMD/SVE code from `do concurrent`
> > 
> > 
> > Good point. I compared the blocked version with a simpler unblocked `parallel do simd` version for these plain array loops. On my benchmark setup, the unblocked version was slightly faster, so I switched those simple loops to that form.
> > I also added the directive stack from #2335 before the OpenMP/SIMD loops.
> 
> Great! 🍻
> 
> Sorry for be that person, but I would suggest that in `euler_res` keep the loops in the main body instead of separate functions. Main reason is that we can then open a larger parallel region containing each sub iteration, and avoid e.g. `s` fork/join of the threads (which is not for free, especially on fugaku).
> 
> Ps. The goal is to rewrite gather-scatter (in-progress) such that it can be called from within a parallel region. So when that is fixed, it should be possible to more or less have one large region for the entire residual calculation.

Makes sense. I am moving the small euler_res loops back into the main RHS/RK routines.

njansson (approved):

timfelle (approved):

PR #2557, opened 2026-05-27
Fix OpenMP directives in projection

njansson:
This PR introduces OpenMP parallelisation to improve the performance of the projection routines. (see #2335 and #2404 )

### DIFF
--- a/src/common/projection.f90
+++ b/src/common/projection.f90
@@ -488,6 +488,7 @@ contains
       call rzero(alpha, this%m)
       this%proj_res = sqrt(glsc3(b, b, coef%mult, n) / coef%volume)
       this%proj_m = this%m
+      !$omp parallel do private(j, k, l, s) reduction(+:alpha)
       do i = 1, n, NEKO_BLK_SIZE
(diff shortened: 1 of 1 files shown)


src/common/projection.f90
```diff
@@ -498,13 +499,15 @@ subroutine cpu_project_on(this, b, coef, n)
...
       end do
+      !$omp end parallel do
 
       !First one outside loop to avoid zeroing xbar and bbar
       call MPI_Allreduce(MPI_IN_PLACE, alpha, this%m, &
            MPI_REAL_PRECISION, MPI_SUM, NEKO_COMM, ierr)
 
       call rzero(work, this%m)
 
+      !$omp parallel do private(j, k, l, s) reduction(+:work)
```
timfelle, line 510: should alpha not be private here?

njansson: Since we only read from it (unless I'm mistaken), it doesn't need to be private in this loop nest

timfelle (approved):

pohm01 (approved):

PR #2563, opened 2026-05-28
Deprecate implicit device usage in tensor operations

timfelle:
Deprecate implicit device usage in tensor operations and protect against zero-sized device tensors. Update the dofmap interface to eliminate dependency on interpolation, ensuring a more robust and clear implementation. Adjust related calls to reflect these changes.

### DIFF
--- a/src/fluid/bcknd/advection/adv_oifs.f90
+++ b/src/fluid/bcknd/advection/adv_oifs.f90
@@ -162,15 +162,27 @@ contains
     n_GL = nel*this%Xh_GL%lxyz
     n = nel*coef%Xh%lxyz
 
-    call this%GLL_to_GL%map(this%coef_GL%drdx, coef%drdx, nel, this%Xh_GL)
-    call this%GLL_to_GL%map(this%coef_GL%dsdx, coef%dsdx, nel, this%Xh_GL)
-    call this%GLL_to_GL%map(this%coef_GL%dtdx, coef%dtdx, nel, this%Xh_GL)
-    call this%GLL_to_GL%map(this%coef_GL%drdy, coef%drdy, nel, this%Xh_GL)
-    call this%GLL_to_GL%map(this%coef_GL%dsdy, coef%dsdy, nel, this%Xh_GL)
-    call this%GLL_to_GL%map(this%coef_GL%dtdy, coef%dtdy, nel, this%Xh_GL)
-    call this%GLL_to_GL%map(this%coef_GL%drdz, coef%drdz, nel, this%Xh_GL)
-    call this%GLL_to_GL%map(this%coef_GL%dsdz, coef%dsdz, nel, this%Xh_GL)
-    call this%GLL_to_GL%map(this%coef_GL%dtdz, coef%dtdz, nel, this%Xh_GL)
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call this%GLL_to_GL%map(this%coef_GL%drdx_d, coef%drdx_d, nel, this%Xh_GL)
+       call this%GLL_to_GL%map(this%coef_GL%dsdx_d, coef%dsdx_d, nel, this%Xh_GL)
+       call this%GLL_to_GL%map(this%coef_GL%dtdx_d, coef%dtdx_d, nel, this%Xh_GL)
+       call this%GLL_to_GL%map(this%coef_GL%drdy_d, coef%drdy_d, nel, this%Xh_GL)
+       call this%GLL_to_GL%map(this%coef_GL%dsdy_d, coef%dsdy_d, nel, this%Xh_GL)
+       call this%GLL_to_GL%map(this%coef_GL%dtdy_d, coef%dtdy_d, nel, this%Xh_GL)
+       call this%GLL_to_GL%map(this%coef_GL%drdz_d, coef%drdz_d, nel, this%Xh_GL)
+       call this%GLL_to_GL%map(this%coef_GL%dsdz_d, coef%dsdz_d, nel, this%Xh_GL)
+       call this%GLL_to_GL%map(this%coef_GL%dtdz_d, coef%dtdz_d, nel, this%Xh_GL)
+    else
+       call this%GLL_to_GL%map(this%coef_GL%drdx, coef%drdx, nel, this%Xh_GL)
+       call this%GLL_to_GL%map(this%coef_GL%dsdx, coef%dsdx, nel, this%Xh_GL)
+       call this%GLL_to_GL%map(this%coef_GL%dtdx, coef%dtdx, nel, this%Xh_GL)
+       call this%GLL_to_GL%map(this%coef_GL%drdy, coef%drdy, nel, this%Xh_GL)
+       call this%GLL_to_GL%map(this%coef_GL%dsdy, coef%dsdy, nel, this%Xh_GL)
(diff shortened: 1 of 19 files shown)


src/sem/interpolation.f90
```diff
@@ -167,12 +182,18 @@ end subroutine interpolator_free
...
+    integer, intent(in) :: nel
+    type(space_t), intent(in) :: to_space
+    real(kind=rp), intent(in) :: x(this%Xh%lx, this%Xh%ly, this%Xh%lz, nel)
+    real(kind=rp), intent(inout) :: y(this%Yh%lx, this%Yh%ly, this%Yh%lz, nel)
+
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call neko_log%deprecated('Interpolator: map implicit device', '2.0.0', &
+            'Please call the generic map or direct interfaces instead.')
+    end if
+
```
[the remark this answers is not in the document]
timfelle: The intention is to swap the generic mapping from pointing to map_old to
map_host when removal takes place.

src/fluid/bcknd/advection/adv_oifs.f90
```diff
@@ -246,9 +264,15 @@ subroutine adv_oifs_init(this, lxd, coef, ctarget, ulag, vlag, wlag, &
...
-    call this%GLL_to_GL%map(this%cz, this%wlag%lf(1)%x, nel, this%Xh_GL)
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call this%GLL_to_GL%map(this%cx_d, this%ulag%lf(1)%x_d, nel, this%Xh_GL)
+       call this%GLL_to_GL%map(this%cy_d, this%vlag%lf(1)%x_d, nel, this%Xh_GL)
+       call this%GLL_to_GL%map(this%cz_d, this%wlag%lf(1)%x_d, nel, this%Xh_GL)
+    else
+       call this%GLL_to_GL%map_host(this%cx, this%ulag%lf(1)%x, nel, this%Xh_GL)
+       call this%GLL_to_GL%map_host(this%cy, this%vlag%lf(1)%x, nel, this%Xh_GL)
+       call this%GLL_to_GL%map_host(this%cz, this%wlag%lf(1)%x, nel, this%Xh_GL)
+    end if
```
njansson, line 275: This is a good example of where I think the old way was actually good. It reduced code duplication, only one interface was necessary. 

If it's worth the potential errors, not sure.

timfelle: You can still get that by storing the arrays as a field, vector or something appropriate for the case.

src/sem/interpolation.f90
```diff
@@ -194,23 +214,74 @@ end subroutine interpolator_map
...
+       call tnsr3d_cpu(y, this%Xh%lx, x, this%Yh%lx, &
+            this%Yh_to_Xh, this%Yh_to_XhT, this%Yh_to_XhT, nel)
     else
        call neko_error('Invalid interpolation')
     end if
   end subroutine interpolator_map_host
 
+  !> Interpolates an array to one of Xh or Yh.
+  !! @param x Original array.
+  !! @param y Interpolated array.
```
vbaconnet:
```suggestion
  !> Interpolates a field to one of Xh or Yh.
  !! @param x Original field.
  !! @param y Interpolated field.
```

src/sem/interpolation.f90
```diff
@@ -60,23 +63,33 @@ module interpolation
...
 
      !> Constructor.
      procedure, pass(this) :: init => interpolator_init
      !> Destructor.
      procedure, pass(this) :: free => interpolator_free
-     !> Interpolate an array to one of Xh or Yh.
-     procedure, pass(this) :: map => interpolator_map
+
+     !> Generic interface to map an array to one of Xh or Yh.
+     generic :: map => map_old, map_device, map_field
```
vbaconnet, line 81: Was this intentional? Maybe I misunderstood something.

```suggestion
     generic :: map => map_old, map_device, map_field, map_host
```

vbaconnet (approved):

src/sem/interpolation.f90
timfelle: Yes, your suggestion would not work as `map_old` and `map_host` have the same input parameters.

timfelle: The intention is to swap out `map_old` with `map_host` when we no longer support the implicit device calls.

adperezm:
A question/comment before I check the code:

Would the idea from now on be that we always use our routines with our custom datatypes (field, vector, array) and in such way things work "automatically"?

I also do not necessarily like what tensor used to do, i.e., retrieve the device pointer, but it avoids a lot of ifelse blocks.

timfelle:
> A question/comment before I check the code:
> 
> Would the idea from now on be that we always use our routines with our custom datatypes (field, vector, array) and in such way things work "automatically"?
> 
> I also do not necessarily like what tensor used to do, i.e., retrieve the device pointer, but it avoids a lot of ifelse blocks.

That is what I am working toward. The ifelse block will be at some place. But we need to be strategic about how high up we want them. It should be high enough to ensure performance, as Niclas mention. But also not so high that every piece of code is littered with it.

One challenge is that we have so many different ways of passing down the same data.
```fortran
real(kind=rp) :: f(lx,ly,lz,ne)
real(kind=rp) :: f(lx,lx,lx,ne)
real(kind=rp) :: f(lx*lx*lx,ne)
real(kind=rp) :: f(lx*lx*lx*ne)
real(kind=rp) :: f(:)
```
All of these are seemingly used randomly and interchangeably. I know a bunch of places it is for performance reasons, but we should be a bit more consistent.

And I am deeply opposed to interfaces not doing what they claim to do:
```fortran
subroutine tnsr3d(v, nv, u, nu, A, Bt, Ct, nelv)
  integer, intent(in) :: nv, nu, nelv
  real(kind=rp), intent(inout) :: v(nv*nv*nv, nelv)
  real(kind=rp), intent(in) :: u(nu*nu*nu, nelv)
  real(kind=rp), intent(in) :: A(nv, nu), Bt(nu, nv), Ct(nu, nv)
```
No modifications are ever made to `v` if we build with a device. And I never asked for the device memory to be touched, yet it was done anyway.

src/sem/bcknd/dofmap_init.f90
```diff
@@ -0,0 +1,82 @@
...
+! BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+! LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+!
+!> Submodule containing the mapping based initialization of dofmaps
+submodule (dofmap) dofmap_init_m
+  use interpolation, only : interpolator_t
```
timofeymukha, line 35: If we are rolling with a _m, I'd just have `dofmap_init_and_map_m` :-).

timofeymukha (approved):

PR #2571, opened 2026-06-01
Low-Mach number fluid scheme (variable-density, ideal-gas EOS) [WIP]

adigh23:
## Summary

**Draft / work-in-progress.** Adds a low-Mach-number formulation of the Pn/Pn
fluid scheme for variable-density flows at negligible acoustic effects, coupling
the density to temperature through an ideal-gas equation of state.

Opening early to share the approach and gather feedback. The branch is based on
an older `develop` (merge base 2026-04-02, ~28 commits behind) and needs a
rebase, the remaining accelerator backends, and quantitative validation before
it is merge-ready — hence draft.

## What it adds

A new fluid scheme `lowmach` (`fluid_lowmach_pnpn`), registered through the fluid
factory, extending the incompressible Pn/Pn solver with:

- **Ideal-gas EOS** — `rho = P0 / (R_specific * T)`; the density field is updated
  each step from the scalar temperature (`P0`, `R_specific`, `T_eos_min/max`,
  `temperature_field` config keys).
- **Thermal divergence `Q_T`** in the pressure-Poisson residual — the velocity is
  no longer solenoidal; `div(u) = Q_T` is computed from the temperature field and
  fed into the pressure solve.
- **`lowmach_residual` module** (+ CPU backend `lowmach_res_cpu`, factory) that
  assembles the momentum and pressure residuals with **field-valued `rho` and
  `mu`** (spatially varying properties) instead of scalar constants.

### Files (relative to the merge base)
- `src/fluid/fluid_lowmach_pnpn.f90` — new scheme (EOS plumbing, `Q_T` from `T`,
  field-valued `rho` in the momentum RHS)
- `src/fluid/lowmach_residual.f90`, `src/fluid/lowmach_res_fctry.f90`,
  `src/fluid/bcknd/cpu/lowmach_res_cpu.f90` — residual module + CPU backend
- `src/fluid/fluid_base_fctry.f90`, `src/Makefile.am`, `src/.depends` — wiring
- `examples/rayleigh_benard/rayleigh_lowmach.{case,f90}`,
  `examples/tgv/tgv_lowmach_scalar.case` — validation/example cases

## TODO before merge-ready
- [ ] Rebase onto current `develop`
- [ ] GPU (CUDA/HIP/OpenCL) and SX backends for the low-Mach residual
- [ ] Quantitative validation (e.g. Rayleigh–Bénard Nu–Ra scaling) + mass/energy conservation diagnostics
- [ ] Unit/regression tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)


LOWMACH_DEV_GUIDE.md
```diff
@@ -0,0 +1,345 @@
...
+`ρ cp DT/Dt = ∇·(λ∇T)` to eliminate `DT/Dt`.)
+
+**Momentum** (full deviatoric stress, because `∇·u ≠ 0`):
+
+    ρ Du/Dt = −∇p + ∇·τ + f ,
+    τ = μ(∇u + ∇uᵀ) − (2/3) μ (∇·u) I
+
+**Energy** — solved by the Neko **scalar** named `temperature`:
+
+    ρ cp (∂T/∂t + u·∇T) = ∇·(λ∇T)
```
timofeymukha, line 45: I guess this is a design decision you want to think about. Whether the fluid type itself should handle the scalar solve or whether the user should setup the scalar for it. The latter is probably easier programming-wise1

LOWMACH_DEV_GUIDE.md
```diff
@@ -0,0 +1,345 @@
...
+                └── fluid_lowmach_pnpn_t (NEW; overrides init + step)
+```
+
+`fluid_lowmach_pnpn_t` (`fluid_lowmach_pnpn.f90`) reuses *everything* from
+`fluid_pnpn` (mesh, dofmap, gs, BCs, projection, the `Ax_vel`/`Ax_prs`
+operators, the krylov solvers, the AB/BDF time scheme) and adds:
+
+| member | meaning |
+|---|---|
+| `low_mach_enabled` | master switch; if false, `step` runs vanilla pnpn |
```
timofeymukha: Probably not needed outside the intial debugging phase.

LOWMACH_DEV_GUIDE.md
```diff
@@ -0,0 +1,345 @@
...
+    gs_op(Q, ADD); Q *= Binv       ! assemble -> strong
+    Q_T = Q / (ρ cp T)             ! divide by rho*cp*T
+
+The decisive detail: `λ` and `cp` are the scalar's **spatially varying fields**,
+resolved lazily from the registry as `"<temperature>_lambda_tot"` (fallback
+`"_lambda"`) and `"_cp"` — i.e. Nek5000's `vdiff`/`vtrans` of the temperature
+field. They are filled by the user `material_properties` hook (e.g. Sutherland
+`λ(T)`). Using a **constant** `k`/`cp` instead (the original code) makes `Q_T`
+inconsistent with the energy balance and was the cause of the heated-channel
+blow-up. If the scalar fields are absent, it falls back to `k_cond`/`cp`.
```
timofeymukha, line 143: Not sure I fully understand this, I guess we can discuss!

LOWMACH_DEV_GUIDE.md
```diff
@@ -0,0 +1,345 @@
...
+Possible future optimisation: replace the fixed-point outer loop with FGMRES
+(use `A_ref` as a flexible preconditioner) to cut the outer count for stiff
+density ratios; cache the `p`-independent part of the residual.
+
+--------------------------------------------------------------------------------
+## 8. Variable-density RHS makers
+
+`lm_rhs_ext_field_rho`, `lm_rhs_bdf_field_rho`, `lm_rhs_oifs_field_rho` are
+variable-ρ counterparts of `rhs_maker_*_cpu`: they assemble the AB-extrapolated
+forcing history and the BDF lagged-velocity term weighting **per node by ρ(x)**
```
timofeymukha: I don't think you need new types here. The rhs makers we have should just treat variable density properly. Density is always an array in the fluid anyway,

LOWMACH_DEV_GUIDE.md
```diff
@@ -0,0 +1,345 @@
...
+EOS `P0`, `R_specific` come from `case.fluid.low_mach`. Gravity/forcing come from
+a user `source_term`.
+
+--------------------------------------------------------------------------------
+## 11. Required configuration
+
+```jsonc
+"fluid": {
+  "scheme": "lowmach",
+  "full_stress_formulation": true,          // REQUIRED (coupled stress operator)
```
timofeymukha: I think you should just force this in the code, since it is required, there is no reason to let the user configure this.

LOWMACH_DEV_GUIDE.md
```diff
@@ -0,0 +1,345 @@
...
+8. optional `vol_flow%adjust` if `flow_rate_force` is set.
+
+Property/coupling timing: the scalar is solved after the fluid in the simulation
+loop, so at step *n* the fluid sees `T`, `λ`, `ρ` from step *n−1* (first-order
+explicit coupling — standard, fine).
+
+--------------------------------------------------------------------------------
+## 10. Material properties / coupling (user side)
+
+Variable `μ(T)`, `λ(T)`, `cp` are supplied through the Neko `material_properties`
```
timofeymukha: It would maybe be good to provide some good default options, i.e. Sutherland's, so that one does not always need to code this in the user file. @tuananhdao Does the compressible solver use these kind of relations? I am thinking that maybe some code can be reused across the two schemes here.

LOWMACH_DEV_GUIDE.md
```diff
@@ -0,0 +1,345 @@
...
+reduction) and the run blows up — *only* once density varies.
+
+Diagnosis was by elimination (all in `examples/heated_channel/`): incompressible
+`pnpn` on the same mesh/BCs converges; low-Mach stalls even at ΔT = 10 K and
+with `Q_T` off but variable ρ on; `phmg` also fails. ⇒ it is the variable-ρ
+operator, not BCs/mesh/`Q_T`.
+
+### The fix — defect correction
+Solve the true variable-ρ system `A_var p = b` by an outer iteration whose inner
+solve uses a **constant reference operator** `A_ref` (hsmg's home turf):
```
timofeymukha, line 185: Is this stuff something you invented, or does this follow the Nek5000 formulation? My thinking is that since we are basically porting the Nek5000 scheme here, we could/should, in the first instance, just match exactly what Nek5000 does and achieve the same solution in both codes for some validation flow problem. That would be a good milestone!

src/fluid/fluid_lowmach_pnpn.f90
```diff
@@ -0,0 +1,624 @@
...
+     !> Thermal divergence source field. Currently held at zero; subsequent
+     !! commits will populate it from the energy equation.
+     type(field_t) :: Q_T_field
+     !> Low-Mach pressure residual operator.
+     class(lowmach_prs_res_t), allocatable :: lm_prs_res
+     !> Low-Mach velocity residual operator.
+     class(lowmach_vel_res_t), allocatable :: lm_vel_res
+   contains
+     procedure, pass(this) :: init => fluid_lowmach_pnpn_init
+     procedure, pass(this) :: step => fluid_lowmach_pnpn_step
```
timofeymukha, line 130: You are clearly missing a `free` routine because you have allocatables and pointers inside the type.

src/fluid/fluid_lowmach_pnpn.f90
```diff
@@ -0,0 +1,624 @@
...
+
+    ! Weak divergence via cdtp, accumulating into Q_T.
+    call cdtp(this%Q_T_field%x, gx%x, this%c_Xh%drdx, this%c_Xh%dsdx, &
+         this%c_Xh%dtdx, this%c_Xh)
+    call cdtp(w1%x, gy%x, this%c_Xh%drdy, this%c_Xh%dsdy, &
+         this%c_Xh%dtdy, this%c_Xh)
+    do concurrent (i = 1:n)
+       this%Q_T_field%x(i,1,1,1) = this%Q_T_field%x(i,1,1,1) &
+            + w1%x(i,1,1,1)
+    end do
```
timofeymukha, line 495: Prefer calling `field_math.f90` / `math.f90` routines where possible, unless you are fusing a big kernel into one loop. This will make GPU porting easier later.

njansson: Since this is the CPU path, I would rather say keep it as is. And in the near future replace `do concurrent` with `do` + directives + `omp do` (see #2335)

PR #2577, opened 2026-06-08
Fix memory leaks and free missing MPI types

timfelle:
Address memory management issues by ensuring proper deallocation of components and adding missing MPI type frees. This improves resource handling and prevents potential memory leaks.


njansson:
Shouldn't there be a corresponding fix for hip and OpenCL as well?

njansson (approved):

timfelle:
> Shouldn't there be a corresponding fix for hip and OpenCL as well?

Agreed, It was trivial to do for HIP as expected. I'll do some searching on OpenCL.

timfelle:
> > Shouldn't there be a corresponding fix for hip and OpenCL as well?
> 
> Agreed, It was trivial to do for HIP as expected. I'll do some searching on OpenCL.

I think we are okay for OpenCL, we already explicitly release the device. I would be surprised if we can do more than that.

```fortran
    if (c_associated(glb_device_id)) then
       if (clReleaseDevice(glb_device_id) .ne. CL_SUCCESS) then
          call neko_error('Failed to release device')
       end if
    end if
```

njansson:
> > > Shouldn't there be a corresponding fix for hip and OpenCL as well?
> > 
> > 
> > Agreed, It was trivial to do for HIP as expected. I'll do some searching on OpenCL.
> 
> I think we are okay for OpenCL, we already explicitly release the device. I would be surprised if we can do more than that.
> 
> ```fortran
>     if (c_associated(glb_device_id)) then
>        if (clReleaseDevice(glb_device_id) .ne. CL_SUCCESS) then
>           call neko_error('Failed to release device')
>        end if
>     end if
> ```

True, should be fine for OpenCL then

timofeymukha (approved):

PR #2579, opened 2026-06-08
Compression for SEM mapping data

GrahamBenHarper:
This is not a final polished product, but this is a feature which is at the perfect stage for discussion. This PR implements an idea that I've put in a few software packages to great success (and tried to publish with great failure).

The main idea is as follows: the 4-dimensional SEM mapping tensors G11, G22, etc. (hereafter collectively Gij) store a lot of redundant information for certain classes of meshes. For example, on a perfectly structured mesh, the Gij array slices are identical along the slowest striding index e. Extruded meshes and many other meshes will also have some repeated structure where this would benefit. On very large meshes, these accesses into the full Gij arrays results in heavy memory paging and cache misses, producing significant memory overhead.

If you want an idea of how much this overhead costs in real simulations, take the poisson example and replace `G11(i,1,1,e)` with `G11(i,1,1,1)` and run it on a really large mesh. I see a **14% speedup** on a CPU-only run using the poisson example. Poisson example, tgv_16M.nmsh, 100 iterations, N=2, Neko v1.99.3 (only one run, so some noise is expected :) )
```
Tot MFlops =   4.6755E+02, MFlops      =   4.6755E+02
Setup Flop =   5.2345E+11, Solver Flop =   2.0334E+11
Start res  =   9.4389E+03, Final res   =   5.1621E+04
Solve Time =   0.1554E+04
```
Results with mappings "compressed" (all mapping data of the form G11(i,1,1,e) replaced with G11(i,1,1,1))
```
Tot MFlops =   5.4573E+02, MFlops      =   5.4573E+02
Setup Flop =   5.2345E+11, Solver Flop =   2.0334E+11
Start res  =   9.4389E+03, Final res   =   5.1621E+04
Solve Time =   0.1332E+04
```

How does this work? At the beginning of a simulation (before the main time-stepping loop), this computes a one-time "compression" of the Gij arrays, by identifying elements of the same shape. It then forms two objects:
1. Gij_compressed, which is the same as its corresponding Gij tensor, but all redundancies in the e dimension are removed
2. compression_inds, which is the random access lookup into Gij_compressed
In theory, you could throw away the original Gij arrays after this point if you want to free up application memory. This doesn't have any device syncing or backend kernels implemented, but I imagine a full feature would support that as well.

To use these in practice, the operator assembly loop is nearly identical, but instead of `c%G11(i,1,1,e)`, you use `c%G11_compressed(i,1,1,c%compression_inds(e))`. If you use `G11 => c%G11_compressed`, then it's even closer semantically. There's probably a smart Fortran-friendly way to do this in a way that's semantically the same without sacrificing performance, but I'm not a Fortran expert.

I'm tagging a few people blindly for feedback, ideas, etc., feel free to also tag others who are interested.

### DIFF
--- a/src/sem/coef.f90
+++ b/src/sem/coef.f90
@@ -74,6 +74,21 @@ module coefs
      !> Geometric factors \f$ G_{23} \f$
      real(kind=rp), allocatable :: G23(:,:,:,:)
 
+     !> Compressed geometric factors \f$ G_{11} \f$
+     real(kind=rp), allocatable :: G11_compressed(:,:,:,:)
+     !> Compressed geometric factors \f$ G_{22} \f$
+     real(kind=rp), allocatable :: G22_compressed(:,:,:,:)
+     !> Compressed geometric factors \f$ G_{33} \f$
+     real(kind=rp), allocatable :: G33_compressed(:,:,:,:)
+     !> Compressed geometric factors \f$ G_{12} \f$
+     real(kind=rp), allocatable :: G12_compressed(:,:,:,:)
+     !> Compressed geometric factors \f$ G_{13} \f$
+     real(kind=rp), allocatable :: G13_compressed(:,:,:,:)
+     !> Compressed geometric factors \f$ G_{23} \f$
+     real(kind=rp), allocatable :: G23_compressed(:,:,:,:)
+     !> Compressed geometric factors lookup indices
+     real(kind=rp), allocatable :: compression_inds(:)
+
      real(kind=rp), allocatable :: mult(:,:,:,:) !< Multiplicity
      !> generate mapping data between element and reference element
      !! \f$ dx/dr, dy/dr, dz/dr \f$
@@ -358,6 +373,8 @@ contains
 
     call coef_generate_geo(this)
 
+    call coef_generate_geo_compressed(this)
+
     call coef_generate_area_and_normal(this)
 
     call coef_generate_mass(this)
@@ -453,6 +470,30 @@ contains
 
     if (allocated(this%G23)) then
        deallocate(this%G23)
+    end if
+
+    if (allocated(this%G11_compressed)) then
+       deallocate(this%G11_compressed)
+    end if
+
+    if (allocated(this%G22_compressed)) then
+       deallocate(this%G22_compressed)
+    end if
+
+    if (allocated(this%G33_compressed)) then
+       deallocate(this%G33_compressed)
+    end if
+
+    if (allocated(this%G12_compressed)) then
+       deallocate(this%G12_compressed)
+    end if
+
+    if (allocated(this%G13_compressed)) then
+       deallocate(this%G13_compressed)
+    end if
+
+    if (allocated(this%G23_compressed)) then
+       deallocate(this%G23_compressed)
     end if
 
     if (allocated(this%mult)) then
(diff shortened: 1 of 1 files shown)


njansson:
Great Stuff! 

I'll give it a try with the operator benchmark (https://gitlab.com/ExtremeFLOW/neko_opr) and see what it would give.

Should one create "compressed" versions of the operators or should this be default?

timfelle:
Love the concept. It seems like a really great idea. How does this impact stuff like user-deformations, ALE and AMR? I would expect some of these would make it impossible to roll out across everything. Especially AMR might be bad for this.

timofeymukha:
Very neat indeed, and I guess it applies to other geometric stuff like the mass matrix? I agree with Tim that we have to be a bit careful to support cases with mesh motion / change. But that should not be impossible, I think!

njansson:
For AMR, I would guess it should only be a matter of regenerating the compressed things. (as everything else after refinement)

timfelle:
Completely agree that it is possible to support in all those cases. I was more concerned with, "is it useful" in those cases or do we need too much recompute for it to be useful.

GrahamBenHarper:
Great questions!

For AMR, it's pretty easy, with one nuance. Given a single parent cell, you know the children on a fixed refinement level will all use a very similar mapping. If the parent cell is a parallelogram, all children on a fixed refinement level will have exactly the same shape and therefore exactly the same mapping. If the parent cell is not a parallelogram, the children on a fixed refinement level will be similar, but off slightly depending on how "close to parallelogram" the parent is. Also, if you're doing fully structured AMR, then everything on a given refinement level has the same mapping, meaning there are only as many unique mappings as there are refinement levels, which would give huge benefits.

For ALE, every time you transport the mesh and evaluate the operator on a new mesh, you would have to recompute the mapping compression, which would make it very expensive. Tim is totally right that the computational tradeoff may not actually help in that case. The compression algorithm that I wrote here is a greedy O(nelv^2) algorithm in the worst case. Even if it were modified to something closer to an O(nelv log(nelv)) algorithm with some automatic sorting, that might be too expensive for ALE.

For mass matrices, as Timofey points out, the compression works exactly the same. In fact, mass matrices are computed using those same mappings, so you could compute the local mass matrices using the compressed Gij mappings and completely avoid the redundant calculations, which would save a ton of setup time.

For default behavior, this likely wouldn't hurt any simulation that takes a reasonable number of time steps (there is a balance between initial cost and savings per time step), but my gut feeling is features with a lot of mesh change like ALE would want it turned off even if the compression is algorithmically optimized.

One other comment I didn't mention earlier: the compression tolerance `real(kind=rp) :: ctol = 1.0E-7_rp` can be configured, but it should remain very small. Making it large would give more compression and faster simulations, but it could case things to become nonphysical very quickly. In an ideal world you would balance that with discretization error estimates.

I need to run this through the linter, I need to configure this using input parameters, and I probably need unit tests. Is there anything else missing that I'm forgetting?

njansson (approved):
Great stuff, I think it would be good to get this this in (and we figure how how to plumb it into the workflow later on). If we have it in develop, more people could test it (it is also not destructive

GrahamBenHarper:
Oh, one other comment: since static meshes are the same from run to run, you could imagine integrating this compression with the mesh file. For example, a material coefficient on each cell which encodes the compression so you don't have to do the compression algorithm during the simulation.

GrahamBenHarper:
Since this is not "on" by default, I have added a print statement which tells you how much redundancy was removed. Hopefully that's okay. I also ran some tests and can confirm this works as intended on basic examples like the poisson example.

GrahamBenHarper:
Using this PR with the poisson example on the TGV mesh `./poisson ../../neko-tgv/tgv_16M.nmsh 2 100` yields
```
Tot MFlops =   5.2703E+02, MFlops      =   5.2703E+02
Setup Flop =   5.2345E+11, Solver Flop =   2.0334E+11
Start res  =   9.4389E+03, Final res   =   5.1621E+04
Solve Time =   0.1379E+04
```
which is an 11.3% speedup, which is close to what I estimated in the PR description when I replaced `G11(i,1,1,e)` with `G11(i,1,1,1)`

njansson (approved):
Instead of letting this hanging, how about we merge it now, and add the develop and compressed version of ax in a separate pr? @timfelle

timfelle (approved):

PR #2580, opened 2026-06-09
Add Valgrind regression test suite

timfelle:
Introduce a Valgrind-based regression test to validate memory leaks for a minimal TGV run. This includes necessary scripts and configurations to execute the test and check for memory issues. The test setup ensures that required dependencies are available and provides options for customizing thresholds. A README file outlines the requirements and usage instructions.


njansson:
I think we should get this is, as is. And then we can setup runners etc afterwards

timofeymukha:
Is the tgv case just a copy from the examples? Then I would maybe add a cp to the shell script, just to have less case / user files to keep track of and update.

njansson (approved):

timofeymukha (approved):

PR #2582, opened 2026-06-10
Removed declaration causing problems with cray compiler

adampep:


njansson:
Nore sure about this one, since we haven't seen an issue with this using CCE. Do you have an example of the issue

adampep:
I've just merged develop into my branch and got error on dardel
          select type (sp => this%recv_dof(src)%data)
                       ^                              
ftn-1871 ftn: ERROR GS_NBWAIT_MPI, File = gs/gs_mpi.f90, Line = 301, Column = 24 
  The selector in a SELECT TYPE statement must be polymorphic.

Line number different due to my changes

adampep:
commenting sp declaration solved issue

njansson:
Strange, which version of CCE do you use?

timfelle:
This makes sense.
When you do sp=> inside the select statement, it is the same as inside an associate. You cannot create an associate of an existing variable.

adampep:
gnu and intel have no problem with it. It is old version loaded with PrgEnv-intel/8.6.0

njansson (approved):

timfelle (approved):

PR #2590, opened 2026-06-12
Fixes to scalar stats and scalar SGS stats

loretet:
PR to fix two issues that occur with scalar statistics when restarting and/or running with multiple scalars:

1. In both `scalar_stats` and `scalar_sgs_stats` the numbering of files after a restart worked only if the name of the output file was the default name.
2. `scalar_sgs_stats_simcomp` had a bug that prevented the default output filename to be assigned correctly when running simulations with multiple scalars

 ------- 

Now:
1. The numbering works also if the output file name includes the name of the scalar field or if the entire filename is provided by the user with `"output_filename"` in the json file.
2. a correct default name based on the scalar name is assigned to the `sgs_stats` CSV files, making it optional to specify the output filename when working with multiple scalars (e.g. if there are two scalars named "pippo" and "pluto",  the SGS stats files that get output will automatically have names `scalar_sgs_stats_pippo0.csv` and `scalar_sgs_stats_pluto0.csv`, just like it would happen with non-SGS statistics (which work correctly)

### DIFF
--- a/src/simulation_components/scalar_sgs_stats_simcomp.f90
+++ b/src/simulation_components/scalar_sgs_stats_simcomp.f90
@@ -65,7 +65,7 @@ module scalar_sgs_stats_simcomp
      !> Time value at which the sampling of statistics is initiated.
      real(kind=rp) :: start_time
      real(kind=rp) :: time
-     logical :: default_fname = .true.
+     character(len=:), allocatable :: base_filename
 
    contains
      generic :: init_from_components => &
@@ -109,19 +109,12 @@ contains
     character(len=:), allocatable :: alphat_field, nut_field
     real(kind=rp) :: pr_turb
     logical :: nut_dependency
-    logical :: sname_provided
-
-    sname_provided = json%valid_path('field')
 
     call json_get_or_default(json, 'field', &
          sname, 's')
-    if (sname_provided) then
-       call json_get_or_default(json, "name", &
-            name, "scalar_sgs_stats_" // trim(sname))
-    else
-       call json_get_or_default(json, "name", &
-            name, "scalar_sgs_stats")
-    end if
+    call json_get_or_default(json, "name", &
+         name, "scalar_sgs_stats_" // trim(sname))
+
     call this%init_base(json, case)
     call json_get_or_default(json, 'avg_direction', &
          hom_dir, 'none')
@@ -149,23 +142,15 @@ contains
           call this%init_from_components(name, sname, coef, &
                start_time, hom_dir, alphat_field, filename)
        end if
-    else if (sname_provided) then
-       if (nut_dependency) then
-          call this%init_from_components(name, sname, coef, &
-               start_time, hom_dir, nut_field, pr_turb, &
-               "scalar_sgs_stats_"//trim(sname) // "0")
-       else
-          call this%init_from_components(name, sname, coef, &
-               start_time, hom_dir, alphat_field, &
-               "scalar_sgs_stats_"//trim(sname) // "0")
-       end if
     else
        if (nut_dependency) then
           call this%init_from_components(name, sname, coef, &
-               start_time, hom_dir, nut_field, pr_turb)
+               start_time, hom_dir, nut_field, pr_turb, &
+               "scalar_sgs_stats_"//trim(sname))
        else
           call this%init_from_components(name, sname, coef, &
-               start_time, hom_dir, alphat_field)
+               start_time, hom_dir, alphat_field, &
+               "scalar_sgs_stats_"//trim(sname))
        end if
     end if
 
@@ -187,7 +172,7 @@ contains
     real(kind=rp), intent(in) :: start_time
     type(coef_t), intent(in), target :: coef
     character(len=*), intent(in) :: alphat_field
-    character(len=*), intent(in), optional :: fname
+    character(len=*), intent(in) :: fname
     character(len=NEKO_FNAME_LEN) :: stats_fname
     character(len=LOG_SIZE) :: log_buf
     character(len=5) :: prefix
@@ -207,13 +192,8 @@ contains
     this%name = name
     this%start_time = start_time
     this%time = start_time
-    if (present(fname)) then
-       this%default_fname = .false.
-       stats_fname = fname
-    else
-       stats_fname = "scalar_sgs_stats_" // trim(sname) // "0"
-       this%default_fname = .true.
-    end if
+    this%base_filename = fname
+    stats_fname = trim(fname) // "0"
 
     call this%stats_output%init(this%stats, this%start_time, &
          hom_dir = hom_dir, name = stats_fname, &
@@ -269,13 +249,8 @@ contains
     this%name = name
     this%start_time = start_time
     this%time = start_time
-    if (present(fname)) then
-       this%default_fname = .false.
-       stats_fname = fname
-    else
-       stats_fname = "scalar_sgs_stats_" // trim(sname) // "0"
-       this%default_fname = .true.
-    end if
+    this%base_filename = fname
+    stats_fname = trim(fname) // "0"
 
     call this%stats_output%init(this%stats, this%start_time, &
          hom_dir = hom_dir, name = stats_fname, &
@@ -314,24 +289,16 @@ contains
     write (prefix, '(I5)') &
          this%stats_output%file_%file_type%get_start_counter()
     call filename_suffix(fname, suffix)
-    call filename_name(fname, basename)
-
-    ! Remove trailing zero if needed
-    i = len(trim(basename))
-    do while (i > 0 .and. basename(i:i) >= '0' .and. basename(i:i) <= '9')
-       i = i - 1
-    end do
-    basename = basename(:i)
 
     last_slash_pos = &
          filename_tslash_pos(fname)
     if (last_slash_pos .ne. 0) then
        fname = &
             trim(fname(1:last_slash_pos))// &
-            trim(basename)// &
+            trim(this%base_filename)// &
             trim(adjustl(prefix))//"."//suffix
     else
-       fname = trim(basename)// &
+       fname = trim(this%base_filename)// &
             trim(adjustl(prefix))//"."//suffix
     end if
     call this%stats_output%init_base(fname)
--- a/src/simulation_components/scalar_sgs_stats_simcomp.f90
+++ b/src/simulation_components/scalar_sgs_stats_simcomp.f90
@@ -149,6 +149,16 @@ contains
           call this%init_from_components(sname, coef, &
                start_time, hom_dir, alphat_field, filename)
        end if
+    else if (sname_provided) then
+       if (nut_dependency) then
+          call this%init_from_components(sname, coef, &
+               start_time, hom_dir, nut_field, pr_turb, &
+               "scalar_sgs_stats_"//trim(sname) // "0")
+       else
+          call this%init_from_components(sname, coef, &
+               start_time, hom_dir, alphat_field, &
+               "scalar_sgs_stats_"//trim(sname) // "0")
+       end if
     else
        if (nut_dependency) then
           call this%init_from_components(sname, coef, &
--- a/src/simulation_components/scalar_sgs_stats_simcomp.f90
+++ b/src/simulation_components/scalar_sgs_stats_simcomp.f90
@@ -44,7 +44,7 @@ module scalar_sgs_stats_simcomp
   use case, only : case_t
   use coefs, only : coef_t
   use utils, only : NEKO_FNAME_LEN, filename_suffix, filename_tslash_pos, &
-       NEKO_VARNAME_LEN
+       filename_name, NEKO_VARNAME_LEN
   use logger, only : LOG_SIZE, neko_log
   use json_utils, only : json_get, json_get_or_default, &
        json_get_or_lookup_or_default, json_get_or_lookup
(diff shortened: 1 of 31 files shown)


linneahuusko:
I don't know why there are changes to so many files, maybe some branch was not up to date with develop... The PR should only include changes to `scalar_stats_simcomp.f90` and `scalar_sgs_stats_simcomp.f90`.

shiyu-du-dsy:
> I don't know why there are changes to so many files, maybe some branch was not up to date with develop... The PR should only include changes to `scalar_stats_simcomp.f90` and `scalar_sgs_stats_simcomp.f90`.

I guess it's some AI tools try to improve the readability of the code and dig into all files...

src/simulation_components/scalar_sgs_stats_simcomp.f90
```diff
@@ -282,25 +300,37 @@ subroutine scalar_sgs_stats_simcomp_restart(this, time)
...
+         this%stats_output%file_%file_type%get_start_counter()
+    call filename_suffix(fname, suffix)
+    call filename_name(fname, basename)
+
+    ! Remove trailing zero if needed
+    i = len(trim(basename))
+    do while (i > 0 .and. basename(i:i) >= '0' .and. basename(i:i) <= '9')
+       i = i - 1
+    end do
+    basename = basename(:i)
```
timofeymukha: This removes any trailing digits, right? Even if the scalar's actual name ends with a number. So if I have a scalar called s1, it will strip off the 1.

linneahuusko: Oh shoot, that's true. I first used another approach that just automatically removed the last character in the string but that doesn't work if the user provides an output file name. I struggled with coming up with something that is general enough but this is clearly not going to work :(

linneahuusko: Could it be an option to not add the 0 to the files initially? Now the default is `scalar_stats0.csv/0.f0000X` but if we instead make it `scalar_stats.csv/0.f0000X` it would be easy to just append a number to whatever the base name is.

timofeymukha: I have to admit that I don't really know how this counter added before the extension works :P.

linneahuusko: I'm not too sure about the counter either, I've just moved stuff around a bit 😅 Now the zero in the file name is hardcoded, and is changed to the value of the counter only during a restart. As far as I understand it the counter number depends on the physical time of the simulation. Perhaps that could be used directly from the start? I'll bring it up at the dev meeting on Wednesday!

timfelle: The counter is used to describe the start index of the current simulation block. So for fluid files you would get filename0.nek5000 for the first iteration, and then for a simulation restarted after 10 writes the file would be called filename10.nek5000.

timofeymukha: Well, the most trivial thing would be to save the name of the file without the prefix to a variable in the type, i.e.   stats_output_stem = "scalar_stats_s". Or if one wants to be really compact, maybe just the number of chars in that stem, and then one can extract it from the `basename`.

src/simulation_components/scalar_sgs_stats_simcomp.f90
```diff
@@ -102,22 +102,31 @@ subroutine scalar_sgs_stats_simcomp_init_from_json(this, json, case)
...
+    sname_provided = json%valid_path('field')
+
+    call json_get_or_default(json, 'field', &
+         sname, 's')
+    if (sname_provided) then
+       call json_get_or_default(json, "name", &
+            name, "scalar_sgs_stats_" // trim(sname))
+    else
+       call json_get_or_default(json, "name", &
+            name, "scalar_sgs_stats")
```
vbaconnet: Can we not have the `"s"` appended to the file name anyways? That way the behavior is always the same, you always get `"scalar_sgs_stats" + name of the scalar field`

vbaconnet: Just checked, and yes it is also done for scalar_stats... so I extend the same comment for scalar stats :D

src/simulation_components/scalar_sgs_stats_simcomp.f90
```diff
@@ -134,35 +143,45 @@ subroutine scalar_sgs_stats_simcomp_init_from_json(this, json, case)
...
-  !! @param s_name The name of the scalar field
+  !! @param sname The name of the scalar field
   !! @param coef sem coefs
   !! @param start_time time to start sampling stats
   !! @param hom_dir directions to average in
   !! @param alphat_field name of the eddy diffusivity field
   !! @param fname name of the outut file
   subroutine scalar_sgs_stats_simcomp_init_from_components_alphat(this, &
-       s_name, coef, start_time, hom_dir, alphat_field, fname)
+       sname, coef, start_time, hom_dir, alphat_field, fname)
```
vbaconnet: `name` should be present as an argument to this routine, so that this%name can be initialized here if it is called on its own (i.e. not from the _init_from_json). See `scalar_stats_simcomp_init_from_components` 
```suggestion
       name, sname, coef, start_time, hom_dir, alphat_field, fname)
```

src/simulation_components/scalar_sgs_stats_simcomp.f90
```diff
@@ -182,16 +201,15 @@ subroutine scalar_sgs_stats_simcomp_init_from_components_alphat(this, &
     write(log_buf, '(A,A)') 'Averaging in direction: ', trim(hom_dir)
     call neko_log%message(log_buf)
 
-
-    call this%stats%init(coef, s_name, alphat_field)
+    call this%stats%init(coef, sname, alphat_field)
 
     this%start_time = start_time
```
vbaconnet, line 193: Assign this%name in case this routine is called on its own.
```suggestion
    this%name = name
    this%start_time = start_time
```

src/simulation_components/scalar_sgs_stats_simcomp.f90
```diff
@@ -209,17 +227,17 @@ end subroutine scalar_sgs_stats_simcomp_init_from_components_alphat
...
+  !! @param sname The name of the scalar field
   !! @param coef sem coefs
   !! @param start_time time to start sampling stats
   !! @param hom_dir directions to average in
   !! @param nut_field name of the eddy diffusivity field
   !! @param pr_turb turbulent Prandtl number
   !! @param fname name of the output file
   subroutine scalar_sgs_stats_simcomp_init_from_components_nut(this, &
-       s_name, coef, start_time, hom_dir, nut_field, pr_turb, fname)
+       sname, coef, start_time, hom_dir, nut_field, pr_turb, fname)
```
vbaconnet: `name` should be present as an argument to this routine, so that this%name can be initialized here if it is called on its own (i.e. not from the _init_from_json)
```suggestion
       name, sname, coef, start_time, hom_dir, nut_field, pr_turb, fname)
```

src/simulation_components/scalar_sgs_stats_simcomp.f90
```diff
@@ -243,15 +261,15 @@ subroutine scalar_sgs_stats_simcomp_init_from_components_nut(this, &
     call neko_log%message(log_buf)
 
 
-    call this%stats%init(coef, s_name, nut_field, pr_turb)
+    call this%stats%init(coef, sname, nut_field, pr_turb)
 
     this%start_time = start_time
```
vbaconnet, line 250: Assign this%name in case this routine is called on its own.
```suggestion
    this%name = name
    this%start_time = start_time
```

vbaconnet (commented):
Thanks for putting in the time, if this fixes your issues I am happy to merge it in. 

I do however have some comments on the design of scalar and SGS stats in general with the file naming.

1. The `name` dummy argument is missing from the sgs stats `_from_components` constructors, causing the this%name to not be initialized if the routines were called on their own. I've made some suggestions accordingly to make your life easier!

2. I think there is an opportunity here to base the output file name on `this%name`. In every `init_from_json` we construct a `this%name` that ends up being exactly the default file name with a trailing `0`. Example:
```.f90
! scalar stats
this%name = "scalar_stats_" // trim(sname)
! ... and later init_from_components is called with
stats_fname = "scalar_stats_" // trim(sname) // "0" ! = trim(this%name) // "0"
```

For SGS stats it is the same now as well. If we based the default filenames on `this%name` every time we would avoid all of these if blocks and make the code much clearer. This would extend to fluid, scalar and sgs stats as well but this may be a bit out of scope if you are urgently wanting to fix your issues.

3. It is a bit strange to me that the sgs_stats constructor take field names and not field pointers like the other stats. I think it would be good at some point in time (in another PR) to do a refactor so all the stats have similar interfaces, unless there is a reason I am not seeing :)

linneahuusko:
@vbaconnet Thank you for the suggestions! I intended to just do an unintrusive quick fix but that turned out to be trickier than I thought, so I agree that we could take the opportunity to do a bit more restructuring to get a solution that is more robust and readable. I'll see what I can do!

src/simulation_components/scalar_sgs_stats_simcomp.f90
linneahuusko: You mean so we can remove the "default" option from the if statements completely?

vbaconnet: yep!

vbaconnet:
> @vbaconnet Thank you for the suggestions! I intended to just do an unintrusive quick fix but that turned out to be trickier than I thought, so I agree that we could take the opportunity to do a bit more restructuring to get a solution that is more robust and readable. I'll see what I can do!

No problem, as I said if things work for you and you want the quick fix to go through we can do that and change things later as well :)

src/simulation_components/scalar_sgs_stats_simcomp.f90
```diff
@@ -209,17 +228,17 @@ end subroutine scalar_sgs_stats_simcomp_init_from_components_alphat
...
   !! @param hom_dir directions to average in
   !! @param nut_field name of the eddy diffusivity field
   !! @param pr_turb turbulent Prandtl number
   !! @param fname name of the output file
   subroutine scalar_sgs_stats_simcomp_init_from_components_nut(this, &
-       s_name, coef, start_time, hom_dir, nut_field, pr_turb, fname)
+       name, sname, coef, start_time, hom_dir, nut_field, pr_turb, fname)
     class(scalar_sgs_stats_simcomp_t), target, intent(inout) :: this
-    character(len=*), intent(in) :: s_name
+    character(len=*), intent(in) :: sname
```
vbaconnet, line 223:
```suggestion
    character(len=*), intent(in) :: name
    character(len=*), intent(in) :: sname
```

src/simulation_components/scalar_sgs_stats_simcomp.f90
```diff
@@ -134,35 +143,45 @@ subroutine scalar_sgs_stats_simcomp_init_from_json(this, json, case)
...
   !! @param start_time time to start sampling stats
   !! @param hom_dir directions to average in
   !! @param alphat_field name of the eddy diffusivity field
   !! @param fname name of the outut file
   subroutine scalar_sgs_stats_simcomp_init_from_components_alphat(this, &
-       s_name, coef, start_time, hom_dir, alphat_field, fname)
+       name, sname, coef, start_time, hom_dir, alphat_field, fname)
     class(scalar_sgs_stats_simcomp_t), target, intent(inout) :: this
-    character(len=*), intent(in) :: s_name
+    character(len=*), intent(in) :: sname
```
vbaconnet, line 170:
```suggestion
    character(len=*), intent(in) :: name
    character(len=*), intent(in) :: sname
```

linneahuusko:
@vbaconnet
> 2. I think there is an opportunity here to base the output file name on `this%name`. In every `init_from_json` we construct a `this%name` that ends up being exactly the default file name with a trailing `0`. Example:
> 
> 
> ```fortran-modern
> ! scalar stats
> this%name = "scalar_stats_" // trim(sname)
> ! ... and later init_from_components is called with
> stats_fname = "scalar_stats_" // trim(sname) // "0" ! = trim(this%name) // "0"
> ```
> 
>  For SGS stats it is the same now as well. If we based the default filenames on `this%name` every time we would avoid all of these if blocks and make the code much clearer. This would extend to fluid, scalar and sgs stats as well but this may be a bit out of scope if you are urgently wanting to fix your issues.
> 

I made an attempt to always construct the output file name based on `sname` but I can't think of a way of doing it without adding a bunch of if-statements to deal with the `"output_filename"` when that is provided. Now I left a `base_filename` in the type since I found that to be the cleanest solution, and as far as I can tell it works robustly. I think we can merge this now since it solves the immediate problem, and then do a more extensive restructuring of this at some later point.

vbaconnet:
> I made an attempt to always construct the output file name based on `sname` but I can't think of a way of doing it without adding a bunch of if-statements to deal with the `"output_filename"` when that is provided. Now I left a `base_filename` in the type since I found that to be the cleanest solution, and as far as I can tell it works robustly. 

Ok I see what you did with the base_filename :) you could also just rely on the contents of `this%name` to build all the names instead, but if this works it's fine.

> I think we can merge this now since it solves the immediate problem, and then do a more extensive restructuring of this at some later point.

I think that sounds good, if you guys are happy with the behavior now. Can I just ask if this has been tested with `fld` files as well?

linneahuusko:
> I think that sounds good, if you guys are happy with the behavior now. Can I just ask if this has been tested with `fld` files as well?

I did actually remember to test on `fld` files, yes :D

njansson:
@timfelle @timofeymukha @vbaconnet can we merge?

vbaconnet:
I'm okay with merging!

vbaconnet (approved):

timfelle (approved):

PR #2591, opened 2026-06-12
Fld_file_write DEVICE_TO_HOST

arash-mp:
Currently, using device, when mesh changes and `output_mesh_in_all_files : true` is needed, `dof` is never copied to `host` for outputting. so the written mesh is always the inital mesh. 
This fixes the issue on device by copying `dof%x_d, dof%y_d, dof%z_d` to `host`.


arash-mp:
@adperezm I also added it in `fld_file_write_masked`. Let me know if it may brake anything there.

adperezm:
> @adperezm I also added it in `fld_file_write_masked`. Let me know if it may brake anything there.

Looks good.

Although for consistency, I would do this outside this call. Note that the velocity is not copied to host in that routine either. So I would copy the mesh at the same instance when the velocity is copied, wherever that is.

adperezm (approved):
As I mentioned in the other comment. I approve, but if we are not consistent it might get confusing.

arash-mp:
> As I mentioned in the other comment. I approve, but if we are not consistent it might get confusing.

@timfelle Adalberto's point is also correct.
In ALE_Device PR, I am copying `dof` to `host` in `fluid_output.f90`, next to all other fields being copied to `host`.
So isn't it better if I keep it there? 
Then if everyone is happy with keeping it there (`fluid_output.f90`), I will close this PR and keep it in ALE_device.

src/io/fld_file.f90
```diff
@@ -311,6 +313,14 @@ subroutine fld_file_write(this, data, t)
...
        msh => dof%msh
        Xh => dof%Xh
+       if (this%write_mesh .and. (NEKO_BCKND_DEVICE .eq. 1)) then
+          call device_memcpy(dof%x, dof%x_d, dof%size(), &
+               DEVICE_TO_HOST, sync = .false.)
+          call device_memcpy(dof%y, dof%y_d, dof%size(), &
+               DEVICE_TO_HOST, sync = .false.)
+          call device_memcpy(dof%z, dof%z_d, dof%size(), &
+               DEVICE_TO_HOST, sync = .true.)
+       end if
```
njansson, line 323: Shouldn't this dtoh copy be done when dof is updated after mesh movement, and not here in the writer (since for none ALE dof and dof_d is the same already)

arash-mp: But we don't need to copy to host every time step that dof is updated. Only when we want to save the snapshots. Right?

njansson: Depends on where the dof is updated (host or device), but regardless would say this is better to kept in coef close to the update code (if it exists). If we do this in fld_writer, it will do a lot of redundant copies (if write mesh is on, and ALE off)

adperezm: I disagree with that, we do not sync the velocity every-time is updated either. Isnt the mesh always moving in ale? It would be something like doing that.

arash-mp: mesh is updated every time step in ALE, but when it is running on device, the mesh is only updated on device. CPU only has the inital mesh in that case. 
So I would also say the copy is just needed only when a snapshot is going to be written. 
Also regarding redundant copies when `output_mesh_in_all_files : true` and no ALE, I would say for a non-ALE simulation, one should always write the mesh only once, and if it is set to `true`, then logically it is safer to copy mesh from device to cpu (for example one does a static mesh deformation on device).
So I think it is best to keep it in `fluid_output.f90`, next to other copies for other fields, which is what is done in ALE_device PR (not here).
The price would be, as Niclas said, the redundant copies in no-ALE sims, but that happens only if one run a static simulation with always_output_mesh, which generally is not the case.

timofeymukha: I think I agree. If one wants the mesh to be written to every file, one must have some reason for that, and we should assume that the dofs may have been changed for one reason or another.

timofeymukha:
We go for syncing in fluid output

PR #2595, opened 2026-06-14
Add Metal backend for Apple Silicon GPUs

njansson:
This PR introduces a Metal compute backend so Neko can run on Apple Silicon GPUs. Adds Objective-C host wrappers (.m) and Metal Shading Language kernels (.metal) covering the device layer and operators across math/mathops, gather–scatter, boundary conditions, Krylov and multigrid smoothers, SEM coefficients, and fluid/scalar residuals.

* Enabled with --with-metal (macOS); kernels are compiled into an embedded metallib at build time.
* Single precision only — Apple GPUs have no FP64, so the backend requires --enable-real=sp.
* Docs: installation guide updated with backend support-tier and communication-backend tables.


timfelle (approved):

timofeymukha:
I am a bit split here. On the one hand it's cool, but adding 16k lines of code to maintain and a new backend to keep track of just to run on a laptop feels like a lot of work for a small win. Particularly with no dp support, since we know pure sp does not really work. 

Is it a big advantage compared to running OpenCL on the Mac?

njansson:
> I am a bit split here. On the one hand it's cool, but adding 16k lines of code to maintain and a new backend to keep track of just to run on a laptop feels like a lot of work for a small win. Particularly with no dp support, since we know pure sp does not really work.
> 
> Is it a big advantage compared to running OpenCL on the Mac?

Agree with the maintenance burden, that is also why I created the backend tier table in the documentation. In general, I think it’s better to allow for experimental backends (Tier-2); we might otherwise lose these in orphaned repositories. As with the other esoteric stuff added (nvshmem, caf etc.), I'm more than happy to try to keep them alive and well in the codebase.

In this case, the metal backend is far better than the clunky OpenCL support in macOS (it’s also deprecated and on life support and stuck on a very old version). Also, it’s not only laptops, but their workstations offer an interesting option, with large memory and high memory bandwidth.  Finally, since these are unified memory, sp limitations can be worked around using the cpu part of the chip.

.github/workflows/check_gnu.yml
```diff
@@ -59,6 +63,9 @@ jobs:
             backend: cuda
           - os: macos-15
             backend: hip
+          - os: macos-15
+            backend: metal
+            precision: dp
```
timofeymukha, line 68: Does dp work?

timofeymukha (commented):
There are many copyright year errors in the files. Probably an easy fix with an llm.

njansson:
> There are many copyright year errors in the files. Probably an easy fix with an llm.

Sure! I'm surprised that the llm failed me with this when doing the OpenCL -> Metal conversion

.github/workflows/check_gnu.yml
njansson: No, that's why we keep it in the exclude list. What I want to try is if the GitHub runners can run the unit tests with the Metal backend

njansson:
> > There are many copyright year errors in the files. Probably an easy fix with an llm.
> 
> Sure! I'm surprised that the llm failed me with this when doing the OpenCL -> Metal conversion

Fixed!

timofeymukha (approved):

PR #2596, opened 2026-06-15
Enhancement to overset simulations

adperezm:
As the title says, there are two things done here:

1. Added some logging to the error at the interface when running simulations.
2. Modified a bit the projections in PnPn so that they only happen at the actual physical time-step. The projections spaces will not be updated during the sub-iterations. I noticed that in some of my simulations, this was destabilizing the solver.

Later I might add projection spaces to only operate during the subiterations but this might not be that useful.


timofeymukha (approved):
Makes sense to me, after the first iteration, the solution itself is hopefully the best approximation for subsequent iterations, and no projection is needed.

njansson (approved):

PR #2597, opened 2026-06-16
Nullify pointers in various subroutines

timfelle:
Enhance memory management by nullifying pointers in multiple subroutines to prevent potential issues with dangling references. This addresses concerns related to pointer safety and improves code stability.


njansson:
Isn't it so that pointers are nullified when they go at out scope, as for allocatable objects (not saying that it is wrong to be formal)

timfelle:
> Isn't it so that pointers are nullified when they go at out scope, as for allocatable objects (not saying that it is wrong to be formal)

You are probably right. But I am kinda loosing fate in the compilers at this point.

njansson:
> > Isn't it so that pointers are nullified when they go at out scope, as for allocatable objects (not saying that it is wrong to be formal)
> 
> You are probably right. But I am kinda loosing fate in the compilers at this point.

Hear hear!

njansson (approved):

timofeymukha (approved):

vbaconnet (approved):

PR #2600, opened 2026-06-17
Throw assertion error for unsupported atomicAdd

timfelle:
Introduce an assertion error in the `masked_atomic_reduction_kernel` when the compute capability is below 6.0, ensuring that the function is only used in supported environments. This change enhances error handling for CUDA architecture compatibility.


timofeymukha (approved):

njansson (approved):
Great! Much needed, maybe we should also add it in `gs_kernel.cu` and related kernels that also uses atomic

pohm01 (approved):

PR #2601, opened 2026-06-19
Use fixed random seed for phmg

pohm01:
Set functions used for phmg (tamg aggregation, Chebyshev's power method) to use a fixed random number seed.


pohm01:
What is the changelog policy these days? This is a change that should be added to the changelog.

timfelle:
I am not sure this will ensure consistent results across restarts.

pohm01:
It may not, but it is a change I have been meaning to add. PHMG being nondeterministic has always been a little annoying, this will at least make it deterministic up to build and launch configuration.

timfelle (approved):

njansson (approved):

PR #2604, opened 2026-06-29
Feature/lagrangian particle tracking

Shiyu-Sandy-Du:
This is a prototype for the lagrangian particle tracking (LPT) simcomp. Particles with/without inertia can be supported. The input file format supports arrays defined in the json file or a csv file, which I will later added an new example. The ownership of the particles has two options:
- Distribute the particles evenly on ranks and stick with the initial owner ship: that will help cases where particles aggregate in a small spatial region
- Migrate the particles such that they are handled by the ranks who own the elements where the particles locate: that will help cases where particles are scattered all over the domain

So basically the computation is based on the kinematic definition of velocity ($\frac{dx}{dt} = v$) and Newton's 2nd law ($\frac{dv}{dt}= a$) if inertia is enabled. The numerical solving is done by the 3rd order explicit Adam-Bashforth (AB3) scheme, _e.g._ $\frac{dx}{dt} = v$ => $$x_{n+1} = x_n + \Delta t L(v_n,v_{n-1},v_{n-2})$$. And the main part of the computation can be divided into two parts:

1. At the end of each time step $n$, call `simcomp%compute_` to update the information on the right hand side, namely 
$v_n$ and $a_n$
2. At the begining of each time step $n+1$, call `simcomp%preprocess_` to compute the location and the velocity (if inertia is enabled) of the particles at time step $n+1$ with AB3.

The procedure is currently supported by CPU and also GPU-aware. And here is several validation cases:

1. (Non-inertial particles) 1D advection with velocity $$u = sin(x) + 1, x\in[-\pi,\pi]$$,
Trajectory:
particle_trajectories_cycling

2. (Non-inertial particles) Rigid body rotation (particles computed every 20 time steps)
Trajectory:
particle_trajectories

3. (Inertial particles) Still fluid with wall-collision (purely-elastic collision), nonlinear drag correction is turned off such that we can have an analytical solution for the first two time steps to build in
Trajectory:
particle_trajectories
And the error scaled well with dt refinement, both on CPU and GPU

CPU error scaling:
err_scaling

GPU error scaling:
image

Apart from potential suggestions on the code structure, there is some potential issues to be solved in the future:

- The migration of the particles are not supported by device MPI, namely now the particles are memcpy to the CPU side, migrate and then memcpy to the devices. 
- The improvement brought by GPU-acceleration in global interpolation is not as much as shown in a [document](https://publications.anl.gov/anlpubs/2021/10/171648.pdf) from Argonne.

TDL:

- [x] A new example
- [x] changelog
- [x] Documentation


njansson (commented):
Great stuff!

How about moving the particles to an own type such that it can be reused in other parts not only the simcomp, eg immersed boundary

shiyu-du-dsy:
> Great stuff!
> 
> How about moving the particles to an own type such that it can be reused in other parts not only the simcomp, eg immersed boundary

yes for sure, but where

njansson:
> > Great stuff!
> > How about moving the particles to an own type such that it can be reused in other parts not only the simcomp, eg immersed boundary
> 
> yes for sure, but where

Like, a type for representing the particles, and then a simcomp for convecting them. Then we can do a lot of reuse with the immersed stuff

src/simulation_components/simulation_component_fctry.f90
```diff
@@ -129,6 +131,8 @@ module subroutine simulation_component_allocator(object, type_name)
        allocate(boundary_operation_t::object)
     case ("boundary_flux")
        allocate(boundary_flux_t::object)
+    case ("lpt")
```
timfelle: Maybe it would be good to use a more verbose name here. Just to make sure the
case file is a bit easy to read.

src/simulation_components/lpt/lpt_wall_collision.f90
```diff
@@ -0,0 +1,235 @@
...
+    type(vector_t), intent(inout) :: acc_xlag, acc_ylag, acc_zlag
+    type(vector_t), intent(inout) :: acc_xlaglag, acc_ylaglag
+    type(vector_t), intent(inout) :: acc_zlaglag
+    type(vector_t), intent(inout) :: u_old, v_old, w_old
+    type(vector_t), intent(inout) :: acc_x, acc_y, acc_z
+    integer, intent(in) :: lag_len
+    type(matrix_t) :: rst_new
+      !  type(vector_t) :: x_t
+      !  type(vector_t) :: y_t
+      !  type(vector_t) :: z_t
```
timfelle:
```suggestion
```

src/simulation_components/lpt/lpt_wall_collision.f90
```diff
@@ -0,0 +1,235 @@
...
+    if (n .eq. 0) return
+
+    el_list_d = C_NULL_PTR
+    wall_facet_mask_d = C_NULL_PTR
+
+    allocate(el_list(n))
+    call rst_new%init(3, n)
+    call resx%init(n)
+    call resy%init(n)
+    call resz%init(n)
```
timfelle: Please use the scratch registry for this, quite sure that would be better long term.

src/simulation_components/lpt/lpt_wall_collision.f90
```diff
@@ -0,0 +1,235 @@
...
+  use matrix, only : matrix_t
+  use vector, only : vector_t
+  use device, only : HOST_TO_DEVICE, DEVICE_TO_HOST, device_map, &
+       device_memcpy, device_deassociate, device_free, device_sync
+  use, intrinsic :: iso_c_binding, only : c_ptr, C_NULL_PTR, c_associated
+  implicit none
+  private
+
+  public :: lpt_handle_elastic_wall_collisions
+  public :: lpt_init_wall_facet_mask
```
timfelle: Maybe these should be part of a submodule to `lpt` instead? It would make it so
you can pass the `particle_t` and `lpt_t` instead of all their components.

src/sem/bcknd/device/device_coef.F90
```diff
@@ -234,6 +261,17 @@ subroutine opencl_coef_generate_area_and_normal(area, nx, ny, nz, &
...
+     subroutine opencl_coef_get_normal(normal_x, normal_y, normal_z, &
+          nx, ny, nz, i_idx, j_idx, k_idx, e_idx, facet, lx, n) &
+          bind(c, name='opencl_coef_get_normal')
+       use, intrinsic :: iso_c_binding
+       implicit none
+       type(c_ptr), value :: normal_x, normal_y, normal_z
+       type(c_ptr), value :: nx, ny, nz
+       type(c_ptr), value :: i_idx, j_idx, k_idx, e_idx, facet
+       integer(c_int) :: lx, n
+     end subroutine opencl_coef_get_normal
```
timfelle, line 274:
```suggestion
     end subroutine opencl_coef_get_normal
  end interface
```

timfelle (changes requested):
First batch of things to update. Particulaly the missing `end interface`

src/global_interpolation/global_interpolation.f90
```diff
@@ -1237,6 +1239,44 @@ subroutine global_interpolation_find_and_redist(this, xyz, n_points)
...
 
+  !> Build a communicator that redistributes point-wise payloads from the
+  !! original point ordering to the local owned-point ordering created by
+  !! find_points_and_redist.
+  subroutine global_interpolation_init_redist_comm(this, redist_comm)
+    class(global_interpolation_t), intent(inout) :: this
+    type(glb_intrp_comm_t), intent(inout) :: redist_comm
+    type(stack_i4_t) :: send_pe
+    type(stack_i4_t) :: recv_pe
+    integer, pointer :: point_ids(:)
```
timfelle: These should be nullified before exiting the code.

src/common/particles.f90
```diff
@@ -0,0 +1,185 @@
...
+  use num_types, only : rp
+  use vector, only : vector_t
+  use utils, only : neko_error
+  implicit none
+  private
+
+  !> Particle positions, velocities, properties, and time-history data.
+  type, public :: particles_t
+     type(vector_t) :: x, y, z
+     type(vector_t) :: u, v, w
```
njansson, line 44: @timfelle should one have SoA or AoS?

src/simulation_components/lpt/lpt.f90
```diff
@@ -0,0 +1,861 @@
...
+             end if
+          else
+             allocate(vels(size(coords)))
+             allocate(diams(n_particles))
+             allocate(densities(n_particles))
+             vels = 0.0_rp
+             diams = 0.0_rp
+             densities = 0.0_rp
+          end if
+          allocate(x(n_particles))
```
timfelle: Please use the scratch registry for these temporaries. We have a `host_array_t`
since you do not need the device part of `vector_t`

src/common/particles.f90
Shiyu-Sandy-Du: Isn't SoA more suitable for vectorisation?

timfelle: Well highly depends on which operations you intend to perform on the particles. If you always intend to run global operations on all paricles, component wise, then it is fine as it is (SoA). But if you intend to treat the particles individually you would want them grouped as AoS.

src/simulation_components/lpt/lpt.f90
```diff
@@ -0,0 +1,900 @@
...
+
+  end subroutine evaluate_velocity
+
+  !> Estimate the local particile acceleration
+  subroutine evaluate_acceleration(this, acc_x, acc_y, acc_z, &
+       u_fluid, v_fluid, w_fluid)
+    class(lpt_t), intent(inout) :: this
+    type(vector_t), intent(in) :: u_fluid, v_fluid, w_fluid
+    type(vector_t), intent(inout) :: acc_x, acc_y, acc_z
+    type(vector_t) :: tau_p, Re_p, f, rho_fluid_local
```
timfelle: Please do not use local `vector_t`. Please use the scratch for all of these.

src/simulation_components/lpt/lpt.f90
```diff
@@ -0,0 +1,900 @@
...
+            u_fluid, v_fluid, w_fluid)
+    else
+       this%particles%u = u_fluid
+       this%particles%v = v_fluid
+       this%particles%w = w_fluid
+    end if
+
+    call u_fluid%free()
+    call v_fluid%free()
+    call w_fluid%free()
```
timfelle: Please reorganize this a bit, it seems to not be required to construct the
temporary arrays if you do not have inertia.

src/simulation_components/lpt/lpt_migration.f90
```diff
@@ -0,0 +1,559 @@
...
+            acc_xlag, acc_ylag, acc_zlag, acc_xlaglag, acc_ylaglag, &
+            acc_zlaglag, n)
+    end if
+  end subroutine initialize_particle_distribution
+
+  !> Update particle ownership according to the selected migration strategy.
+  subroutine migrate_particles(this, global_interp, periodic_bc, inertia, &
+       x, y, z, ids, u_lag, v_lag, w_lag, u_laglag, v_laglag, w_laglag, &
+       u, v, w, acc_x, acc_y, acc_z, d, rho, acc_xlag, acc_ylag, acc_zlag, &
+       acc_xlaglag, acc_ylaglag, acc_zlaglag, n, n_global)
```
timfelle: Essentially all of the information here exists in the particles object. There is
no reason to pass all of these parameters individually.

src/simulation_components/lpt/lpt_migration.f90
```diff
@@ -0,0 +1,559 @@
...
+    integer :: n_particles_local
+    integer, allocatable :: particle_ids_local(:)
+    type(vector_t) :: x_local, y_local, z_local
+    type(vector_t) :: u_local, v_local, w_local
+    type(vector_t) :: acc_xlocal, acc_ylocal, acc_zlocal
+    type(vector_t) :: d_local, rho_local
+    type(vector_t) :: u_lag_local, v_lag_local, w_lag_local
+    type(vector_t) :: u_laglag_local, v_laglag_local, w_laglag_local
+    type(vector_t) :: acc_xlag_local, acc_ylag_local, acc_zlag_local
+    type(vector_t) :: acc_xlaglag_local, acc_ylaglag_local, acc_zlaglag_local
```
timfelle: All of these should be temporaries made with the scratch registries.

src/simulation_components/lpt/lpt_migration.f90
```diff
@@ -0,0 +1,559 @@
...
+       call acc_y%free()
+       call acc_z%free()
+       call d%free()
+       call rho%free()
+       call acc_xlag%free()
+       call acc_ylag%free()
+       call acc_zlag%free()
+       call acc_xlaglag%free()
+       call acc_ylaglag%free()
+       call acc_zlaglag%free()
```
timfelle: Why are all the inputs provided to the subroutine freed here. That does not seem safe.

timfelle (changes requested):

src/simulation_components/lpt/lpt.f90
```diff
@@ -0,0 +1,913 @@
...
+          wall_zone_indices)
+       logical, allocatable, intent(inout) :: wall_facet_mask(:, :)
+       type(mesh_t), intent(in) :: msh
+       integer, intent(in) :: wall_zone_indices(:)
+     end subroutine lpt_init_wall_facet_mask
+
+     module subroutine lpt_handle_elastic_wall_collisions(global_interp, &
+          msh, dm_Xh, coef, wall_facet_mask, x_old, y_old, z_old, x, y, z, &
+          d, u, v, w, u_lag, v_lag, w_lag, u_laglag, v_laglag, w_laglag, &
+          acc_xlag, acc_ylag, acc_zlag, acc_xlaglag, acc_ylaglag, &
```
timfelle: Could you not remove all of these arguments by just passing `class(lpt_t), intent(inout) :: this`

src/simulation_components/lpt/lpt.f90
Shiyu-Sandy-Du: it's still needed. The logic here is if inertia is turned on, `uvw_fluid` are used to estimate the acceleration, otherwise they will be the particles velocity at the current time step

src/simulation_components/lpt/lpt.f90
```diff
@@ -0,0 +1,913 @@
...
+         this%particles%n, this%particles%n_global)
+    call this%sync_time_controller(case%time)
+    call this%update_current_rhs()
+
+    call json_get_or_default(json, "output_filename", output_filename, &
+         trim(this%name))
+    call json_get_or_default(json, "output_format", output_format, "csv")
+    select case (trim(output_format))
+    case ("csv", "h5", "hdf5")
+    case default
```
timfelle: You already do this in the outputter, so there is no reason to do it here.

src/simulation_components/lpt/lpt_migration.f90
Shiyu-Sandy-Du: now they are removed

src/simulation_components/lpt/lpt.f90
Shiyu-Sandy-Du: yes it's removed now

src/simulation_components/lpt/lpt.f90
Shiyu-Sandy-Du: solved

src/simulation_components/lpt/lpt.f90
```diff
@@ -0,0 +1,850 @@
...
+          call neko_scratch_registry%request_host_array(y, ind(2), n_particles, &
+               .false.)
+          call neko_scratch_registry%request_host_array(z, ind(3), n_particles, &
+               .false.)
+          call neko_scratch_registry%request_host_array(u, ind(4), n_particles, &
+               .false.)
+          call neko_scratch_registry%request_host_array(v, ind(5), n_particles, &
+               .false.)
+          call neko_scratch_registry%request_host_array(w, ind(6), n_particles, &
+               .false.)
```
timfelle:
```suggestion
          call neko_scratch_registry%request(x, ind(1), n_particles, .false.)
          call neko_scratch_registry%request(y, ind(2), n_particles, .false.)
          call neko_scratch_registry%request(z, ind(3), n_particles, .false.)
          call neko_scratch_registry%request(u, ind(4), n_particles, .false.)
          call neko_scratch_registry%request(v, ind(5), n_particles, .false.)
          call neko_scratch_registry%request(w, ind(6), n_particles, .false.)
```

src/simulation_components/lpt/lpt.f90
```diff
@@ -0,0 +1,850 @@
...
+          call neko_scratch_registry%request_host_array(y, ind_inertia(2), &
+               n_particles, .false.)
+          call neko_scratch_registry%request_host_array(z, ind_inertia(3), &
+               n_particles, .false.)
+          call neko_scratch_registry%request_host_array(u, ind_inertia(4), &
+               n_particles, .false.)
+          call neko_scratch_registry%request_host_array(v, ind_inertia(5), &
+               n_particles, .false.)
+          call neko_scratch_registry%request_host_array(w, ind_inertia(6), &
+               n_particles, .false.)
```
timfelle:
```suggestion
          call neko_scratch_registry%request(x, ind_inertia(1), n_particles, &
               .false.)
          call neko_scratch_registry%request(y, ind_inertia(2), n_particles, &
               .false.)
          call neko_scratch_registry%request(z, ind_inertia(3), n_particles, &
               .false.)
          call neko_scratch_registry%request(u, ind_inertia(4), n_particles, &
               .false.)
          call neko_scratch_registry%request(v, ind_inertia(5), n_particles, &
               .false.)
          call neko_scratch_registry%request(w, ind_inertia(6), n_particles, &
               .false.)
```

src/lpt/lpt.f90
```diff
@@ -0,0 +1,850 @@
...
+               n_particles, .false.)
+          call neko_scratch_registry%request_host_array(z, ind_inertia(3), &
+               n_particles, .false.)
+          call neko_scratch_registry%request_host_array(u, ind_inertia(4), &
+               n_particles, .false.)
+          call neko_scratch_registry%request_host_array(v, ind_inertia(5), &
+               n_particles, .false.)
+          call neko_scratch_registry%request_host_array(w, ind_inertia(6), &
+               n_particles, .false.)
+          x%x = mat_in%x(:, 1)
```
timfelle, line 384: You assume the file reader synchronizes data to the CPU. Is this expected to.

Shiyu-Sandy-Du: Don't we read the csv from cpu?

src/simulation_components/lpt/lpt_periodic_bc.f90
```diff
@@ -0,0 +1,559 @@
...
+    if (present(w_laglag)) call w_laglag%copy_from(DEVICE_TO_HOST, .false.)
+    if (present(acc_xlag)) call acc_xlag%copy_from(DEVICE_TO_HOST, .false.)
+    if (present(acc_ylag)) call acc_ylag%copy_from(DEVICE_TO_HOST, .false.)
+    if (present(acc_zlag)) call acc_zlag%copy_from(DEVICE_TO_HOST, .false.)
+    if (present(acc_xlaglag)) &
+         call acc_xlaglag%copy_from(DEVICE_TO_HOST, .false.)
+    if (present(acc_ylaglag)) &
+         call acc_ylaglag%copy_from(DEVICE_TO_HOST, .false.)
+    if (present(acc_zlaglag)) &
+         call acc_zlaglag%copy_from(DEVICE_TO_HOST, .false.)
```
timfelle: Do not do single line if syntax if you statement does not fit on one line. Then
do `if (..) then ..... end if`

src/lpt/lpt.f90
```diff
@@ -0,0 +1,844 @@
...
+    call vector_col3(acc_x, u_rel, f)
+    call vector_col3(acc_y, v_rel, f)
+    call vector_col3(acc_z, w_rel, f)
+    call vector_invcol2(acc_x, tau_p)
+    call vector_invcol2(acc_y, tau_p)
+    call vector_invcol2(acc_z, tau_p)
+
+    call neko_scratch_registry%relinquish(ind)
+
+  end subroutine evaluate_acceleration
```
timfelle, line 520: Please reorganize this into chunks of concern instead. That way you can reduce
the number of temporary arrays you need at a given time quite drastically. You
should still retain the proper variable naming but generally do 
- request
- compute
- relinquish

examples/rebounding_particles/box.nmsh
timofeymukha, line 1: Do not commit box meshes, include the genmeshbox comand in the readme instead (and in run.sh).

src/simulation_components/lpt/lpt.f90
```diff
@@ -0,0 +1,844 @@
...
+            case%fluid%name // "_mu")
+       this%rho_fluid => neko_registry%get_field_by_name( &
+            case%fluid%name // "_rho")
+    end if
+    this%u_field => neko_registry%get_field_by_name("u")
+    this%v_field => neko_registry%get_field_by_name("v")
+    this%w_field => neko_registry%get_field_by_name("w")
+
+    if (json%valid_path("wall_zone_indices")) then
+       call json_get(json, "wall_zone_indices", this%wall_zone_indices)
```
timofeymukha: This is something we need to think about storing in the fluid type maybe. I feel like a lot of modeling relies on knowing which boundaries are walls.

src/simulation_components/lpt/lpt_periodic_bc.f90
Shiyu-Sandy-Du: fixed

src/lpt/lpt.f90
Shiyu-Sandy-Du: fixed

src/simulation_components/lpt/lpt.f90
Shiyu-Sandy-Du: yes that's a good idea

src/common/particles.f90
```diff
@@ -0,0 +1,190 @@
...
+    call this%x%copy_from(memdir, .false.)
+    call this%y%copy_from(memdir, .false.)
+    call this%z%copy_from(memdir, .false.)
+    call this%u%copy_from(memdir, .false.)
+    call this%v%copy_from(memdir, .false.)
+    call this%w%copy_from(memdir, .true.)
+    if (this%inertia) then
+       call this%d%copy_from(memdir, .false.)
+       call this%rho%copy_from(memdir, .true.)
+    end if
```
timfelle: What about all the other components of the particles. Lags etc. Why are those not synced?

Shiyu-Sandy-Du: This subroutine is only called in the output procedure. Later we can add checkpointing to particles, and at that time we can add device sync of lags

timfelle: For now yes. But when someone need to do a sync of the particles later for some reason, they will call this routine and it will not do what is expected.

Either sync the particles object, or do the sync of the things you need inside the output routine and delete this function if you do lot want to sync extra arrays.

timofeymukha (commented):
Since this will probably be active work for a long while still, I think we should merge sooner than later. Also this has 0 interactions with the main solver, so safe. What I would really like is to get the following right:

- Create an lpt_t type and an lpt_simcomp_t that just wraps it and adds the simcomp boiloder plate, like all the from-component constructors. The lpt code can be under src/lpt directly
- Add the docstrings for the types and routines that are missing. E.g. lpt_periodic_bc.f90 is not well-documented. This will for sure save time later when / if we need to change things.

Generally, great work, this opens many more research avenues for Neko!

timfelle (approved):
This is a good first step, as @timofeymukha mention, it might be worth to just get this in and then we can update later as long as it runs and does not interfere with other parts of the code.

timofeymukha (approved):
Nice. As future work, we should really try to have the constructor in  lpt_t to be from components rather than from json. Then the simcom can have a json one + from componetnts one.

PR #2607, opened 2026-06-30
Scale gather-scatter and mesh setup to high MPI rank counts

njansson:
At high MPI rank counts (targeting >10⁵ ranks), the one-time setup exchanges in gs_schedule and the mesh connectivity routines overflow the MPI runtime's internal unexpected-message buffers, faulting with SIGBUS/SIGSEGV (mostly on Fugaku). The personalized MPI exchanges were also O(P) (gather-scatter, mesh edge/facet renumbering) or O(P²) (mesh point connectivity).

This PR removes both the failure and the scaling cost:
- New crystal_router module (src/comm/crystal_router.f90) — a generic all-to-some personalized MPI exchange in ⌈log₂P stages, correct for non-power-of-two P, with per-stage size negotiation so no O(P) buffer is ever allocated. crystal_router_pack owns the packed record format.
- gs_schedule and mesh_generate_external_point_conn rewritten as canonical-owner rendezvous (mod(gid, P)) routed through the crystal router. This drops the O(P²)/O(P) cost and the buffer-exhaustion crash. The neigh array is now symmetric by construction, and gather-scatter's per-peer dof lists are ordered by global id so both ranks of a pair agree on ordering without transmitting a sequenced key list.
- Remaining mesh neigh_order exchanges (external facet connectivity, shared edge/facet renumbering) converted from standard MPI_Isend to synchronous MPI_Issend, so they can't eager-buffer either. These stay point-to-point — they loop actual neighbours, not all ranks.

### DIFF
--- a/src/comm/crystal_router.f90
+++ b/src/comm/crystal_router.f90
@@ -1,0 +1,285 @@
+! Copyright (c) 2026, The Neko Authors
+! All rights reserved.
+!
+! Redistribution and use in source and binary forms, with or without
+! modification, are permitted provided that the following conditions
+! are met:
+!
+!   * Redistributions of source code must retain the above copyright
+!     notice, this list of conditions and the following disclaimer.
+!
+!   * Redistributions in binary form must reproduce the above
+!     copyright notice, this list of conditions and the following
+!     disclaimer in the documentation and/or other materials provided
+!     with the distribution.
+!
+!   * Neither the name of the authors nor the names of its
+!     contributors may be used to endorse or promote products derived
+!     from this software without specific prior written permission.
+!
+! THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+! "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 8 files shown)


src/comm/crystal_router.f90
```diff
@@ -0,0 +1,285 @@
...
+  !> Route packed records to their destination ranks.
+  !! @param buf  Packed records `[dest, len, payload...]` repeated. On entry,
+  !!             arbitrary destinations in @f$ [0, P) @f$; on exit, only the
+  !!             records addressed to `pe_rank` (may be reallocated).
+  !! @param n    Used length of @a buf (number of `integer(i8)` words). Updated
+  !!             on exit to the length of the received set.
+  subroutine crystal_router_transfer(buf, n)
+    integer(i8), allocatable, intent(inout) :: buf(:)
+    integer, intent(inout) :: n
+    integer(i8), allocatable :: keep(:), snd(:), rcv(:), extra(:)
```
timfelle, line 124: Why are these not deallocated again?

njansson: sure, these are routine local, but added some explicit dealloc calls

timfelle (approved):

tuananhdao (approved):

PR #2619, opened 2026-07-07
Update NVIDIA HPC SDK version to 26.5

njansson:


.github/workflows/develop.yml
```diff
@@ -178,10 +177,10 @@ jobs:
           fi
 
           # NVidia check is skipped since the compiler is broken.
```
timfelle: Probably good to update the comment as well

timfelle (approved):

PR #2625, opened 2026-07-08
Fix pack/unpack race in the NVSHMEM gs backend

njansson:
At large rank counts the NVSHMEM gather-scatter backend diverges. The fused pack-and-push kernel gates its pack on the peer's ready signal, so the read of the shared buffer for a slow peer can be delayed past an unpack from a fast peer, which reduces into the same buffer on a concurrent stream. A dof shared with both peers then reaches the slow peer already partially reduced, double-counting contributions. The race needs dofs shared with several peers plus round-level skew between ranks, so it only shows up at scale.

This change packs all peer slabs with one bulk kernel on the main stream in `nbsend`, before any per-peer work, ordering every read of the shared buffer ahead of any unpack write. Send buffers become round-parity double-buffered so the pack needs no remote gate (stream ordering guarantees the parity slab has drained), and the ready signal now gates only the put. Signal protocol, recv buffering, and the single-block put path are unchanged.

### DIFF
--- a/src/gs/bcknd/device/cuda/gs_kernels.h
+++ b/src/gs/bcknd/device/cuda/gs_kernels.h
@@ -293,7 +293,7 @@ template<typename T>
 __device__ T atomicMinFloat(T* address, T val);
 
 template<>
-__device__ float atomicMinFloat<float>(float* address, float val) {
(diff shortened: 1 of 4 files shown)


timfelle (approved):
LGTM

timofeymukha (approved):

PR #2628, opened 2026-07-08
Fix/variable dt doc

### LINKED ISSUE #2575: Documentation: adaptive time settings
Our current documentation duplicates some settings related to the CFL and
adaptive timestepping. We should make sure it is not actually a duplication and
correct the documentation.

https://neko.cfd/docs/release/dd/d33/case-file.html#autotoc_md47

`cfl_running_avg_coeff` - `running_avg_coeff`

Shiyu-Sandy-Du:
Addressing issue #2575


njansson (approved):

timfelle (changes requested):
This does not address the complete issue. There are still duplicated settings in the time object. Only the explicitly mentioned setting was touched, that setting was only an example of the type of issue. Please be thorough and fix the entire issue.

shiyu-du-dsy:
> issue. There are still duplicate

Now it should be fine

timfelle (approved):

PR #2629, opened 2026-07-08
Add json_get_or_lookup to json_utils usage

adperezm:
Added json_get_or_lookup to the neko module so we can use it in user file.


timfelle (approved):
You could also just do `use json_utils, only : json_get_or_lookup` in your user file. You can access almost everything in the userfile without ever doing `use neko`

adperezm:
Hi!, yeah, that is what I am doing now. 

I only added this one since the other json_utils that one would need are there. Since the lookup option is more recent, I thought maybe it was supposed to be there and forgotten.

njansson (approved):

PR #2634, opened 2026-07-08
Fix device resource leaks and destructor gaps reported in #2632

### LINKED ISSUE #2632: Potential memory leaks in the code: way above my paygrade
# Neko memory-leak & resource-bug hunt

**Scope:** `develop` branch, whole codebase (Fortran + CUDA/HIP/OpenCL/Metal backends).
**Motivating symptom:** the run gets **slower over time** and **dies after a certain number
of time steps** — the classic signature of *unbounded per-step growth* (something allocated
or created every step/iteration and never released).

Method: 12 discovery agents (hot path, destructor coverage, and each GPU backend) produced
44 candidate findings; 9 of the most dangerous were re-checked by adversarial verification
(which **refuted 3**); a synthesis ranked everything against the symptom. The load-bearing
claims below were then re-verified by hand against the source.

---

## 0. Read this first: the cause depends on your backend

The per-step growth that best explains your symptom is **backend-specific**, because the
leak is in GPU *driver-handle* management, not in the Fortran physics:

| Backend | Prime per-step leak | Verdict |
|---|---|---|
| **OpenCL** | `cl_event` leaked in **every Krylov inner product** (`glsc3`/`glsc3_many`), plus `opr_cfl` (per step) and `absval` | **Strongly explains it.** Thousands of leaked events per step. |
| **Metal (Apple)** | `MTLCommandBuffer`s autoreleased with **no `@autoreleasepool`** — drain only at process exit | **Strongly explains it.** Every sync / halo exchange leaks. |
| **CUDA / HIP** | The per-iteration reduction/aux buffers are **cached correctly** (verified) — *not* leaking | These do **not** explain it. See §2/§3 for the CUDA-relevant suspects. |

So: **if you run OpenCL or Metal, jump to §1** — the fix is a handful of one-line additions.
**If you run CUDA/HIP**, the per-step growth is most likely (a) runtime-statistics if you
enabled it (§2), or (b) an object being rebuilt mid-run whose destructor leaks (§4), and you
should also look hard at the correctness bugs in §3 (the `fusedcg` out-of-bounds write can
corrupt memory and cause a delayed death). Tell me your backend and I'll zero in.

---

### 1a. OpenCL: `cl_event` leaked in every inner product  ⚠️ CRITICAL (OpenCL)
`opencl_glsc3` (`math.c:~1276`) and `opencl_glsc3_many` (`math.c:~1345`) create a
`cl_event` (`kern_wait`) for the reduction kernel, use it as the read-back wait event, then
return calling only `clReleaseKernel` — **never `clReleaseEvent(kern_wait)`**. The sibling
`glsc2` (`:1391`), `glsum`, `glmax`, `glmin` all *do* release it, which is the tell.

- `glsc3` is the weighted inner product `aᵀ b c` used **at least once per CG/GMRES
  iteration**; `glsc3_many` is the batched product in GMRES orthogonalization. → thousands
  of `cl_event`s leak per timestep. Event bookkeeping grows (slowdown) and host/driver
  memory exhausts (death).
- **Fix:** add `CL_CHECK(clReleaseEvent(kern_wait));` before the `clReleaseKernel` in both
  functions. The read-back is blocking (`CL_TRUE`), so the event is already complete.

### 1b. OpenCL: `opr_cfl` leaks a `cl_event` every timestep  ⚠️ HIGH (OpenCL)
`opencl_opr_cfl` (`opr_cfl.c:~137`) frees `cfl`, `cfl_d`, and the kernel but not
`kern_wait`. `device_cfl` runs once per timestep → exactly one `cl_event` leaked per step.
- **Fix:** add `CL_CHECK(clReleaseEvent(kern_wait));` alongside the existing releases.

### 1c. OpenCL: `absval` leaks a `cl_kernel` every call  ⚠️ MEDIUM (OpenCL)
`opencl_absval` (`math.c:~1618`) does `clCreateKernel` (`:1605`) with no `clReleaseKernel`.
Called per step in the entropy-viscosity path (`entropy_viscosity.f90:249`) and
gradient-jump-penalty (`gradient_jump_penalty.f90:975`).
- **Fix:** add `CL_CHECK(clReleaseKernel(kernel));` before the closing brace.

### 1d. Metal: command buffers leak — no `@autoreleasepool`  ⚠️ CRITICAL (Metal)
`metal_device_sync` (`metal.m:253`), `metal_stream_sync` (`:265`), `metal_event_record`
(`:318`), `metal_stream_wait_event` (`:343`) each do `id<MTLCommandBuffer> cb = [queue
commandBuffer]` — an **autoreleased** object. There is no `@autoreleasepool` anywhere in
`metal.m`, and a Fortran program has no Cocoa run loop, so the top-level pool never drains
until exit. `event_record`/`stream_wait_event` run on **every gs halo exchange** (many per
step). Command buffers are heavyweight → host memory climbs monotonically (death) and the
autorelease list lengthens (slowdown). The sibling `math/bcknd/device/metal/math.m` already
wraps its command buffer in `@autoreleasepool`, proving the intended pattern.
- **Fix:** wrap each of the four function bodies in `@autoreleasepool { … }`.

### 1e. Backend-agnostic, but OFF by default: runtime-statistics sample stack  ⚠️ (only if enabled)
`runtime_stats_end_region` (`runtime_statistics.f90:205`) does `…%push(elapsed_time)` on
every region close; the `Time-Step` region closes every step and inner solver regions close
per iteration. Each push appends to a stack freed only at shutdown → O(N-steps) host growth,
plus doubling-copy on each regrowth (extra slowdown). **Gated behind
`case.runtime_statistics.enabled`, which defaults to `.false.`** — so it only bites if you
turned runtime statistics on. If so, this alone can produce exactly your symptom on *any*
backend.
- **Fix:** keep a constant-memory running accumulator (count/sum/sumsq/min/max) per region
  instead of pushing every sample; only grow a per-sample buffer when a profile output is
  requested, and cap/flush it.

---

## 2. Serious correctness bugs (can cause crashes / "results that die")

These are not growth leaks, but they can make a run die or produce wrong results.

### 2a. `fusedcg_device` — out-of-bounds heap write at init  ⚠️ HIGH, backend-agnostic
`p` and `p_d` are sized `DEVICE_FUSEDCG_P_SPACE = 10` (`fusedcg_device.F90:219-220`), but the
init loop runs `do i = 1, DEVICE_FUSEDCG_P_SPACE+1` (`:231`) and at `i=11` writes `p_d(11)`
and `device_map(p(:,11), …)` — **past the end of both arrays**. Undefined behavior / heap &
descriptor corruption (with bounds-checking off, as in production). `fusedcg_cpld_device`
(`:323`, arrays `p1/p2/p3`) has the **same** bug. `pipecg` avoided it by sizing its arrays
`P_SPACE+1`.
- **Fix:** change the loop bound to `DEVICE_FUSEDCG_P_SPACE` (the solve never uses column
  11), or size the arrays `+1` if an 11th slot is genuinely intended. Same for `fusedcg_cpld`.

### 2b. `opr_device_convect_scalar` — uninitialized size in `device_map`  ⚠️ HIGH (OIFS scalar + device)
`opr_device.F90:784` assigns `n_GLL` a *second* time (copy-paste) where it meant `n_GL`, so
`n_GL` is **never initialized** and is then used at `:787` `call device_map(ud, ud_d, n_GL)`.
The GL work buffer is allocated with a garbage size → `cudaMalloc` failure/crash, or a
too-small buffer → kernel out-of-bounds write. Reached per step in OIFS dealiased scalar
advection on a device backend.
- **Fix:** line 784 → `n_GL = coef_GL%msh%nelv * Xh_GL%lxyz`.

### 2c. Metal: `metal_event_sync` never actually waits  ⚠️ HIGH (Metal correctness)
`metal.m:329`: the spin target is set to `e.signaledValue`, so the loop
`while (e.signaledValue < target)` is immediately false — host/device synchronization is
**defeated** (use-before-ready race → intermittent wrong results / corruption on Metal).
- **Fix:** capture the value `metal_event_record` will signal (`signaledValue+1`) and wait on
  that, not the current value.

### 2d. NVSHMEM: `opr_cfl` device buffer under-allocated  ⚠️ MEDIUM (NVSHMEM builds)
`cuda/opr_cfl.cu:77` (and `hip/opr_cfl.hip`) allocate `nvshmem_malloc(sizeof(real))` (1
element) but the kernel writes/reads `nel` elements → device-side OOB.
- **Fix:** `nvshmem_malloc((*nel) * sizeof(real))`, matching the non-NVSHMEM branch.

### 2e. `field` assignment — latent double-free of a dofmap  ⚠️ LATENT (currently unreachable)
`field_assign_field` (`field.f90:251`): when the LHS owns an *internal* dofmap and is
repointed to `g`'s *external* dofmap, `internal_dofmap` is left `.true.`, so a later
`field_free` frees a dofmap owned by `g`. Verification showed the triggering assignment does
not occur in the current hot path (hence not your symptom), but it is a real latent defect.
- **Fix:** free+deallocate the owned internal dofmap and clear the flag before repointing.

---

## 3. Confirmed resource leaks (per-object / per-setup)

Real leaks, but they only accumulate if the owning object is **created/destroyed repeatedly**
during a run (AMG/mesh rebuild, restart, moving mesh, or many transient objects). Still worth
fixing; a couple leak *device* memory and stale `device_addrtbl` entries.

| # | Leak | Location | Fix |
|---|---|---|---|
| 1 | `gs_device_free` never frees `shared_gs_d` / `shared_dof_gs_d` / `shared_gs_dof_d` (device buffers **and** addrtbl entries) | `gs/bcknd/device/gs_device.F90:190-238` (mapped at :324/334/340) | add `device_free` (better: `device_unmap`) for the three shared buffers |
| 2 | **`phmg_free` is an empty no-op** — the whole p-multigrid hierarchy leaks | `multigrid/phmg.f90:302-304` | implement it: free each level's cheby/jacobi/schwarz/work/bc + allocated Xh/dm/gs/coef, `amg_solver`, `ax`, `intrp` |
| 3 | `hsmg_free` never frees `amg_solver` (tamg has no `FINAL`) — tree-AMG device hierarchy leaks with `coarse solver = tamg` | `krylov/pc_hsmg.f90:403-472` | `call this%amg_solver%free(); deallocate(this%amg_solver)` |
| 4 | `gs_device_mpi_free` never destroys/deallocates the per-neighbour **event** array (DEFAULT CUDA/HIP comm backend) — main cause of the 19-create/16-destroy imbalance | `gs/bcknd/device/gs_device_mpi.F90:313-332` (created :301-304) | add event-destroy loop + `deallocate(this%event)` |
| 5 | `pipecg_device_free` frees `u_d(1..P_SPACE)` but init maps `1..P_SPACE+1` → leaks `u_d(11)` | `krylov/bcknd/device/pipecg_device.F90:283` | free loop bound → `P_SPACE+1` |
| 6 | `field_dirichlet_vector_free` never calls `free_base()` — leaks `msk`/`facet` + `msk_d`/`facet_d` + addrtbl | `bc/field_dirichlet_vector.f90:130-142` | add `call this%free_base()` |
| 7 | `neumann_free` never frees `flux(:)` (`vector_t`, no `FINAL`) — each component's device buffer leaks; ×3 inside every shear-stress / wall-model BC | `bc/neumann.f90:331-336` | loop-free `flux(:)` then deallocate |
| 8 | `spalding_free` never frees `nu` (`vector_t`) device buffer | `wall_models/spalding.f90:203-209` | `call this%nu%free()` |
| 9 | `fluid_scheme_compressible_free` never frees `bcs_vel`/`bcs_prs` bc lists (facet masks carry device mem) | `fluid/fluid_scheme_compressible.f90:228-292` | iterate+free both lists (mirror incompressible) |
| 10 | `gs_device_nccl_free` / `gs_device_shmem_free` leak per-neighbour event arrays (+ NVSHMEM notify buffers) | `gs_device_nccl.F90:240`, `gs_device_shmem.F90:270` | add event-destroy loop + deallocate (non-default backends) |
| 11 | `gs_device_free` releases local mapped pointers with `device_free` instead of `device_unmap` → stale addrtbl entries (dangling `device_get_ptr` risk) | `gs/bcknd/device/gs_device.F90:193-219` | use `device_unmap` |
| 12 | `scalar_pnpn` adds redundant registry fields `<name>_abx1/_abx2` that `free()` never removes | `scalar/scalar_pnpn.f90:215-220` | drop the redundant `add_field` (abx1/abx2 already own the data) |
| 13 | `runge_kutta_scheme_coeffs_init` doesn't `free()` before re-alloc/`device_map` → re-init aborts / leaks | `time_schemes/runge_kutta_scheme.f90:75-90` | `call this%free()` first |

Plus minor: `jit.c:66` leaks the build-log `malloc` on kernel-build failure (one-off, fatal
path); `vreman_free` doesn't deallocate `scalar_name` (auto-frees on scope exit; cosmetic).

**Performance (not a leak, but explains a high OpenCL baseline):** the OpenCL math/operator
wrappers `clCreateKernel`+`clReleaseKernel` on **every call** (`math.c:345`, `add2s2`, `col3`,
`glsc3`, `ax_helm`, …) instead of caching the kernel once like `math_program`. Constant
per-call overhead — worth fixing for OpenCL throughput.

---

## 4. What we ruled out (verification did its job)

Three findings a naive audit would have shipped were **refuted** by adversarial re-check:

- **`field_assign_field` as a per-step dofmap leak** — the leaking branch is unreachable in
  the hot path (the only internal-dofmap fields are always assigned from internal-dofmap
  sources, which take the safe branch). Latent only (kept in §2e).
- **HDF5 dataset/property-list handle imbalance** (`io/hdf5_file.F90`) — handles are balanced
  on every reachable path; the raw count double-counts mutually-exclusive overwrite branches,
  and `h5close_f` reclaims per call. Cosmetic at most, opt-in build.
- **runtime-statistics as a *default* cause** — the growth is real but gated off by default
  (§1e).

---

## 5. Recommended immediate actions

1. **Fix the two OOB/uninitialized correctness bugs regardless** (§2a `fusedcg`/`fusedcg_cpld`
   loop bound, §2b `n_GL`): these are latent memory corruption that can cause delayed,
   hard-to-reproduce deaths on any backend.
2. If you enabled `runtime_statistics`, disable it or apply the accumulator fix (§1e) as a
   quick test — if the slowdown vanishes, that was it.
3. The §3 destructor leaks matter if your case rebuilds gs/AMG/preconditioners mid-run
   (AMG/`hsmg`+`tamg`, moving mesh/ALE, frequent restarts). `phmg_free` (empty no-op) and the
   `hsmg` `amg_solver` leak are the biggest of these.

All fixes above are 1–15 lines each and none change numerics.

njansson:
Addresses the memory-leak and resource-bug audit in #2632. All findings were re-verified against `develop` before fixing; claims already resolved by #2541 (glsc3 events) or stale against current code (gs backend event/buffer frees) are not touched.

**Correctness**
- Fix out-of-bounds init loops in `fusedcg_device` / `fusedcg_cpld_device` (wrote past the size-10 p-space arrays)
- Fix uninitialized `n_GL` in `opr_device_convect_scalar` (copy-paste double assignment of `n_GLL`)
- Fix under-allocated NVSHMEM CFL buffer: local work buffer is now `cudaMalloc`'d (`nel` is not symmetric across PEs), with a separate one-element symmetric buffer for the reduction
- Rework Metal event protocol: event handles now carry the recorded signal target, so `event_sync` / `stream_wait_event` wait on the recorded value instead of racing on stale `signaledValue` reads
- Fix latent dofmap leak/double-free in `field_assign_field` when repointing an internal dofmap to an external one

**Resource leaks**
- OpenCL: release `kern_wait` in `opencl_cfl` (one event per step), release kernel in `opencl_absval`, free build log in `jit.c`
- Metal: wrap command-buffer sync/record/wait paths in `@autoreleasepool`
- Implement the empty `phmg_free`; free `amg_solver` in `hsmg_free`
- Free missed members in destructors: pipecg `u_d(P_SPACE+1)`, neumann `flux(:)`, spalding `nu`, `free_base` in `field_dirichlet_vector`, bc objects in the compressible scheme (`bcs_vel`/`bcs_prs`/`bcs_density`)
- Remove stale device address-table entries for the lazily-mapped gs arrays in `gs_free`
- Remove duplicate `_abx1`/`_abx2` registry fields in `scalar_pnpn` (checkpointing uses the member fields directly)
- Guard `runge_kutta_scheme_coeffs_init` against re-init

No changes to numerics. Fixes #2632.


src/krylov/bcknd/device/pipecg_device.F90
```diff
@@ -280,14 +280,17 @@ subroutine pipecg_device_free(this)
     end if
     if (allocated(this%u)) then
        if (allocated(this%u_d)) then
-          do i = 1, DEVICE_PIPECG_P_SPACE
+          do i = 1, DEVICE_PIPECG_P_SPACE+1
```
Mr-2c, line 283: is this correct? :-) the previous one was the opposite

Mr-2c (approved):
really above my pay grade to judge, but all of it makes sense to me

src/krylov/bcknd/device/pipecg_device.F90
njansson: It is kind of a mess, in this case yes, since `u_d`  had size (`P_SPACE+1`) while in `fused_cg` it was `P_SPACE`. I'll give it another look if this can be unified to reduce confusion.

njansson: Ok, it should be different between both methods, `pipecg` is 11. so All good

timfelle (approved):

PR #2637, opened 2026-07-09
Add workaround patch for Fujitsu Fortran tp-procedure memleaks

njansson:
This adds a workaround patch to resolve a compiler related memory leak when Neko has been compiled with Fujitsu Fortran. (see #2638 )

### DIFF
--- a/patches/fujitsu_memleak.patch
+++ b/patches/fujitsu_memleak.patch
@@ -1,0 +1,370 @@
+diff --git a/src/.depends b/src/.depends
+index fc5aa5bb01b..1c36e89dbde 100644
+--- a/src/.depends
++++ b/src/.depends
+@@ -463,4 +463,5 @@ wall_models/richardson.lo : wall_models/richardson.f90 device/device.lo math/bck
+ wall_models/bcknd/cpu/richardson_cpu.lo : wall_models/bcknd/cpu/richardson_cpu.f90 math/math.lo common/log.lo common/utils.lo config/num_types.lo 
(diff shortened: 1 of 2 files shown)


patches/fujitsu_memleak.patch
```diff
@@ -0,0 +1,370 @@
...
++    class(stack_2i4r8t3_t), pointer :: tmp_stack_2i4r8t3_t
++    class(stack_curve_t), pointer :: tmp_stack_curve_t
++    class(stack_nq_t), pointer :: tmp_stack_nq_t
++    class(stack_nh_t), pointer :: tmp_stack_nh_t
++    class(stack_nz_t), pointer :: tmp_stack_nz_t
++    class(stack_nc_t), pointer :: tmp_stack_nc_t
++    class(stack_pt_t), pointer :: tmp_stack_pt_t
++    class(uset_i4_t), pointer :: tmp_uset_i4_t
++    class(uset_i8_t), pointer :: tmp_uset_i8_t
++    class(uset_r8_t), pointer :: tmp_uset_r8_t
```
timfelle, line 145: I would probably add this in a stack routine similar to the htable one, but up to you.

timfelle (approved):

patches/fujitsu_memleak.patch
njansson: Let's keep it here, the type specific things are special cases (as is this entire mess...)

njansson:
I need to add it to the dist list as well

PR #2641, opened 2026-07-10
fixed stale facet-normals in pressure surface terms (ALE-rotation)

arash-mp:
In facet_normal_t, normal vectors are only initialized once, and they are not linked to normals from coef. 
So for ALE problems involving rotation, it was giving wrong results since `this%bc_prs_surface` in `prs_res%compute` was using stale normal vectors. This PR  fixes the problem. 

I also put the recompute_normals for `symmetry`. Right now, symmetry BCs should be axis aligned, so their normal vectors do not change. But I put it their so in future we won't forget to fix that when we have non-axis aligned BCs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Develop
 
+- Fixed stale facet-normals in fluid_pnpn pressure surface terms.
+  This was only affecting ALE simulations containing rotations.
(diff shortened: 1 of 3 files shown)


timofeymukha (approved):

timofeymukha:
symmetry seems untouched, which I am a bit happy about. I would really want to merge the mixed bc pr before we touch those again :D

timfelle (approved):

PR #2642, opened 2026-07-10
Fix cuda issue when using dealiasing for P > 9

Shiyu-Sandy-Du:
Algorithmically it makes no difference. However, cuda has some memory management issues when the local shwork2 gets too large (maybe @njansson knows much more). The fix here is to allocate a global array and use it with offset for each thread. For the sake of code consistency, work1 is also allocated in this way instead of declared in the kernel. The result is shown by a cylinder case at Re=100 and t=2.0
Before the fix:
before0000

 After the fix:
after0000

### DIFF
--- a/src/math/bcknd/device/cuda/tensor.cu
+++ b/src/math/bcknd/device/cuda/tensor.cu
@@ -63,12 +63,18 @@ extern "C" {
     break
 
 #define CASE_LARGE(N)                                                        \
-    case N:                                                                  \
+    case N: {                                                                \
+    const size_t shmem_size = 2 * N * N * N * sizeof(real);                  \
+    CUDA_CHECK(cudaFuncSetAttribute(tnsr3d_kernel_large<real, N>,            \
+                                    cudaFuncAttributeMaxDynamicSharedMemorySize, \
+                                    shmem_size));                            \
     tnsr3d_kernel_large<real, N>                                             \
-      <<<nblcks, nthrds, 0, stream>>>((real *) v, *nv,                       \
-                                      (real *) u, *nu,                       \
-                                      (real *) A, (real *) Bt, (real *) Ct); \
+      <<<nblcks, nthrds, shmem_size, stream>>>((real *) v, *nv,              \
+                                               (real *) u, *nu,              \
+                                               (real *) A, (real *) Bt,      \
+                                               (real *) Ct);                 \
     CUDA_CHECK(cudaGetLastError());                                          \
+    }                                                                        \
     break
 
     switch(n) {
--- a/src/math/bcknd/device/cuda/tensor.cu
+++ b/src/math/bcknd/device/cuda/tensor.cu
@@ -42,6 +42,17 @@
      __typeof__ (b) _b = (b);   \
      _a > _b ? _a : _b; })
 
+template<const int N>
+static void set_tnsr3d_large_shmem_attr(const size_t shmem_size) {
+  static bool is_set = false;
+
+  if (!is_set) {
+    CUDA_CHECK(cudaFuncSetAttribute(tnsr3d_kernel_large<real, N>,
+                                    cudaFuncAttributeMaxDynamicSharedMemorySize,
+                                    shmem_size));
+    is_set = true;
+  }
+}
 
 extern "C" {
 
@@ -65,9 +76,7 @@ extern "C" {
 #define CASE_LARGE(N)                                                        \
     case N: {                                                                \
     const size_t shmem_size = 2 * N * N * N * sizeof(real);                  \
-    CUDA_CHECK(cudaFuncSetAttribute(tnsr3d_kernel_large<real, N>,            \
-                                    cudaFuncAttributeMaxDynamicSharedMemorySize, \
-                                    shmem_size));                            \
+    set_tnsr3d_large_shmem_attr<N>(shmem_size);                              \
     tnsr3d_kernel_large<real, N>                                             \
       <<<nblcks, nthrds, shmem_size, stream>>>((real *) v, *nv,              \
                                                (real *) u, *nu,              \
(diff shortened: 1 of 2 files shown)


njansson:
No this needs more work, we shouldn't not default to slow memory for cases where it fits in shared memory.

Let me check how it can be fix

njansson (changes requested):
As mentioned in the comment, we shouldn't default to slow memory unless necessary.

Options forward is to either special case that or the best option to increase shared memory limits

njansson:
One obvious issue with the (old) code code is that one of the arrays is not marked shared.

shiyu-du-dsy:
Now it uses dynamic shared-memory to handle the issue.

src/math/bcknd/device/cuda/tensor_kernel.h
```diff
@@ -159,8 +159,8 @@ __global__ void tnsr3d_kernel_large(T  * __restrict__  v,
                                     const T * __restrict__ A,
                                     const T * __restrict__ Bt,
                                     const T * __restrict__ Ct) {
-  __shared__ T shwork[N*N*N];
-  T shwork2[N*N*N];
+  extern __shared__ T shwork[];
+  T *shwork2 = shwork + N*N*N;
```
njansson: To improve readability, I would suggest something along the line of:
```suggestion
  extern __shared__ T shmem[];
  T *shwork = shmem;
  T *shwork2 = shmem + N*N*N;
```

src/math/bcknd/device/cuda/tensor.cu
```diff
@@ -63,12 +63,18 @@ extern "C" {
     break
 
 #define CASE_LARGE(N)                                                        \
-    case N:                                                                  \
+    case N: {                                                                \
+    const size_t shmem_size = 2 * N * N * N * sizeof(real);                  \
+    CUDA_CHECK(cudaFuncSetAttribute(tnsr3d_kernel_large<real, N>,            \
+                                    cudaFuncAttributeMaxDynamicSharedMemorySize, \
+                                    shmem_size));                            \
```
njansson: Since it is persistent per kernel, it should be done once to avoid the costly overhead (remember these tnsr kernels are pretty small, this the overhead might be almost as much as the kernel it self)

Shiyu-Sandy-Du: okay now it's done by a static variable

njansson (approved):

arash-mp (approved):

src/math/bcknd/device/cuda/tensor.cu
```diff
@@ -42,6 +42,17 @@
...
+static void set_tnsr3d_large_shmem_attr(const size_t shmem_size) {
+  static bool is_set = false;
+
+  if (!is_set) {
+    CUDA_CHECK(cudaFuncSetAttribute(tnsr3d_kernel_large<real, N>,
+                                    cudaFuncAttributeMaxDynamicSharedMemorySize,
+                                    shmem_size));
+    is_set = true;
+  }
+}
```
timfelle, line 55: This will leak with no option to actually fix it. Please add a deallocator and
make sure it is called when neko finalizes.

timfelle (changes requested):

src/math/bcknd/device/cuda/tensor.cu
Shiyu-Sandy-Du: now it's fixed

timfelle (approved):

njansson (commented):
Hate to be that person, but there must be a better way than this wiring, now we are mixing CUDA interface code with direct calls to our kernels.

shiyu-du-dsy:
> Hate to be that person, but there must be a better way than this wiring, now we are mixing CUDA interface code with direct calls to our kernels.

Maybe we could remove the most recent commit? Since the shared memory `shmem` seems like to be automatically managed by cuda itself... and the previous PR is not related to memory management anyway

njansson:
> > Hate to be that person, but there must be a better way than this wiring, now we are mixing CUDA interface code with direct calls to our kernels.
> 
> Maybe we could remove the most recent commit? Since the shared memory `shmem` seems like to be automatically managed by cuda itself... and the previous PR is not related to memory management anyway

Yes, sounds good. But we still have to make sure that we don't call the set attribute function before each kernel launch.

shiyu-du-dsy:
> > > Hate to be that person, but there must be a better way than this wiring, now we are mixing CUDA interface code with direct calls to our kernels.
> > 
> > 
> > Maybe we could remove the most recent commit? Since the shared memory `shmem` seems like to be automatically managed by cuda itself... and the previous PR is not related to memory management anyway
> 
> Yes, sounds good. But we still have to make sure that we don't call the set attribute function before each kernel launch.

Yes it is called each time we call the kernel but only be executed once for the first time

njansson:
> > > > Hate to be that person, but there must be a better way than this wiring, now we are mixing CUDA interface code with direct calls to our kernels.
> > > 
> > > 
> > > Maybe we could remove the most recent commit? Since the shared memory `shmem` seems like to be automatically managed by cuda itself... and the previous PR is not related to memory management anyway
> > 
> > 
> > Yes, sounds good. But we still have to make sure that we don't call the set attribute function before each kernel launch.
> 
> Yes it is called each time we call the kernel but only be executed once for the first time

Yeah, but we should add some checks such that we don't call the API for each kernel invocation.

njansson:
> > > > > Hate to be that person, but there must be a better way than this wiring, now we are mixing CUDA interface code with direct calls to our kernels.
> > > > 
> > > > 
> > > > Maybe we could remove the most recent commit? Since the shared memory `shmem` seems like to be automatically managed by cuda itself... and the previous PR is not related to memory management anyway
> > > 
> > > 
> > > Yes, sounds good. But we still have to make sure that we don't call the set attribute function before each kernel launch.
> > 
> > 
> > Yes it is called each time we call the kernel but only be executed once for the first time
> 
> Yeah, but we should add some checks such that we don't call the API for each kernel invocation.

my bad all good now, lgtm

PR #2643, opened 2026-07-10
Flush log output during initialization, around file output, and on errors

Mr-2c:
## Motivation

When a large case crashes during initialization, the log is often empty or
truncated: all output written via `neko_log` sits in the Fortran runtime's
I/O buffer, which is only drained on a clean exit. With stdout redirected to
a file (the normal situation in a batch job), or with `NEKO_LOG_FILE` set, a
segfault, OOM-kill, `error stop`, or scheduler kill discards everything still
buffered — precisely the lines needed to diagnose the failure. The code base
currently contains no `flush` statement at all; the only existing mitigation
is launcher-specific (e.g. `srun --unbuffered`), which does not help
`NEKO_LOG_FILE` and is easy to forget.

This PR makes the log durable at the moments that matter, using deterministic,
event-based flushing only. There is **no timer-based or per-time-step
flushing**: the time loop is unaffected except when an output file is written.

A Fortran `FLUSH` pushes buffered output out of the process into the OS, which
is exactly what is needed to survive a process-level death (crash, abort,
kill). It is not an `fsync` and does not protect against a node hard-crash or
power loss; that is out of scope.

### 1. Flushing infrastructure in the logger (`src/common/log.f90`)

- New type-bound procedure `neko_log%flush()`: flushes the log unit. It is
  local to rank 0 (the only rank that ever writes the log) and never
  collective, so it is safe to call from any rank and cannot deadlock.
  The unit guard uses the `-1` sentinel set by `log_free` rather than a
  positivity test, since `NEWUNIT`-assigned units (used for `NEKO_LOG_FILE`)
  are negative.
- **Startup section-flushing**: a new internal flag (`section_flush_`,
  enabled from `log_init`) makes the logger flush whenever a section header
  is printed and whenever a section ends. Since every initialization step
  (mesh reading, gather-scatter setup, boundary conditions, solver setup,
  initial conditions, simulation components, restart, ...) logs its own
  section, the log on disk always shows what the code was attempting and
  what had completed, at every nesting depth — including deep inside
  `fluid%init`, where large cases most often fail.
- New type-bound procedure `neko_log%end_startup()`: performs a final flush
  and disables section flushing. It is called at the end of
  `simulation_init`, so the complete startup log (banner through
  `Preprocessing` and the first output) is on disk right before the time
  loop starts, and **zero** section flushes occur inside the loop.
- The banner (`log_header`) is flushed as soon as it is written.
- `log_error` now flushes the buffered log *before* writing to stderr (so
  errors can no longer appear before earlier log lines in a combined
  stream) and flushes stderr afterwards; `log_warning` flushes after the
  warning. The C-interface entry points (`log_error_c`, `log_warning_c`)
  received the same treatment.

### 2. Flush before aborting (`src/common/utils.f90` and direct abort sites)

All four error routines (`neko_error_plain`, `neko_error_msg`,
`neko_type_error`, `neko_type_registration_error`) flush `output_unit`
before writing the error and flush `error_unit` before `error stop`.
Whether `ERROR STOP` itself flushes open units is processor-dependent:
GNU Fortran happens to flush, but the standard does not require it, so
other compilers (e.g. Cray) cannot be relied upon — and no runtime can
flush after a segfault, OOM-kill, or a launcher-initiated kill triggered
by another rank aborting. Flushing `output_unit` first also guarantees
buffered log lines appear *before* the error in a combined stream.
`neko_warning` flushes after the warning.

The boundary-condition validation errors in `fluid_pnpn.f90` (3 sites) and
`scalar_pnpn.f90` (4 sites) write to `error_unit` and `error stop` directly,
bypassing `neko_error` — and these are exactly the "misconfigured case dies
in init" errors. They received the same flush treatment, as did the usage
message in `neko.f90`.

### 3. Flush around file output (`src/io/output_controller.f90`,
`src/simulation.f90`)

- `output_controller_execute` flushes right before each output's `sample()`
  call (after the `File name`/`Output number` messages), so the log shows
  which file was being written if the run dies during a write, and once
  after the final `Writing at time ...` message, confirming completion.
  This covers field snapshots, checkpoints, and the statistics field files
  registered with the controller, including the forced final write in
  `simulation_finalize`. The post-batch flush sits after the pre-existing
  `MPI_Barrier`; no synchronization is added anywhere.
- The joblimit checkpoint (`simulation_joblimit_chkp`) now logs
  `! saving checkpoint >>>` *before* writing (previously after) and flushes
  around the write, so a job killed while dumping its final checkpoint
  leaves evidence of what it was doing.

Small CSV outputs written directly by simulation components (e.g.
`boundary_flux`) are deliberately not flushed — they are small, frequent
appends and not worth the noise.

## Behavior and performance

- **Time loop**: unchanged, except one flush before/after each file-output
  event. On steps with no output, no flushing happens at all.
- **Initialization**: a few dozen additional one-syscall flushes on rank 0
  in total (two per log section). Not measurable.
- **Errors/warnings**: two flushes per event; errors abort anyway and
  warnings are rare.
- No new options, environment variables, or log content changes (the only
  textual change is that the joblimit checkpoint message now appears before
  the write instead of after).
- Non-root ranks never flush the log (the method self-guards on
  `pe_rank == 0`); nothing added is collective.

## Known limitation

If a run crashes mid-time-loop *between* file outputs, buffered step
summaries since the last output are still lost (the log is intact up to and
including the last output event, and the error itself still reaches stderr).
Covering that window would require some form of periodic flushing inside the
loop, which is intentionally left out of this PR.

## Verification

- Full build with GNU + OpenMPI against json-fortran 8.4.0 using the CI
  flags (`-g -O2 -pedantic -std=f2008`): clean.
- Unit tests (`make check` with pFUnit v4.15.0): 43/43 pass.
- `findent` formatting check (CI settings) clean on all changed files;
  `flint` scores unchanged or improved on every changed file.
- Runtime checks with the small integration cylinder case (2 ranks),
  each scenario run on both this branch and unpatched `develop`:
  - **Normal run**: identical log content and outputs, `Normal end.`
  - **`SIGKILL` during initialization** (the reported symptom, simulated
    with a higher polynomial order and a kill 3 s in, `NEKO_LOG_FILE`
    set): `develop` leaves a **0-byte log**; this branch leaves the full
    startup log up to the exact operation in flight when the kill landed.
  - **`SIGKILL` mid-time-loop**: on `develop` the log ends at an
    arbitrary buffer cut and records *fewer* completed writes than
    checkpoint files actually on disk (17 vs 18 in the test); on this
    branch every completed write's confirmation is in the log (19/19)
    and the log ends at a clean output-event boundary.
  - **Invalid BC zone index** (controlled `error stop` in init): full
    startup log up to the `Fluid boundary conditions` step in the log
    file, error on stderr. (With GNU Fortran the runtime happens to
    flush on a controlled `error stop`, so `develop` also retains this
    log; the added flushes make the behavior guaranteed rather than
    processor-dependent, which matters for other compilers and for the
    stdout/stderr ordering.)


src/common/log.f90
```diff
@@ -154,13 +162,40 @@ subroutine log_free(this)
...
+  !> Mark the end of the startup (initialisation) phase of a simulation
+  !! @details Flushes the log and disables the per-section flushing used
+  !! during startup, keeping the time loop free of any flushing overhead.
+  subroutine log_end_startup(this)
+    class(log_t), intent(inout) :: this
+
+    call this%flush()
+    this%section_flush_ = .false.
+
+  end subroutine log_end_startup
```
timfelle: Remove, based on conversation earlier.

src/common/log.f90
```diff
@@ -154,13 +162,40 @@ subroutine log_free(this)
     this%indent_ = 0
     this%level_ = NEKO_LOG_INFO
     this%unit_ = -1
+    this%section_flush_ = .true.
```
timfelle: Remove, based on conversation

src/common/log.f90
```diff
@@ -127,6 +133,8 @@ subroutine log_init(this)
        this%unit_ = stdout
     end if
 
+    this%section_flush_ = .true.
```
timfelle: Remove based on conversation

src/common/log.f90
```diff
@@ -71,6 +75,8 @@ module logger
      procedure, pass(this) :: warning => log_warning
      procedure, pass(this) :: deprecated => log_deprecated
      procedure, pass(this) :: end_section => log_end_section
+     procedure, pass(this) :: flush => log_flush
+     procedure, pass(this) :: end_startup => log_end_startup
```
timfelle: Remove, update

src/common/log.f90
```diff
@@ -54,6 +54,10 @@ module logger
      integer, private :: tab_size_
      integer, private :: level_
      integer, private :: unit_
+     !> Flush the log at section boundaries (enabled during startup, disabled
+     !! once the time loop begins, such that a crash during initialisation
+     !! cannot swallow buffered log output)
+     logical, private :: section_flush_ = .true.
```
timfelle: Remove, based on conversation today

src/simulation.f90
```diff
@@ -94,6 +94,10 @@ subroutine simulation_init(C, dt_controller)
     call neko_log%end_section()
     call neko_log%newline()
 
+    ! Initialisation complete; persist the startup log and disable
+    ! per-section flushing before entering the time loop
+    call neko_log%end_startup()
```
timfelle: Just call flush

src/fluid/fluid_pnpn.f90
```diff
@@ -1032,21 +1032,25 @@ subroutine fluid_pnpn_setup_bcs(this, user, params)
...
 
              if (global_zone_size .eq. 0) then
+                flush(output_unit)
                 write(error_unit, '(A, A, I0, A, A, I0, A)') "*** ERROR ***: ",&
                      "Zone index ", zone_indices(j), &
                      " is invalid as this zone has 0 size, meaning it ", &
                      "is not in the mesh. Check fluid boundary condition ", &
                      i, "."
+                flush(error_unit)
                 error stop
```
timfelle: Just swap to neko_error

src/fluid/fluid_pnpn.f90
```diff
@@ -1032,21 +1032,25 @@ subroutine fluid_pnpn_setup_bcs(this, user, params)
...
              if (marked_zones(zone_indices(j))) then
+                flush(output_unit)
                 write(error_unit, '(A, A, I0, A, A, A, A)') "*** ERROR ***: ", &
                      "Zone with index ", zone_indices(j), &
                      " has already been assigned a boundary condition. ", &
                      "Please check your boundary_conditions entry for the ", &
                      "fluid and make sure that each zone index appears only ", &
                      "in a single boundary condition."
+                flush(error_unit)
                 error stop
```
timfelle: Swap to neko_error

src/fluid/fluid_pnpn.f90
```diff
@@ -1133,8 +1137,10 @@ subroutine fluid_pnpn_setup_bcs(this, user, params)
        do i = 1, size(this%msh%labeled_zones)
           if ((this%msh%labeled_zones(i)%size .gt. 0) .and. &
                (.not. marked_zones(i))) then
+             flush(output_unit)
              write(error_unit, '(A, A, I0)') "*** ERROR ***: ", &
                   "No fluid boundary condition assigned to zone ", i
+             flush(error_unit)
              error stop
```
timfelle: swap to neko error

adperezm:
How is this one going?  If the conversations are resolved, then you can click on them so it lets you merge.


Is Nvidia giving problems again? Maybe someone can rerun the workflows? It seems PRs after this one work. Perhaps also merge develop into this one.

njansson:
> How is this one going? If the conversations are resolved, then you can click on them so it lets you merge.
> 
> Is Nvidia giving problems again? Maybe someone can rerun the workflows? It seems PRs after this one work. Perhaps also merge develop into this one.

I think Nvidia workflows will be happy if this is rebased/sync. with latest develop

timofeymukha:
- Needs to be cleaned-up, merged with develop.
- Do not flush every section. Manual flush at specific points is a better way to do it.
- Flushing to stdout on error is not good?

timfelle (approved):
Update, log will flush:

- Before freeing the log.
- Before and after the main simulation loop.
- Before calling output sample.
- After sampling is complete.
- When joblimit triggers a checkpoint write.
- On errors.

timofeymukha (approved):

PR #2644, opened 2026-07-10
Add `*_light` contrib tools: low-memory, standalone mesh utilities

Mr-2c:
## What

This PR adds five standalone, CPU-only companions to the existing `contrib/`
tools, each in its own self-contained folder (tool + validation harness +
Makefile + doc), plus a top-level README with a dependency matrix:

| Tool | What it does | Needs |
|---|---|---|
| `rea2nbin_light` | Serial `.re2` → `.nmsh` converter, byte-exact with `rea2nbin`; periodic BCs and curved elements fully supported | gfortran |
| `genmeshbox_light` | O(1)-memory box-mesh generator, byte-exact with `genmeshbox`; streams the `.nmsh` without building a `mesh_t` | gfortran |
| `mesh_checker_light` | Low-memory `.nmsh` diagnostics: sizes (matching `glb_mfcs`/`glb_meds` incl. the periodic merge), BC zones, unlabeled external faces, `--jacobian`, `--write-zone-indices` | gfortran |
| `genmap_light` | Spectral-bisection mesh partitioner (Nek5000 `genmap` algorithm) that reorders the `.nmsh` like `prepart`; deterministic, dependency-free | gfortran |
| `prepart_light` | Mesh partitioner in scientific Python with three backends: spectral (default), METIS (`pip install pymetis`), and geometric RCB for very large meshes | python3 + numpy/scipy |

## Why

The existing `contrib` tools build a full `mesh_t` through Neko's
general-purpose data structures. Those are robust and reused everywhere, but on
large meshes (10⁷–10⁸ elements) the generously-sized polymorphic hash tables
and connectivity structures need hundreds of GB — for tasks (convert, check,
generate, partition) that only need to stream records. The recently merged
`lgenc` fix removes the unwanted connectivity build; these tools go further and
skip the `mesh_t` entirely: plain typed arrays, no Neko initialisation, the
same output in a small fraction of the memory.

Because they do not link Neko, they also need no Neko build at all — in
particular no separate CPU rebuild of a GPU-configured Neko (the `contrib`
tools are CPU-only), no MPI, and no ParMETIS for the partitioners.

## Validation

Each folder ships a validation harness (`make check`, designed to run from
`contrib/<tool>/` inside the source tree):

- **rea2nbin_light** — byte-exact against the bundled `hemi` golden pair and a
  real Neko-produced pipe mesh (the only differing bytes are the type-7 zone
  fields Neko's writer leaves uninitialised); round-trip geometry/topology
  43/43 meshes, periodic identification 19/19, curved records bit-exact 11/11.
- **genmeshbox_light** — 9/9 shipped boxes byte-exact, head-to-head byte-exact
  against a freshly built `genmeshbox`, Neko-compatible CSV distribution files.
- **mesh_checker_light** — all reported quantities match an independent
  recomputation on 43/43 meshes, including the periodic-merged face/edge counts
  (`glb_mfcs`/`glb_meds`); broken-mesh, Jacobian and `.fld`-layout tests.
- **genmap_light / prepart_light** — output verified to be a pure permutation
  of the input with curves and zones carried through (element references
  renumbered, payloads verbatim) and contiguous per-partition blocks, so a
  linear read reproduces the partition (the `prepart` contract); partition
  quality checked against an exact-eigensolver reference; exact balance;
  deterministic output.

All tools reject malformed/truncated inputs with clean one-line errors and
leave no partial output behind.

## Notes for reviewers

- Purely additive: no changes to any existing Neko source.
- All files carry the standard BSD-3 header (The Neko Authors).
- The four Fortran tools build with `gfortran -O2 -Wall` (warning-free); the
  Python tool targets python3 + numpy/scipy, with `pymetis`/`pyamg` optional.
- Each `.md` documents the algorithm, the validation evidence, and the scope /
  limitations (e.g. 3D hex only; `mesh_checker_light`'s Jacobian is exact for
  straight-sided elements), and notes that users should validate on a case they
  trust before relying on a tool in production.

### DIFF
--- a/contrib/README_neko_light_tools.md
+++ b/contrib/README_neko_light_tools.md
@@ -1,0 +1,44 @@
+```text
+     _  __  ____  __ __  ____
+    / |/ / / __/ / //_/ / __ \
+   /    / / _/  / ,<   / /_/ /
+  /_/|_/ /___/ /_/|_|  \____/
+```
+
+# Neko light tools
+
+Five standalone, CPU-only companions to Neko's `contrib/` tools. Each folder is
+self-contained (tool + validation harness + Makefile + doc) and is designed to
+sit at `contrib/<tool>/` inside a Neko source checkout, so that `make check` can
+find the meshes shipped with Neko (`../../examples`, `../../tests`). Using the
+tools on your own meshes needs no Neko tree at all.
+
+The common idea: Neko's own `contrib` tools build a full `mesh_t` through Neko's
+general-purpose, production data structures — robust and reused everywhere, but
+memory-hungry on large meshes, and CPU-only (so a GPU-configured Neko must be
+rebuilt for CPU just to run them). These reimplementations use plain typed
+arrays and no Neko initialisation: the same results in a small fraction of the
+memory, needing only a Fortran compiler or a Python interpreter.
+
+| Tool | What it does | Build / run needs | `make check` needs |
+|---|---|---|---|
+| `rea2nbin_light` | Serial `.re2` → `.nmsh` converter, byte-exact with `contrib/rea2nbin` (periodic BCs, curved elements) | gfortran | python3 |
+| `genmeshbox_light` | O(1)-memory box-mesh generator, byte-exact with `contrib/genmeshbox` | gfortran | python3 |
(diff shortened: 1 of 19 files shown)


Mr-2c:
I am closing this to be superseded by another one that is entirely python, addresses what we discussed, and has more functionalities.

PR #2646, opened 2026-07-13
Wall model sampling enhancements

timofeymukha:
Adds the possibility to set the sampling based on arbitrary distance in addition to the existing functionality, which samples from a gll node. Adds a user routine to set index / distance per-node.

Implemented by adding a wall_sampler type, which takes care of grabbing the data and a routine to pack sampled data for a given field into a vector_t.

Simple unit test coverage added. I also tested running a channel flow (in sp, but the way!) and the results are good.

On the JSON side there is now a `sampling` subdict to set, but the now legacy `h_index` remains supported.

I have deliberately designed the interface to support sampling from multiple nodes for each wall node. No wall model currently supports this, but I think this is the way to go. The data is in a 1D array, for multiple samples on just relies on idexing.

Previously, to check h, one could write the tau field from the registry, where the gll nodes sampled from were marked with -1. Now that we do not necessarily sample form a node this will not do. Instead, we save a dedicated field, which stores h as the distance (whatever sampling method). The output of the field can be suppressed via json.


src/fluid/bcknd/device/compressible_ops_device.F90
```diff
@@ -89,7 +89,7 @@ end subroutine hip_update_mxyz_p_ruvw
 
   interface
      subroutine hip_update_e(E_d, p_d, ruvw_d, gamma, n) &
-       bind(c, name = 'hip_update_e')
+          bind(c, name = 'hip_update_e')
```
Shiyu-Sandy-Du, line 92: irrelevant question to this PR, why are these influenced by formatting? (I assume the code would not pass the formatting pass in some other PRs, would they?

src/wall_models/wall_gll_sampler.f90
```diff
@@ -0,0 +1,334 @@
...
+             call neko_error('Wall GLL sampling index lies outside element')
+          end if
+
+          sample_idx(p) = linear_index(sample(1), sample(2), sample(3), &
+               sample(4), lx, ly, lz)
+
+          dx = coef%dof%x(sample(1), sample(2), sample(3), sample(4)) - xw
+          dy = coef%dof%y(sample(1), sample(2), sample(3), sample(4)) - yw
+          dz = coef%dof%z(sample(1), sample(2), sample(3), sample(4)) - zw
+          this%h%x(p) = -(dx*n_x%x(i) + dy*n_y%x(i) + dz*n_z%x(i))
```
Shiyu-Sandy-Du, line 281: maybe I missed it in the previous wall model code, but why it's not (dx*n_x)^2 + (dy*n_y)^2 + (dz*n_z)^2?

src/fluid/bcknd/device/compressible_ops_device.F90
timofeymukha: That is a great question, and I don't know. I think there something about how the formatter detects what is "touched", which can get funky?... Not sure

src/wall_models/wall_gll_sampler.f90
timofeymukha: dx, dy, dz is the distance vector from the wall to the sampling point, component wise, so h is the projection of that vector on to the normal direction, obtained with a dot product. The minus is due to the fact that the normal points outside the domain. So I think everything is correct, or? How do you get your formula with the squares?

src/wall_models/wall_sampler.f90
```diff
@@ -0,0 +1,203 @@
...
+  !! when samplers support multiple points per wall node.
+  !! @param coef SEM coefficients defining the output field.
+  !! @param msk Mask selecting local wall nodes.
+  !! @param bc_name Name of the owning boundary condition.
+  subroutine wall_sampler_output_h(this, coef, msk, bc_name)
+    class(wall_sampler_t), intent(inout) :: this
+    type(coef_t), intent(in) :: coef
+    integer, intent(in) :: msk(0:)
+    character(len=*), intent(in) :: bc_name
+    type(field_t), pointer :: h_field
```
Shiyu-Sandy-Du, line 169: I think it's good for now, but maybe we need later have the output for the 2D surface?

timofeymukha: Yeah, Arash had some plans to implement that I remember. Dumping a field is of course less than optimal...

njansson (approved):
🍻 Would be nice with a Metal backend as well🤘

shiyu-du-dsy (approved):

PR #2650, opened 2026-07-14
Add device reduction buffer type

### LINKED ISSUE #2210: Global interpolation unit tests fails for OpenCL backend
The test `glob_interp_par_suite.test_glob_interp_0_pts` fails using OpenCL, throwing,

`CL_INVALID_MEM_OBJECT in math/bcknd/device/opencl/math.c:1067`

Which is related to the reduction buffer in `opencl_glsc3`!?

njansson:
This PR adds a device reduction buffer type, removing several stale reductions buffers that were never freed during teardown.

Fixes #2210

### DIFF
--- a/src/device/cuda/buffer.cu
+++ b/src/device/cuda/buffer.cu
@@ -52,8 +52,9 @@ extern "C" {
     if (size <= buf->size)
       return;
 
-    if (buf->host != NULL) {
-      CUDA_CHECK(cudaFreeHost(buf->host));
+    if (buf->dev != NULL) {
+      if (buf->host != NULL)
+        CUDA_CHECK(cudaFreeHost(buf->host));
 #ifdef HAVE_NVSHMEM
       if (buf->symm)
         nvshmem_free(buf->dev);
@@ -64,7 +65,8 @@ extern "C" {
 #endif
     }
 
-    CUDA_CHECK(cudaMallocHost(&buf->host, size));
+    if (!buf->dev_only)
+      CUDA_CHECK(cudaMallocHost(&buf->host, size));
 #ifdef HAVE_NVSHMEM
     if (buf->symm)
       buf->dev = nvshmem_malloc(size);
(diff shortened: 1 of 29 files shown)


timfelle (commented):
I very much aplaude the initiative on this one.

njansson:
> I very much aplaude the initiative on this one.

Has been on the todo list since ages...the failing OpenCL and CUDA reductions was what pushed me to get it done with

src/device/cuda/buffer.cu
```diff
@@ -0,0 +1,111 @@
...
+    CUDA_CHECK(cudaMalloc(&buf->dev, size));
+#endif
+    buf->size = size;
+
+    if (!buf->registered) {
+      buf->next = buffer_registry;
+      buffer_registry = buf;
+      buf->registered = 1;
+    }
+  }
```
timfelle, line 85: You already provide a registry of the buffers to keep track of this. Is there a
reason why you keep allocating and deallocating the same buffer every time you
need a reduction of a different size, instead of having multiples of different
lengths which can actually be reused?

njansson: we can't safely reuse buffers since some of the consumer reductions are running concurrently (i.e. different hsmg streams)

njansson: But something to cleanup later as well, post 1.1 I would say

timfelle: But you do? I can make a list of 5 calls to vlsc3 and then the buffer gets reallocated at every call if i am unlucky with the size ordering.

Then again, it is quite minimal since it is only on expansion that the buffer gets resized.

njansson: > But you do? I can make a list of 5 calls to vlsc3 and then the buffer gets reallocated at every call if i am unlucky with the size ordering.
> 

There are some ksp reductions also incoming that could be executed at the same time as math reductions

> Then again, it is quite minimal since it is only on expansion that the buffer gets resized.

timfelle (approved):
Well it runs on my system :thumbs_up:

adperezm (approved):
Cool!

PR #2657, opened 2026-07-15
Fix target lifetime issues in ALE checkpoint field registration

timfelle:
Address a target lifetime issue in the `register_checkpoint_fields` subroutine by adding the `target` attribute to the `this` dummy argument. This change prevents undefined behavior in ALE simulations that utilize checkpointing. Document this bug class in the PR-review checklist to enhance awareness and prevent future occurrences.


njansson (approved):

arash-mp (approved):
Thanks for finding this👍👍

Issue #2633, opened 2026-07-08
Reducing host memory footprint as a way to optimize memory usage on unified memory architectures


Mr-2c:
# Neko host/device memory-duplication audit

**Scope:** the `develop` branch. **Goal:** find host memory that is duplicated with a
GPU device copy and can be reclaimed, with an emphasis on systems that share physical
CPU/GPU memory (NVIDIA GH200 Grace-Hopper, AMD MI300A, Apple Silicon/Metal, Jetson,
integrated GPUs).

This document is the write-up of an automated, exhaustive sweep of `src/` (228
`device_map` sites across 45 files; 208 individual duplicated arrays inventoried and
classified). Every non-trivial claim below is backed by a `file:line` citation.

---

## 1. Bottom line

On a device build, **every array that Neko maps to the GPU is stored twice**: a full
Fortran `allocatable` on the host *and* an identical device buffer. On a discrete GPU
that is unavoidable (separate physical memories). **On shared-memory hardware the host
twin is pure waste** — it shadows byte-for-byte identical device data in the *same*
physical RAM.

- **208 duplicated arrays** were inventoried. **124 of them are never read on the host
  after the initial staging** — they are pure device-only data on a GPU build. A further
  **29 are used only as transient I/O staging buffers**. Only ~47 are genuinely read on
  the host during time-stepping (the "keep/managed" set).
- At a representative size (fp64, polynomial order `p=7` so `lx=ly=lz=8`, `nelv=32768`
  elements/GPU), **one per-DOF field = 128 MiB**. A full-featured run (pnpn + 1 scalar +
  GMRES pressure + solution projection + fluid statistics) duplicates on the order of
  **40–60 GiB of host RAM per GPU** that merely mirrors device data.
- **The single easiest big win for shared-memory systems** is a *central* unified-memory
  zero-copy seam in `src/device/device.F90`: make `device_map` wrap the existing host
  allocation as the device buffer instead of allocating a second buffer and copying.
  This de-duplicates **all 228 map sites with zero call-site edits** and reclaims
  essentially the entire duplicated working set on GH200/MI300A/Metal.
- **Metal (Apple Silicon) is already fully unified memory but still pays the 2× tax**
  today (`newBufferWithLength` + `memcpy`, `metal.m:161-206`). Fixing just the Metal
  backend is a self-contained quick win.
- **The portable, all-GPU win** is migrating the 124 device-only arrays to the new
[...]

PR #2658, opened 2026-07-15
Add support for unified memory in the Metal backend

njansson:
This PR add support for unified memory in the Metal backend. On supported platforms (Apple Silicon) the host pointer will be wrapped to a device pointer, with zero copies, while on unsupported platforms (Intel Macs), it follows the normal behaviour of replicating the arrays on the device. This reduces the host memory footprint (ref. #2633).


timfelle (approved):
LGTM

timofeymukha (approved):

PR #2659, opened 2026-07-16
Fix HDF5 context initialization

timfelle:
Remove the HDF5 session management and directly call the HDF5 open and close functions. This change simplifies the initialization and finalization of the HDF5 library within the application.

Acording to my reading, the h5open_f and h5close_f will keep track of an internal counter for the context and only close everything at 0. So it should be safe to have a Neko wide pair and then have it in each of our function so they can be used standalone.


src/io/hdf5_file.F90
```diff
@@ -714,9 +694,10 @@ subroutine hdf5_file_open(this, mode)
     counter = this%get_counter() - this%get_start_counter()
 
     ! Set the configuration for MPI IO
+    call h5open_f(ierr)
+
     mpi_info = MPI_INFO_NULL%mpi_val
-    mpi_comm = NEKO_COMM%mpi_val
-    call hdf5_session_start(ierr)
+    i_comm = NEKO_COMM%mpi_val
```
njansson:
```suggestion
    mpi_comm = NEKO_COMM%mpi_val
```

Typo?

njansson:
Great, seems to fix the issues I had locally. Will try on Fugaku once it's back running normal jobs

timfelle:
> Great, seems to fix the issues I had locally. Will try on Fugaku once it's back running normal jobs

Great, can ypu also have a look at the preprocessor errors from the ci?

njansson:
> > Great, seems to fix the issues I had locally. Will try on Fugaku once it's back running normal jobs
> 
> Great, can ypu also have a look at the preprocessor errors from the ci?

Yes, will do

src/io/hdf5_file.F90
timfelle: Deleted too much 😅

adperezm (approved):

njansson (approved):
All good on Fugaku, LGTM

PR #2662, opened 2026-07-16
Add support for unified memory on MI300A and GH200

### LINKED ISSUE #2633: Reducing host memory footprint as a way to optimize memory usage on unified memory architectures

njansson:
This PR add support for unified memory in the CUDA and HIP backends. On systems where CPU and GPU share coherent memory, `device_map` now aliases the host array directly instead of allocating a replicated device buffer. Host–device copies of mapped arrays become no-ops, roughly halving the memory footprint of mapped data and eliminating map-related transfer traffic.

- On APUs with XNACK (e.g. MI300A) the device pointer is the host pointer.
- On coherent memory platforms (e.g. GH200) mapped arrays are prefetched to HBM with preferred-location advice, staying host-accessible over NVLink-C2C.

Other discrete or PCIe-attached GPUs keeps the current replicated buffers.

The mapping, can be disabled via `NEKO_HIP_ZEROCOPY=0` / `NEKO_CUDA_ZEROCOPY=0`

Fixes #2633

### DIFF
--- a/src/multigrid/tree_amg.f90
+++ b/src/multigrid/tree_amg.f90
@@ -43,7 +43,7 @@ module tree_amg
   use ax_product, only: ax_t
   use bc_list, only: bc_list_t
   use gather_scatter, only : gs_t, GS_OP_ADD
-  use device, only: device_map, device_unmap, &
+  use device, only: device_map, device_unmap, device_sync, &
        device_stream_wait_event, glb_cmd_queue, glb_cmd_event
   use neko_config, only: NEKO_BCKND_DEVICE
   use, intrinsic :: iso_c_binding
@@ -204,6 +204,9 @@ contains
        call device_cfill(tamg_lvl%wrk_in_d, 0.0_rp, ndofs)
        call device_map(tamg_lvl%wrk_out, tamg_lvl%wrk_out_d, ndofs)
        call device_cfill(tamg_lvl%wrk_out_d, 0.0_rp, ndofs)
+       ! Order the async fills against host writes; on unified memory
+       ! the device pointers may alias the work arrays
+       call device_sync()
     end if
   end subroutine tamg_lvl_init
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Develop
 
+- Added zero-copy `device_map` on unified memory architectures (Apple Silicon
+  Metal, MI300A HIP, Grace Hopper CUDA): mapped arrays alias their host
+  allocation instead of being replicated, halving their footprint; discrete
+  GPUs are unaffected. Disable with `NEKO_{METAL,HIP,CUDA}_ZEROCOPY=0`.
 - Removed the hardcoded `CUDA_ARCH="-arch sm_60"` fallback used for
   `--enable-device-mpi` CUDA builds, which CUDA >= 13 toolkits can no longer
   compile (Pascal support was dropped). `configure` now requires `CUDA_ARCH`
--- a/doc/pages/appendices.md
+++ b/doc/pages/appendices.md
@@ -23,6 +23,7 @@ of the code. But can be useful for users and developers alike.
 | `NEKO_GS_COMM`           | Gather-scatter communication backend                                  | Unset         |
 | `NEKO_GS_CAF_SIGNALING`  | Coarray Fortran gather-scatter signaling mode                         | Unset         |
 | `NEKO_COMM_ID`           | Communicator id for this process (non-negative integer)               | 0             |
+| `NEKO_METAL_ZEROCOPY`    | Zero-copy host/device mapping on unified memory (Metal), 0 disables   | 1             |
 | `NEKO_MPI_THREAD_LEVEL`  | Requested MPI (and SHMEM) thread support level                        | Unset         |
 | `NEKO_DEPRECATION_ERROR` | Whether to treat deprecated features as errors (boolean)              | Unset         |
 
--- a/doc/pages/appendices.md
+++ b/doc/pages/appendices.md
@@ -23,6 +23,7 @@ of the code. But can be useful for users and developers alike.
 | `NEKO_GS_COMM`           | Gather-scatter communication backend                                  | Unset         |
 | `NEKO_GS_CAF_SIGNALING`  | Coarray Fortran gather-scatter signaling mode                         | Unset         |
 | `NEKO_COMM_ID`           | Communicator id for this process (non-negative integer)               | 0             |
+| `NEKO_CUDA_ZEROCOPY`     | Zero-copy host/device mapping on unified memory (CUDA), 0 disables    | 1             |
 | `NEKO_HIP_ZEROCOPY`      | Zero-copy host/device mapping on unified memory (HIP), 0 disables     | 1             |
 | `NEKO_METAL_ZEROCOPY`    | Zero-copy host/device mapping on unified memory (Metal), 0 disables   | 1             |
 | `NEKO_MPI_THREAD_LEVEL`  | Requested MPI (and SHMEM) thread support level                        | Unset         |
--- a/doc/pages/appendices.md
+++ b/doc/pages/appendices.md
@@ -23,6 +23,7 @@ of the code. But can be useful for users and developers alike.
 | `NEKO_GS_COMM`           | Gather-scatter communication backend                                  | Unset         |
 | `NEKO_GS_CAF_SIGNALING`  | Coarray Fortran gather-scatter signaling mode                         | Unset         |
 | `NEKO_COMM_ID`           | Communicator id for this process (non-negative integer)               | 0             |
+| `NEKO_HIP_ZEROCOPY`      | Zero-copy host/device mapping on unified memory (HIP), 0 disables     | 1             |
 | `NEKO_METAL_ZEROCOPY`    | Zero-copy host/device mapping on unified memory (Metal), 0 disables   | 1             |
 | `NEKO_MPI_THREAD_LEVEL`  | Requested MPI (and SHMEM) thread support level                        | Unset         |
 | `NEKO_DEPRECATION_ERROR` | Whether to treat deprecated features as errors (boolean)              | Unset         |
--- a/doc/pages/user-guide/installation.md
+++ b/doc/pages/user-guide/installation.md
@@ -245,7 +245,7 @@ $ ./configure  --with-cuda=/usr/local/cuda CUDA_CFLAGS=-O3  CUDA_ARCH=-arch=sm_8
   cluster's GPUs when cross-compiling.
 * Build using `make && make install`
 
-@note On cache-coherent CPU-GPU platforms (e.g. NVIDIA Grace Hopper), mapped arrays alias their host allocation zero-copy: host and device share a single allocation instead of keeping replicated copies, and host-device transfers become no-ops. Mapped arrays are prefetched to device memory (HBM) and stay resident there, remaining accessible from the host over the coherent interconnect. Zero-copy mapping can be disabled at runtime by setting the environment variable `NEKO_CUDA_ZEROCOPY=0`, which restores fully replicated buffers. Conventional PCIe-attached GPUs always use replicated buffers.
+@note On cache-coherent CPU-GPU platforms (e.g. NVIDIA Grace Hopper), mapped arrays alias their host allocation zero-copy: host and device share a single allocation instead of keeping replicated copies, and host-device transfers become no-ops. Mapped arrays are prefetched to device memory (HBM) and stay resident there, remaining accessible from the host over the coherent interconnect. Zero-copy mapping can be disabled at runtime by setting the environment variable `NEKO_CUDA_ZEROCOPY=0`, which restores fully replicated buffers. Conventional PCIe-attached GPUs always use replicated buffers. Since host and device share one allocation under zero-copy, host code (e.g. user routines) must not write to a mapped array while device work touching it may be in flight; synchronize with `device_sync` first.
 
 #### Compiling Neko for AMD GPUs
 To compile Neko for AMD GPUs
@@ -259,7 +259,7 @@ $ ./configure  --with-hip=/opt/rocm/hip
 $ ./configure  --with-hip=/opt/rocm/hip HIP_HIPCC_FLAGS=-O3  HIPCC=/opt/rocm/hip/bin/hipcc
 ```
 
-@note On APUs with unified physical memory (e.g. AMD Instinct MI300A) and XNACK enabled (`HSA_XNACK=1`), mapped arrays alias their host allocation zero-copy: host and device share a single allocation instead of keeping replicated copies, and host-device transfers become no-ops. This roughly halves the memory footprint of mapped data. Zero-copy mapping can be disabled at runtime by setting the environment variable `NEKO_HIP_ZEROCOPY=0`, which restores fully replicated buffers. Discrete GPUs (e.g. MI250X) always use replicated buffers, regardless of the XNACK setting.
+@note On APUs with unified physical memory (e.g. AMD Instinct MI300A) and XNACK enabled (`HSA_XNACK=1`), mapped arrays alias their host allocation zero-copy: host and device share a single allocation instead of keeping replicated copies, and host-device transfers become no-ops. This roughly halves the memory footprint of mapped data. Zero-copy mapping can be disabled at runtime by setting the environment variable `NEKO_HIP_ZEROCOPY=0`, which restores fully replicated buffers. Discrete GPUs (e.g. MI250X) always use replicated buffers, regardless of the XNACK setting. Since host and device share one allocation under zero-copy, host code (e.g. user routines) must not write to a mapped array while device work touching it may be in flight; synchronize with `device_sync` first.
 
 #### Compiling Neko for Apple Silicon GPUs
 To compile Neko for Apple Silicon GPUs (macOS only)
--- a/doc/pages/user-guide/installation.md
+++ b/doc/pages/user-guide/installation.md
@@ -257,6 +257,8 @@ $ ./configure  --with-hip=/opt/rocm/hip
 $ ./configure  --with-hip=/opt/rocm/hip HIP_HIPCC_FLAGS=-O3  HIPCC=/opt/rocm/hip/bin/hipcc
 ```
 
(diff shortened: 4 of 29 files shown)


src/multigrid/tree_amg.f90
```diff
@@ -204,6 +204,9 @@ subroutine tamg_lvl_init(tamg_lvl, lvl, nnodes, ndofs)
        call device_cfill(tamg_lvl%wrk_in_d, 0.0_rp, ndofs)
        call device_map(tamg_lvl%wrk_out, tamg_lvl%wrk_out_d, ndofs)
        call device_cfill(tamg_lvl%wrk_out_d, 0.0_rp, ndofs)
+       ! Order the async fills against host writes; on unified memory
+       ! the device pointers may alias the work arrays
+       call device_sync()
```
timofeymukha, line 209: Seems like something to be aware of when coding. When do we need to do this exactly?

njansson: During init routines, similar to that we had to sync memcopies

njansson: In this case, it's more of a double safety thing, if the object is used very close to it being initialised

njansson:
Seems like there's a too large performance regression on NVIDIA with this one...thus on hold for now

njansson:
> Seems like there's a too large performance regression on NVIDIA with this one...thus on hold for now

Even on AMD MI300A there is a difficult to pinpoint ~20% overhead cost of this... !@#$ => we skip this for v1.1

njansson:
We will continue this in a separate GH200 branch

PR #2666, opened 2026-07-20
Add opt-in zero-copy unified memory mapping for AMD MI300A (HIP)

njansson:
On unified-memory APUs (MI300A) with XNACK, `device_map` can alias the host allocation instead of keeping a separate replicated device buffer, halving the memory footprint of mapped data. This adds that path for the HIP backend, mirroring the existing Metal support.

**Off by default.** Enable with `NEKO_HIP_ZEROCOPY=1`. It is correct but currently ~1.3x slower per step than replicated buffers on MI300A (overhead still under investigation), so it ships as an opt-in capacity mode for running larger cases, not a speedup.

refs: #2633


Mr-2c (approved):

timofeymukha (approved):
My jobs is to make Siavash happy

PR #2667, opened 2026-07-21
fixed only_facets in bc%finalize in force_torque

arash-mp:
force_torque reads `bc%facet` together with `bc%msk` in `setup_normals`, but the bc (which is `dirichlet_t`, with no guard on `only_facets`) was finalized without `only_facets`. In that case `bc_finalize_base` rebuilds msk and leaves facet stale, so facet(i) no longer belongs to msk(i). This could sometimes lead to wrong and garbage force/torque values.

### DIFF
--- a/src/qoi/drag_torque.f90
+++ b/src/qoi/drag_torque.f90
@@ -67,7 +67,7 @@ module drag_torque
   use math, only : rzero, col3, vdot3, col2
   use space, only : space_t
   use num_types, only : rp
-  use utils, only : nonlinear_index
+  use utils, only : nonlinear_index, neko_error
   use iso_c_binding, only : c_ptr
   use neko_config, only : NEKO_BCKND_DEVICE
   use device, only : HOST_TO_DEVICE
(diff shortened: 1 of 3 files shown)


src/qoi/drag_torque.f90
```diff
@@ -471,17 +470,21 @@ subroutine setup_normals(coef, mask, facets, n1, n2, n3, n_pts)
...
     do i = 1, n_pts
        fid = facets(i)
+       if ( (fid .lt. 1) .or. (fid .gt. 6)) then
+          call neko_error('setup_normals: invalid facet id. ' // &
+               'Finalize the bc with only_facets = .true.')
+       end if
        idx = nonlinear_index(mask(i), coef%Xh%lx, coef%Xh%lx, coef%Xh%lx)
        normal = coef%get_normal(idx(1), idx(2), idx(3), idx(4), fid)
        area = coef%get_area(idx(1), idx(2), idx(3), idx(4), fid)
-       n1%x(i) = normal(1)*area(1)
```
adperezm, line 482: Interesting, so what was happening before? Was there just unintialized stuff in the area array?

adperezm (approved):

njansson (approved):

src/qoi/drag_torque.f90
arash-mp: I think it was working fine and copying area to all area(1), area(2), area(3) (telling this since force/torque values were correct in most cases, and when they were wrong, it was related to only_facets).

PR #2668, opened 2026-07-21
Fix the release doc. workflow

njansson:
This fixed an issue with the documentation workflow for releases, which nuked the entire homepage when publishing the data.


pohm01 (approved):

GrahamBenHarper (approved):
🫪

GrahamBenHarper (commented):
it has come to my attention some people do not have Unicode 17.0, so the emoji appears as a unicode symbol on ancient devices

PR #2670, opened 2026-07-23
Add precon and ksp user type injection

timofeymukha:
Adds the possibility to inject custom preconditioner and krylov solver types.

Renamed the precon factory to a precon allocator, because it does not init. This is the split we have almost everywhere: an allocator just executes allocate, a factory calls the allocator and then calls the deffered `%init`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Develop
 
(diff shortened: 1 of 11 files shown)


njansson (approved):
🍻

Shiyu-Sandy-Du (approved):

PR #2672, opened 2026-07-23
Fix logged filenames and user_stats outptut

timofeymukha:
user_stats does not respect the output_directory setting, this fixes that.

We have a bug with logged filenames. The reason is that the name is printed before the write routine is executed, and for many file formats the write routine increments the counter that forms the filename. Here, we add a new routine ` get_next_output_fname` to the `generic_file`, which can be overridden to set the right name for each type. This is now done, so the loggings should be correct.

The AI also says that the .bp output has been bugged and the output counter started with -1, this is fixed here, but I never used .bp so please check if you are familiar with it.


njansson:
Not sure I understand the bug. Aren't we now printing the name (to be written ) before writing

vbaconnet:
> Not sure I understand the bug. Aren't we now printing the name (to be written ) before writing

If the file name is modified in the writer what is logged will be lagging behind, at least that's my understanding.

```
  --------Writer output---------
  File name     : field00004.fld
  Output number :     5
```

I'm surprised to not have noticed this earlier to be honest!

vbaconnet (approved):
look okay to me!

njansson (approved):

PR #2676, opened 2026-07-27
Generalize checkpoints to support arbitrary payloads

timofeymukha:
Dep #2674 

This overhauls the chkpt_t type and related functionality. 

**Old checkpoints should remain valid**

## Main change and motivation
Previously the type stores a large number of pointers corresponding to various restart fields, like u, v, w, etc. The main issue is that freezes what a checkpoint can contain, unless one goes in and modifies the type. This is a big issue for neko's extensibility, with roots in the legacy custom-format .chkpt format. At the same time the HDF5 checkpoint file, fundamentally, supports a free layout!

In this PR, the chkpt_t now consists of an array of `checkpoint_payload`s, which are collections of fields, field series, mesh arrays (a 4D array living on an SEM geometry, used by ALE currently), and ordinary arrays. Generic arrays may be replicated or explicitly distributed. All of these are non-owning of course, and just point to data, like in the original type. Schemes and stuff like ALE simply add payloads to the checkpoint.

The hdf5_file, which writes checkpoints, maps each payload to hdf5 group, and the stores the rest as datasets under it.

To load or write a legacy checkpoint file, a compatibility type `legacy_checkpoint_view_t` is added that stores all the pointers that were previously in the main type checkpoint type. This structure is populated from the new payload layout in a helper routine called `chkp_build_legacy_view`.

## Additional
- Support for interpolation across p is added to the hdf5 checkpoints.
- ALE now supports p-interpolation @arash-mp At least in principle, I am not sure how well an interpolated mass matrix will work and stuff like that.
- New tests.

The code was written mostly with AI, with iterations on the design from my side. I checked everything manually in the end, with the exception of the low-level hdf5 library calls. I am not versed in that, yet it seems those were very sensitive based on the vtkhdf5 stuff. So if @timfelle, @adperezm or @njansson can take an extra look at that during review it would be great.


arash-mp:
Very nice @timofeymukha. 
regarding ALE restart with a different poly order, it was already possible with `.chkp`, but with the (accuracy) cost of setting Blag and Blaglag = B at the restart. Considering that changing poly order will anyway introduce some error and is not considered exact restart. I haven't tested interpolating mass matrix either (not sure even if that's a good idea)
I haven't checked the PR yet, but if it now interpolates the Blag and Blaglag if poly order is changed, I think it's better to leave that as an option, if possible.
also a general question, does it affect both `chkp` and `h5` formats, or works only with h5?

njansson (approved):
Great stuff 🍻 🐈‍⬛ 🐈

timfelle (approved):

PR #2679, opened 2026-07-28
Boundary_data_t, wall_shear_stress_simcomp and boundary_data_wrtier_simcomp

arash-mp:
This PR adds and new `boundary_data_t` and 2 new simulation components based on  `boundary_data_t`.

### `boundary_data_t`

Collects the boundary points on labelled zones and gives access to their coordinates, normals and surface weights. It can sample any field at those points, and compute surface quantities over the zones: integrals, averages, surface area, centroid, flux, and the wall-normal and wall-tangential parts of a vector.

### `wall_shear_stress_simcomp`

Computes the wall shear stress (tau_x, tau_y, tau_z, tau_mag) on selected zones, and adds them to fields registries.

### `boundary_data_writer_simcomp`

Samples a list of fields at the boundary points and writes them to CSV or HDF5.
I think it would be very good if in future we add spatial averaging to this simcomp. It would make many post processings much easier and less memoery intensive.
I am not sure if map_1D and map_2D would lead to correct averaging for surface values (haven't checked them thoroughly).

wall_data_osccyl_new2_pressure_arrows


src/simulation_components/wall_shear_stress_simcomp.f90
```diff
@@ -0,0 +1,644 @@
...
+    character(len=:), allocatable :: viscosity_field
+    integer, allocatable :: zone_indices(:)
+    character(len=NEKO_VARNAME_LEN), allocatable :: fields(:)
+    character(len=NEKO_VARNAME_LEN), allocatable :: components(:)
+    logical :: want_x, want_y, want_z, want_mag
+    integer :: i
+
+    call json_get_or_default(json, "name", name, "wall_shear_stress")
+    call json_get_or_default(json, "computed_field", computed_field, "tau")
+    call json_get_or_default(json, "fluid_name", fluid_name, "fluid")
```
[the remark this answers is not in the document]
arash-mp: added

src/simulation_components/boundary_data_writer_simcomp.f90
```diff
@@ -0,0 +1,732 @@
...
+    class(case_t), intent(inout), target :: case
+    character(len=:), allocatable :: name, output_filename
+    character(len=NEKO_VARNAME_LEN), allocatable :: fields(:)
+    integer, allocatable :: zone_indices(:)
+    real(kind=rp) :: start_time
+    logical :: output_normals
+    logical :: user_set_compute
+
+    call json_get_or_default(json, "name", name, "boundary_data_writer")
+
```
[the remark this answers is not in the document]
arash-mp: added

src/simulation_components/wall_shear_stress_simcomp.f90
```diff
@@ -0,0 +1,650 @@
...
+    class(case_t), intent(inout), target :: case
+    integer, intent(in) :: order
+    type(time_based_controller_t), intent(in) :: preprocess_controller
+    type(time_based_controller_t), intent(in) :: compute_controller
+    type(time_based_controller_t), intent(in) :: output_controller
+    character(len=*), intent(in) :: computed_field
+    character(len=*), intent(in) :: viscosity_field
+    integer, intent(in) :: zone_indices(:)
+    type(coef_t), intent(inout), target :: coef
+    logical, intent(in), optional :: want_x, want_y, want_z, want_mag
```
timofeymukha: No need for optional, I think

src/simulation_components/wall_shear_stress_simcomp.f90
```diff
@@ -0,0 +1,650 @@
...
+    call wall_shear_stress_build_field_list(computed_field, wx, wy, wz, wm, &
+         fields)
+
+    call this%init_base_from_components(case, order, preprocess_controller, &
+         compute_controller, output_controller)
+
+    do i = 1, size(fields)
+       call neko_registry%add_field(coef%dof, trim(fields(i)), &
+            ignore_existing = .true.)
+    end do
```
timofeymukha: Is there a reason this stuff cannot be in the common constructor? Similar parsing seems to be repeated in several constructors.

src/simulation_components/wall_shear_stress_simcomp.f90
```diff
@@ -0,0 +1,650 @@
...
+     !> Masked dynamic viscosity.
+     type(vector_t) :: mu_msk
+     !> Masked pressure.
+     type(vector_t) :: p_msk
+     !> Masked strain rate components.
+     type(vector_t) :: s11, s22, s33, s12, s13, s23
+     !> Viscous traction components at the masked points.
+     type(vector_t) :: t1, t2, t3
+     !> Pressure traction.
+     type(vector_t) :: pt1, pt2, pt3
```
timofeymukha: Please consider whether some or all of these can be requested from scratch (it supports vectors!). On quick look it seems like you use these to compute stuff and then scatter the result into the 3d fields.

doc/pages/user-guide/simcomps.md
```diff
@@ -259,6 +263,49 @@ Optional fields for this simcomp are:
...
+- `start_time`: only write samples after this time. Default `0.0`.
+
+~~~~~~~~~~~~~~~{.json}
+{
+  "type": "boundary_data_writer",
+  "name": "wall_data",
+  "zone_indices": [7],
+  "fields": ["p", "u", "v", "w", "tau_mag"],
+  "output_filename": "wall_data.csv",
+  "output_normals": true,
```
vbaconnet, line 289: Is there a reason why normals are optional but not `area`?

doc/pages/user-guide/simcomps.md
```diff
@@ -259,6 +263,49 @@ Optional fields for this simcomp are:
...
+#### Output files
+
+The geometry (coordinates, optional normals and `area`) is written differently for a static and a moving mesh, so that it is stored only where it is actually needed.
+
+On a **static mesh**, the geometry is constant and is written once to a companion file, while the main file holds only `time` and the sampled fields:
+- CSV: geometry in `<name>_mesh.csv` (columns `x, y, z`, then `n_x, n_y, n_z` if `output_normals`, then `area`); data in `<name>.csv` (columns `time` then the sampled fields).
+- HDF5: geometry in the `coordinates` dataset; the samples are appended under the `boundary_data` group.
+
+On a **moving mesh (ALE)**, the geometry changes every step, so it is written inline as leading columns of every sample; in addition, the initial (undeformed) geometry is written once as a reference:
+- CSV: `<name>.csv` has columns `time, x, y, z, [n_x, n_y, n_z,] area`, then the sampled fields; the initial geometry is in `<name>_initial_mesh.csv`.
```
vbaconnet: I would maybe append the normals instead of putting them in between, so the `area` array is always located at the same column index regardless of if normals are output or not. Or maybe there was a reason for that?

```suggestion
- CSV: `<name>.csv` has columns `time, x, y, z, area, [n_x, n_y, n_z,]`, then the sampled fields; the initial geometry is in `<name>_initial_mesh.csv`.
```

src/common/boundary_data.f90
```diff
@@ -0,0 +1,789 @@
...
+       call vector_copy(v, this%area)
+       call boundary_data_sync_host(v, to_host)
+    case default
+       if (.not. neko_registry%field_exists(trim(name))) then
+          call neko_error("boundary_data: '" // trim(name) // &
+               "' is neither a geometry attribute (x, y, z, n_x, n_y, " // &
+               "n_z, area) nor a field in the registry")
+       end if
+       f => neko_registry%get_field_by_name(trim(name))
+       call this%get(f, v, to_host)
```
vbaconnet: maybe you can explicit what is being called here for clarity.
```suggestion
       call this%get_vector_by_field(f, v, to_host)
```

src/simulation_components/boundary_data_writer_simcomp.f90
```diff
@@ -0,0 +1,738 @@
...
+    call json_get(json, "zone_indices", zone_indices)
+    call json_get(json, "output_filename", output_filename)
+    call json_get_or_default(json, "output_normals", output_normals, .true.)
+    call json_get_or_default(json, "start_time", start_time, 0.0_rp)
+
+    if (json%valid_path("fields")) then
+       call json_get(json, "fields", fields)
+    else
+       allocate(fields(0))
```
vbaconnet: Why not let json throw the error if it doesn't find "fields"?

src/simulation_components/boundary_data_writer_simcomp.f90
```diff
@@ -0,0 +1,738 @@
...
+
+    this%ale_enabled = .false.
+    if (associated(neko_ale)) then
+       if (neko_ale%active) this%ale_enabled = .true.
+    end if
+
+    this%geometry_in_data = this%ale_enabled
+
+    ! `only_facets`
+    call this%bdata%init(this%coef, this%zone_indices)
```
vbaconnet, line 354: can you not pass `ale_enabled` into the constructor of `boundary_data_t` here? Then boundary data doesn't need to do the check again in its own constructor.

src/simulation_components/boundary_data_writer_simcomp.f90
```diff
@@ -0,0 +1,738 @@
...
+            "'.csv', '.h5' or '.hdf5'")
+    end if
+
+    call this%fout%init(this%case%output_directory // trim(output_filename))
+
+    n_geom = 4
+    if (this%output_normals) n_geom = 7
+
+    select type (ft => this%fout%file_type)
+    type is (csv_file_t)
```
vbaconnet, line 459: maybe you can put the writers in separate routines? This constructor is quite long already imo

src/simulation_components/boundary_data_writer_simcomp.f90
```diff
@@ -0,0 +1,738 @@
...
+       call neko_error("boundary_data_writer requires at least one entry " // &
+            "in 'fields'")
+    end if
+
+    call this%fields%init(size(fields))
+    do i = 1, size(fields)
+       if (.not. neko_registry%field_exists(trim(fields(i)))) then
+          call neko_error("boundary_data_writer: the field '" // &
+               trim(fields(i)) // "' is not in the registry")
+       end if
```
vbaconnet: I'm pretty sure that `get_fields_by_name` will throw an error if it doesn't find the field, no?

src/simulation_components/boundary_data_writer_simcomp.f90
```diff
@@ -0,0 +1,738 @@
...
+     !> Output vector, used for the HDF5.
+     type(vector_t) :: vec_out
+     !> The output file.
+     type(file_t) :: fout
+     !> Time after which to start writing.
+     real(kind=rp) :: start_time = 0.0_rp
+     !> Whether ALE or not.
+     logical :: ale_enabled = .false.
+     !> Whether the geometry is part of every sample.
+     logical :: geometry_in_data = .false.
```
vbaconnet, line 112: do you need this flag if it ends up mirroring `ale_enabled` anyways?

src/common/boundary_data.f90
```diff
@@ -0,0 +1,789 @@
...
+     type(vector_t) :: n_x, n_y, n_z
+     !> Surface quadrature weights at the boundary points.
+     type(vector_t) :: area
+     !> Work vector.
+     type(vector_t), private :: work
+     !> Whether the stored normals point out of the wall into the fluid. When
+     !! false they keep the `coef` convention, pointing out of the fluid.
+     logical :: outward_normals = .true.
+     !> Whether ALE is enabled or not.
+     logical :: ale_enabled = .false.
```
vbaconnet: I don't think this type should have any visibility on whether ale (or moving/changing mesh in general) is enabled. In the end this flag is only used inside `update_geometry` to enable copying of geometry data, which imo is not necessary, i.e. the caller should determine itself if the routine needs to be called.

src/common/boundary_data.f90
```diff
@@ -0,0 +1,789 @@
...
+    logical, intent(in), optional :: to_host
+    logical :: forced, copy_to_host
+    integer :: n
+
+    forced = .false.
+    if (present(force)) forced = force
+    copy_to_host = .false.
+    if (present(to_host)) copy_to_host = to_host
+
+    if (.not. (forced .or. this%ale_enabled)) return
```
vbaconnet: see my previous comment on `this%ale_enabled`, do you really need to have these flags here? I would let the caller determine when to call the subroutine and not bother with more internal logic.

src/common/boundary_data.f90
```diff
@@ -0,0 +1,789 @@
...
+    integer :: n
+
+    if (.not. associated(f%dof, this%coef%dof)) then
+       call neko_error("boundary_data: the destination field is on a " // &
+            "different dofmap than the boundary mask")
+    end if
+
+    n = this%coef%dof%size()
+
+    call field_rzero(f)
```
vbaconnet: this may not be desired depending on the use case, if a zeroing is required I would let that be done separately.

src/simulation_components/boundary_data_writer_simcomp.f90
```diff
@@ -0,0 +1,738 @@
...
+     !> Output matrix.
+     type(matrix_t) :: mat_out
+     !> Output vector, used for the HDF5.
+     type(vector_t) :: vec_out
+     !> The output file.
+     type(file_t) :: fout
+     !> Time after which to start writing.
+     real(kind=rp) :: start_time = 0.0_rp
+     !> Whether ALE or not.
+     logical :: ale_enabled = .false.
```
vbaconnet: Maybe this is too picky, but could we generalize this to "mesh has changed" instead? In the end this simcomp only needs to know if the geometry has changed, not if ALE is enabled... Perhaps there will be other features (not ALE) in the future that will also have an effect on the mesh?

src/simulation_components/wall_shear_stress_simcomp.f90
```diff
@@ -0,0 +1,650 @@
...
+
+    call wall_shear_stress_build_field_list(computed_field, want_x, want_y, &
+         want_z, want_mag, fields)
+
+    call this%init_base(json, case)
+
+    ! Register the selected fields so they are available in the registry.
+    do i = 1, size(fields)
+       call neko_registry%add_field(case%fluid%c_Xh%dof, trim(fields(i)), &
+            ignore_existing = .true.)
```
vbaconnet: Should this be `false`? What if you have two `wall_shear_stress` simcomps?

src/simulation_components/wall_shear_stress_simcomp.f90
```diff
@@ -0,0 +1,650 @@
...
+    if (want_x) n = n + 1
+    if (want_y) n = n + 1
+    if (want_z) n = n + 1
+    if (want_mag) n = n + 1
+
+    allocate(fields(n))
+    k = 0
+    if (want_x) then
+       k = k + 1
+       fields(k) = trim(computed_field) // "_x"
```
vbaconnet, line 228: this will generate a non-unique name if several `wss` simcomps are in effect, maybe consider prepending the simcomp `name` to the field name?

src/simulation_components/wall_shear_stress_simcomp.f90
```diff
@@ -0,0 +1,650 @@
...
+
+    call neko_scratch_registry%request_field(fx, work_indices(1), .false.)
+    call neko_scratch_registry%request_field(fy, work_indices(2), .false.)
+    call neko_scratch_registry%request_field(fz, work_indices(3), .false.)
+
+    ! Scatter the vectors into fields.
+    call this%bdata%scatter(this%t1, fx)
+    call this%bdata%scatter(this%t2, fy)
+    call this%bdata%scatter(this%t3, fz)
+
```
vbaconnet: what is the reason for scattering back the shear stresses into fields? Can the `math` operations below not be done on vectors?

src/common/boundary_data.f90
```diff
@@ -0,0 +1,789 @@
...
+  !! @param to_host Whether to copy to the host.
+  subroutine boundary_data_sync_host(v, to_host)
+    type(vector_t), intent(inout) :: v
+    logical, intent(in), optional :: to_host
+    logical :: copy_to_host
+
+    copy_to_host = .false.
+    if (present(to_host)) copy_to_host = to_host
+
+    if ( (NEKO_BCKND_DEVICE .eq. 1) .and. copy_to_host) then
```
vbaconnet: `copy_from` already makes the check in `vector_t`
```suggestion
    if (copy_to_host) then
```

src/common/boundary_data.f90
```diff
@@ -0,0 +1,789 @@
...
+            trim(f%name) // "' is on a different dofmap than the boundary " // &
+            "mask")
+    end if
+
+    n = this%coef%dof%size()
+
+    call field_rzero(f)
+
+    call this%get(trim(name), this%work, to_host = .false.)
+    call vector_masked_scatter_copy_0(f%x, this%work, this%bc%msk, n, &
```
vbaconnet: I don't really understand the purpose of this routine... Is this not essentially a `masked_copy`?

src/common/boundary_data.f90
```diff
@@ -0,0 +1,789 @@
...
+
+    call vector_masked_gather_copy_0(v, f%x, this%bc%msk, &
+         this%coef%dof%size(), this%n_local)
+
+    copy_to_host = .false.
+    if (present(to_host)) copy_to_host = to_host
+
+    if ( (NEKO_BCKND_DEVICE .eq. 1) .and. copy_to_host) then
+       call v%copy_from(DEVICE_TO_HOST, .true.)
+    end if
```
vbaconnet:
```suggestion
    call boundary_data_sync_host(v, to_host)
```

src/common/boundary_data.f90
```diff
@@ -0,0 +1,789 @@
...
+
+  end function boundary_data_is_geometry
+
+  !> Sample a named quantity at the boundary points into a vector.
+  !! @details The name is either one of the geometry attributes `x`, `y`, `z`,
+  !! `n_x`, `n_y`, `n_z`, `area`, or the name of a field in the registry.
+  !! @param name The requested quantity.
+  !! @param v The vector to fill.
+  !! @param to_host Whether to copy `v` to the host.
+  subroutine boundary_data_get_vector_by_name(this, name, v, to_host)
```
vbaconnet: Is this routine actually useful? The arrays x,y,z,nx,ny,nz,area can be accessed directly by `this%x` (as done in `boundary_data_writer`). The last part calls `get_vector_by_field`, so that second subroutine can exist separately.

src/common/boundary_data.f90
vbaconnet, line 1: General comments on this file:
1. I am personally not a big fan of the getter routines that have integrated memcopies in them. If you think about this type on its own, without any relation to the simcomps, there is no apparent why memcopies would be needed when getters (essentially wrappers around `vector_masked_gather_0`) are called. The simcomps themselves need the data on the host for I/O operations, therefore in my opinion the memcopies should be done by the simcomps as needed. Maybe you can consider refactoring the code a bit so that operations are better encapsulated within their respective routines.

2. The getters are a bit confusing to me because they do slightly different things, not just "getting" the attributes but also doing some masked gathering, and even calling each other. I think the core of the code in this type should be a `get_vector_from_field_name` and `get_vector_from_field_pointer` (or something less verbose...), which just wrap around vector_masked_gather.

3. It is a shame that all of the extra code for averaging, integration, etc is not reused in the boundary operations / boundary flux simcomps.. I suspect these simcomps would greatly benefit from the `boundary_data_t` helper type!

vbaconnet (commented):
I have some conceptual comments, especially on the boundary data type, but nothing that affects the features themselves. My main complaint is about the memcopies embedded inside the getters. Otherwise it is mostly design choices and (a little bit of) personal preference :) happy to discuss, and please push back if you feel like things are too picky!

src/common/boundary_data.f90
timofeymukha: Just to chip in a bit: I also thought about the existing boundary simcomps here. It feels like there should be synergy here, and, perhaps, separation of concerns (geometric data vs operations on data?). Not sure what the best way to go here, just looking from my phone. But for example for fields operations are mostly just functions in operator.f90, so one *could* have a similar module with functions that eat boundary data, and the boundary_operation simcomp would hold the data and just call the relevant operators.

doc/pages/user-guide/simcomps.md
arash-mp: now it's optional too (tho both defaults to true since they are mostly needed for postprocessing)

doc/pages/user-guide/simcomps.md
arash-mp: it's optional now, so no fixed position.

src/common/boundary_data.f90
arash-mp: done.

src/simulation_components/boundary_data_writer_simcomp.f90
arash-mp: fixed.

src/simulation_components/boundary_data_writer_simcomp.f90
arash-mp: I removed the ale awarness in bounary_data_t.

src/simulation_components/boundary_data_writer_simcomp.f90
arash-mp: done.

src/simulation_components/boundary_data_writer_simcomp.f90
arash-mp: fixed

src/simulation_components/boundary_data_writer_simcomp.f90
arash-mp: so I kept it but now it mirros `this%mesh_has_changed`, so if mesh changed due to something else than ALE, it will be aware of this.

src/common/boundary_data.f90
arash-mp: Agree. fixed.

src/common/boundary_data.f90
arash-mp: I added an optional `clear`, similar to `neko_scratch_registry%request_field`. It's default to `false`, so harmless. But convenient to have it in the routine.

src/simulation_components/boundary_data_writer_simcomp.f90
arash-mp: done

src/simulation_components/wall_shear_stress_simcomp.f90
arash-mp: Good catch. Fixed. thanks

src/simulation_components/wall_shear_stress_simcomp.f90
arash-mp: Isn't it fixed with `ignore_existing = .false.`?  No two `wss` simcomps can now have the same `computed_field`, so the fields would be unique.

src/simulation_components/wall_shear_stress_simcomp.f90
arash-mp: Fixed. Now the math is on `vectors` and then scattered to fields.
Also the reason for scattering to fields is to have it in field_registry and also for writing as `fld` (for some visulasiations it can be very usefull to have it as `fld`, and the wall can be extracted etc).
Not sure if we can output a surfae as `fld`, if possible that is the way to do it I think.

src/common/boundary_data.f90
arash-mp: fixed.

src/common/boundary_data.f90
arash-mp: it's for putting boundary values directly into a field by name. similar to scatter.
so it's not used right now, but also harmless to have. 
*update* I changed it to scatter.

src/common/boundary_data.f90
arash-mp: I think it can be useful (or more convenient) in some cases :) it's harmless and simple, so I prefer keeping it.

src/common/boundary_data.f90
arash-mp: Very good comments Victor thanks.

I removed the `to_host` from `boundary_data_t`, and also it's ale awarness.

*update*:  so I changed `get_field_by_name` to `scatter_to_field_by_name`. Now:

for `get` we have  `get_vector_by_name`, `get_vector_by_field`, which we can use to put boundary values for a zone_id using field or name of the field into a vector

and for `scatter` we have: `scatter_to_field_by_vector` and `scatter_to_field_by_name` which can be used to put a vector into a full field (destination field can be zeroed before the copy), or put the boundary values of a field using filed_name directly into another field (can be zeroed before the copy).

vbaconnet (approved):
Very happy with this! Super cool :)

timofeymukha (commented):
Very nice. We need to think how to integrate boundary_operation into this a little bit better. But we can probably do that without intruding on the user side of things later on.

src/simulation_components/wall_shear_stress_simcomp.f90
arash-mp: fixed.

src/simulation_components/wall_shear_stress_simcomp.f90
arash-mp: fixed.

src/simulation_components/wall_shear_stress_simcomp.f90
arash-mp: fixed.

timofeymukha (approved):

arash-mp:
@njansson This one is ready to be merged.

PR #2683, opened 2026-07-29
More forceful regen.sh

timofeymukha:
Adds -f and --force to a few places in regen.sh so it stops complaining about files being newer like

``` 
libtoolize: error: './config.guess' is newer: use '--force' to overwrite libtoolize: error: './config.sub' is newer: use '--force' to overwrite l

libtoolize: error: './install-sh' is newer: use '--force' to overwrite
```

### DIFF
--- a/regen.sh
+++ b/regen.sh
@@ -2,31 +2,31 @@
 
 echo "Updating configuration..."
 echo "Running libtoolize"
-if which libtoolize > /dev/null 2>&1; then
-    libtoolize -i
-elif which glibtoolize > /dev/null 2>&1; then
-    glibtoolize -i
+if command -v libtoolize >/dev/null 2>&1; then
+    libtoolize --install --force
+elif command -v glibtoolize >/dev/null 2>&1; then
+    glibtoolize --install --force
 else
-    echo "No libtoolize found on your system"
+    echo "No libtoolize or glibtoolize found on your system" >&2
     exit 1
 fi
 
-rm -fr autom4te.cache
-rm aclocal.m4 
+rm -rf autom4te.cache
+rm -f aclocal.m4
 
 echo "Running aclocal"
 if which aclocal > /dev/null 2>&1; then
-  aclocal -I m4 --install
+    aclocal -I m4 --install
 else
-  echo "No aclocal found on your system"
-  exit 1
+    echo "No aclocal found on your system" >&2
+    exit 1
 fi
 
 echo "Running autoconf"
 autoconf
+
 echo "Running automake"
-automake -a
+automake --add-missing
 
 echo "Deleting autom4te.cache directory"
-rm -r autom4te.cache
-
+rm -rf autom4te.cache
--- a/regen.sh
+++ b/regen.sh
@@ -2,9 +2,9 @@
 
 echo "Updating configuration..."
 echo "Running libtoolize"
-if command -v libtoolize >/dev/null 2>&1; then
+if which libtoolize > /dev/null 2>&1; then
     libtoolize --install --force
-elif command -v glibtoolize >/dev/null 2>&1; then
+elif which glibtoolize > /dev/null 2>&1; then
     glibtoolize --install --force
 else
     echo "No libtoolize or glibtoolize found on your system" >&2


regen.sh
```diff
@@ -3,30 +3,30 @@
...
+    echo "No aclocal found on your system" >&2
+    exit 1
 fi
 
 echo "Running autoconf"
 autoconf
+
 echo "Running automake"
-automake -a
+automake --add-missing
```
njansson, line 29:
```suggestion
automake -a
```
Let's follow the normal option set, no need to add the silly gnu flags

regen.sh
```diff
@@ -3,30 +3,30 @@
 echo "Updating configuration..."
 echo "Running libtoolize"
 if which libtoolize > /dev/null 2>&1; then
-    libtoolize -i
+    libtoolize --install --force
 elif which glibtoolize > /dev/null 2>&1; then
-    glibtoolize -i
+    glibtoolize --install --force
```
njansson, line 8:
```suggestion
    libtoolize -i
elif which glibtoolize > /dev/null 2>&1; then
    glibtoolize -i
```

Same as below let's keep these option sane, no need for verbose gnusim.

Also what problem are we trying to solve here? Why should we always reset a couple of links, and cause rather verbose output? IFF the system is setup correctly, the state warnings would never occur.

regen.sh
```diff
@@ -3,30 +3,30 @@
...
 else
-    echo "No libtoolize found on your system"
+    echo "No libtoolize or glibtoolize found on your system" >&2
```
njansson, line 10:
```suggestion
    echo "No libtoolize found on your system" >&2
```

It is the same program, `glibtoolize` is an alias on some none GNU systems

njansson (changes requested):
I rather see that people fix their setups, than adding these `force` flags, causing a rather verbose output, especially for us that runs `regen.sh` on a daily basis.

Also, can we be sane people, and please avoid the horrible GNU variants of the options.

regen.sh
timofeymukha: They seem to occur on most computers I had rather consistently, I am not sure why. Cannot blame it on WSL anymore because the last one is a proper linux :)).

Isn't the output less verbose with this change?

There is also this stuff, which this PR does not touch, becaues it is above my autotools paygrade

```
configure.ac:36: warning: The macro 'AC_HELP_STRING' is obsolete.
configure.ac:36: You should run autoupdate.
./lib/autoconf/general.m4:204: AC_HELP_STRING is expanded from...
configure.ac:36: the top level
configure.ac:39: warning: The macro 'AC_HELP_STRING' is obsolete.
configure.ac:39: You should run autoupdate.
./lib/autoconf/general.m4:204: AC_HELP_STRING is expanded from...
configure.ac:39: the top level
configure.ac:57: warning: The macro 'AC_TRY_COMPILE' is obsolete.
configure.ac:57: You should run autoupdate.
./lib/autoconf/general.m4:2845: AC_TRY_COMPILE is expanded from...
m4/ax_mpi.m4:69: AX_MPI is expanded from...
configure.ac:57: the top level
configure.ac:65: warning: AC_REQUIRE: 'AC_PROG_CXX' was expanded before it was required
configure.ac:65: https://www.gnu.org/software/autoconf/manual/autoconf.html#Expanded-Before-Required
./lib/autoconf/c.m4:660: AC_LANG_COMPILER(C++) is expanded from...
./lib/autoconf/lang.m4:335: AC_LANG_COMPILER_REQUIRE is expanded from...
./lib/autoconf/general.m4:2909: AC_LINK_IFELSE is expanded from...
lib/m4sugar/m4sh.m4:690: _AS_IF_ELSE is expanded from...
lib/m4sugar/m4sh.m4:697: AS_IF is expanded from...
./lib/autoconf/general.m4:2249: AC_CACHE_VAL is expanded from...
m4/libtool.m4:1183: _LT_SYS_MODULE_PATH_AIX is expanded from...
m4/libtool.m4:6480: _LT_LANG_CXX_CONFIG is expanded from...
m4/libtool.m4:842: _LT_LANG is expanded from...
m4/libtool.m4:824: LT_LANG is expanded from...
m4/ax_mpi.m4:69: AX_MPI is expanded from...
configure.ac:65: the top level
configure.ac:65: warning: The macro 'AC_TRY_COMPILE' is obsolete.
configure.ac:65: You should run autoupdate.
./lib/autoconf/general.m4:2845: AC_TRY_COMPILE is expanded from...
m4/ax_mpi.m4:69: AX_MPI is expanded from...
```

timofeymukha: Oh is the latter thing fixed in 2686?

timofeymukha: It is! Oh, that's very nice

timofeymukha:
I can give this up, the 2686 thing was the bigger annoyance.

PR #2684, opened 2026-07-29
Die when a dep is not found during configure

timofeymukha:
Currently, when specifying additional capabilities in configure, it may silently not find them (like hdf5 for example). This is quite annoying, basically one has to inspect the config log to make sure things really worked.

This PR changes that, so configure will actually fail when a requested dep is not found.

Now, the issue is that this is **100% vibe-coded** because I am really not versed in autotools that much ^^. However, due to the pretty small diff and the huge quality of life bonus this would provide, I dare put it up for review by competent people.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -82,8 +82,12 @@ if test "x${enable_openmp}" != xno; then
       AX_OPENMP([have_openmp=yes], [have_openmp=no])
       if test "x${have_openmp}" != xno; then
          CFLAGS="$CFLAGS $OPENMP_CFLAGS"
+      else
+         AC_MSG_ERROR([OpenMP requested but not supported by the C compiler])
       fi
       AC_LANG_POP([C])
+   else
+      AC_MSG_ERROR([OpenMP requested but not supported by the Fortran compiler])
    fi
 fi
 
@@ -335,6 +339,9 @@ if (test "x${have_cuda}" = xyes ||
       AC_SUBST(device_mpi, .false.)
    fi
 else
+   if test "x${enable_device_mpi}" = xyes; then
+      AC_MSG_ERROR([Device-aware MPI requires a CUDA, HIP, or OpenCL backend])
+   fi
    AC_SUBST(device_mpi, .false.)
 fi
 
@@ -368,6 +375,11 @@ if (test "x${is_cray}" = xyes || test "x${is_hpe_cray}" = xyes); then
    AX_CRAY_SHMEM
 fi
 
+if test -n "${with_openshmem}" &&
+   test "x${with_openshmem}" != xno &&
+   test "x${have_cray_shmem}" != xyes; then
+   AC_MSG_ERROR([OpenSHMEM requested but it is only available through a Cray environment])
+fi
 
 
 # JSON-Fortran
--- a/m4/ax_blas.m4
+++ b/m4/ax_blas.m4
@@ -98,6 +98,14 @@ if test "x$BLAS_LIBS" != x; then
 fi
 fi
 
+# An explicitly selected BLAS library is a requirement, not a hint.
+case $with_blas in
+	yes | no | "") ;;
+	*) if test $ax_blas_ok != yes; then
+		AC_MSG_ERROR([Requested BLAS library '$with_blas' was not found])
+	   fi ;;
+esac
+
 # BLAS linked to by default?  (happens on some supercomputers)
 if test $ax_blas_ok = no; then
 	save_LIBS="$LIBS"; LIBS="$LIBS"
--- a/m4/ax_lapack.m4
+++ b/m4/ax_lapack.m4
@@ -104,6 +104,14 @@ if test "x$LAPACK_LIBS" != x; then
         fi
 fi
 
+# An explicitly selected LAPACK library is a requirement, not a hint.
+case $with_lapack in
+        yes | no | "") ;;
+        *) if test $ax_lapack_ok != yes; then
+                AC_MSG_ERROR([Requested LAPACK library '$with_lapack' was not found])
+           fi ;;
+esac
+
 # LAPACK linked to by default?  (is sometimes included in BLAS lib)
 if test $ax_lapack_ok = no; then
         save_LIBS="$LIBS"; LIBS="$LIBS $BLAS_LIBS $FLIBS"
--- a/m4/ax_metis.m4
+++ b/m4/ax_metis.m4
@@ -23,13 +23,18 @@ AC_DEFUN([AX_METIS],[
 	AS_HELP_STRING([--with-metis-libdir=LIBDIR],
 	[Directory for metis library]),
 	[
-	if test -d "$withval"; then
+	if test "x${withval}" = xno; then
+	   ac_metis_libdir=""
+	elif test -d "$withval"; then
 	   ac_metis_libdir="$withval"
+	else
+	   AC_MSG_ERROR([METIS library directory '$withval' was not found])
 	fi
 	],)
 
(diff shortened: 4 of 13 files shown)


njansson:
I think one should fail only for requirements not optional dependencies (or optional flags). This is was most other codes does.

configure.ac
```diff
@@ -368,6 +375,11 @@ if (test "x${is_cray}" = xyes || test "x${is_hpe_cray}" = xyes); then
    AX_CRAY_SHMEM
 fi
 
+if test -n "${with_openshmem}" &&
+   test "x${with_openshmem}" != xno &&
+   test "x${have_cray_shmem}" != xyes; then
+   AC_MSG_ERROR([OpenSHMEM requested but it is only available through a Cray environment])
+fi
```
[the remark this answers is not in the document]
njansson: Yea and this is the intended behaviour

njansson:
How about we list enabled options at the end of configure? Including cflags and friends (then one can also see which package configure picked up)

timofeymukha:
> I think one should fail only for requirements not optional dependencies (or optional flags). This is was most other codes does.

But why? If I set --with-hdf5, then clearly I want it. I'd say it is optional until explicitly requested.

configure.ac
```diff
@@ -82,8 +82,12 @@ if test "x${enable_openmp}" != xno; then
       AX_OPENMP([have_openmp=yes], [have_openmp=no])
       if test "x${have_openmp}" != xno; then
          CFLAGS="$CFLAGS $OPENMP_CFLAGS"
+      else
+         AC_MSG_ERROR([OpenMP requested but not supported by the C compiler])
```
njansson, line 86: No, this is not safe since the macro doesn't cover every compiler

configure.ac
```diff
@@ -82,8 +82,12 @@ if test "x${enable_openmp}" != xno; then
...
       fi
       AC_LANG_POP([C])
+   else
+      AC_MSG_ERROR([OpenMP requested but not supported by the Fortran compiler])
```
njansson, line 90: Same here we don't have the necessary coverage by the macros

configure.ac
```diff
@@ -335,6 +339,9 @@ if (test "x${have_cuda}" = xyes ||
       AC_SUBST(device_mpi, .false.)
    fi
 else
+   if test "x${enable_device_mpi}" = xyes; then
+      AC_MSG_ERROR([Device-aware MPI requires a CUDA, HIP, or OpenCL backend])
+   fi
```
njansson, line 344: Also bogus, platform dependent.

configure.ac
njansson, line 382: Better not to throw an error here, since we can't safely say it is a failure...sallad it is, better not to be clever here.

m4/ax_pfunit.m4
```diff
@@ -12,21 +12,28 @@ AC_DEFUN([AX_PFUNIT], [
...
-			if test -d "$withval"; then
+			if test "x${withval}" = xno; then
+			   PFUNIT_DIR=""
+			elif test "x${withval}" = xyes; then
+			   PFUNIT_DIR="/usr"
+			elif test -d "$withval"; then
 			   PFUNIT_DIR="$withval";
 			else
-			   PFUNIT_DIR="/usr";
+			   AC_MSG_ERROR([pFUnit directory '$withval' was not found])
```
njansson, line 22: This goes against all best practices with auto tools...

m4/ax_parmetis.m4
```diff
@@ -26,13 +26,18 @@ AC_DEFUN([AX_PARMETIS],[
 	AS_HELP_STRING([--with-parmetis-libdir=LIBDIR],
 	[Directory for parmetis library]),
 	[
-	if test -d "$withval"; then
+	if test "x${withval}" = xno; then
+	   ac_parmetis_libdir=""
+	elif test -d "$withval"; then
 	   ac_parmetis_libdir="$withval"
+	else
+	   AC_MSG_ERROR([ParMETIS library directory '$withval' was not found])
```
njansson, line 34: Goes against the best practices

m4/ax_metis.m4
```diff
@@ -23,13 +23,18 @@ AC_DEFUN([AX_METIS],[
 	AS_HELP_STRING([--with-metis-libdir=LIBDIR],
 	[Directory for metis library]),
 	[
-	if test -d "$withval"; then
+	if test "x${withval}" = xno; then
+	   ac_metis_libdir=""
+	elif test -d "$withval"; then
 	   ac_metis_libdir="$withval"
+	else
+	   AC_MSG_ERROR([METIS library directory '$withval' was not found])
```
njansson, line 31: Against best practices with auto tools

m4/ax_lapack.m4
```diff
@@ -104,6 +104,14 @@ if test "x$LAPACK_LIBS" != x; then
...
 fi
 
+# An explicitly selected LAPACK library is a requirement, not a hint.
+case $with_lapack in
+        yes | no | "") ;;
+        *) if test $ax_lapack_ok != yes; then
+                AC_MSG_ERROR([Requested LAPACK library '$with_lapack' was not found])
+           fi ;;
+esac
+
```
njansson, line 114: This is a well tested macro from the autoconf archives...we really don't need to mess with it, and also we do this check on the outside (AS IT SHOULD BE DONE)

m4/ax_blas.m4
```diff
@@ -98,6 +98,14 @@ if test "x$BLAS_LIBS" != x; then
...
 fi
 
+# An explicitly selected BLAS library is a requirement, not a hint.
+case $with_blas in
+	yes | no | "") ;;
+	*) if test $ax_blas_ok != yes; then
+		AC_MSG_ERROR([Requested BLAS library '$with_blas' was not found])
+	   fi ;;
+esac
+
```
njansson, line 108: This is a well tested macro from the autoconf archives...we really don't need to mess with it, and also we do this check on the outside (AS IT SHOULD BE DONE)

njansson (changes requested):
Strongly argue against it, we should strive to follow best practices and behave as other open source projects using autotools.

m4/ax_parmetis.m4
timofeymukha: Which best practice are you referring to here?

timofeymukha:
> Strongly argue against it, we should strive to follow best practices and behave as other open source projects using autotools.

The implementation details aside, I cannot see how what we do can be best practice. If we set --with-cuda and it doesn't find it, an error is thrown (good!). How is --with-hdf5 fundamentally different?  We could also just silently configure a CPU build then, when a requested backend is not found, but it is clearly not nice for the user. If I request a build with pfunit and parmetis, to me it is clear that an error or at least a giant warning should be thrown at me, if this requested configuration has failed.

timfelle (commented):
> > Strongly argue against it, we should strive to follow best practices and behave as other open source projects using autotools.
> 
> The implementation details aside, I cannot see how what we do can be best practice. If we set --with-cuda and it doesn't find it, an error is thrown (good!). How is --with-hdf5 fundamentally different?  We could also just silently configure a CPU build then, when a requested backend is not found, but it is clearly not nice for the user. If I request a build with pfunit and parmetis, to me it is clear that an error or at least a giant warning should be thrown at me, if this requested configuration has failed.  

Lets add this one to the meeting docket. I strongly agree with Timofey here, and I would expect that Niclas might have missed the "Error for optionals only when explicitly requested" part of the argument. I cannot see how that would not be useful.

Of course we need to make sure we actually handle this correctly in terms of implementation.

PR #2685, opened 2026-07-29
Fix 2d mesh extrusion for curved boundaries

timofeymukha:
This fixes the issue spotted a while ago by @adperezm : when we do the automatic extrusion of a 2d nmsh, the curvature of the eges is not replicated on the extruded side.

This PR fixes that. The idea is simple, we add an optional arg to the routine that reads the edges, which signals that we are in 2d extrusion mode. For that case, it copies the curve data to the opposite of the element. Just ast he extrusion itself, it is hardcoded to extruding in z.


njansson (approved):
🍻

arash-mp (approved):
Great!

PR #2688, opened 2026-07-30
Ale integration test

arash-mp:
Two ale tests, which tests most of ALE update_metrics, kinematics, mesh motion stuff, force/torque,  ale_restart.
Most of the critical things are covered in the two tests.
I tested them on cpu (sp/dp) and cuda (sp,dp) on 2 different workstations, and they all passed.
If there is any suggestions, I would be happy to apply them.

The test itself is written by AI,  but I checked it myself .


timofeymukha (approved):
Skimmed through it. The .py seems a bit too verbose, but one can understand what is going on without problems if one needs to.

njansson (approved):

PR #2695, opened 2026-08-07
Remove memcpy after material_properties in the compressible solver

vbaconnet:
Removes memcopies for mu and kappa in the compressible solver. This is done for two reasons:
1. Bugfix: If mu and kappa are constant, they are populated by field_cfill(). With GPUs, field_cfill populates the device arrays but the device_memcpy that follows overwrites the values back to zero, effectively reverting back to Euler instead of NS.
2. Performance: if mu and kappa are computed in the user file (say following Sutherland's law), allows for the user to call device_math routines directly, avoiding the need for memcopies. 

@njansson would be nice if we could cherrypick this into 1.1!

### DIFF
--- a/examples/cns_2d_shock_tube/2d_shock_tube.case
+++ b/examples/cns_2d_shock_tube/2d_shock_tube.case
@@ -20,6 +20,8 @@
         "fluid": {
             "scheme": "compressible",
             "gamma": 1.4,
+            "Re": 1000.0,
+            "Pr": 0.73,
             "initial_condition": {
                 "type": "user"
             },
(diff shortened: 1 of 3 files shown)


tuananhdao:
Thanks! The solver change looks right, but the example `cns_2d_shock_tube` still writes `mu` and `kappa` only on the CPU. Could you update that example to handle the GPU arrays too?

njansson (approved):
Great, I'll cherry pick it into 1.1

vbaconnet:
> Could you update that example to handle the GPU arrays too?

Sure! Should be done now :) Added a few things so Re and Pr could be specified from the case file as well

tuananhdao (approved):

PR #2696, opened 2026-08-08
phmg_update

arash-mp:
This PR implements `phmg_update`, which propagates mesh changes down to the coarse levels of the multigrid hierarchy. The implementation is based on code and an idea that @pohm01 shared with me a while back.

Motivation: in ALE simulations the pressure solve sometimes fails to converge, because the coarse levels are still carrying the geometry from initialization (original mesh). `msh` itself doesn't change, so we need to re-interpolate the coarse-grid geometry from the updated `dof` after the mesh moves.

From the tests I ran (and from @pohm01's earlier ones), refreshing the coarse-grid geometry is enough to recover convergence in most cases. On top of that we can also recompute the Chebyshev eigenvalue estimates, but that's a bit expensive, so I added options to control how often it happens instead of doing it every time step.

Outside phmg, `coef_t` gains a `metrics_version` counter (incremented each time `coef_recompute_metrics` is called) so the update can tell whether the mesh actually moved. Also, with `smoother_cheby_acc = "jacobi"` the diagonal is rebuilt from the new geometry; with "schwarz" the local solves aren't refreshed.

Off by default, so nothing changes unless enabled. Even when enabled, the update only triggers when `coef_recompute_metrics` has been called, so it is a no-op on a static mesh.

``` json
"update": {
  "enabled": true,
  "eigs": {
    "enabled": true,
    "frequency": 10,
    "warm_start": true,
    "warm_start_iterations": 20
  }
}
``` 

The plot below shows pressure iteration counts for an ALE case with ~400k elements. Without the update the solve eventually stops converging and hits the iteration cap; with the update it runs smoothly.

pressure_iters_raw_ext


njansson (approved):
Overall looks good, but I guess @pohm01 should give the phmg part a look before we merge it.

A small thing to think about is the `update` block, I think the `enabled` might be redundant. For similar settings I recall we do a check if the entry is present or not instead of storing a boolean. Not a blocker, but something to think about.

pohm01 (approved):
Looks good. Thanks for putting this together

PR #2697, opened 2026-08-08
Fix missing edge terms in pc jacobi for deformed elements

njansson:
This fixes some missing edge terms in the Jacobi preconditioner for deformed elements. With this fix, the CPU backend is consistent with the device backends.


timfelle (changes requested):
Test golden files need to be updated.

njansson:
> Test golden files need to be updated.

Also, I start to wonder if this is correct even ;)

PR #2702, opened 2026-08-10
phmg/tamg improvements

njansson:
This pull request significantly improves the performance and scalability of the TreeAMG multigrid solver, particularly for CPU parallelism. The main changes revolve around optimizing the coarse-level matvec (matrix-vector multiplication), restriction, and prolongation operations to avoid atomic operations and enable efficient, disjoint writes. New data structures are introduced to drive these operations from the aggregate side, and OpenMP parallelism is applied more effectively. Additionally, the setup routines are optimized for speed and correctness, and memory management is updated to handle new allocations.


pohm01 (approved):
Looks good. Should probably eventually try to do something similar on the device backend as well.

timofeymukha (approved):
Since Peter approves, I'll just facilitate.

PR #2704, opened 2026-08-11
Correct material property treatment in scalar_pnpn

timofeymukha:
This PR fixes the material property behaviour in scalar_pnpn. There were several issues that are now addressed.

- The user material property user hook was called after the RHS assembly, which uses material properties.
- The Neumann bc contributions were multiplied by rho*cp (inside the rhs_maker). This is wrong, we assume the user input to be the actual flux. The fix requires multiplying sources and convection by rho x cp, and not doing so for the weak bc contributions, and then passing the sum to the rhs_maker. Therefore, rho and cp are removed from the rhs_maker%makeext interface.
- The bdf kernel did not use cp at all, only rho. This is now implicitly fixed by transitioning to correct treatment of rho and cp as fields in the scheme: we form a rho_cp in a scratch field, and then use that everywhere with pointwise multiplication instead of the previous `%x(1,1,1,1)` hack.

This obviously comes at the price of reduced performance, but I think it is negligible in the context of the big time consumers (linear solves) and the benefit of having everything consistent outweighs the costs of a few extra field multiplication. That said, I haven't done any real timings.


njansson (approved):
🐈

arash-mp (approved):
Great.
Just what about `Changelog`?

njansson:
@timofeymukha are we good to merge?

PR #2705, opened 2026-08-12
fix data race in adv_dealias

arash-mp:
fixes a data race in `compute_scalar_advection_dealias` and `compute_ale_advection_dealias` openMP block.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Develop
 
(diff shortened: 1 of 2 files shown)


arash-mp:
@njansson If it look's ok, I think it should go to v1.1.

njansson (approved):

timofeymukha (approved):

PR #2715, opened 2026-08-18
Add a framework for spectral vanishing viscosity

shiyu-du-dsy:
Add a framework to perform spectral vanishing viscosity (SVV) for under-resolved simulation for incompressible flows. (The method is proposed by [Tadmor 1989](https://math.umd.edu/~tadmor/spectral_viscosity/Maday-Tad_SV_SINUM1989.pdf) and introduced into SEM framework by [Kirby and Sherwin 2006](https://www.sciencedirect.com/science/article/pii/S0045782505002434))

This PR is a preliminary establishment of SVV regularization framework in Neko which will be followed by the following enhancement (to be added in the future):
1. Add SVV to viscous regularization;
2. Add a dynamic approach for the SVV amplitude coefficient;

It seems to be over-dissipative now but it's expected with the current constant coefficient setup

A simple test with constant amplitude coefficient has been done on an advecting scalar hump (left: before, right: after)
hump_ref hump_smooth
And also on a crappy DNS of the turb_channel  (left: before, right: after)
channel_refchannel_stab


src/math/ax_helm_svv_full.f90
```diff
@@ -0,0 +1,76 @@
...
+! INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+! BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+! LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+! CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+! LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+! ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+! POSSIBILITY OF SUCH DAMAGE.
+!
+module ax_helm_svv_KS_full
+  use ax_product, only : ax_t
```
timofeymukha, line 35: Not a fan of the capitals in names overall, neither in code nor filenames. Particularly since fortran is case-insensitive. I would either use ks or just spell out kirby_sherwin. Or name it by some property of the method instead of the inventors.

Otherwise this looks good to me. I will run an AI check with a good model tomorrow morning to avoid any gotchas!

timofeymukha:
@Shiyu-Sandy-Du  I fixed some minor stuff with codex. Biggest thing is that axhelm needs a destructor interface now so that we can handle the svv pointer there.

A more substantial thing is this

"Both schemes update material properties before assembling/solving, but call svv%update only after the solve: fluid_pnpn.f90:833, fluid_pnpn.f90:974, scalar_pnpn.f90:404,
     scalar_pnpn.f90:482.

     Therefore:
      - A time-variable viscosity field is one step stale during residual assembly and the linear solve.
      - For nu.type = "value", h1 = rho * nu is calculated only during initialisation; svv_update_h1 returns immediately, so time-varying density is never reflected.

Recompute svv%h1 immediately after material-property updates and before any call using Ax. Value mode also needs to retain nu_val so it can refresh the density product.
"
What do you think?

shiyu-du-dsy:
> @Shiyu-Sandy-Du I fixed some minor stuff with codex. Biggest thing is that axhelm needs a destructor interface now so that we can handle the svv pointer there.
> 
> A more substantial thing is this
> 
> "Both schemes update material properties before assembling/solving, but call svv%update only after the solve: fluid_pnpn.f90:833, fluid_pnpn.f90:974, scalar_pnpn.f90:404, scalar_pnpn.f90:482.
> 
> ```
>  Therefore:
>   - A time-variable viscosity field is one step stale during residual assembly and the linear solve.
>   - For nu.type = "value", h1 = rho * nu is calculated only during initialisation; svv_update_h1 returns immediately, so time-varying density is never reflected.
> ```
> 
> Recompute svv%h1 immediately after material-property updates and before any call using Ax. Value mode also needs to retain nu_val so it can refresh the density product. " What do you think?

That's a good catch, yes we need to move it together with material property update

timofeymukha (approved):

shiyu-du-dsy:
The latest push change the naming from Kirby-Sherwin formulation into one-sided formulation since the filtering is a bit different w.r.t. the way to perform basis mapping: they minimize the mapping error in an integral sense, while we minimize the error on each GLL nodes

njansson (approved):

PR #2719, opened 2026-08-20
mesh_checker: Add axis alignment check for labeled zones

timofeymukha:
Make mesh_checker report whether zone normals are axis-algined. Uses a proxy symmetry bc, so the exact same check as internally is performed. This helps make sure that no geometric mistake crept into the boundary zone definitions and geometry during meshing. 

The downside is that gs and coef are always built now, so that makes the tool slower.

Example:

```
 Labeled zones: 
    Zone  1: 1440 faces. Normal alignment: none
    Zone  2: 1000 faces. Normal alignment: x
    Zone  3: 650 faces. Normal alignment: y
    Zone  4: 650 faces. Normal alignment: y
    Zone  5: 970 faces. Normal alignment: none
    Zone  6: 970 faces. Normal alignment: none
```

### DIFF
--- a/contrib/mesh_checker/mesh_checker.f90
+++ b/contrib/mesh_checker/mesh_checker.f90
@@ -1,4 +1,4 @@
-! Copyright (c) 2024-2025, The Neko Authors
+! Copyright (c) 2024-2026, The Neko Authors
 ! All rights reserved.
 !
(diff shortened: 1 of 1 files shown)


arash-mp (approved):

adperezm (approved):
Very nice!

PR #2720, opened 2026-08-20
Fix serial loops on the CPU critical path

njansson:


src/scalar/bcknd/cpu/scalar_residual_cpu.f90
```diff
@@ -53,18 +52,30 @@ subroutine scalar_residual_cpu_compute(Ax, s, s_res, f_Xh, c_Xh, msh, Xh, &
     integer, intent(in) :: n
     integer :: i
 
-    call copy(c_Xh%h1, lambda%x, n)
+    !OCL NORECURRENCE, NOVREC, NOALIAS
+    !DIR$ CONCURRENT
+    !DIR$ IVDEP
+    !GCC$ ivdep
+    !$omp parallel do
     do i = 1, n
```
timofeymukha, line 60: out of curiosity: could the loop inside copy and similar CPU math kernels just be openmp:ed as well?

njansson: They are already openmpified, but the point here is to merge parallel regions ot avoid the penalty of fork/join threads

timfelle (approved):

shiyu-du-dsy (approved):

PR #2724, opened 2026-08-24
Add matrix/tensor cores versions of Ax for AMD and NVIDIA accelerators

njansson:
This PR adds matrix/tensor core variants of Ax to the autotune. On AMD the kernels utilise MFMA and on NVIDIA it uses DMMA. Furthermore, both MFMA and DMMA sweeps various element packing and warp/wavefront variants during each autotune invocation. 

Example on a GH200,

```shell
    ---Autotune Ax helm (lx: 8)---
    1D    ch=1024:    138.30 us/call
    1D    ch=512 :    108.47 us/call
    1D    ch=256 :    100.68 us/call
    1D    ch=128 :    107.62 us/call
    KSTEP eb=1   :    114.23 us/call
    KSTEP eb=2   :    121.32 us/call
    KSTEP eb=4   :    129.73 us/call
    DMMA  2w 1e  :    100.14 us/call
    DMMA  4w 1e  :     91.31 us/call
    DMMA  8w 1e  :     95.63 us/call
    Chose        : 3 (DMMA, 4 warps, 1 elem/blk)
```

### DIFF
--- a/doc/pages/user-guide/performance.md
+++ b/doc/pages/user-guide/performance.md
@@ -108,7 +108,7 @@ provided (see #case-file).
 On the CUDA and HIP backends the spectral element operators --- the
 Helmholtz operator `Ax`, and `opgrad`, `dudxyz`, `cdtp`, `conv1`,
 `convect_scalar` and `lambda2` --- each ship two kernel formulations,
-and the best one depends on the polynomial order, the element count and
+and `Ax` a third, and the best one depends on the polynomial order, the element count and
 the hardware. Rather than fix a choice at build time, Neko benchmarks
 them on the first call of each operator, using the real element count of
 the running case, and caches the winner for the rest of the run.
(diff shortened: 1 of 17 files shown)


src/math/bcknd/device/cuda/ax_helm.cu
```diff
@@ -548,16 +738,302 @@ int tune_padded(void *w, void *u, void *dx, void *dy, void *dz,
...
+            NEKO_DMMA_NW(best3), NEKO_DMMA_PACK(LX));
+  }
+  log_message(neko_log_buf);
+  log_end_section();
+  return retval;
+}
+
+
+/*
+ * Autotuner for the vector Ax
```
timfelle, line 764: Are we getting to a point where it might make sense to create a folder for the
autotuning system? Instead of it being scattered around the code?

doc/pages/user-guide/performance.md
```diff
@@ -108,7 +108,7 @@ provided (see #case-file).
 On the CUDA and HIP backends the spectral element operators --- the
 Helmholtz operator `Ax`, and `opgrad`, `dudxyz`, `cdtp`, `conv1`,
 `convect_scalar` and `lambda2` --- each ship two kernel formulations,
-and the best one depends on the polynomial order, the element count and
+and `Ax` a third, and the best one depends on the polynomial order, the element count and
 the hardware. Rather than fix a choice at build time, Neko benchmarks
 them on the first call of each operator, using the real element count of
 the running case, and caches the winner for the rest of the run.
```
timfelle: Line out of bounds.
```suggestion
On the CUDA and HIP backends the spectral element operators --- the
Helmholtz operator `Ax`, and `opgrad`, `dudxyz`, `cdtp`, `conv1`,
`convect_scalar` and `lambda2` --- each ship two kernel formulations,
and `Ax` a third, and the best one depends on the polynomial order, the element 
count and the hardware. Rather than fix a choice at build time, Neko benchmarks
them on the first call of each operator, using the real element count of
the running case, and caches the winner for the rest of the run.
```

timfelle (approved):

timofeymukha (approved):

src/math/bcknd/device/cuda/ax_helm.cu
njansson: Not yet I would say, it's not general enough. But we have made steps to move boilerplate redundant code into a common header for each backend.

PR #2726, opened 2026-08-24
Freshen up bicgstab

timofeymukha:
My goal is to get a reference implementation blessed, and then ask a private consultant to port to the device and make a coupled version. 

Here:
 - Bring the documentation of the type up to standard.
 - Add scale-aware breakdown checks for rho, alpha, and omega. The code is now generally more defensive / explicit than the other solvers at the cost of some `if`s. If it is not acceptable, I can remove it.
 - Permit negative omega. A friend says that negative omega is valid for non-symmetric systems.
 - Handle initially converged systems without entering the iteration.


src/krylov/bcknd/cpu/bicgstab.f90
```diff
@@ -251,7 +334,110 @@ function bicgstab_solve(this, Ax, x, f, n, coef, blst, gs_h, niter) &
...
+    product = reductions(1)
+    norm_squared = reductions(2)
+
+  end subroutine bicgstab_product_and_norm
+
+  !> Return a vector norm from its weighted squared norm.
+  !! @param norm_squared Weighted squared norm.
+  !! @param quantity Name of the vector for error reporting.
+  !! @return The non-negative vector norm.
+  function bicgstab_vector_norm(norm_squared, quantity) result(vector_norm)
```
pohm01: I think this function name is misleading. Maybe just call it `bicgstab_sqrt`?

src/krylov/bcknd/cpu/bicgstab.f90
```diff
@@ -175,72 +211,119 @@ function bicgstab_solve(this, Ax, x, f, n, coef, blst, gs_h, niter) &
...
+         alpha = rho_1 / alpha_denominator
+         if (.not. ieee_is_finite(alpha)) then
+            call neko_error('BiCGStab failure: non-finite alpha')
+         end if
+
          call copy(s, r, n)
          call add2s2(s, v, -alpha, n)
-         rtr = glsc3(s, coef%mult, s, n)
-         rnorm = sqrt(rtr) * norm_fac
+         str = glsc3(s, coef%mult, s, n)
```
pohm01: Maybe I've just never understood the notation, but wouldn't `str` make more sense as `sts`, as in `transpose(s)*s`, which is what I thought `rtr` was a stand-in for?

(suppose the same logic would also rename `vtr` -> `ftv` and `ttr` -> `stt`)

pohm01 (changes requested):

src/krylov/bcknd/cpu/bicgstab.f90
timofeymukha: done

src/krylov/bcknd/cpu/bicgstab.f90
timofeymukha: - str → sts for (s^T s)
  - alpha_denominator → ftv for (f^T v)
  - vtr → vtv for (v^T v)
  - omega_numerator → stt for (s^T t)
  - ttr → ttt for (t^T t)

pohm01 (approved):

njansson (approved):
Great, remaining omp tuning we can do in a different pr 🐈

PR #2727, opened 2026-08-24
Quality of life fixes in coef, reduced memory footprint and element distortion check

njansson:
This PR adds some quality of life and improvements to coefficients. First it allows a reduce coefficient object to be built, with only the arrays necessary for our operators (e.g. multigrid levels in `hsmg` and `phmg`), thus reducing our memory footprint (according to a friend, for `phmg` coefficient related memory in coarse level will be reduce by 70% (*^-^*)v )

Furthermore, it also measure the worst element distortion, and throws a warning if the measure is larger than what the current working precision (`rp`) can handle.


pohm01:
Could the small eig_sym helper functions be useful elsewhere such that one might want them somewhere like `math` instead of `coef`?

njansson:
> Could the small eig_sym helper functions be useful elsewhere such that one might want them somewhere like `math` instead of `coef`?

True, good point, I'll move them to math instead.

njansson:
> > Could the small eig_sym helper functions be useful elsewhere such that one might want them somewhere like `math` instead of `coef`?
> 
> True, good point, I'll move them to math instead.

@pohm01 Fixed!

pohm01 (approved):

timfelle (approved):

PR #2728, opened 2026-08-25
Allow independent log_t instances with custom env-var prefixes

timfelle:
Introduce an optional `env_prefix` argument to `log_init`, enabling separate configuration for multiple `log_t` instances through distinct environment variables. 
Move the `deprecated_list` to a private component to prevent interference between instances.


timofeymukha:
So this is for using this in a library setting only, right? Where you init a logger yourself by passing this argument? Neko itself will still always just uses NEKO_ I guess. Out of curiosity, isn't it easier at this point to just pass the parameters to the logger directly instead of reading them from env variables? Maybe I misunderstand how you use this stuff :-)

timfelle (commented):
> So this is for using this in a library setting only, right? Where you init a logger yourself by passing this argument? Neko itself will still always just uses NEKO_ I guess. Out of curiosity, isn't it easier at this point to just pass the parameters to the logger directly instead of reading them from env variables? Maybe I misunderstand how you use this stuff :-) 

I completely agree, but we have users who insists that the log should go to stdout. Therefore it must be configurable at runtime, and I think it is the sort of thing which belong in an ENV rather than the case file, but maybe not.

Mr-2c:
> > So this is for using this in a library setting only, right? Where you init a logger yourself by passing this argument? Neko itself will still always just uses NEKO_ I guess. Out of curiosity, isn't it easier at this point to just pass the parameters to the logger directly instead of reading them from env variables? Maybe I misunderstand how you use this stuff :-)
> 
> I completely agree, but we have users who insists that the log should go to stdout. Therefore it must be configurable at runtime, and I think it is the sort of thing which belong in an ENV rather than the case file, but maybe not.

do I know those users? :-) 

is this an alternative to flushing, or is it complementary to it because flushing only always works if it was sent to stdout? or will the output be always shown if sent to stdout so there is no need for flushing anymore?

timfelle:
> > > So this is for using this in a library setting only, right? Where you init a logger yourself by passing this argument? Neko itself will still always just uses NEKO_ I guess. Out of curiosity, isn't it easier at this point to just pass the parameters to the logger directly instead of reading them from env variables? Maybe I misunderstand how you use this stuff :-)
> > 
> > I completely agree, but we have users who insists that the log should go to stdout. Therefore it must be configurable at runtime, and I think it is the sort of thing which belong in an ENV rather than the case file, but maybe not.
> 
> do I know those users? :-) 
> 
> is this an alternative to flushing, or is it complementary to it because flushing only always works if it was sent to stdout? or will the output be always shown if sent to stdout so there is no need for flushing anymore?

This has nothing to do with flushing. And flushing will also affect buffers which do not target stdout. This allow me a clean stdout for job related information, neko.log for logs related to neko and the simulation and optimizer.log for things related to optimization iterations and such.

timofeymukha (approved):

timofeymukha:
@Mr-2c merge?

Mr-2c:
> @Mr-2c merge?

It looks ok to me and I can approve, but I felt I was the least qualified person in the reviewers list to approve so I didn't comment :D

adperezm (approved):

PR #2730, opened 2026-08-25
OIFS fixes

timofeymukha:
This is a second set of OIFS fixes.

- The OpenCL code launched the kernel with incorrect argument numbering and also used the wrong program name.
- OIFS + advection=false did not work since OIFS high jacked correct BDF term treatment. We now check for advection to be true when checking for oifs.
- A subtle issue for OIFS in the scalar scheme. The fluid runs first, so it advances the fluid solution arrays by the time it reaches the scalar. This makes the scalar use an updated convecting velocity, which it should not. So we now pick up ulag, vlag, and wlag instead.


njansson (approved):
Great stuff! 🍻

timfelle (approved):

PR #2731, opened 2026-08-25
Probes moving mesh

arash-mp:
This PR makes probes work for ALE simulations + optimisations in AABB_tree sort and initialization.

In an ALE simulation, typically 2 types of probes may be needed.
First, a fix probe in the lab-frame (`fixed` mode), so like measuring velocity at a fixed point in the domain for vortex shedding, etc.
And second, probes which are fixed relative to a moving body, and need to move rigidly with respect to a particular body (`body_attached` mode), for BL measurements etc.

In the second case, we can simply find the current positions of the probes from their initial positions, and current rotation matrices and pivot locations, given by neko_ale.

The main challenge is that the interpolation mapping, the rank, element and `rst` each probe was found at, is only valid for the mesh it was found on. When the mesh moves, it goes stale in both modes, so a new mapping has to be found, and only when needed. On top of that, when a probe is fixed in the lab frame, it can go out of the domain and get lost (e.g. when the body passes over the probe), but it needs to be refound when it comes back into the physical domain. These also needs to be considered.

The way that I trigger this process is by measuring how stale the current mapping is. Basically, I find the `target` positions of the probes at the current time step (the initial position for `fixed` mode, the current position for `body_attached` mode), and compare them with the location the current mapping will sample at (this is just by evaluating the current `dof%x,y,z` at the probe locations). If the max distance between the target position and the evaluated position is larger than `max_position_drift`, the global interpolation is initialised again (without any redistribution of points) and the new mapping is found. All of this is a no-op on a static mesh.

**EDIT:**
**The follwoing is removed from the current PR. I will open sepearte PRs for them, but I will keep the description below.**

The main bottleneck in this process _was_ the initialization of the global interpolator, which was very slow on meshes that I was trying (660k elements, with cartesian PE/EL finder it was taking order of minutes, and with AABB order of 2-3 seconds). Considering that this has to be called frequently during a run, it was not practical. So, here it comes the second part of this PR, which is some optimisation for AABB_tree. The optimisations are mainly implemented by Claude (Opus 5), but I checked them, and they seems to be totally ok.

The main things changed are (a) the `sort` function in `aabb_tree.f90`, and (b) `aabb_pe_finder` now builds its local tree with a single `build_from_aabb` instead of per-element `insert_object` (which is now similar to what is happening in `aabb_el_finder`). The existing sort function was O(E^2), which essentially was doing (maybe a bug?)

```.f90
do i = 1, size(array)
   indices(i) = i
end do
````
at the cost of O(E^2). Now, it's replaced with a Morton sort, which is O(ElogE) and the resulting tree also seems to improve the performance of global_interpolation.

Based on the tests that I did (all with `load_balancing=true`, 96 GPUs, Lumi dev-g), for a mesh with 660k elements:

```
Cartesian EL/PE finder initialized in \approx. 157 seconds
AABB EL/PE finder (develop branch) initialized in \approx. 2.0 seconds
AABB EL/PE finder (current PR) initialized in \approx. 0.3 seconds
```

### CODE REVIEW

**Comment by njansson** [comment]
Great stuff! Does it still work as before on both CPU and Device backends?

File: src/global_interpolation/legendre_rst_finder.f90 [hunk 1]
```diff
@@ -100,21 +101,16 @@ subroutine legendre_rst_finder_init(this, x, y, z, nelv, Xh, tol, max_iter)
...
-    call tnsr3d_cpu(this%z_hat%x, Xh%lx, z, &
+    call tnsr3d(this%z_hat%x, Xh%lx, z, &
          Xh%lx, Xh%vinv, &
          Xh%vinvt, Xh%vinvt, nelv)
 
-    !> Copy the data to the device (if device exists)
-    call this%x_hat%copy_from(HOST_TO_DEVICE, .false.)
-    call this%y_hat%copy_from(HOST_TO_DEVICE, .false.)
-    call this%z_hat%copy_from(HOST_TO_DEVICE, .false.)
-
```
njansson, line 117: Does this removal mean that legendre_rst_finder is doing it's work on a gpu ?

src/global_interpolation/legendre_rst_finder.f90
```diff
@@ -100,21 +101,16 @@ subroutine legendre_rst_finder_init(this, x, y, z, nelv, Xh, tol, max_iter)
     call this%y_hat%init(nelv*Xh%lxyz)
     call this%z_hat%init(nelv*Xh%lxyz)
 
-    call tnsr3d_cpu(this%x_hat%x, Xh%lx, x, &
+    call tnsr3d(this%x_hat%x, Xh%lx, x, &
```
njansson: If we have compiled for GPUs, doesn't we need `_cpu` in the call to avoid getting down on the gpu backend?

arash-mp: tnsr3d dispatches to the right backend, no? @njansson 
````
  subroutine tnsr3d(v, nv, u, nu, A, Bt, Ct, nelv)
    integer, intent(in) :: nv, nu, nelv
    real(kind=rp), intent(inout) :: v(nv*nv*nv, nelv)
    real(kind=rp), intent(in) :: u(nu*nu*nu, nelv)
    real(kind=rp), intent(in) :: A(nv, nu), Bt(nu, nv), Ct(nu, nv)
    type(c_ptr) :: v_d, u_d, A_d, Bt_d, Ct_d

    if (nelv .eq. 0) return

    if (NEKO_BCKND_SX .eq. 1) then
       call tnsr3d_sx(v, nv, u, nu, A, Bt, Ct, nelv)
    else if (NEKO_BCKND_XSMM .eq. 1) then
       call tnsr3d_xsmm(v, nv, u, nu, A, Bt, Ct, nelv)
    else if (NEKO_BCKND_DEVICE .eq. 1) then
       v_d = device_get_ptr(v)
       u_d = device_get_ptr(u)
       A_d = device_get_ptr(A)
       Bt_d = device_get_ptr(Bt)
       Ct_d = device_get_ptr(Ct)
       call tnsr3d_device(v_d, nv, u_d, nu, A_d, Bt_d, Ct_d, nelv)
    else
       call tnsr3d_cpu(v, nv, u, nu, A, Bt, Ct, nelv)
    end if

  end subroutine tnsr3d
```

njansson: exactly. But how I remember things, the rst_finder works on host data, and to avoid the automatic selection, one had to call the cpu backend to make sure the operator saw the correct data.

arash-mp: hmmm, I tried it on both CPU and GPU, both were the same. 
Let me check it more carefully to be sure nothing goes wrong, but in my tests everything was ok.

njansson: Interesting, if it does work without moving data even better!

vbaconnet (commented):
Would it be possible for you to separate the contributions in two different PRs, since they are not dependent on each other? I think we can all agree that we want to be careful with changes in the global interpolation, I believe it deserves its own PR.

arash-mp:
> Would it be possible for you to separate the contributions in two different PRs, since they are not dependent on each other? I think we can all agree that we want to be careful with changes in the global interpolation, I believe it deserves its own PR.

Yess sure.

adperezm:
How does the search scale?

The tree should be in general slower. Thats why Martin later added the Cartesian one.

arash-mp:
> How does the search scale?
> 
> The tree should be in general slower. Thats why Martin later added the Cartesian one.

@adperezm 
find points you mean? that is slower than Cartesian but initialization is much faster. But the modificaitons in AABB apparantly have made it faster in all aspects comapred to the previous AABB version. Below are the timings for the case that I tested (with and without load_balancing. 660k elements, 96GPUs, 7500 probe points)

WITHOUT load_balancing:

THIS PR AABB:
  -----Global Interpolation-----
  Using AABB element finder
  Using AABB PE finder
  Global interpolation initialized (s):  **0.3117360E+00**
  -----Global Interpolation-----
  Found PE candidates time since start of findpts (s):  0.3228967E+00
  Sent to points to PE candidates, time since start of find_points (s):  0.3539621E+00
  Element candidates found, now time for finding rst, time since start of find_points (s):  0.4165271E+00
  Found rst with Newton iteration, time (s):  0.3614382E-02
  Global interpolation find points done, time (s):  **0.4412741E+00**

Develop AABB:
 -----Global Interpolation-----
  Using AABB element finder
  Using AABB PE finder
  Global interpolation initialized (s):  **0.4489095E+01**
  -----Global Interpolation-----
  Found PE candidates time since start of findpts (s):  0.3361845E+00
  Sent to points to PE candidates, time since start of find_points (s):  0.3830756E+00
  Element candidates found, now time for finding rst, time since start of find_points (s):  0.4646846E+00
  Found rst with Newton iteration, time (s):  0.3605234E-02
  Global interpolation find points done, time (s):  **0.4936018E+00**

Cartesian:
  -----Global Interpolation-----
  Using Cartesian element finder
  Using Cartesian PE finder
  Global interpolation initialized (s):  **0.9865453E+02**
  -----Global Interpolation-----
  Found PE candidates time since start of findpts (s):  0.1000553E-01
  Sent to points to PE candidates, time since start of find_points (s):  0.1277952E-01
  Element candidates found, now time for finding rst, time since start of find_points (s):  0.1520111E-01
  Found rst with Newton iteration, time (s):  0.5482570E-02
  Global interpolation find points done, time (s):  **0.3460378E-01**

WITH load_balancing:

THIS PR AABB:

 -----Global Interpolation-----
  Initializing global interpolation
  Using AABB element finder
  Using AABB PE finder
  Global interpolation initialized (s):  **0.3269140E+00**
  -----Global Interpolation-----
  Found PE candidates time since start of findpts (s):  0.2242019E+00
  Sent to points to PE candidates, time since start of find_points (s):  0.2396847E+00
  Element candidates found, now time for finding rst, time since start of find_points (s):  0.2576056E+00
  Found rst with Newton iteration, time (s):  0.2575603E-02
  Global interpolation find points done, time (s):  **0.2712421E+00**

Develop AABB:
  -----Global Interpolation-----
  Initializing global interpolation
  Using AABB element finder
  Using AABB PE finder
  Global interpolation initialized (s):  0.2066817E+01
  -----Global Interpolation-----
  Found PE candidates time since start of findpts (s):  0.1668122E+00
  Sent to points to PE candidates, time since start of find_points (s):  0.1737702E+00
  Element candidates found, now time for finding rst, time since start of find_points (s):  0.1868817E+00
  Found rst with Newton iteration, time (s):  0.2805317E-02
  Global interpolation find points done, time (s):  **0.1978969E+00**

Cartesian:
  -----Global Interpolation-----
  Initializing global interpolation
  Using Cartesian element finder
  Using Cartesian PE finder
  Global interpolation initialized (s):  **0.1545593E+03**
  -----Global Interpolation-----
  Found PE candidates time since start of findpts (s):  0.2469615E-01
  Sent to points to PE candidates, time since start of find_points (s):  0.2497041E-01
  Element candidates found, now time for finding rst, time since start of find_points (s):  0.2509514E-01
  Found rst with Newton iteration, time (s):  0.4382190E-03
  Global interpolation find points done, time (s):  **0.2680044E-01**

I tried with 100k prob points also and  Global interpolation find points done, time (s) will increase to about **0.9 s**.

So it when global interpolation needs to be initialized during simulation, the overall gain is quite significant.

I will make this PR into 3 sepearte PRs, so one can experiemnt with them better. Part of the imprrovment is also due to the changein rst_finder%init. The interpolation results also remiain unchanged between different setups, and different backends that I tired.

arash-mp:
@vbaconnet This is now only the probes PR.

src/global_interpolation/global_interpolation.f90
```diff
@@ -1230,10 +1240,19 @@ subroutine global_interpolation_find_and_redist(this, xyz, n_points)
...
+          call device_unmap(this%el_owner0, this%el_owner0_d)
+       end if
+    end if
     this%el_owner0 = this%el_owner0_local
     if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_map(this%el_owner0, this%el_owner0_d, &
+            size(this%el_owner0))
        call device_memcpy(this%el_owner0, this%el_owner0_d, &
-            this%n_points, HOST_TO_DEVICE, sync = .true.)
+            size(this%el_owner0), HOST_TO_DEVICE, sync = .true.)
```
arash-mp, line 1255: This part actually fixes a bug which leads to Inconsistent host/device mapping state in device_unmap when a simulation with probes ends (only more than 1 GPU). I faced this on current develop branch (not sure if others have also experienced it).

PR #2732, opened 2026-08-25
Don't assume wave-size to be 64 on AMD

### LINKED ISSUE #2729: Consumer-grade AMD cards currently not supported
Probably due to different warp size.

njansson:
Fixes #2729

### DIFF
--- a/src/math/bcknd/device/hip/elem_block.h
+++ b/src/math/bcknd/device/hip/elem_block.h
@@ -34,12 +34,15 @@
  POSSIBILITY OF SUCH DAMAGE.
 */
 
+#include "wave.h"
+
 /**
  * Elements per thread block for the SEM operator kstep kernels (HIP)
  *
- * A kstep block is one (LX,LX) thread plane per element. Against a 64 wide
- * wavefront that masks off a lot of lanes for every LX whose square is not a
- * multiple of the wave: 25% useful at LX = 4, 39% at LX = 5, 63% at LX = 9.
+ * A kstep block is one (LX,LX) thread plane per element. Against a wavefront
+ * that masks off a lot of lanes for every LX whose square is not a multiple
+ * of the wave: at the 64 lanes of CDNA, 25% useful at LX = 4, 39% at LX = 5,
+ * 63% at LX = 9.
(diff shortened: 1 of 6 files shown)


timfelle (approved):

src/math/bcknd/device/hip/elem_block.h
```diff
@@ -34,34 +34,37 @@
...
+ * not yet understood.
  *
  * Candidate 0 is one element per block. Candidate C > 0 fits as many whole
  * elements as it can into 2^(C+1) wavefronts.
+ *
+ * NEKO_EB_WAVE is deliberately NEKO_WAVE_SIZE_UNIFORM and not NEKO_WAVE_SIZE:
+ * it sizes a kernel template argument and the matching launch geometry, which
+ * the host and device passes have to agree on, and only the command line form
+ * of the width is visible to both. On RDNA, pass -DNEKO_WAVE_SIZE=32 to size
+ * the candidates for a 32 lane wave; getting it wrong costs occupancy, not
```
timofeymukha, line 63: -DNEKO_WAVE_SIZE=32 looks like a cmake argument, or? If we want to to document this here, we should provide the right way to do it in our `configure`. Better still, we should really mention this in install.md !

njansson: The new `wave.h` misses some macros, passing `-DNEKO_WAVE_SIZE` should be seen as a fallback if all fails. I've fixed this now, hopefully.

timofeymukha: I am just confused regarding what we pass this to. Do i write `./configure -DNEKO_WAVE_SIZE=32 `?

njansson: sorry my bad, it should be pass under the hipcc compiler flags

njansson: fair point, let me check with claude what a suitable doc entry would be

njansson: @timofeymukha Claude typed up some nice stuff, please have a look

timofeymukha (approved):

PR #2734, opened 2026-08-26
Add templates for user-defined types

timofeymukha:
I thought it would be nice to have ready-made templates for all the possible user-injectable types. This way instead of "cleaning" up a copied existing type in neko, one can start with a fresh, empty, but compilable template.
The template compilation is added to the CI.


vbaconnet (approved):
nice :)

timfelle (approved):

PR #2735, opened 2026-08-26
Modify the axhelm_factory to accept a typename and user-defined types

timofeymukha:
This is partially prompted by the SVV PR by @Shiyu-Sandy-Du , where new Ax types are introduced and the factory has to add an extra optional argument and the `if`s start looking scary ^^. After a bit of discussion with @njansson, we now change the factory to accept a type name instead (like the other factories).  This means that any messy selection based on configuration is the burden of the callee now, and the factory is clean. In more detail the changes are:

- ax_helm_factory -> ax_helm_allocator. The routine does not run a constructor (there isn't one for this type), so by convention this is an allocator.
- The allocator now takes the name of the type: currently "standard" or "full". No other arguments besides for the object currently taken.
- Added the usual machinery of user-type injection and a corresponding template. 

@Shiyu-Sandy-Du if you don't mind let's merge this in first, and then you can adapt the SVV PR to this.


timfelle (approved):

shiyu-du-dsy (approved):
Nice! Now the code for SVV could be clearner!

src/math/ax_helm_fctry.f90
```diff
@@ -77,9 +75,66 @@ module subroutine ax_helm_factory(object, full_formulation)
        else
           allocate(ax_helm_cpu_t::object)
        end if
-    end if
+    case ("full")
+       if (NEKO_BCKND_SX .eq. 1 .or. NEKO_BCKND_XSMM .eq. 1) then
```
njansson: For SX we should fallback on CPU (thus only throw error on xsMM)

njansson:
```suggestion
       if (NEKO_BCKND_XSMM .eq. 1) then
```

njansson:
Great stuff indeed, will simplify things a lot with custom bilinear forms in the future 🍺

PR #2738, opened 2026-08-26
Global interpolator improvments

arash-mp:
This PR makes the `rst_finder%init` to run on device, when compiled on device.
This helps a lot when global_interpolator needs to be re-initialzed frequently during a run.

All the inputs to the `rst_finder%init` are already synced to device before the call, so running `tnsr3d_device` is safe.
The outputs of `legendre_rst_finder_init` are only `x_hat`, `y_hat`, `z_hat`, which before they were copied to dev memory. The only place which  `x_hat`, `y_hat`, `z_hat` are used is in `rst_finder%find`, which already was being done on `find_rst_legendre_device`, so nothing changed here.
Also, since  `x_hat`, `y_hat`, `z_hat` on the host side are only used in `find_rst_legendre_cpu`, there is no need to have them on host (during rst_finder%init) when we're on device.
I compared the interpolation results with the original code and everything looks good, but if I'm missing anything, I'm happy to fix.

Furthurmore, I added the optional `max_iterations` in Newton iteration to case file (similar to padding and tolerance), as it was not possible to change it, and in some cases one may need to increase it.  If Newton iteration is not converged in any rank, that rank will print a warning. The default is kept 10, as before.

It's better if this PR get merged after #2731.


timfelle (approved):

adperezm:
Wasn't there a max iteration before on rst?

src/global_interpolation/legendre_rst_finder.f90
```diff
@@ -216,6 +226,14 @@ subroutine find_rst_legendre_device(this, rst, pt_x, pt_y, pt_z, &
        if( iter .ge. this%max_iter) converged = .true.
     end do
 
+    ! A nonzero count here means max_iter cut the iteration off.
+    if (nint(conv_sum) .gt. 0) then
+       write(warn_buf, '(I0,A,I0,A,I0,A)') nint(conv_sum), ' of ', n_pts, &
+            ' points did not converge in ', this%max_iter, &
+            ' Newton iterations (rst finder, device)'
+       call neko_warning(trim(warn_buf))
```
adperezm, line 234: I am not sure any type of logging should happen in the rst_finder.

It should, in my opinion, be controlled by the global interpolator object. Ourside the rst calls.

Particularly if the interpolator sends the points to more than one rank. It might not be in one rank but in another. This type of log needs to happen just at the end of the process.

arash-mp:
> Wasn't there a max iteration before on rst?

the variable was there, but hardcoded.

adperezm (commented):
There is one thing. 

There are some options in the interpolator or in probes that allow to do some things in host. I do not think anyone uses those, but does this change affect in any way that path?

(see line 1077 that has the option to evaluate in host)

If it is verified that this does not change that, then I agree all is fine.

adperezm:
> > Wasn't there a max iteration before on rst?
> 
> the variable was there, but hardcoded.

Yes thanks! now I saw that we had to set it with env variables before. Martin really liked those :D. I also prefer to have it in the case file.

PR #2739, opened 2026-08-27
Add coupled_bicgstab on the CPU

timofeymukha:
Adds the coupled version of bicgstab, which should be suitable for the full-stress formulation when the system is non-symmetric.

Implemented by Sol 5.6 High, but looks to me like a reasonable extenstion of the baseline version, similar to coupled_cg.


njansson (approved):

pohm01 (approved):

src/krylov/bcknd/cpu/bicgstab_coupled.f90
```diff
@@ -0,0 +1,602 @@
...
+     !> Three-component residual \f$r\f$.
+     real(kind=rp), allocatable :: r(:, :)
+     !> Intermediate residual \f$s = r - \alpha v\f$.
+     real(kind=rp), allocatable :: s(:, :)
+     !> Preconditioned intermediate residual \f$\hat{s} = M^{-1}s\f$.
+     real(kind=rp), allocatable :: s_hat(:, :)
+     !> Coupled operator action \f$t = A\hat{s}\f$.
+     real(kind=rp), allocatable :: t(:, :)
+     !> Coupled operator action \f$v = A\hat{p}\f$.
+     real(kind=rp), allocatable :: v(:, :)
```
timfelle: Do you really need all of these to be stored throughout? Or can some of them be
substituted to scratch arrays. It seems like you fill r with new values at every
call for example.

timofeymukha: It seems pretty much everything is scratch material!!

timofeymukha: I have now completely removed persistent storage for both the coupled and non-coupled solver1

njansson: How about keeping the pointers for the scratch arrays in the type (and reassign at each solve).

It will eat some bytes but would make the type a bit more readable for a linear solvers guy imo

njansson: Is this (scratch things mentioned by @timfelle ) fixed now? such that we can merge

src/krylov/bcknd/cpu/bicgstab.f90
```diff
@@ -214,8 +167,31 @@ function bicgstab_solve(this, Ax, x, f, n, coef, bc_projector, gs_h, niter) &
     end if
     norm_fac = 1.0_rp / sqrt(coef%volume)
 
-    associate(r => this%r, t => this%t, s => this%s, v => this%v, &
-         p => this%p, s_hat => this%s_hat, p_hat => this%p_hat)
+    block
+      type(host_array_t), pointer :: p_tmp, p_hat_tmp, r_tmp
+      type(host_array_t), pointer :: s_hat_tmp, t_tmp, v_tmp
+      real(kind=rp), pointer :: p(:), p_hat(:), r(:), s_hat(:), t(:), v(:)
```
njansson: can we move these pointers up in the type. As we have done with the other ksp types. (to ease readability for linear solver people)

PR #2744, opened 2026-08-30
Refactor mesh representation, reducing memory footprint by ~80%

njansson:
This PR refactors the mesh representation, replacing hash tables with static arrays where possible. This results in a reduced memory footprint by approximately 80%.


src/mesh/mesh.f90
```diff
@@ -162,13 +185,10 @@ module mesh
...
+     !! @attention only valid while the mesh is being built, see
+     !! @a mesh_get_local_point
      !! @todo Add similar mappings for element ids
-     generic :: get_local => get_local_point, get_local_edge, get_local_facet
-     !> Get global id for a mesh entity
-     !! @todo Add similar mappings for element ids
-     generic :: get_global => get_global_edge, get_global_facet
-     !> Check if a mesh entity is shared
-     generic :: is_shared => is_shared_point, is_shared_edge, is_shared_facet
+     generic :: get_local => get_local_point
```
timfelle: Should this not be private then? Is it not only constructed as part of this module?

njansson: Indeed, will fix

njansson: it was actually a left over, not called anymore ;)

njansson:
Turns out that this fixes various issues with CCE, taking a long vacation when cleaning some internal mesh structures, i.e. connectivity and finalise is now fast

timofeymukha (approved):

tuananhdao (approved):

PR #2747, opened 2026-08-31
Add mention of rp kind to agents.md

timofeymukha:
I feel like AI setting a tight tolerance in tests that then fail in the single-precision CI runs is a recurring theme. Hopefully this addition to AGENTS.md will help.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -39,7 +39,9 @@ simulation case, use the project-level skill in
     validation tests checking that important cases produce the expected output.
   - *Important*: Since Neko can be run in both double and single precision,
     numerical assertions in tests should adjust their tolerance based on the
-    kind of `rp`. 
+    kind of `rp`. The `math` module contains the `NEKO_EPS` parameter that is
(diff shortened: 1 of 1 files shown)


njansson (approved):

AGENTS.md
```diff
@@ -37,6 +37,9 @@ simulation case, use the project-level skill in
     supercomputer via a gitlab pipeline. The tests are written using
     [reframe](https://reframe-hpc.readthedocs.io/en/stable/). These are
     validation tests checking that important cases produce the expected output.
+  - *Important*: Since Neko can be run in both double and single precision,
+    numerical assertions in tests should adjust their tolerance based on the
+    kind of `rp`.
```
timfelle: Somehow we should probably say the tolerances should be based on NEKO_EPS, since
that one adjusts with precision, and already represent what we believe is a
minimal possible tolerance to use.

timofeymukha: Good point, see update

timfelle (approved):

PR #2749, opened 2026-09-01
Add tensor3_t and tensor4_t container types

timfelle:
This is the first PR on the path to simplifying the device array handling. This have significant potential across modules like coef.

Introduce tensor3_t and tensor4_t as rank-3 and rank-4 container types, respectively, mirroring the functionality of vector_t and matrix_t. Update field_t to extend tensor4_t, simplifying its implementation. Include unit tests for both new tensor types, verified on CPU and CUDA backends.


src/math/tensor4.f90
```diff
@@ -0,0 +1,315 @@
...
+       if (NEKO_BCKND_DEVICE .eq. 1) then
+          call device_map(t%x, t%x_d, t%n)
+       end if
+
+    end if
+
+    if (NEKO_BCKND_DEVICE .eq. 1) then
+       call device_copy(t%x_d, w%x_d, t%n)
+    else
+       t%x = w%x
```
njansson, line 264: OpenMP would help here (and in similar places)

src/field/field.f90
```diff
@@ -154,88 +163,78 @@ subroutine field_init_external_dof(this, dof, fld_name)
...
-         end block
-      end if
-
-      ! Zero on the host after the device-side memset: under zero-copy
-      ! the device then faults the pages first (device first touch),
-      ! which gives contiguous physical mappings and thus better GPU
-      ! TLB utilisation; rewriting the zeros on the host is benign
-      if (fresh) then
-         this%x = 0.0_rp
+         call this%tensor4_t%init(lx, ly, lz, nelv, "Field")
```
njansson: Would be good if one can retain that first touch on GPUs if possible (helps on unified archs)

njansson (commented):
As long as one can access `x` and `x_d` via the field (yes `f%t%x` is not the best for performance, but it works), I think this is a good improvement for encapsulation.

Left some general comments, that is valid in several places

timfelle:
> As long as one can access `x` and `x_d` via the field (yes `f%t%x` is not the best for performance, but it works), I think this is a good improvement for encapsulation.
> 
> Left some general comments, that is valid in several places

It is better than `f%t%x`, the original `f%x` is 100% retained. It is not that fields now includes a tensor, they are a special type of tensor. `type, extends(tensor4_t) :: field_t`

src/field/field.f90
timfelle: It is explicitly retained in the `tensor4_t`. So first touch is still there.

timfelle:
```fortran
  type, public, extends(tensor4_t) :: field_t
     ! ...
   contains
     procedure, private, pass(this) :: init_common => field_init_common
     procedure, private, pass(this) :: init_external_dof => &
          field_init_external_dof
     procedure, private, pass(this) :: init_internal_dof => &
          field_init_internal_dof
     generic :: init => init_external_dof, init_internal_dof
  end type field_t

  contains

  subroutine field_init_common(this, fld_name)
    class(field_t), intent(inout) :: this
    character(len=*), optional :: fld_name !< Name of the field

    ! ...
         call this%tensor4_t%init(lx, ly, lz, nelv, fld_name)
    ! ... 

  end subroutine field_init_common
  ```

One subtle thing I discovered was that an extending type can overwrite an operator `init` in this case, and still call the parent one through `this%parent_t%init`. That should help avoid all of the `init_base` methods we have everywhere. I have not tested this on a cray machine yet though, so it might not resolve in a great way. We really need to test this before merging.

njansson:
> ```fortran
>   type, public, extends(tensor4_t) :: field_t
>      ! ...
>    contains
>      procedure, private, pass(this) :: init_common => field_init_common
>      procedure, private, pass(this) :: init_external_dof => &
>           field_init_external_dof
>      procedure, private, pass(this) :: init_internal_dof => &
>           field_init_internal_dof
>      generic :: init => init_external_dof, init_internal_dof
>   end type field_t
> 
>   contains
> 
>   subroutine field_init_common(this, fld_name)
>     class(field_t), intent(inout) :: this
>     character(len=*), optional :: fld_name !< Name of the field
> 
>     ! ...
>          call this%tensor4_t%init(lx, ly, lz, nelv, fld_name)
>     ! ... 
> 
>   end subroutine field_init_common
> ```
> 
> One subtle thing I discovered was that an extending type can overwrite an operator `init` in this case, and still call the parent one through `this%parent_t%init`. That should help avoid all of the `init_base` methods we have everywhere. I have not tested this on a cray machine yet though, so it might not resolve in a great way. We really need to test this before merging.

Oooh happy days 🍻

timfelle:
> > ```fortran
> >   type, public, extends(tensor4_t) :: field_t
> >      ! ...
> >    contains
> >      procedure, private, pass(this) :: init_common => field_init_common
> >      procedure, private, pass(this) :: init_external_dof => &
> >           field_init_external_dof
> >      procedure, private, pass(this) :: init_internal_dof => &
> >           field_init_internal_dof
> >      generic :: init => init_external_dof, init_internal_dof
> >   end type field_t
> > 
> >   contains
> > 
> >   subroutine field_init_common(this, fld_name)
> >     class(field_t), intent(inout) :: this
> >     character(len=*), optional :: fld_name !< Name of the field
> > 
> >     ! ...
> >          call this%tensor4_t%init(lx, ly, lz, nelv, fld_name)
> >     ! ... 
> > 
> >   end subroutine field_init_common
> > ```
> > 
> > One subtle thing I discovered was that an extending type can overwrite an operator `init` in this case, and still call the parent one through `this%parent_t%init`. That should help avoid all of the `init_base` methods we have everywhere. I have not tested this on a cray machine yet though, so it might not resolve in a great way. We really need to test this before merging.
> 
> Oooh happy days 🍻 

Well it does not work. All we do there is shadow the original init with a generic, which i am not sure is a good choise, so lets not.

njansson (commented):
Great Stuff! 
🐈‍⬛ 🍻 🍻 🐈

njansson (approved):

timfelle:
@timofeymukha All done.

timofeymukha (approved):

PR #2754, opened 2026-09-02
Use scratch arrays in CG, coupled CG, and GMRES on the CPU

timofeymukha:
Should remove quite a bit of memory overhead on the CPU, since CG and GMRES are typically used in tandem.

The types retain pointers (instead of persistent storage) that are redirected to scratch arrays in the solve routine.


njansson (commented):
Same as for bicgstab, can we keep the pointers with the same name in the type, it eats some bytes but greatly improves readability for linear solver people looking at the code

timofeymukha:
> Same as for bicgstab, can we keep the pointers with the same name in the type, it eats some bytes but greatly improves readability for linear solver people looking at the code

We do!

src/krylov/bcknd/cpu/cg.f90
```diff
@@ -108,37 +116,23 @@ subroutine cg_init(this, n, max_iter, M, rel_tol, abs_tol, monitor)
...
+    nullify(this%p)
+    nullify(this%z)
+    nullify(this%alpha)
 
     nullify(this%M)
 
   end subroutine cg_free
 
-  !> Standard PCG solve
+  !> Solve a linear system with the CPU PCG method.
```
njansson:
```suggestion
  !> Solve a linear system with the Standard PCG method.
```

I would retain standard here, since there're so many variants of cg. If we want to emphasize cpu, then we can add it on a new line or maybe better at the module comment

src/krylov/bcknd/cpu/gmres.f90
```diff
@@ -405,14 +424,16 @@ function gmres_solve(this, Ax, x, f, n, coef, bc_projector, gs_h, niter) &
...
+    call neko_scratch_registry%relinquish_host_array(temp_indices)
     call this%monitor_stop()
     ksp_results%res_final = rnorm
     ksp_results%iter = iter
     ksp_results%converged = this%is_converged(iter, rnorm)
 
   end function gmres_solve
 
-  !> Standard GMRES coupled solve
+  !> Solve three independent systems with the CPU GMRES method.
```
njansson: Same here as for cg. In this setting people might confuse CPU as some gmres flavor

njansson (approved):

src/krylov/bcknd/cpu/cg.f90
```diff
@@ -162,6 +158,23 @@ function cg_solve(this, Ax, x, f, n, coef, bc_projector, gs_h, niter) &
...
+         temp_indices(5), CG_P_SPACE, .false.)
+
+    this%w => w_tmp%x
+    this%r => r_tmp%x
+    this%p(1:n, 1:CG_P_SPACE) => p_tmp%x
+    this%z => z_tmp%x
+    this%alpha => alpha_tmp%x
+
     associate(w => this%w, r => this%r, p => this%p, &
          z => this%z, alpha => this%alpha)
```
timfelle:
```suggestion
    call neko_scratch_registry%request(this%w, temp_indices(1), n, .false.)
    call neko_scratch_registry%request(this%r, temp_indices(2), n, .false.)
    call neko_scratch_registry%request(this%z, temp_indices(4), n, .false.)
    call neko_scratch_registry%request(this%alpha, temp_indices(5), CG_P_SPACE, .false.)
    call neko_scratch_registry%request(p_tmp, temp_indices(3), &
         n * CG_P_SPACE, .false.)

    this%p(1:n, 1:CG_P_SPACE) => p_tmp%x

    associate(w => this%w%x, r => this%r%x, p => this%p%x, &
         z => this%z%x, alpha => this%alpha%x)
```

timfelle (commented):
This can be done much cleaner with out having the subroutine temporaries if you use the host_array_t in the class.

src/krylov/bcknd/cpu/cg.f90
```diff
@@ -162,6 +158,23 @@ function cg_solve(this, Ax, x, f, n, coef, bc_projector, gs_h, niter) &
...
+    call neko_scratch_registry%request_host_array(p_tmp, temp_indices(3), &
+         n * CG_P_SPACE, .false.)
+    call neko_scratch_registry%request_host_array(z_tmp, temp_indices(4), &
+         n, .false.)
+    call neko_scratch_registry%request_host_array(alpha_tmp, &
+         temp_indices(5), CG_P_SPACE, .false.)
+
+    this%w => w_tmp%x
+    this%r => r_tmp%x
+    this%p(1:n, 1:CG_P_SPACE) => p_tmp%x
```
njansson, line 168: This makes me a bit nervous. Can't we add a rank-2 registry as well (@timfelle )?

timfelle: Array reshaping like this seem to be completely standard in the fortran community, But sure we can easily add another host_array_t to the system. But explicitly for a host implementation the only difference between host_array_t and vector_t is a %x_d=C_NULL_PTR, so I actually don't think the host array thing makes a difference at all since it should never be active in a device build.

timfelle: So my point is, you should probably just use a matrix_t here.

njansson: > Array reshaping like this seem to be completely standard in the fortran community, But sure we can easily add another host_array_t to the system. But explicitly for a host implementation the only difference between host_array_t and vector_t is a %x_d=C_NULL_PTR, so I actually don't think the host array thing makes a difference at all since it should never be active in a device build.

standard, and working on compilers are different things, as we know ;) But sure let's give it a try

timfelle: Sure, but we do it all over the place already, we just do it at subroutine interfaces, with out of bound indexing and such.

timfelle (approved):

src/krylov/bcknd/cpu/cg.f90
njansson: @timofeymukha use `matrix_t` 🐈

PR #2757, opened 2026-09-03
Refactor GS tuner with device backends

njansson:
This pr refactors the GS backend (init + tuner), to allow for auto tuner sweeps of all backends e.g. HOST + GPU for a GPU backend, or all possible device backends e.g. `NCCL`, `MPIGPU`,`CRYSTALGPU`.

However, the default is that the tuner is disabled for device backends (Controllable by the usual env. variables)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,38 @@
 
 ## Develop
 
+- The gather-scatter comm. backend autotuning can now benchmark the
+  device-resident backends. `MPIGPU`, `NCCL` and `CRYSTALGPU` join the
+  candidates on CUDA and HIP builds (`NVSHMEM` only when asked for), measured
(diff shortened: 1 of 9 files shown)


timfelle (commented):
Why not make it default tune for devices?

njansson:
> Why not make it default tune for devices?

true, I can split it into all device backends (I first had it to test everything host and device which didn't make sense)

timfelle:
> > Why not make it default tune for devices?
> 
> true, I can split it into all device backends (I first had it to test everything host and device which didn't make sense)

Unless you want to keep diving into the heterogenous systems. Then you would probably need to identify the best version for each hardware type.

timfelle (approved):

tuananhdao (approved):
LGTM!

PR #2759, opened 2026-09-04
Use scratch arrays in dealias adv

njansson:
This changes the advection (dealiasing) backend to use scratch arrays instead of allocated arrays. The main benefit from this work is to reuse the dealiasing memory for both fluid and scalars, roughly saving 17% memory per scalar added to the simulation.

### DIFF
--- a/src/fluid/bcknd/advection/adv_dealias.f90
+++ b/src/fluid/bcknd/advection/adv_dealias.f90
@@ -49,22 +49,78 @@ module adv_dealias
   implicit none
   private
 
-  !> Type encapsulating advection routines with dealiasing
-  type, public, extends(advection_t) :: adv_dealias_t
-     !> Coeffs of the higher-order space
-     type(coef_t) :: coef_GL
-     !> Coeffs of the original space in the simulation
-     type(coef_t), pointer :: coef_GLL
-     !> Interpolator between the original and higher-order spaces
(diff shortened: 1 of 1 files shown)


timfelle:
Why was the scratch registry not used instead of reimplementing it as a local subroutine?

njansson:
> Why was the scratch registry not used instead of reimplementing it as a local subroutine?

since the scratch space is tied to a specific function space, and keeping more registries active would eat more memory (at least that's the idea)

timfelle:
> > Why was the scratch registry not used instead of reimplementing it as a local subroutine?
> 
> since the scratch space is tied to a specific function space, and keeping more registries active would eat more memory (at least that's the idea)

But all you store are 1D arrays, so if you use a scratch vector for those you will reuse a lot of them

timfelle (commented):
> > Why was the scratch registry not used instead of reimplementing it as a local subroutine?
> 
> since the scratch space is tied to a specific function space, and keeping more registries active would eat more memory (at least that's the idea) 

Also I just realized a misunderstanding here.
It is only `field_t` objects requested which is tied to a function space (which we should also somehow fix so that they are not). Arrays, matrices and such are requested by their size. So an array of 1000 elements is just an array of 1000 elements, regardless of what it is used for.

PR #2761, opened 2026-09-04
Fix HDF5 session management

timfelle:
Implement a session management system for HDF5 to maintain a reference count, preventing premature closure of the library. This change introduces functions to initialize and finalize the session, ensuring proper handling of the library's state during file operations.

Also fixes the checkpoint mismatch between HDF5 and CHKP.


timofeymukha (approved):

njansson (approved):
Great. 

Let's remember to check on Fugaku some time in the future

PR #2764, opened 2026-09-05
Improvements to the time based write controller

Mr-2c:
When an output is written was derived from how many writes had been performed
so far: `check` returned true once `t` reached `nexecutions * time_interval`,
and a restart guessed `nexecutions` back from the time. Several ways of losing
or duplicating a file followed from that, and from the conditions around it.

`time_based_controller_t` now holds a schedule that is a property of the case
rather than of the run — for `simulationtime` the whole multiples of the
interval, for `nsamples` the division of the simulated interval — plus a
cursor into it. `check` is free of observable side effects and idempotent
within a time step, and a restart moves the cursor to the first scheduled time
not yet reached using exactly the predicate the schedule itself uses, so the
two runs together write what the uninterrupted run would have written.

This does not address file overwrites if run within the same folder. 
That will be something for a future PR.

## The issues this intends to fix

- `output_at_end` duplicated a scheduled write landing on the last time step,
  giving two files for one output time and, for the statistics, a second file
  of zeros ([#2278](https://github.com/ExtremeFLOW/neko/issues/2278)).
- A restart skipped everything scheduled within `0.1` time units of it: the
  write counter was reconstructed with a tolerance of `0.1 * dt`, where `dt`
  at that point is the placeholder a variable time step run starts from.
- Checkpoints and statistics were written at the start of a simulation, where
  a checkpoint holds nothing the initial condition does not and an average has
  nothing to average over ([#2378](https://github.com/ExtremeFLOW/neko/issues/2378)).
- The write scheduled for `end_time` was dropped when the last time step
  overshoots it, which it does whenever the step does not divide the interval.
- The schedule was anchored at `start_time`, so a run started from a field at
  t = 10.5 writing every 1.0 wrote at 11.5, 12.5, ... rather than 11.0, 12.0.
- `start_time` had no effect on an output: the condition gating it cancelled
  out algebraically, so the output was written from the beginning of the run.
- An interval smaller than the time step built up a backlog of writes instead
  of skipping the times the step jumped over, leaving the file counter behind.
- A `tsteps` output wrote at the step a restart resumed from, duplicating the
  file the checkpoint was taken with.
- An output starting after the end of the simulation took its direction of
  time from its own window, which let `output_at_end` force a write out of it.
- `time_state_t%is_done` took one time step more than asked for when the
  accumulated time fell a few ulps short of `end_time`, and never ended a
  simulation marching backwards in time.

## What is new

- `case.output_at_start` decides whether the initial state is written. It
  defaults to `true`, and to `false` when the fluid initial condition is a
  field file, where it would only copy that file.
- A restart warns when the first file an output is about to write already
  exists, which happens when a run repeats an interval or the output frequency
  was changed at the restart. The file numbering follows the schedule, so the
  same numbers then stand for different times.

## Testing

- The unit tests for `time_based_controller_t` are rewritten around the
  schedule: idempotency within a step, no duplicate at the end, the write at
  `end_time` when the last step overshoots, restart continuity from every step
  of a run, a restart of a variable time step run, anchoring, `nsamples`,
  `tsteps`, `never`, and an output that starts past the end.
- New integration test `test_output_schedule`, which checks the files a run
  and a restart of it produce, and that `output_at_start` suppresses the
  initial one.
- Off-tree, a harness compiling the real controller against stub modules and
  comparing it write by write to an independently enumerated schedule:
  34 000 scenarios over interval-to-timestep ratios from 1e-4 to 1e4, start
  times from 0 to 1e9, end times on and around a scheduled write, variable
  time steps, `nsamples` 1 to 60, `tsteps` 1 to 40, restarts after every step
  and up to three runs in a row, restarts with a changed time step, interval
  or start time, degenerate runs, hundred-thousand-step runs and backwards
  time. Happy to add it to `tests/` if it is wanted.

The reference logs of `test_cylinder` and `test_scalar_restart` lose their
last row: it was produced by the extra time step `is_done` used to take. The
`user_stats` test now compares its average of a random field against a
two-sided tolerance covering the sampling noise; the one-sided `1e-4` it used
passed or failed roughly at random.

## Acknowledgements
Implemented by Claude Code (Opus 5)


src/common/time_state.f90
```diff
@@ -148,11 +148,18 @@ subroutine time_state_status(this)
...
+  !! adding one more time step, which would overshoot the end time by almost
+  !! `dt` and produce an output past it.
   pure function time_state_is_done(this) result(is_done)
     class(time_state_t), intent(in) :: this
     logical :: is_done
+    real(kind=dp) :: span, tol
 
-    is_done = this%t - this%start_time .ge. this%end_time - this%start_time &
+    span = this%end_time - this%start_time
+    tol = 1.0e-9_dp * max(span, this%dt)
```
timfelle: This one does not retain the sign of the timeline.

njansson (approved):
🐈‍⬛

src/common/time_state.f90
timofeymukha: Tim, stepping back in time is now back in the PR.

Mr-2c:
How is this PR doing? :-)

timofeymukha:
> How is this PR doing? :-)

waiting for Tim to re-review with the go back in time stuff.

timfelle (approved):

PR #2769, opened 2026-09-08
Return region ID if not enabled

vbaconnet:
As the title says, if runtime stats were disabled but find_region_id was called anywhere in the code, the unallocated array `rt_stats_id` is referenced, throwing segfaults.

### DIFF
--- a/src/common/runtime_statistics.f90
+++ b/src/common/runtime_statistics.f90
@@ -300,6 +300,8 @@ contains
     integer, intent(out) :: region_id
     integer :: i
 
+    if (.not. this%enabled) return
+
     region_id = -1
 
     ! Look for the region name first
--- a/src/common/runtime_statistics.f90
+++ b/src/common/runtime_statistics.f90
@@ -300,11 +300,9 @@ contains
     integer, intent(out) :: region_id
     integer :: i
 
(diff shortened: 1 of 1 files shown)


timfelle (approved):

njansson (dismissed):

src/common/runtime_statistics.f90
```diff
@@ -300,6 +300,8 @@ subroutine runtime_stats_find_region_id(this, name, region_id)
     integer, intent(out) :: region_id
     integer :: i
 
+    if (.not. this%enabled) return
+
```
njansson, line 306: Wait, since region_id is intent(out), we have to assign a value

src/common/runtime_statistics.f90
```diff
@@ -300,6 +300,8 @@ subroutine runtime_stats_find_region_id(this, name, region_id)
...
     region_id = -1
```
njansson:
```suggestion

region_id = -1

    if (.not. this%enabled) return

```

Something like this to avoid violating the standards

vbaconnet:
good catch indeed

njansson (approved):

PR #2770, opened 2026-09-09
Enhance diagnostic of geometric factors

njansson:
This pr enhances the diagnostic of geometric factors (G) to also consider pertubation errors caused by G. Each check will throw a warning if condition numbers etc are above certain thresholds. Furthermore, the checks will also report wether it is safe or not to store factors in fp32. This is only used as diagnostic information for now, but in the future we might want to use fp32 G for fp64 runs on device backends, for example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 ## Develop
 
+- Added a setup-time conditioning diagnostic for the geometric factors, on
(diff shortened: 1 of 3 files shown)


timfelle (approved):
Some of those docstrings are getting close to essays. Maybe some of it is better suited as extra details in the docs rather than docstrings in code.

tuananhdao (approved):

PR #2772, opened 2026-09-09
Enhancements to overset

adperezm:
This PR does 2 things:

1. Adds a relaxation term in case one sees that subiterations oscilate a lot. The relaxation would hopefully smooth things out, but it affects convergence so it is a trade off.
2. Conserves high order restarts for the interfaces. I personally do not think that this is important. But I wanted to keep it.

Something here that maybe @timofeymukha can comment: I added an extra routine to the bc type that handels the restarts. It does nothing in all bcs except overset. This was the simplest solution I could think, but it was annoying I had to touch BC_t


timofeymukha (approved):
I think restart would have to appear soon anyway. Once I finally polish the checkpoint PR, it will also be trivial to add stuff there from bcs or whereever.

njansson (approved):

adperezm:
It seems like this one has been accepted but merging is blocked.

timfelle:
> It seems like this one has been accepted but merging is blocked.

Nah just nobody enabled auto-merge 😅

PR #2773, opened 2026-09-09
Introduce tensor4 type to the dofmap

timfelle:
This PR shows the effect of introducing the tensor4 type on the dofmap, `x`, `y` and `z`. This is the First attempt at doing this to remove explicit handling of the x and x_d pair in a complex class.

This PR depends on #2749


adperezm:
This is nice.

The idea would eventually to also have field own a tensor to contain the data, right? Very nice

adperezm (approved):

njansson (commented):
I really like the idea with the tensor type, but have a difficult time with the `%x%x` `%y%x` thingy. 

Crazy suggestion, how about we retain 3 rank-4 pointers (marked with contiguous) inside `dofmap_t`, let's call them x, y, z and these points to the tensor's %x components for each direction.

It would then become transparent for users, and smaller fallout of these changes

adperezm:
> the `%x%x` `%y%x` thingy.
> It would then become transparent for users, and smaller fallout of these changes


As a user, to me it actually makes sense. The fields and things we use already do the u%x. A bunch of other stuff does that as I remember. So it was in fact dofmap that was inconsistent with the rest.

To me it makes sense. y is the variable and %x is the container that has the data.

timfelle:
> I really like the idea with the tensor type, but have a difficult time with the `%x%x` `%y%x` thingy.
> 
> Crazy suggestion, how about we retain 3 rank-4 pointers (marked with contiguous) inside `dofmap_t`, let's call them x, y, z and these points to the tensor's %x components for each direction.
> 
> It would then become transparent for users, and smaller fallout of these changes

Right would you then also want a x_d y_d and z_d for the device pointers, which kinda defeats the entire purpose of trying to make this consistent with how fields are defined.

src/sem/dofmap.f90
```diff
@@ -792,11 +792,9 @@ subroutine dofmap_generate_xyz(this)
...
     end if
 
-    if (NEKO_BCKND_DEVICE .eq. 0) then
-       call this%x%copy_from(HOST_TO_DEVICE, .false.)
-       call this%y%copy_from(HOST_TO_DEVICE, .false.)
-       call this%z%copy_from(HOST_TO_DEVICE, .true.)
-    end if
+    call this%x%copy_from(HOST_TO_DEVICE, .false.)
+    call this%y%copy_from(HOST_TO_DEVICE, .false.)
+    call this%z%copy_from(HOST_TO_DEVICE, .true.)
```
njansson, line 797: Doesn't we still need the check, or will copy_from just silently drop the copy if no devs are present?

timfelle: Copy_from does have the check
```
 subroutine tensor4_copy_from(t, memdir, sync)
    class(tensor4_t), intent(inout) :: t
    integer, intent(in) :: memdir
    logical, intent(in) :: sync

       call device_memcpy(t%x, t%x_d, t%n, memdir, sync)
    end if

  end subroutine tensor4_copy_from
```

njansson (approved):

PR #2779, opened 2026-09-11
Add a factorised SVV formulation

shiyu-du-dsy:
Add a factorized SVV formulation. Different from directly inheriting SVV from spectral method, the factorized SVV is built upon the Galerkin construction and unconditionally symmetric.


njansson (commented):
The new ACCUMULATE boolean has to be added to all kernel variants, otherwise SVV is not guaranteed to work (Unless we force to user to disable autotuning for when SVV is used (bad!))

shiyu-du-dsy:
> The new ACCUMULATE boolean has to be added to all kernel variants, otherwise SVV is not guaranteed to work (Unless we force to user to disable autotuning for when SVV is used (bad!))

now it's fixed

PR #2780, opened 2026-09-11
Fix bug in SVV

### LINKED ISSUE #2777: SVV Bug
On lumi we get the following compile error for HIP compilation.

```
module ax_helm_svv_one_sided_full_cpu
       ^                              
ftn-855 ftn: ERROR AX_HELM_SVV_ONE_SIDED_FULL_CPU, File = math/bcknd/cpu/ax_helm_svv_one_sided_full_cpu.f90, Line = 34, Column = 8 
  The compiler has detected errors in module "AX_HELM_SVV_ONE_SIDED_FULL_CPU".  No module information file will be created for this module.

         coef%jacinv, Xh%w3, this%svv%h1, this%svv%filter%fh, &
                                                          ^     
ftn-1711 ftn: ERROR AX_HELM_SVV_ONE_SIDED_FULL_COMPUTE_VECTOR, File = math/bcknd/cpu/ax_helm_svv_one_sided_full_cpu.f90, Line = 84, Column = 59 
  An actual argument must be definable when associated with a dummy argument that has the intent(out) or intent(inout) attribute.  Argument "THIS" has the INTENT(IN) attribute.

         this%svv%filter%fht, this%svv%direction, this%svv%ident, &
                         ^                                          
ftn-1711 ftn: ERROR AX_HELM_SVV_ONE_SIDED_FULL_COMPUTE_VECTOR, File = math/bcknd/cpu/ax_helm_svv_one_sided_full_cpu.f90, Line = 85, Column = 26 
  An actual argument must be definable when associated with a dummy argument that has the intent(out) or intent(inout) attribute.  Argument "THIS" has the INTENT(IN) attribute.
                                                           ^        
ftn-1711 ftn: ERROR AX_HELM_SVV_ONE_SIDED_FULL_COMPUTE_VECTOR, File = math/bcknd/cpu/ax_helm_svv_one_sided_full_cpu.f90, Line = 85, Column = 60 
  An actual argument must be definable when associated with a dummy argument that has the intent(out) or intent(inout) attribute.  Argument "THIS" has the INTENT(IN) attribute.

```

njansson:
Fix compilation bug in SVV, fixes #2777

### DIFF
--- a/src/math/bcknd/cpu/ax_helm_svv_one_sided_full_cpu.f90
+++ b/src/math/bcknd/cpu/ax_helm_svv_one_sided_full_cpu.f90
@@ -156,9 +156,9 @@ contains
     real(kind=rp), intent(in) :: Dyt(lx,lx)
     real(kind=rp), intent(in) :: Dzt(lx,lx)
     real(kind=rp), intent(in) :: svv_h1(lx, lx, lx, n)
-    real(kind=rp), intent(inout) :: svv_Q(lx, lx), svv_Qt(lx, lx)
+    real(kind=rp), intent(in) :: svv_Q(lx, lx), svv_Qt(lx, lx)
     character(len=*), intent(in) :: svv_direction
-    real(kind=rp), intent(inout) :: ident(lx, lx)
+    real(kind=rp), intent(in) :: ident(lx, lx)
 
     real(kind=rp) :: s11_h, s22_h, s33_h, s12_h, s13_h, s23_h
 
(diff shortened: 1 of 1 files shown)


shiyu-du-dsy (approved):

shiyu-du-dsy:
I don't understand why the old code can pass the the compiler both on my workstation and the PR check with such a bug

timfelle:
> I don't understand why the old code can pass the the compiler both on my workstation and the PR check with such a bug

Very easy, your compilers help you just like Overleaf does, "I think you did something wrong so I will fix it for you and never tell you."

njansson:
It is a bit strange that our CI complained, since we do the GNU build with pedantic flags

timfelle:
> It is a bit strange that our CI complained, since we do the GNU build with pedantic flags

`-g -O2 -pedantic -std=f2008 -w`

Sure, but then we suppress all warnings afterward 😅 
Technically we should tailor the warnings and the do `-Werror` to throw if we hit them.

njansson:
> > It is a bit strange that our CI complained, since we do the GNU build with pedantic flags
> 
> `-g -O2 -pedantic -std=f2008 -w`
> 
> Sure, but then we suppress all warnings afterward 😅 Technically we should tailor the warnings and the do `-Werror` to throw if we hit them.

that will fail our fast formulation (I,1,1,1) ;), but pedantic will fail the std violating things. But that said, gfortran is forgiving

timfelle:
> > > It is a bit strange that our CI complained, since we do the GNU build with pedantic flags
> > 
> > 
> > `-g -O2 -pedantic -std=f2008 -w`
> > Sure, but then we suppress all warnings afterward 😅 Technically we should tailor the warnings and the do `-Werror` to throw if we hit them.
> 
> that will fail our fast formulation (I,1,1,1) ;), but pedantic will fail the std violating things. But that said, gfortran is forgiving

Maybe we want `-pedantic-errors` then?

timfelle (approved):

PR #2782, opened 2026-09-11
Use the kstep layout for the coef device kernels

timofeymukha:
The three per-element coef kernels gave each element a block of 1024 threads. Registers are allocated per block, so with a high enough per thread count the block asks for more than the 65536 registers it can be given and the launch fails with "too many resources requested for launch". Built for sm_120 with nvcc 13 in double precision, the area and normal kernel takes 70 registers per thread at lx = 2 and the dxyz kernel 126 at lx = 16, both past the limit at 1024 threads. The lx = 2 case is reached by the coarsest level of the HSMG pressure preconditioner whatever the case's polynomial order is, and the same 1024 threads leave most of a block idle at low order.

The CUDA and HIP kernels now use the kstep layout of the SEM operator kernels: one (lx,lx) thread plane per element walking the element in k, with several elements stacked along threadIdx.z, which holds a block between 64 and 256 threads for every supported order. Elements per block are fixed at compile time rather than picked by an autotuner as in elem_block.h, since coef runs once per mesh. Results are bitwise identical to the old kernels for lx = 2..16 in both precisions.

The Metal backend hard-codes the same 1024, where the limit surfaces as the pipeline's maxTotalThreadsPerThreadgroup rather than as a register budget; its three dispatches now clamp to what the pipeline reports. Those kernels already stride with threads_per_threadgroup, so they need no change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Develop
 
+- Fixed `too many resources requested for launch` from the CUDA and HIP `coef`
+  kernels, which gave each element 1024 threads. Registers are allocated per
+  block, and at 1024 threads the per-thread count of two of these kernels
+  exceeded the 65536 registers a block can be given: 70 at `lx = 2` in the
+  area and normal kernel, which the coarsest level of the HSMG pressure
+  preconditioner reaches whatever the case's polynomial order is, and 126 at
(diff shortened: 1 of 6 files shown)


timofeymukha:
@njansson is this what you had in mind? Otherwise feel free to guide the machine to the right fix off this branch.

PR #2792, opened 2026-09-17
Allow scratch request of direct real array

timfelle:
EDIT: Also extended the approach to device arrays.

This change allows for more intuitive management of real scratch arrays. Now you can just request a real pointer to a 1d array from the registry.

This removes the need to do the extra %x indexing when using a 1d scratch array.

Usage:
```fortran
real(kind=rp), pointer :: x(:)
type(c_ptr) :: x_d
integer :: temp_index(2), n

! Assign size
n = 10
call neko_scratch_registry%request(x, temp_index(1), n, clear = .false.)
call neko_scratch_registry%request(x_d, temp_index(2), n, clear = .false.)

! Perform work

nullify(x)
x_d = C_NULL_PTR
call neko_scratch_registry%relinquish(temp_index)
```


src/registries/scratch_registry.f90
```diff
@@ -359,7 +368,63 @@ subroutine request_host_array(this, v, index, n, clear)
...
+  end subroutine request_host_array_t
+
+  !> Get a host array from the registry by assigning it to a pointer.
+  !! @param v Pointer to the requested host_array.
+  !! @param index Index of the host array in the registry (for
+  !! relinquishing later).
+  !! @param n Size of the requested host_array.
+  !! @param clear If true, the host_array values are set to zero upon request.
+  subroutine request_real_array(this, v, index, n, clear)
+    class(scratch_registry_t), target, intent(inout) :: this
```
timofeymukha, line 325: Maybe call this host_array now that the other is host_array_t? I think with a proper docstring that signals the intent a bit better. It is breaking, but I think the host array stuff is so fresh that it is ok

timfelle: That is a good idea.

timfelle: done

njansson (approved):

timofeymukha (approved):

PR #2796, opened 2026-09-18
rework the logging of sponge source term

vbaconnet:
Some changes to the logging of the sponge.

If anyone wants minimal changes to the source term done, please write here and we can piggyback on this PR to get things in.


loretet (approved):
This all looks good to me, useful addition! For me the source term is already very much okay, nothing to add comes to mind.

src/source_terms/sponge_source_term.f90
```diff
@@ -134,7 +134,7 @@ subroutine sponge_init_from_json(this, json, fields, coef, variable_name)
     type(json_file) :: baseflow_subdict
     integer :: i, izone
 
-    call neko_log%section("SPONGE SOURCE TERM", LVL = NEKO_LOG_DEBUG)
+    call neko_log%section("Sponge source term", LVL = NEKO_LOG_INFO)
```
timfelle, line 137: Remember start and end section should probably share a logging level.

src/source_terms/sponge_source_term.f90
```diff
@@ -290,12 +294,11 @@ subroutine sponge_init_field(this, fields, coef, start_time, end_time, &
          amplitudes, fringe_registry_name, bf_registry_pref, dump_fields, &
          dump_fname)
 
+    call neko_log%message("Baseflow   : field")
+
```
timfelle: Maybe log the filename as well.

src/source_terms/sponge_source_term.f90
```diff
@@ -359,21 +364,35 @@ subroutine sponge_init_common(this, fields, coef, start_time, end_time, &
...
     this%dump_fname = trim(dump_fname)
 
-    call neko_log%message("Initializing sponge", lvl = NEKO_LOG_DEBUG)
+    call neko_log%message("Fringe name: " // trim(fringe_registry_name), &
+         lvl = NEKO_LOG_INFO)
+    call neko_log%message("Baseflow prefix  : " // trim(bf_registry_pref), &
+         lvl = NEKO_LOG_DEBUG)
+
+    write (log_buf, "(A,L1)") "Dump fields: ", this%dump_fields
+    call neko_log%message(log_buf, lvl=NEKO_LOG_INFO)
```
timfelle: Would it make sense to align output at ":"?

timfelle (approved):

src/source_terms/sponge_source_term.f90
vbaconnet: it is logged right after when import fields is called, seemed a bit redundant

nick-morse (approved):

src/source_terms/sponge_source_term.f90
vbaconnet: did some changes, you can resolve this conversation if/when you're happy with them :)

PR #2800, opened 2026-09-20
lightweight python tools

Mr-2c:
This covers only for some of the simpler tools that don't require the full neko. It is basically the fully python implementation of the previously closed PR that had a mixture of fortran and python implementations. The main python implementation is inside the subfolder `nekolight` with the python code inside `pylight` itself is mostly command line interface. 

This is an older version of the tools (1-2 months) which I have tested myself. A newer version with some additional functionalities (like direct conversion to nmsh from gmsh, etc.) will be coming later (maybe another month or two). 

And as usual, it was written by my advanced friend :-)


timofeymukha:
To me it makes most sense to have this as a separate Python package (nmshtools or whatever) in its own repo, which I would add under contrib here as a git submodule. This way the connection to Neko is clear, but it is a bit at arms length. Furthermore, having it as its own package means one can easily install it, reuse the functionality and add things to it, without touching Neko proper.

PR #2803, opened 2026-09-21
Fix evaluation of Navier-Stokes fluxes on device backend

njansson:
This PR fixes the evaliation of the NS fluxes on the device backend. Currently, this is purely evaluated on the host arrays, which is both wrong (not up to date with the device, and also slow...)


timfelle (approved):

tuananhdao (approved):
LGTM!

PR #2804, opened 2026-09-21
Land the time step exactly on the scheduled output and sampling times

Mr-2c:
## Summary

Adds the opt-in option `case.time.exact_output_time`. When it is set, the time
step is shortened where needed so that the simulation reaches the scheduled
times of every time based controller in the case exactly, instead of at the
first step past them, and ends exactly at `end_time`. This covers the fluid,
checkpoint and simulation component outputs, and the `preprocess_control`,
`compute_control` and `output_control` of the simulation components, for the
`simulationtime` and `nsamples` controls. It works with a fixed and with a
variable time step, and the shortened steps never feed back into the CFL
controller.

Without the option nothing changes for a run, apart from the two items
described under "Other changes".

## Motivation

The reworked time based controller (#2764) fixed *which* scheduled times are
executed, but an execution still happens at the first time step that reaches
the scheduled time, so the time actually written or sampled lies anywhere in
`[t_k, t_k + dt)`. 
The step has to land exactly on the scheduled times
for field files that are to be used for POD/FFT, for probes,
for statistics to avoid the question of how to weight the samples,
for post-processing that compares runs, and in many other situaltions.

## User facing behaviour

In the `time` object of the case file:

| Name                   | Description                                                                          | Default |
| ---------------------- | ------------------------------------------------------------------------------------ | ------- |
| `exact_output_time`    | Whether to shorten `dt` so that the sampling and output times are reached exactly    | `false` |
| `output_landing_steps` | Steps ahead of a sampling or output time over which `dt` is shortened to land on it  | `10`    |

Once the next scheduled time is within `output_landing_steps` steps, the
remaining time is divided into the smallest whole number of equal steps that
are no longer than the step asked for. The step therefore only ever gets
shorter, never longer, and it is back to the step asked for right after the
scheduled time. With the default of ten steps, a scheduled time approached from
further away than that changes the step by at most a tenth. With a fixed time
step the step asked for is `timestep`; with a variable time step it is the one
the CFL controller settles on, which keeps working on it as if the step had
not been shortened.

Guarantees and limits, as documented in `case-file.md`:

- No step is shorter than a tenth of the step asked for or, with a variable
  time step, than `min_timestep`. Two schedules with unrelated intervals can
  put scheduled times closer to each other than that; the first is landed on
  and the second is executed within that shortest step of its time. A
  scheduled time closer than that to `end_time` is executed at `end_time`,
  and a run that comes closer than that to `end_time` without landing on it,
  such as one started there, ends within that step past it.
- An interval shorter than the time step executes at every step, as without
  the option, and does not shorten the step to the interval. An interval
  between one and two steps is landed on with two steps each, so the run
  then takes up to twice the steps.
- `tsteps` controls count steps and are not affected.
- Even when the step divides the interval, the last step before a scheduled
  time absorbs the round-off of the accumulated time and differs from the
  step asked for by round-off.
- The schedules are a property of the case, so a restart lands on the same
  times as the uninterrupted run.
- A change of the step invalidates the projection spaces of the velocity and
  pressure solves, so they are cleared and rebuilt around each scheduled
  time; expect a few more solver iterations there.
- In an MPMD run the option has to be set in all the coupled cases (they take
  the shortened steps together); a mismatch is an error at setup.

## How it works

The landing is built on the schedules the controllers already hold; nothing
is duplicated or predicted outside them.

1. `time_based_controller_t%time_to_next(time, dt)` returns the time until
   the next scheduled time the step should land on, or `huge` if there is
   none. It uses the controller's own cursor (`next_time`), its start time,
   direction and end time, so `simulationtime` (anchored at zero), `nsamples`
   (anchored at the output's start) and an output with its own `start_time`
   are all handled. Two details matter:
   - An execution that is already due at the current time (which happens for
     the simulation components, which are not checked before the time loop,
     and for a scheduled time reached within the check's tolerance) is
     performed at the coming step wherever it lands, so the time to land on is
     the scheduled time *after* it, which would otherwise be passed by. When
     the landing takes a single step, the two are executed together at the
     later one.
   - An interval shorter than the step returns `huge`.
2. `output_controller_t%time_to_next`, `simulation_component_t%time_to_next`
   (overridden by `lpt_simcomp_t`, whose particle tracker holds a controller
   of its own) and `simcomp_executor_t%time_to_next` take the minimum over
   their controllers.
3. `time_step_controller_t%land(time, time_to_next)` is called from
   `simulation_step` right after `set_dt`. It also considers `end_time`,
   decides whether the landing is engaged (within `landing_steps` steps, and
   never later than one step plus the shortest step allowed ahead, so that a
   full step cannot leave less than that shortest step), splits the remaining
   time into equal steps, applies the floor, and in an MPMD run takes the
   minimum over the coupled simulations. The shortened step is flagged in
   `dt_is_landing`, and the next `set_dt` restores the step asked for and
   rescales the CFL number it was given, since that was computed with the
   shortened step. `simulation_step` recomputes the logged CFL number
   whenever the step changed.
4. A change of the step, whatever its cause, resets `dt_last_change`, and
   the projection spaces now consult `dt_may_change()` (variable time step
   *or* landing) instead of `is_variable_dt` when deciding whether to clear
   and rebuild. Without that, a landing step in a fixed time step run would
   reuse a projection basis built for a different Helmholtz operator.

## Other changes

- **First step of a variable time step run.** `simulation_init` now sets the
  first step, through the new `first_dt`, before the initial output is
  checked. The initial output used to be checked against the placeholder
  step of one time unit, so an output scheduled within a tenth of its
  interval after the start of the run was written at the start instead. The
  first step of the loop is unchanged up to round-off, as `set_dt` computes
  the same value.
- **MPMD.** Both a variable time step and the landing take a collective at
  every step, so the coupled cases have to agree on `variable_timestep` and
  on `exact_output_time`. A mismatch is now an error at setup; for
  `variable_timestep` it used to hang at the first step.

## Files changed

| File | Change |
| --- | --- |
| `src/common/time_step_controller.f90` | New options, `land`, `first_dt`, `dt_may_change`, the restore of the step in `set_dt`, the MPMD agreement check in `init`. |
| `src/common/time_based_controller.f90` | `time_to_next`; `check` is now `pure`. |
| `src/io/output_controller.f90`, `src/simulation_components/simulation_component.f90`, `simcomp_executor.f90`, `lpt_simcomp.f90` | `time_to_next` aggregation. |
| `src/simulation.f90` | Calls `land` after `set_dt`, recomputes the logged CFL when the step changed, sets the first variable step before the initial output, logs the option. |
| `src/common/projection.f90`, `projection_vel.f90` | `dt_may_change()` instead of `is_variable_dt`. |
| `src/.depends` | Dependencies of the files with new `use` statements. |
| `tests/unit/time_step_controller/` | New suite (13 tests): fixed and variable step, one landing step, `min_timestep`, close targets and the floor, interval shorter than the step, exact end, a scheduled time just before the end, restart, backwards in time, `dt_last_change` bookkeeping, reading the options. Added to `configure.ac` and `tests/unit/Makefile.am`. |
| `tests/unit/time_based_controller/test_time_based_controller.pf` | Tests for `time_to_next`, including an output's own start time and a run backwards in time. |
| `tests/integration/tests/test_output_schedule/test_output_schedule.py` | A part with a step of 0.007 that does not divide the output interval of 0.05, checking that the steps land on 0.05 and the run ends at 0.1. |
| `doc/pages/user-guide/case-file.md`, `statistics-guide.md`, `doc/schemas/time.schema.json`, `CHANGELOG.md` | Documentation of the option and the fixes. |

## Suggested order for reviewing

1. `time_based_controller.f90`: `time_to_next` (the rules above).
2. `time_step_controller.f90`: `land`, then the changes to `set_dt` and
   `init`.
3. `simulation.f90` for the two call sites, then the aggregation in the
   output controller and the simulation components.
4. The projection files, together with `dt_may_change`.
5. The tests, which double as worked examples of the behaviour.

## Testing

- The two unit suites (22 and 13 tests) and the integration tests
  `test_output_schedule`, `test_scalar_restart` and `test_cylinder` pass;
  the latter two do not use the option and are unchanged.
- The real controller modules were also driven off-tree through a copy of
  the time loop against an independent enumeration of the schedules, over
  more than 80 000 randomised and gridded scenarios: steps from 1e-6 to
  1e-1, intervals from a hundredth of a step to a thousand steps and
  unrelated to each other, up to four schedules of all control modes,
  outputs with their own start time, start times up to 1000, variable steps,
  restarts, runs backwards in time and 1 to 50 landing steps. Every
  scheduled time of a participating schedule is hit to round-off, the step
  bounds and the rules for the end hold, no run stalls, and the number of
  steps is at most one more per scheduled time than without the option.
- The repository lint score is unchanged.

## Acknowledgements

Implemented with Claude Code.


src/common/time_step_controller.f90
```diff
@@ -99,92 +138,309 @@ subroutine time_step_controller_init(this, params)
...
 
+  !> Whether this is one of several simulations coupled in an MPMD run, in
+  !! which case the time step is agreed on across them at every step.
+  pure function is_mpmd() result(mpmd)
+    logical :: mpmd
+
+    mpmd = global_pe_size .gt. 0 .and. pe_size .ne. global_pe_size
+
+  end function is_mpmd
+
```
timofeymukha, line 182: This is a good-to-have function, @Mr-2c can we put it into the utils module?

src/common/time_step_controller.f90
```diff
@@ -99,92 +138,309 @@ subroutine time_step_controller_init(this, params)
...
+       call MPI_Allreduce(MPI_IN_PLACE, flags_any, 2, MPI_INTEGER, &
+            MPI_MAX, NEKO_GLOBAL_COMM, ierr)
+       call MPI_Allreduce(MPI_IN_PLACE, flags_all, 2, MPI_INTEGER, &
+            MPI_MIN, NEKO_GLOBAL_COMM, ierr)
+       if (any(flags_any .ne. flags_all)) then
+          call neko_error('variable_timestep and exact_output_time must &
+               &each be set in all the coupled cases of an MPMD run, or in &
+               &none')
+       end if
+    end if
```
timofeymukha, line 170: @adperezm This looks reasonable to me, but wanted to check with you as the MPMD master!

PR #2808, opened 2026-09-22
Add backend for sponge + allow for shared baseflows

vbaconnet:
Two contributions:
- Add proper cpu/gpu backends for the sponge, similar to other source terms 
- Allow for sponge objects to use a baseflow field from other sponge objects, essentially by doing nothing and just pointing at an already existing baseflow field in the registry. This is needed when using multiple sponge objects that need the same baseflow, in which case the user had to create 3 new fields (u,v,w) for every object which would contain the same exact field values. It's not a common situation but good to have I believe (I've been using this extensively in my cases).


src/source_terms/bcknd/device/sponge_source_term_device.f90
```diff
@@ -0,0 +1,92 @@
...
+    type(field_t), pointer :: wk
+
+    call neko_scratch_registry%request_field(wk, tmp_index, .false.)
+
+    ! The RHS components
+    fu => fields%get_by_index(1)
+    fv => fields%get_by_index(2)
+    fw => fields%get_by_index(3)
+
+    n = fu%size()
```
timfelle: You can avoid allocating a new field on the CPU by just requesting a device array directly instead
```suggestion
    type(c_ptr) :: wk

    ! The RHS components
    fu => fields%get_by_index(1)
    fv => fields%get_by_index(2)
    fw => fields%get_by_index(3)

    n = fu%size()
    call neko_scratch_registry%request(wk, tmp_index, n, .false.)
```

src/source_terms/bcknd/device/sponge_source_term_device.f90
```diff
@@ -0,0 +1,92 @@
...
+
+    call device_sub3(wk%x_d, v_bf%x_d, v%x_d, n)
+    call device_col2(wk%x_d, fringe%x_d, n)
+    call device_add2s2(fv%x_d, wk%x_d, a_y, n)
+
+    call device_sub3(wk%x_d, w_bf%x_d, w%x_d, n)
+    call device_col2(wk%x_d, fringe%x_d, n)
+    call device_add2s2(fw%x_d, wk%x_d, a_z, n)
+
+    call neko_scratch_registry%relinquish_field(tmp_index)
```
timfelle:
```suggestion
    call neko_scratch_registry%relinquish(tmp_index)
```

src/source_terms/bcknd/device/sponge_source_term_device.f90
```diff
@@ -0,0 +1,92 @@
...
+    ! fu = fu + a_x*wk = fu + a_x*fringe*(u_bf - u)
+    call device_add2s2(fu%x_d, wk%x_d, a_x, n)
+
+    call device_sub3(wk%x_d, v_bf%x_d, v%x_d, n)
+    call device_col2(wk%x_d, fringe%x_d, n)
+    call device_add2s2(fv%x_d, wk%x_d, a_y, n)
+
+    call device_sub3(wk%x_d, w_bf%x_d, w%x_d, n)
+    call device_col2(wk%x_d, fringe%x_d, n)
+    call device_add2s2(fw%x_d, wk%x_d, a_z, n)
```
timfelle: For now, fair enough. But maybe these should be merged instead of lining up such
a long queue for the gpu.

vbaconnet: yea that was my intention, but then I saw the other source terms and thought i'd stop here for now. didn't feel like implementing the kernels... :)

src/source_terms/bcknd/device/sponge_source_term_device.f90
vbaconnet: great!

src/source_terms/bcknd/device/sponge_source_term_device.f90
```diff
@@ -0,0 +1,97 @@
...
+       u_bf, v_bf, w_bf, fringe, a_x, a_y, a_z)
+    type(field_list_t), intent(inout) :: fields
+    type(field_t), intent(in) :: u, v, w, fringe, u_bf, v_bf, w_bf
+    real(kind=rp), intent(in) :: a_x, a_y, a_z
+    integer :: n
+    type(field_t), pointer :: fu, fv, fw
+    integer :: tmp_index
+    type(c_ptr), pointer :: wk
+
+    call neko_scratch_registry%request(wk, tmp_index, n, .false.)
```
timfelle: n is not defined yet, this should be moved after `n = `

src/source_terms/bcknd/device/sponge_source_term_device.f90
```diff
@@ -0,0 +1,97 @@
...
+  !! @param a_z The amplitude in the z-direction
+  subroutine sponge_source_term_compute_device(fields, u, v, w, &
+       u_bf, v_bf, w_bf, fringe, a_x, a_y, a_z)
+    type(field_list_t), intent(inout) :: fields
+    type(field_t), intent(in) :: u, v, w, fringe, u_bf, v_bf, w_bf
+    real(kind=rp), intent(in) :: a_x, a_y, a_z
+    integer :: n
+    type(field_t), pointer :: fu, fv, fw
+    integer :: tmp_index
+    type(c_ptr), pointer :: wk
```
timfelle: No pointer specification needed.
```suggestion
    type(c_ptr) :: wk
```

timfelle (changes requested):

timofeymukha (approved):
Looks good sans Tim's comments.

src/source_terms/bcknd/device/sponge_source_term_device.f90
vbaconnet: yikes... thanks

timfelle (approved):

src/source_terms/bcknd/device/sponge_source_term_device.f90
```diff
@@ -0,0 +1,98 @@
...
+       u_bf, v_bf, w_bf, fringe, a_x, a_y, a_z)
+    type(field_list_t), intent(inout) :: fields
+    type(field_t), intent(in) :: u, v, w, fringe, u_bf, v_bf, w_bf
+    real(kind=rp), intent(in) :: a_x, a_y, a_z
+    integer :: n
+    type(field_t), pointer :: fu, fv, fw
+    integer :: tmp_index
+    type(c_ptr) :: wk
+
+
```
timfelle:
```suggestion

```

src/source_terms/bcknd/device/sponge_source_term_device.f90
```diff
@@ -0,0 +1,98 @@
...
+    type(c_ptr) :: wk
+
+    
+    ! The RHS components
+    fu => fields%get_by_index(1)
+    fv => fields%get_by_index(2)
+    fw => fields%get_by_index(3)
+    
+    n = fu%size()
+
```
timfelle:
```suggestion

    n = fu%size()

```

timfelle (approved):

PR #2809, opened 2026-09-23
Update variable time initial timestep

timfelle:
Previous logic was flawed. If no timestep was specified and flow was uniform 0, initial timestep was infinite.


timfelle:
This also does not work

timofeymukha:
@timfelle I think the PR draft from Siavash does something to the initial timestep as well. But I don't remmeber what :P

